PR #31, opened 2019-09-17
Moved provider_name values to the correct section

kgriffiths:
The list of provider values should be in the provider_name section


marinsalinas (approved):
LGTM!! Thanks!!

PR #32, opened 2019-09-18
Add Private IP Mode Resource.

### LINKED ISSUE #24: Unable to create project with peering only connections
When creating a GCP peered cluster, the project configuration doesn't allow you to set it for peer networking only.

In order to create a project with peer networking you need to:

 - create the project (terraform)
 - switch to the UI and enable this
 - create your cluster(terraform)

This is supported by the Atlas API (https://docs.atlas.mongodb.com/reference/api/set-private-ip-mode-for-project).

Looking through the source, it may be possible relatively easily to add a field to the project stanza (https://github.com/terraform-providers/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_project.go#L21-L38) which can then fire off the appropriate API call.

It's possible I've missed some documentation so thoughts would be very welcome

marinsalinas:
closes #24 

## Added:

- mongodbatlas_private_ip_mode resource
- mongodbatlas_private_ip_mode resource acceptance testing
- mongodbatlas_private_ip_mode resource website documentation
- Minor documentation's format fixes.

## Example Configuration:

```hcl
resource "mongodbatlas_private_ip_mode" "my_private_ip_mode" {
    project_id = "<YOUR PROJECT ID>"
	enabled    = true
}
```

### DIFF
--- a/website/docs/r/private_ip_mode.html.markdown
+++ b/website/docs/r/private_ip_mode.html.markdown
@@ -1,0 +1,50 @@
+---
+layout: "mongodbatlas"
+page_title: "MongoDB Atlas: private_ip_mode"
+sidebar_current: "docs-mongodbatlas-resource-private-ip-mode"
+description: |-
+    Provides a Private IP Mode resource.
+---
+
+# mongodbatlas_private_ip_mode
+
+`mongodbatlas_private_ip_mode` provides a Private IP Mode resource. This allows one to enable/disable Connect via Peering Only mode for a MongoDB Atlas Project.
+
+
+~> **IMPORTANT**: <br>**What is Connect via Peering Only Mode?** <br>Connect via Peering Only mode prevents clusters in an Atlas project from connecting to any network destination other than an Atlas Network Peer. Connect via Peering Only mode applies only to **GCP** and **Azure-backed** dedicated clusters. This setting disables the ability to: <br><br>• Deploy non-GCP or Azure-backed dedicated clusters in an Atlas project, and
+<br>• Use MongoDB Stitch with dedicated clusters in an Atlas project.
+
+
+-> **NOTE:** You should create one private_ip_mode per project.
+
+## Example Usage
+
+```hcl
+resource "mongodbatlas_private_ip_mode" "my_private_ip_mode" {
+    project_id = "<YOUR PROJECT ID>"
+	enabled    = true
+}
+```
+
+## Argument Reference
+
+* `project_id` - (Required) The unique ID for the project to enable Only Private IP Mode.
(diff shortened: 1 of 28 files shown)


PacoDw (approved):
Very nice! 👍

thetonymaster:
Just a question, `enabled` can be toggled or does it make sense to be toggled? If not, remove it and always send `true` to enable this functionality.

thetonymaster:
Just went through the code, for me it makes more sense to have this resource as:

````
resource "mongodbatlas_private_ip_mode" "my_private_ip_mode" {
    project_id = "<YOUR PROJECT ID>"
}
````

Since deleting it will disable the private IP functionality

File: website/docs/r/private_ip_mode.html.markdown [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+sidebar_current: "docs-mongodbatlas-resource-private-ip-mode"
+description: |-
+    Provides a Private IP Mode resource.
+---
+
+# mongodbatlas_private_ip_mode
+
+`mongodbatlas_private_ip_mode` provides a Private IP Mode resource. This allows project to enable Connect via Peering Only mode.
+
+
```
themantissa, line 13: Add in box (so we call out here only applicable to GCP/Azure): 

What is Connect via Peering Only Mode?

Connect via Peering Only mode prevents clusters in an Atlas project from connecting to any network destination other than an Atlas Network Peer. **Connect via Peering Only mode applies only to GCP and Azure-backed dedicated clusters**. This setting disables the ability to:

- Deploy non-GCP or Azure-backed dedicated clusters in an Atlas project, and
- Use MongoDB Stitch with dedicated clusters in an Atlas project.

website/docs/r/private_ip_mode.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+layout: "mongodbatlas"
+page_title: "MongoDB Atlas: private_ip_mode"
+sidebar_current: "docs-mongodbatlas-resource-private-ip-mode"
+description: |-
+    Provides a Private IP Mode resource.
+---
+
+# mongodbatlas_private_ip_mode
+
+`mongodbatlas_private_ip_mode` provides a Private IP Mode resource. This allows project to enable Connect via Peering Only mode.
```
themantissa: Modify:

provides a Private IP Mode resource. This allows one to enable/disable Connect via Peering Only mode for a MongoDB Atlas Project.

themantissa (changes requested):
A few docs changes but looks good otherwise - just need to be able to enable/disable and I think this more than covers it!

PR #45, opened 2019-10-03
fix-#40: added true value by defualt on provider_encrypt_ebs_volume attr

### LINKED ISSUE #40: AWS clusters default provider_encrypt_ebs_volume to false
Hey folks,

If I omit `provider_encrypt_ebs_volume` in my resource definition for an AWS cluster, this value defaults to `false`. However, the [Atlas API](https://docs.atlas.mongodb.com/reference/api/clusters-create-one/) defaults this to `true`.

I think it would also be beneficial to remove `provider_encrypt_ebs_volume: false` from the Example AWS cluster in the [**mongodbatlas_cluster** docs](https://www.terraform.io/docs/providers/mongodbatlas/r/cluster.html) as encrypting your volume is the preferred config.

Thanks!

PacoDw:
closes #40

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -625,7 +625,10 @@ func expandProviderSetting(d *schema.ResourceData) matlas.ProviderSettings {
 
 	if d.Get("provider_name") == "AWS" {
 		providerSettings.DiskIOPS = pointy.Int64(cast.ToInt64(d.Get("provider_disk_iops")))
-		providerSettings.EncryptEBSVolume = pointy.Bool(cast.ToBool(d.Get("provider_encrypt_ebs_volume")))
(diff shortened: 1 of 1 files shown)


themantissa (approved):
Thank you!  I'll submit a PR for the example change in a bit!

marinsalinas (commented):
LGTM but, why did you choose to fix it like this instead of using `Default: true` in the schema definition like we've done here:

https://github.com/terraform-providers/terraform-provider-mongodbatlas/blob/26f10ee1c268910ae659a80e2760f5c88c4e099d/mongodbatlas/resource_mongodbatlas_cluster.go#L50

PacoDw:
@marinsalinas it didn´t work because the schema doesn´t allow you when you have set the computed attribute

themantissa (approved):
I'm good so approving but Marin is the final word on this one per his question.  Thanks!

marinsalinas (approved):
LGTM!

PR #60, opened 2019-11-06
Cluster advanced configuration Options

### LINKED ISSUE #52: specify oplog size?
hello,

when will the `mongodbatlas_cluster` resource have support for setting the oplog size?  i dont see anything in the docs atm.

i also stood up a cluster via tf,  set the oplog size manually in the console, and then ran a `terraform plan` which results in `No changes. Infrastructure is up-to-date.` .  This would imply oplog size isn't supported at all yet.

thanks.

PacoDw:
- The MongoDB client was updated, incorporating the advanced configuration options.
- Added the advanced configuration options to the cluster resource.
- Added test case using the advanced configuration options.
- The documentation was improved, adding a cluster resource example with the advanced configuration options.

#### Example Advanced Configuration Options

```hcl
  project_id   = "<YOUR-PROJECT-ID>"
  name         = "cluster-advanced-configuration"
  disk_size_gb = 10

  // Provider Settings "block"

  // Advanced Configuration Options
  advanced_configuration = {
    fail_index_key_too_long              = true
    javascript_enabled                   = true
    minimum_enabled_tls_protocol         = "TLS1_2"
    no_table_scan                        = true
    oplog_size_mb                        = 1000
    sample_size_bi_connector             = 310
    sample_refresh_interval_bi_connector = 110
  }
}
```

closes #52


marinsalinas (approved):

website/docs/r/cluster.html.markdown
```diff
@@ -167,6 +167,39 @@ resource "mongodbatlas_cluster" "cluster-test" {
...
+  mongo_db_major_version       = "4.0"
+
+  // Provider Settings "block"
+  provider_name               = "AWS"
+  provider_disk_iops          = 100
+  provider_encrypt_ebs_volume = false
+  provider_instance_size_name = "M10"
+  provider_region_name        = "EU_CENTRAL_1"
+
+  // Advanced Configuration Options
```
themantissa: I'd prefer we document the options but not include them in a sample.  People are often used to just copying examples and these should only be used by those in the know.

website/docs/r/cluster.html.markdown
```diff
@@ -275,6 +308,21 @@ Physical location of the region.
 
     If you do not specify this option, no analytics nodes are deployed to the region.
 
+### Advanced Configuration Options
+
```
themantissa, line 279: Add beneath the header but before the options:  Prior to setting these options please ensure you read https://docs.atlas.mongodb.com/cluster-config/additional-options/.

themantissa (commented):
Two minor changes and one question below.    

These are only valid for M10s or above (M0/M2/M5 cant' take these) - not sure if we've checked that before but fyi.  I'm assuming we just let the API respond w/ the error in that case?

website/docs/r/cluster.html.markdown
PacoDw: Oh, Thank you, It's a good point, okay I will remove the example, and for another hand, it's not available for M0/M2/M5 and if you make a request with some M0/M2/M5 the server responds with the following error:

`400 (request "Bad Request") An invalid enumeration value M5 was specified`

website/docs/r/cluster.html.markdown
PacoDw: okay, thanks @themantissa 👍

themantissa (approved):
Thank you for the quick corrections!  LGTM!

marinsalinas (approved):

PR #63, opened 2019-11-07
Added format function to handle the mongo_db_major_version attr

### LINKED ISSUE #58: Syntax are not mandatory and creates confusion
To launch a cluster using terraform,
mongo_db_major_version       = “4.0” // this is in documentation and requires the version to be in string, but with version 4.2
mongo_db_major_version       = 4.2 // this will work without the “”. 

Please see if any changes need to be made to make this mandatory.

PacoDw:
- Added a format function to allow string and float values to be able to set for the mongo_db_major_version attr.

For example, now you can use as a following:

`mongo_db_major_version = “4.0”` // with ""
`mongo_db_major_version = 4.0`   // without ""
`mongo_db_major_version = 4`      // even you can use it of this way

closes #58

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -98,9 +98,10 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Required: true,
 			},
 			"mongo_db_major_version": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:      schema.TypeString,
+				Optional:  true,
+				Computed:  true,
+				StateFunc: formatMongoDBMajorVersion,
 			},
 			"num_shards": {
 				Type:     schema.TypeInt,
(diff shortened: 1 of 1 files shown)


themantissa:
Since the underlying API requires a string shouldn't we not allow numerical values?  My thinking around #58 was the enforce it to be a string more than opening it up to more options.   Not sure it matters as long as we do, as you've added, convert it properly but I can see taking it in multiple ways being more confusing than enforcing a type.

PacoDw:
@themantissa, we can't enforce it to be string type in Terraform because if you set 4.2 (type float), Terraform automatically converts it to a string, so I can't validate it because it is already a string. But if you set it to 4.0 (type float), Terraform converts it to string but removes the 0 value, so Terraform sends a wrong value ("4") instead of "4.0". For this reason, Terraform allows 4.2 (type float) without a problem, so I enabled to use of type float to avoid this problem. Behind the scenes, I convert and corroborate the value to be the correct string value that accepts the MongoDB server.

themantissa:
Makes total sense.  Thank you for the explanation @PacoDw

themantissa (approved):

PR #66, opened 2019-11-11
Clarify Azure Option in Doc

themantissa:
-Clarify provider_disk_type_name is azure only and what it accepts.  Clarify disk_size_db is for GCP/Azure.


themantissa:
Had an internal request from customer to better clarify these options in doc.

PacoDw (approved):
Very nice! 👍

PR #77, opened 2019-11-25
Added default Disk Size when it doesn't set up on cluster resource

### LINKED ISSUE #72: Creating M2 cluster without specifying disk_size_gb results in 400 Bad Request
Excerpt from my main.tf file:

``` 
resource "mongodbatlas_cluster" "cluster-test" {
  project_id   = mongodbatlas_project.testing.id
  name         = "testing-cluster"

  //Provider Settings "block"
  provider_name               = "TENANT"
  backing_provider_name       = "AWS"
  provider_instance_size_name = "M2"
  provider_region_name        = "eu-west-1"
}
```

According to the documentation, disk_size_gb is optional and should default to the minimum (2GB for an M2) but I get an error: `400 (request "Bad Request") The cluster's disk size of 0.0 GB is invalid"`.

If I specify `disk_size_gb = 2` it works as expected.

PacoDw:
Added:

- Now you don't need to specify the `disk_size_gb` and `provider_disk_iops` attrs. The last one will set it up by the MongoDB server automatically.
- Test cases improved making a test for each provider with its tier
- The `auto_scaling_disk_gb_enabled` attribute will default to false when the provider name sets to TENANT.

closes #72

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -52,7 +52,6 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"auto_scaling_disk_gb_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
 			},
 			"backup_enabled": {
 				Type:     schema.TypeBool,
@@ -296,6 +295,16 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 	projectID := d.Get("project_id").(string)
 	providerName := d.Get("provider_name").(string)
 
+	autoScaling := matlas.AutoScaling{
+		DiskGBEnabled: pointy.Bool(true),
+	}
+
+	if diskGBEnabled, ok := d.GetOkExists("auto_scaling_disk_gb_enabled"); ok {
+		autoScaling = matlas.AutoScaling{
+			DiskGBEnabled: pointy.Bool(diskGBEnabled.(bool)),
+		}
+	}
+
 	//validate cluster_type conditional
 	if _, ok := d.GetOk("replication_specs"); ok {
 		if _, ok1 := d.GetOk("cluster_type"); !ok1 {
@@ -331,6 +340,15 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 		}
 	}
 
+	if providerName == "TENANT" {
+		if diskGBEnabled := d.Get("auto_scaling_disk_gb_enabled"); diskGBEnabled.(bool) {
+			return fmt.Errorf("`auto_scaling_disk_gb_enabled` cannot be true when provider name is TENANT")
+		}
+		autoScaling = matlas.AutoScaling{
+			DiskGBEnabled: pointy.Bool(false),
+		}
+	}
+
 	// We need to validate the oplog_size_mb attr of the advanced configuration option to show the error
 	// before that the cluster is created
 	if oplogSizeMB, ok := d.GetOk("advanced_configuration.oplog_size_mb"); ok {
@@ -347,13 +365,8 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 	providerSettings := expandProviderSetting(d)
 
 	replicationSpecs, err := expandReplicationSpecs(d)
-
 	if err != nil {
 		return fmt.Errorf(errorCreate, err)
-	}
-
-	autoScaling := matlas.AutoScaling{
-		DiskGBEnabled: pointy.Bool(d.Get("auto_scaling_disk_gb_enabled").(bool)),
 	}
 
 	// set the default disk size gb for the provider and tier
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -743,10 +743,13 @@ func expandProviderSetting(d *schema.ResourceData) matlas.ProviderSettings {
 	providerSettings := matlas.ProviderSettings{}
 
 	if d.Get("provider_name") == "AWS" {
-		// Asks if the Provider Disk IOS was set if it didn't the server will set it up automatically
+
+		// Check if the Provider Disk IOS sets in the Terraform configuration.
+		// If it didn't, the MongoDB Atlas server would set it to the default for the amount of storage.
 		if v, ok := d.GetOk("provider_disk_iops"); ok {
 			providerSettings.DiskIOPS = pointy.Int64(cast.ToInt64(v))
 		}
+
 		providerSettings.EncryptEBSVolume = pointy.Bool(true)
 		if encryptEBSVolume, ok := d.GetOkExists("provider_encrypt_ebs_volume"); ok {
 			providerSettings.EncryptEBSVolume = pointy.Bool(cast.ToBool(encryptEBSVolume))
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -369,24 +369,20 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf(errorCreate, err)
 	}
 
-	// set the default disk size gb for the provider and tier
-	diskSizeGB := matlas.DefaultDiskSizeGB[providerName][d.Get("provider_instance_size_name").(string)]
-	// But if it's specify the disk size will set up
-	if v, ok := d.GetOk("disk_size_gb"); ok {
-		diskSizeGB = v.(float64)
-	}
-
 	clusterRequest := &matlas.Cluster{
 		Name:                     d.Get("name").(string),
 		EncryptionAtRestProvider: d.Get("encryption_at_rest_provider").(string),
 		ClusterType:              cast.ToString(d.Get("cluster_type")),
 		BackupEnabled:            pointy.Bool(d.Get("backup_enabled").(bool)),
-		DiskSizeGB:               &diskSizeGB,
 		ProviderBackupEnabled:    pointy.Bool(d.Get("provider_backup_enabled").(bool)),
 		AutoScaling:              autoScaling,
 		BiConnector:              biConnector,
 		ProviderSettings:         &providerSettings,
 		ReplicationSpecs:         replicationSpecs,
+	}
+
+	if v, ok := d.GetOk("disk_size_gb"); ok {
+		clusterRequest.DiskSizeGB = pointy.Float64(v.(float64))
 	}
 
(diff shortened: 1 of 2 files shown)


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -346,13 +365,15 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
...
-
 	if err != nil {
 		return fmt.Errorf(errorCreate, err)
 	}
 
-	autoScaling := matlas.AutoScaling{
-		DiskGBEnabled: pointy.Bool(d.Get("auto_scaling_disk_gb_enabled").(bool)),
+	// set the default disk size gb for the provider and tier
+	diskSizeGB := matlas.DefaultDiskSizeGB[providerName][d.Get("provider_instance_size_name").(string)]
+	// But if it's specify the disk size will set up
```
themantissa: Just be aware that if someone does specify it and it's lower than the default Atlas automatically sets it to the default.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -719,7 +740,10 @@ func expandProviderSetting(d *schema.ResourceData) matlas.ProviderSettings {
 	providerSettings := matlas.ProviderSettings{}
 
 	if d.Get("provider_name") == "AWS" {
-		providerSettings.DiskIOPS = pointy.Int64(cast.ToInt64(d.Get("provider_disk_iops")))
+		// Asks if the Provider Disk IOS was set if it didn't the server will set it up automatically
```
themantissa: Just checking but you are noting in the comment here that if it's not included in the terraform config Atlas just sets it to the default for the amount of storage, correct?

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -12,51 +12,226 @@ import (
...
+	"github.com/spf13/cast"
 )
 
 func TestAccResourceMongoDBAtlasCluster_basicAWS(t *testing.T) {
 	var cluster matlas.Cluster
 
 	resourceName := "mongodbatlas_cluster.test"
 	projectID := os.Getenv("MONGODB_ATLAS_PROJECT_ID")
-	name := fmt.Sprintf("test-acc-%s", acctest.RandString(10))
+	tiers := []string{"M10", "M20", "M30", "M40"}
```
themantissa: For cost it might make sense to just do M10 and M30, those are the two levels where we have different capabilities.

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -12,51 +12,226 @@ import (
...
+
+	resourceName := "mongodbatlas_cluster.test"
+	projectID := os.Getenv("MONGODB_ATLAS_PROJECT_ID")
+	var region string
+
+	for tier, size := range matlas.DefaultDiskSizeGB["GCP"] {
+		name := fmt.Sprintf("testAcc-%s-%s-%ggb-%s", "GCP", tier, size, acctest.RandString(1))
+
+		switch tier {
+		case "M200":
```
themantissa, line 124: These are really expensive.  Can we test lower?

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -12,51 +12,226 @@ import (
...
+				},
+			})
+		})
+	}
+}
+
+func TestAccResourceMongoDBAtlasCluster_tenant(t *testing.T) {
+	var cluster matlas.Cluster
+
+	resourceName := "mongodbatlas_cluster.tenant"
```
themantissa: this might be confusing but while we use the word Tenant in the API this is what we refer to as our "shared" tier.  Might make more sense to use shared instead of tenant for readability.

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -471,24 +521,23 @@ func testAccMongoDBAtlasClusterConfigAzure(projectID, name, backupEnabled string
 			
 			replication_factor           = 3
 			backup_enabled               = %s
-			auto_scaling_disk_gb_enabled = true
```
themantissa, line 474: just to note the default is true.

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -498,10 +547,44 @@ func testAccMongoDBAtlasClusterConfigGCP(projectID, name, backupEnabled string)
...
-			provider_instance_size_name = "M30"
-			provider_region_name        = "US_EAST_4"
+			provider_instance_size_name = "%s"
+			provider_region_name        = "%s"
 		}
-	`, projectID, name, backupEnabled)
+	`, projectID, name, backupEnabled, tier, region)
+}
+
+func testAccMongoDBAtlasClusterConfigTenant(projectID, name, autoScaling, size, tier, region string) string {
```
themantissa: Shared again makes more sense for a name.

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/clusters.go
```diff
@@ -114,6 +114,61 @@ type clustersResponse struct {
 	TotalCount int       `json:"totalCount,omitempty"`
 }
 
+// DefaultDiskSizeGB represents the Tier and the default disk size for each one
+// it can be use like: DefaultDiskSizeGB["AWS"]["M10"]
+var DefaultDiskSizeGB map[string]map[string]float64 = map[string]map[string]float64{
```
themantissa: This will be fun to keep updated!! :)

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/maintenance.go
```diff
@@ -0,0 +1,136 @@
+package mongodbatlas
```
themantissa: Skipping reviewing maintenance related as that's in a separate PR.  Please ensure each PR just contains changes related to task at hand please.

themantissa (commented):
A few comments/questions/suggestions but looking good.

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: yes, it's correct

mongodbatlas/resource_mongodbatlas_cluster_test.go
PacoDw: okay, thank you I will change the test case :)

mongodbatlas/resource_mongodbatlas_cluster_test.go
PacoDw: It tests the lower tiers, but this switch statement only waits for each case and then sets the specific region for each one, I can remove them if you want

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/maintenance.go
PacoDw: sorry I will remove it

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: Great, perhaps clarify in the comment for others?

mongodbatlas/resource_mongodbatlas_cluster_test.go
themantissa: Sounds good!

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/maintenance.go
themantissa: Thank you!

mongodbatlas/resource_mongodbatlas_cluster_test.go
themantissa: Did you see this @PacoDw ? Any objections to calling it Shared?

mongodbatlas/resource_mongodbatlas_cluster_test.go
themantissa: So it's more just a random - if we test w/ this level we try a different region?  I'm just not sure why M200/M300 specifically since both are available in US_EAST_4?  I'm guessing I'm missing something?

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: Yes, it's a great idea, thanks.

mongodbatlas/resource_mongodbatlas_cluster_test.go
PacoDw: No, everything is ok for me 😅. Sorry, I was flying yesterday

mongodbatlas/resource_mongodbatlas_cluster_test.go
PacoDw: just to make a test with a different region

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/maintenance.go
PacoDw: I can't remove because is the vendor files that are dependencies

themantissa: Ah, that's fine.  At least we know it's not been reviewed yet.

mongodbatlas/resource_mongodbatlas_cluster_test.go
themantissa: Makes sense - we can use these as long as the test immediately cleans them up, if not probably best to top at at a M30 or M40.

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/maintenance.go
PacoDw: Yes, for this reason, exists the vendor files, sorry for not explain you this before

themantissa:
@PacoDw  - I'm unclear why another PR was made after my comments as no commits in response to them seem to have been made?

themantissa (approved):
Saw the commits I missed, looks like all was addressed - LGTM.

vendor/github.com/mongodb/go-client-mongodb-atlas/mongodbatlas/clusters.go
PacoDw: I will remove it due that it isn't already necessary in the Terraform part, I tested again without this change verifying if the MongoDB server set up correctly the  Disk Size by default, and everything is ok.

marinsalinas (changes requested):
Just please fix the conflicts :) - LGTM!

marinsalinas (approved):
LGTM!

PR #79, opened 2019-11-26
Feat: Team Resource and Data Source

marinsalinas:
Added:

- `mongodbatlas_team` resource.
- `mongodbatlas_team` resource acceptance testing.
- `mongodbatlas_team` resource website documentation.
- `mongodbatlas_team` data source.
- `mongodbatlas_team` data source acceptance testing.
- `mongodbatlas_team` data source website documentation.

## Example Configuration:
```hcl
resource "mongodbatalas_teams" "test" {
  org_id     = "<ORGANIZATION-ID>"
  name       = "myNewTeam"
  usernames  = ["user1", "user2", "user3"]
}

data "mongodbatlas_teams" "test" {
  org_id     = mongodbatlas_teams.test.org_id
  team_id    = mongodbatlas_teams.test.team_id
}

```


go.mod
```diff
@@ -6,7 +6,9 @@ require (
 	github.com/Sectorbob/mlab-ns2 v0.0.0-20171030222938-d3aa0c295a8a
 	github.com/hashicorp/hcl v0.0.0-20180404174102-ef8a98b0bbce // indirect
 	github.com/hashicorp/terraform v0.12.1
-	github.com/mongodb/go-client-mongodb-atlas v0.1.1
+	github.com/mongodb/go-client-mongodb-atlas v0.1.2
 	github.com/mwielbut/pointy v1.1.0
 	github.com/spf13/cast v1.3.0
 )
+
+replace github.com/mongodb/go-client-mongodb-atlas v0.1.2 => ../go-client-mongodb-atlas
```
PacoDw: could you remove the replace?

marinsalinas: yes, I will remove until we run a new release for the client.

mongodbatlas/provider.go
```diff
@@ -44,6 +44,7 @@ func Provider() terraform.ResourceProvider {
 			"mongodbatlas_network_peerings":                     dataSourceMongoDBAtlasNetworkPeerings(),
 			"mongodbatlas_cloud_provider_snapshot_restore_job":  dataSourceMongoDBAtlasCloudProviderSnapshotRestoreJob(),
 			"mongodbatlas_cloud_provider_snapshot_restore_jobs": dataSourceMongoDBAtlasCloudProviderSnapshotRestoreJobs(),
+			"mongodbatlas_team":                                 dataSourceMongoDBAtlasTeam(),
```
themantissa, line 55: Add s to team?

mongodbatlas/resource_mongodbatlas_team.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if err := d.Set("usernames", usernames); err != nil {
+		return fmt.Errorf("error setting `usernames` for team (%s): %s", d.Id(), err)
+	}
+
+	return nil
+}
+
+func resourceMongoDBAtlasTeamUpdate(d *schema.ResourceData, meta interface{}) error {
```
themantissa, line 117: Looks like this is just update for the team name?  https://docs.atlas.mongodb.com/reference/api/teams-rename-one/. Perhaps makes sense to clarify that in the name?  There will need to be an update for roles, users, etc too.

mongodbatlas/resource_mongodbatlas_team.go
```diff
@@ -0,0 +1,190 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"usernames": {
+				Type:     schema.TypeList,
+				Optional: true,
```
themantissa: These needs to be existing Atlas users (cannot be pending either).  May want to try adding a username that doesn't exist - I'll make a note about this in the docs below.

website/docs/d/team.html.markdown
```diff
@@ -0,0 +1,64 @@
+---
+layout: "mongodbatlas"
+page_title: "MongoDB Atlas: team"
+sidebar_current: "docs-mongodbatlas-datasource-team"
+description: |-
+    Describes a Team.
+---
+
+# mongodbatlas_team
```
themantissa: mongodbatalas_teams as suggested before

website/docs/d/team.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+`mongodbatlas_team` describes a Team. The resource requires your Organization ID and Team ID.
```
themantissa: slight edit:
describes a Team. The resource requires a MongoDB Atlas Organization Id and a Team Id.

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+    name      = "myNewTeam"
+    usernames = ["user1", "user2", "user3"]
+}
+```
+
+## Argument Reference
+
+* `org_id` - (Required) The unique identifier for the organization you want to associate the team with.
+* `name` - (Required) 	The name of the team you want to create.
+* `usernames` - (Optional) Usernames to add to the new team.
```
themantissa: Add: 
You can only add Atlas users who are part of the organization.  Users who have not accepted an invitation to join the organization cannot be added as team members.   There is a maximum of 250 Atlas users per team.

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+page_title: "MongoDB Atlas: team"
+sidebar_current: "docs-mongodbatlas-resource-team"
+description: |-
+    Provides a Team resource.
+---
+
+# mongodbatlas_team
+
+`mongodbatlas_team` provides a Team resource. The resource lets you create, edit and delete teams. The resource requires your Organization ID.
+
```
themantissa, line 14: Add a note: MongoDB Atlas Team limits: max 250 teams in an organization and max 100 teams per project.

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+layout: "mongodbatlas"
+page_title: "MongoDB Atlas: team"
+sidebar_current: "docs-mongodbatlas-resource-team"
+description: |-
+    Provides a Team resource.
+---
+
+# mongodbatlas_team
+
+`mongodbatlas_team` provides a Team resource. The resource lets you create, edit and delete teams. The resource requires your Organization ID.
```
themantissa: There is a missing bit of functionality.   Once a team is created it can be assigned to project and giving a project role, e.g. I can assign a team, TEAM1 to project PRO1 with the role Project Owner.  That way all the team members in the role get that access to the project, make sense?

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+## Example Usage
+
+```hcl
+resource "mongodbatlas_team" "test" {
+    org_id    = "<Your Organization ID>"
+    name      = "myNewTeam"
+    usernames = ["user1", "user2", "user3"]
+}
+```
+
```
themantissa, line 28: We'll then likely need a resource to assign a team to a project with a specific role(s)

mongodbatlas/data_source_mongodbatlas_team.go
```diff
@@ -0,0 +1,154 @@
...
+
+						"last_name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"username": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"roles": {
```
themantissa: The roles should be at the actual team level I believe, not users.  https://docs.atlas.mongodb.com/reference/api/teams-update-roles/

themantissa (commented):
If I'm reading this correctly I think there may be some missing functionality or maybe a misunderstanding of what teams does.   Basically a team is a list of existing Atlas users in an organization. They get grouped into a team by username.  Then the team can be added to a Project with a role or role(s) and now the users in the team get that role on the Project.    Hope this helps!!

themantissa (changes requested):
Should have set the PR to request changes, see comments.

mongodbatlas/provider.go
PacoDw: Thanks, @themantissa I will attend all the comments 👍

mongodbatlas/resource_mongodbatlas_team.go
PacoDw: Also, this attribute must be required, I will change it

website/docs/r/team.html.markdown
PacoDw: I was thinking of creating another attribute like optional to assign the roles for the team. I mean, first, create the team, assign the users, later if exists roles for the team attach them to it (using the update API entry point to do that), it would be more comfortable than has another resource. However, we have a bit problem that doesn't exist a GET method or the way to retrieve the team roles (when you create or get a team by id, the roles don't get, I don't know why if the roles is part of the team), just updating it.

mongodbatlas/data_source_mongodbatlas_team.go
PacoDw: This roles belong to each user, but If you want a team roles as I commented above there is not a way to retrieve the team roles.

website/docs/r/team.html.markdown
PacoDw: Thank you 👍

mongodbatlas/data_source_mongodbatlas_team.go
themantissa: Ah, I follow - this is for the result set of https://docs.atlas.mongodb.com/reference/api/teams-get-all-users/.  And you are right, there doesn't seem to be a method to read team roles.   I'll look into that.

website/docs/r/team.html.markdown
themantissa: Ah, okay I think it will help to explain Teams more.   I'll try here and I think this will show us what is (potentially) missing:
A Team is created at the organization level.  Users are added to the team at the organization level.
But then a Team is given access to a Project (so this is at the project level) and then the Team is given roles for that project.   

So the issue seems that we do not have an API method documented for adding a team to a project or for reading what what roles are assigned (but we can add roles to a team via PATCH), does that sound right?  If so I'll follow up internally to figure this out.

website/docs/r/team.html.markdown
themantissa: This sentence did not get added, will clarify:  In the usernames, after "Usernames to add to the new team." Please add the following:   
You can only add Atlas users who are part of the organization. Users who have not accepted an invitation to join the organization cannot be added as team members. There is a maximum of 250 Atlas users per team.

themantissa (changes requested):
One missed task and some questions to you Paco to clarify what we need to solve this.   Thanks!

website/docs/r/team.html.markdown
themantissa: I still think the core is missing here is that a team can be created before assigning to a project.  With this resource as is it's not clear how to create a team and assign it to more than one project w/ the provider.  For example if I have a team called Support and I create it for my org I might want to add it to 3 different projects, I don't see how that would happen here?  Would you submit the same org and each project in 3 different resource blocks?  If so then the users in each makes even less sense.  Basically it's 1 team to, possibly, 1 or more projects.  The team, however, would have same users - the roles could be different for each project. Happy to setup time to talk about this if we need to @marinsalinas  and @PacoDw      

And just to cover our bases of what we covered in slack quite a few of the team/project related resources are in projects within our documentation https://docs.atlas.mongodb.com/reference/api/project-add-team/  and https://docs.atlas.mongodb.com/reference/api/project-get-teams/

themantissa (changes requested):
Still don't think it's quite clear on how teams work.  Perhaps we should talk about?   Will message the team tomorrow in a chat.

website/docs/r/team.html.markdown
PacoDw: Make sense to me @themantissa

website/docs/r/team.html.markdown
```diff
@@ -45,10 +41,10 @@ In addition to all arguments above, the following attributes are exported:
 
 ## Import
 
-Clusters can be imported using the organization ID and team id, in the format `ORGID-TEAMID-PROJECTID`, e.g.
+Clusters can be imported using the organization ID and team id, in the format `ORGID-TEAMID`, e.g.
```
themantissa: One minor change! 
Teams can be imported using the organization ID and team id, in the format `ORGID-TEAMID`, e.g.

themantissa (changes requested):
One tiny change, otherwise LGTM!

themantissa (approved):
LGTM - thanks!!

PR #85, opened 2019-12-09
fix: validate if mongo_db_major_version is set in cluster resource

### LINKED ISSUE #81: Cluster fails to build on 0.3.1 when mongo_db_major_version is not specified
Hey folks,

Using version `0.3.1`, if you omit `mongo_db_major_version` on the `mongodbatlas_cluster` resource then you'll hit an API error as the provider is sending along the following:

```
2019-11-27T22:06:32.115-0500 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.3.1_x4:  "mongoDBMajorVersion": "0.0",
```

This doesn't seem to be an issue with `0.3.0`, so is likely something introduced in `0.3.1`.

The docs for the [**mongodbatlas_cluster** resource](https://www.terraform.io/docs/providers/mongodbatlas/r/cluster.html) indicate that this argument is optional, so this does appear to be a bug.

Thanks for having a look!

--Matt

marinsalinas:
closes #81

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -358,7 +358,6 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 	clusterRequest := &matlas.Cluster{
 		Name:                     d.Get("name").(string),
 		EncryptionAtRestProvider: d.Get("encryption_at_rest_provider").(string),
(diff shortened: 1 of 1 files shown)


PacoDw (approved):
LGTM! 👍

themantissa (approved):
LGTM

mattt416:
Tested and works, tyvm @marinsalinas !

PR #91, opened 2019-12-13
Update project.html.markdown

themantissa:
Update to Project Resource to warn/clarify behavior regarding name changes of a Project in Terraform due to difference between API and UI


PacoDw (approved):
LGTM! 👍

PR #92, opened 2019-12-16
Network Peering RS: remove provider_name=AWS as default, use Required=true instead i…

### LINKED ISSUE #86: Cannot update GCP network peer
**What I did**: Import a mongodbatlas_network_peering resource and run terraform apply to apply the `route_table_cidr_block` change it thinks it needs to make (asserting the same value)

**What I expected**: The change to be successful

**What happened**: 
```Error: error updating MongoDB Network Peering Connection (our-group-id): PATCH https://cloud.mongodb.com/api/atlas/v1.0/groups/our-group-id/peers/our-peer-id: 404 (request "Not Found") No peer with ID our-peer-id exists in project our-group-id```

Looking at the MongoDB API and the code, it looks like the default provider of AWS is being applied and thus the peering connection isn't found since we are using GCP.

Plugin version: 0.3.1
Terraform version: v0.12.17

### PR DESCRIPTION
**Pull request by marinsalinas**
Closes #86

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_network_peering.go
+++ b/mongodbatlas/resource_mongodbatlas_network_peering.go
@@ -390,7 +390,7 @@ func resourceMongoDBAtlasNetworkPeeringUpdate(d *schema.ResourceData, meta inter
 			peer.AzureDirectoryID = d.Get("azure_directory_id").(string)
 		}
 		if d.HasChange("azure_subscription_id") {
-			peer.AzureSubscriptionId = d.Get("azure_subscription_id").(string)
+			peer.AzureSubscriptionID = d.Get("azure_subscription_id").(string)
 		}
 		if d.HasChange("resource_group_name") {
 			peer.ResourceGroupName = d.Get("resource_group_name").(string)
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: mongodbatlas/resource_mongodbatlas_network_peering.go [hunk 1]
```diff
@@ -61,9 +61,8 @@ func resourceMongoDBAtlasNetworkPeering() *schema.Resource {
 
 			"provider_name": {
 				Type:         schema.TypeString,
-				Optional:     true,
+				Required:     true,
```
Uvedale, line 64: I don't think this is being used when sending the update request. Line 426 doesn't include provider_name

themantissa: It should be within peer, also see line 389, is this correct @marinsalinas ?

themantissa (commented):
Left a comment in response to reporter, once we resolve this we should be good to approve.

mongodbatlas/resource_mongodbatlas_network_peering.go
Uvedale: If I'm reading line 389 correctly, providerName will only be included if it changed but it should always be included.

marinsalinas: > I don't think this is being used when sending the update request. Line 426 doesn't include provider_name

That's right, let me add it to the request.

marinsalinas: I've added it on the last commit

themantissa (approved):
LGTM

themantissa:
Okay to go ahead and merge I think @marinsalinas ?

Issue #68, opened 2019-11-12
Whitelisted Project IPs when manually deleted causes failure at next plan/apply


richardj-bsquare:
**Version affected**
terraform-provider-mongodbatlas 0.3.1

**How to reproduce**
Use mongodbatlas_project_ip_whitelist to whitelist a list of IPs
Terraform plan & apply
Delete *all* of the IP whitelisted addresses.
Run terraform plan & apply

**Expected Behaviour**
The whitelisted IP is recreated.

**Actual behaviour**
Upon re-running the terraform plan this is reported (example):

`Error: error getting project IP whitelist information: GET https://cloud.mongodb.com/api/atlas/v1.0/groups/5dc3051579358e9725c677e8/whitelist/139.14.250.168%2F29: 404 (request "Not Found") IP Address 139.14.250.168/29 not on Atlas whitelist for group 5dc3021579358e9745c677e8.`

**Workaround**
The above error will be reported individually for each IP that is missing, but only one error per plan until each is manually re-added (which is quite tedious, to say the least, especially if some are dynamically generated).

**Notes**
The resource also does not notice changes in the comment field and therefore does not modify them to match the terraform.

PacoDw:
Hi @richardj-bsquare! 
Thank you so much for your review, on your steps to reproduce how do you remove all the whitelist? Due you can't remove them and run the plan or apply because Terraform whitelist is a required field. On another hand, I made a test removing them, always leaving one or updating them, and I didn't have errors.

If you have another comment or concern plz just let me know.

richardj-bsquare:
Hi,

Sorry for the confusion. I removed them **all manually** from the Mongo DB Atlas whitelist via the Atlas web UI.

As indicated in this scenario, the provider should detect entries are missing and re-add them, this prevents inadvertent deletion or other manual changes from being missed, and would be consistent behaviour with other providers (i.e. if you delete a subnet in AWS via the console, and its in your terraform configuration it gets re-created by the plan and apply cycle).

themantissa:
[...]

Issue #67, opened 2019-11-12
Support for AWS security groups in mongodbatlas_project_ip_whitelist


sjauld:
The Atlas API supports adding security groups to a Project IP Whitelist, by specifying the security group ID in the `cidr_block` field.

However since we validate that `cidr_block` is a CIDR block, we can't input security groups here.

I'll throw together a quick PR.

sjauld:
Actually, on further inspection, the Atlas API *returns* security group IDs in the cidr_block field, but doesn't *accept* them when you post it. I'll raise a ticket with them.

sjauld:
No ticketing system exists for the API, but feel free to +1 this tweet if you want to support my request 👍  https://twitter.com/sjogrady/status/1194116135551029251

themantissa:
Security group support was actually very recently added to the MongoDB Atlas IP Whitelist resource.  We have it on our roadmap to add support for it via the Provider as soon as we can schedule engineering resources to tackle it.   

For future reference,  per the README, feature requests can be made via https://feedback.mongodb.com for both MongoDB Atlas (and its API) as well as the Terraform provider.  

I'll leave this open until we add support and we'll post here when it's added.

PR #94, opened 2019-12-17
Whitelist

### LINKED ISSUE #51: Modifying project ip whitelist destroy and re-create all resources
Hi,

This seems to be a bug into the `mongodbatlas_project_ip_whitelist` resource.
Exemple :
```
resource "mongodbatlas_project_ip_whitelist" "whitelist" {
  project_id = data.mongodbatlas_project.project.id
  whitelist {
    cidr_block = "10.10.0.0/24"
    comment    = "cidr 1"
  }
}
```
Add a cidr :
```
resource "mongodbatlas_project_ip_whitelist" "whitelist" {
  project_id = data.mongodbatlas_project.project.id
  whitelist {
    cidr_block = "10.10.0.0/24"
    comment    = "cird 1"
  }
  whitelist {
    cidr_block = "10.10.1.0/24"
    comment    = "cidr 2"
  }
}
```
Plan :
```
  # module.atlas.mongodbatlas_project_ip_whitelist.whitelist must be replaced
-/+ resource "mongodbatlas_project_ip_whitelist" "whitelist" {
      ~ id         = "cHJvamVjdF9pZA==:NWQ4OWM4MzJmMmEzMGI4ZTc4Mzc5NDEx-ZW50cmllcw==:MTAuMTAwLjk4LjAvMjM=" -> (known after apply)
        project_id = "xxxxxxxxxxxxxxxxxxxxxxxxxx"

      + whitelist { # forces replacement
          + cidr_block = "10.10.0.0/24"
          + comment    = "cidr 1"
          + ip_address = (known after apply)
        }
      - whitelist { # forces replacement
          - cidr_block = "10.10.0.0/24" -> null
          - comment    = "cidr 1" -> null
        }
      + whitelist { # forces replacement
          + cidr_block = "10.10.1.0/24"
          + comment    = "cidr 2"
          + ip_address = (known after apply)
        }
    }
```
It happens if i add or delete or modify a cidr_block.  
Thanks

PacoDw:
The Project IP Whitelist resource changed to be used as an independent resource block for each cidr block or ip address instead of the handle them in the same resource.

#### Using Cidr Block
```hcl
resource "mongodbatlas_project_ip_whitelist" "test" {
  project_id = "<PROJECT-ID>"
  cidr_block = "1.2.3.4/32"
  comment    = "cidr block for tf acc testing"
}
```

#### Using IP Address
```hcl
resource "mongodbatlas_project_ip_whitelist" "test" {
  project_id = "<PROJECT-ID>"
  ip_address = "2.3.4.5"
  comment    = "ip address for tf acc testing"
}
```

#### Using AWS Security Group
```hcl
resource "mongodbatlas_network_container" "test" {
  project_id       = "<PROJECT-ID>"
  atlas_cidr_block = "192.168.208.0/21"
  provider_name    = "AWS"
  region_name      = "US_EAST_1"
}

resource "mongodbatlas_network_peering" "test" {
  project_id             = "<PROJECT-ID>"
  container_id           = mongodbatlas_network_container.test.container_id
  accepter_region_name   = "us-east-1"
  provider_name          = "AWS"
  route_table_cidr_block = "172.31.0.0/16"
  vpc_id                 = "vpc-0d93d6f69f1578bd8"
  aws_account_id         = "232589400519"
}

resource "mongodbatlas_project_ip_whitelist" "test" {
  project_id         = "<PROJECT-ID>"
  aws_security_group = "sg-0026348ec11780bd1"
  comment            = "TestAcc for awsSecurityGroup"

  depends_on = ["mongodbatlas_network_peering.test"]
}
```

closes #51 #67 #68


mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
```diff
@@ -25,226 +28,209 @@ func resourceMongoDBAtlasProjectIPWhitelist() *schema.Resource {
...
 			},
+			"ip_address": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"aws_security_group", "cidr_block"},
+				ValidateFunc:  validation.SingleIP(),
+			},
+			"aws_security_group": {
```
themantissa, line 74: Should probably call out in a comment here in the code and in the doc that one has to already have VPC peering configured to be able to use this.

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
```diff
@@ -25,226 +28,209 @@ func resourceMongoDBAtlasProjectIPWhitelist() *schema.Resource {
...
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"ip_address", "cidr_block"},
+			},
+			"comment": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
```
themantissa, line 85: Updating a comment shouldn't cause the entry to be recreated - per the doc "For each whitelist entry specified in the request body whose awsSecurityGroup, ipAddress, or cidrBlock value match an existing entry, Atlas only updates the specified matching entries. Atlas will not set the whitelist to only contain the entries specified in the request."  So a comment can be updated but as of now we have no way to update the IP/CIDR/ASG since we don't have a patch/put and id value.  The UI allows a user to update a comment without removing the IP/CIDR/ASG so if we can't do this in a terraform way without creating/deleting perhaps we should just now allow changing the comment after creation?

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
```diff
@@ -25,226 +28,209 @@ func resourceMongoDBAtlasProjectIPWhitelist() *schema.Resource {
...
-}
-
 func resourceMongoDBAtlasProjectIPWhitelistCreate(d *schema.ResourceData, meta interface{}) error {
-	//Get the client connection.
 	conn := meta.(*matlas.Client)
-
 	projectID := d.Get("project_id").(string)
+	cirdBlock := d.Get("cidr_block").(string)
+	ipAddress := d.Get("ip_address").(string)
+	awsSGroup := d.Get("aws_security_group").(string)
```
themantissa: I'd do awsSecurityGroup - awsSGroup is hard to read and out of flow for the others where you removed the underscore and camel cased.

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
```diff
@@ -25,226 +28,209 @@ func resourceMongoDBAtlasProjectIPWhitelist() *schema.Resource {
...
 
-	req := expandProjectIPWhitelist(d)
-	_, _, err := conn.ProjectIPWhitelist.Create(context.Background(), projectID, req)
-	if err != nil {
-		return fmt.Errorf("error creating project IP whitelist: %s", err)
+	if cirdBlock == "" && ipAddress == "" && awsSGroup == "" {
+		return errors.New("cidr_block, ip_address or aws_security_group needs to be setted")
```
themantissa: Grammar error: 
return errors.New("cidr_block, ip_address or aws_security_group needs to be **set**")
or perhaps more clearly
return errors.New("cidr_block, ip_address or aws_security_group needs to **contain a value**")

website/docs/r/project_ip_whitelist.html.markdown
```diff
@@ -18,34 +18,58 @@ When you remove an entry from the whitelist, existing connections from the remov
```
themantissa, line 18: On line 11 we'll need to add in AWS Security Groups, i.e.:
mongodbatlas_project_ip_whitelist provides an IP Whitelist entry resource. The whitelist grants access from IPs, CIDRs or AWS Security Groups (if VPC Peering is enabled) to clusters within the Project.
@themantissa

website/docs/r/project_ip_whitelist.html.markdown
```diff
@@ -18,34 +18,58 @@ When you remove an entry from the whitelist, existing connections from the remov
 
 ## Example Usage
 
+### Using Cidr Block
```
themantissa: CIDR should in text be capitalized.

website/docs/r/project_ip_whitelist.html.markdown
```diff
@@ -18,34 +18,58 @@ When you remove an entry from the whitelist, existing connections from the remov
...
+### Using IP Address
+```hcl
+resource "mongodbatlas_project_ip_whitelist" "test" {
+  project_id = "<PROJECT-ID>"
+  ip_address = "2.3.4.5"
+  comment    = "ip address for tf acc testing"
+}
+```
+
+### Using AWS Security Group
```
themantissa: Slight change:
Using an AWS Security Group

website/docs/r/project_ip_whitelist.html.markdown
```diff
@@ -18,34 +18,58 @@ When you remove an entry from the whitelist, existing connections from the remov
...
+```hcl
+resource "mongodbatlas_network_container" "test" {
```
themantissa, line 41: Add comment above this 
// In order to use AWS Security Group(s) VPC Peering must be enabled.

website/docs/r/project_ip_whitelist.html.markdown
```diff
@@ -18,34 +18,58 @@ When you remove an entry from the whitelist, existing connections from the remov
...
 ```
 
 ## Argument Reference
 
 * `project_id` - (Required) The ID of the project in which to add the whitelist entry.
-* `cidr_block` - (Optional) The whitelist entry in Classless Inter-Domain Routing (CIDR) notation. Mutually exclusive with `ip_address`.
-* `ip_address` - (Optional) The whitelisted IP address. Mutually exclusive with `cidr_block`.
+* `aws_security_group` - (Optional) ID of the whitelisted AWS security group. Mutually exclusive with `cidr_block` and `ip_address`.
+* `cidr_block` - (Optional) Whitelist entry in Classless Inter-Domain Routing (CIDR) notation. Mutually exclusive with `aws_security_group` and `ip_address`.
+* `ip_address` - (Optional) Whitelisted IP address. Mutually exclusive with `aws_security_group` and `cidr_block`.
```
themantissa, line 74: One of these is required for it to work so perhaps a comment above the three that one of the following must be included?

themantissa (changes requested):
A few suggestions for improvements and a concern around comments.  Thank you!!

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
PacoDw: okay thanks

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
PacoDw: I tried to achieve that, but like we can create multiple whitelist entries at the same time, the way to do that doesn't allow us to update the comments. Let me explain how to works when you want to create multiple whitelist entries. Imagine you want to create ten whitelist entries, so you apply in Terraform to sent the ten requests (note that the MongoDB server doesn't support multiple requests at the same time). The MongoDB server will return multiple errors (internal error 500 code) when this happens; the resource will send the requests again until the server accepts them. So this turns complicated because if we want to update the comment, we need to handle it like this, but extra more complication due to we have an updating function, and it needs to handle the state or the other attributes, it becomes unstable. For this reason I decided to remove.

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
PacoDw: Thanks let me change that

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
PacoDw: Thank you so much 👍

website/docs/r/project_ip_whitelist.html.markdown
PacoDw: Thanks 👍

themantissa (approved):
LGTM!  Hopefully will be a better solution till we have a change in the API!

PR #96, opened 2019-12-18
update containers and ip whitelist doc

themantissa:
Fix formatting issue in container doc and added important information on connections for ipwhitelist doc.


PacoDw (approved):
LGTM! Thank you so much :)

PR #98, opened 2019-12-23
Update database_user.html.markdown

themantissa:
Cleared up exampled and clarified that password is required to create the user but optional after creation.


PacoDw (approved):
Thank you so much! 👍

PR #100, opened 2020-01-06
Added labels attr for Database User resource

PacoDw:
- Added `labels` attribute to allow create multiple labels for each database user resource
- Added Test case using dynamic labels attribute
- Updated the documentation and vendor files

### Example Usage 
```hcl
resource "mongodbatlas_database_user" "test" {
  username      = "test-acc-username"
  password      = "test-acc-password"
  project_id    = "<PROJECT-ID>"
  database_name = "admin"

  roles {
    role_name     = "readWrite"
    database_name = "dbforApp"
  }
  roles {
    role_name     = "readAnyDatabase"
    database_name = "admin"
  }

  labels {
    key   = "key 1"
    value = "value 1"
  }
  labels {
    key   = "key 2"
    value = "value 2"
  }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ website/node_modules
 *.iml
 *.test
 *.iml
-log.txt
+log.*
 
 website/vendor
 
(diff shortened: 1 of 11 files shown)


themantissa (changes requested):
Looks good but missing the documentation updates for the data source.   That's all I think we need.

marinsalinas (changes requested):
Could you please update from master and get the latest version of the client code?

PacoDw:
yes, thank you! 👍

themantissa (approved):
LGTM

PR #106, opened 2020-01-15
Fix #68: Added the ability to re-create the whitelist entry when it's remove manually

### LINKED ISSUE #68: Whitelisted Project IPs when manually deleted causes failure at next plan/apply

PacoDw:
- Also, the export function was added due that it before doesn't work well

closes #68

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
+++ b/mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
@@ -173,10 +173,15 @@ func resourceMongoDBAtlasProjectIPWhitelistRead(d *schema.ResourceData, meta int
 
 		whitelist, _, err := conn.ProjectIPWhitelist.Get(context.Background(), ids["project_id"], ids["entry"])
 		if err != nil {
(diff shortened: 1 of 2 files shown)


themantissa (approved):
LGTM

marinsalinas (approved):
LGTM!

PR #107, opened 2020-01-17
Fix peering resource

PacoDw:
- The client merged the branch https://github.com/mongodb/go-client-mongodb-atlas/pull/38 which contains some renamed variables changes that break the peering resource. So this pr fixed those issues.


marinsalinas (approved):
LGTM! Thanks @PacoDw

themantissa (commented):
LGTM

PR #112, opened 2020-01-24
Update cluster.html.markdown

themantissa:
Include more explanation/example of where/how to set advanced options.


PacoDw (approved):
LGTM! 👍  thanks

PR #117, opened 2020-01-30
Update cluster doc

themantissa:
Removed 3.4 (EoL) and added 4.2. M2/M5 also now defaults 4.2

Upgraded instance size to call out M0 not supported.


PacoDw (approved):
LGTM! Thanks

PR #119, opened 2020-01-31
Shared tier doc edits

themantissa:
*Added a shared tier creation example
*Clarified M0 not supported
*Clarified cannot upgrade M2/M5.


PacoDw (approved):
LGTM! Thanks

PR #122, opened 2020-02-04
Database users: added x509_type attribute

PacoDw:
Added:
- `x509_type` attribute was added in the resource and the data source in the database resource, this allows us to configure it with the future X509 resource.
- The GET and DELETE Methods were refactored due to the client changed, so when the `x509_type` is enabled the database name must be `$external` and to be able to Get or Delete it we need to pass the database name to avoid errors.
- The documentation was updated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/go-test/deep v1.0.1
 	github.com/gobuffalo/packr v1.30.1
 	github.com/hashicorp/terraform v0.12.1
-	github.com/mongodb/go-client-mongodb-atlas v0.1.3-0.20200124011041-e364211a87d6
+	github.com/mongodb/go-client-mongodb-atlas v0.1.4-0.20200205164654-a46da013820e
 	github.com/mwielbut/pointy v1.1.0
 	github.com/spf13/afero v1.2.2 // indirect
 	github.com/spf13/cast v1.3.0
--- a/mongodbatlas/data_source_mongodbatlas_database_user.go
+++ b/mongodbatlas/data_source_mongodbatlas_database_user.go
@@ -3,7 +3,6 @@ package mongodbatlas
 import (
 	"context"
 	"fmt"
-	"log"
 
 	"github.com/hashicorp/terraform/helper/schema"
 
@@ -22,10 +21,13 @@ func dataSourceMongoDBAtlasDatabaseUser() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-
-			"database_name": {
+			"auth_database_name": {
 				Type:     schema.TypeString,
 				Required: true,
+			},
+			"x509_type": {
+				Type:     schema.TypeString,
+				Computed: true,
 			},
 			"roles": {
 				Type:     schema.TypeList,
@@ -72,7 +74,7 @@ func dataSourceMongoDBAtlasDatabaseUserRead(d *schema.ResourceData, meta interfa
 	conn := meta.(*matlas.Client)
 	projectID := d.Get("project_id").(string)
 	username := d.Get("username").(string)
-	databaseName := d.Get("database_name").(string)
+	databaseName := d.Get("auth_database_name").(string)
 
 	dbUser, _, err := conn.DatabaseUsers.Get(context.Background(), databaseName, projectID, username)
 	if err != nil {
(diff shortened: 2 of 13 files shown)


mongodbatlas/resource_mongodbatlas_database_user_test.go
```diff
@@ -220,7 +248,28 @@ func testAccMongoDBAtlasDatabaseUserConfig(projectID, roleName, username, keyLab
 			password      = "test-acc-password"
 			project_id    = "%[1]s"
 			database_name = "admin"
-			
+
+			roles {
+				role_name     = "%[2]s"
+				database_name = "admin"
```
themantissa, line 263: just as an FYI this doesn't have to be admin.  I think it's clear in our doc, but this could be any database that this user should have a specific role access on.  I know it can be confusing to have both roles.database and databaseName that's for the authentication database.

website/docs/d/database_user.html.markdown
```diff
@@ -54,14 +54,14 @@ data "mongodbatlas_database_user" "test" {
 
 * `username` - (Required) Username for authenticating to MongoDB.
 * `project_id` - (Required) The unique ID for the project to create the database user.
+* `database_name` - (Required) The user’s authentication database. A user must provide both a username and authentication database to log into MongoDB. In Atlas deployments of MongoDB, the authentication database is always the admin database.
```
themantissa: Update: 
`database_name` - (Required) The user’s authentication database. A user must provide both a username and authentication database to log into MongoDB. In Atlas deployments of MongoDB, the authentication database is almost always the admin database, for X509 it is $external.

website/docs/r/database_user.html.markdown
```diff
@@ -76,10 +76,10 @@ In addition to all arguments above, the following attributes are exported:
 
 ## Import
 
-Database users can be imported using project ID and username, in the format `PROJECTID-USERNAME`, e.g.
+Database users can be imported using project ID and username, in the format `PROJECTID-USERNAME-DATABASENAME`, e.g.
```
themantissa: This is going to be confusing. Do we have to have the authentication database here?  If so can we make it clear, more like: PROJECTID-USERNAME-AUTHENTICATIONDATABASENAME

website/docs/d/database_user.html.markdown
```diff
@@ -22,7 +22,7 @@ resource "mongodbatlas_database_user" "test" {
 	password      = "test-acc-password"
 	project_id      = "<PROJECT-ID>"
 	database_name = "admin"
-	
+
```
themantissa: Shouldn't we have a SCRAM Authentication example (what is here now) and x509 Certificate example now (e.g. admin and auth)?

website/docs/r/database_user.html.markdown
```diff
@@ -76,10 +76,10 @@ In addition to all arguments above, the following attributes are exported:
```
themantissa, line 101: Missing some needed changes to database_name, same as my comments in the data source docs and also would be great to have both a SCRAM authentication and X509 example.   Also doesn't this need the X509_type or did you set that w/ the X509 resource (I think it is needed here?) ?

mongodbatlas/resource_mongodbatlas_database_user.go
```diff
@@ -38,9 +38,15 @@ func resourceMongoDBAtlasDatabaseUser() *schema.Resource {
...
 			"password": {
-				Type:      schema.TypeString,
-				Optional:  true,
-				Sensitive: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Sensitive:     true,
+				ConflictsWith: []string{"x509_type"},
+			},
+			"x509_type": {
```
themantissa, line 46: Missing from doc (see notes below)

themantissa (changes requested):
Missing a few things, especially in docs. See comments.  Thanks!

marinsalinas (approved):

go.mod
```diff
@@ -16,3 +16,5 @@ require (
 	golang.org/x/sys v0.0.0-20190804053845-51ab0e2deafa // indirect
 	google.golang.org/grpc v1.23.0 // indirect
 )
+
+replace github.com/mongodb/go-client-mongodb-atlas => ../go-client-mongodb-atlas
```
marinsalinas: Plz remove when client's branch is merged.

marinsalinas (approved):

website/docs/r/database_user.html.markdown
PacoDw: Yes, It would be great, but I want first to merge the x509 resource to be able to create a test with both resources and later put the documentation examples.

go.mod
PacoDw: thanks @marinsalinas :)

website/docs/d/database_users.html.markdown
```diff
@@ -68,6 +67,8 @@ In addition to all arguments above, the following attributes are exported:
 * `roles` - List of user’s roles and the databases / collections on which the roles apply. A role allows the user to perform particular actions on the specified database. A role on the admin database can include privileges that apply to the other databases as well. See [Roles](#roles) below for more details.
 * `database_name` - The user’s authentication database. A user must provide both a username and authentication database to log into MongoDB. In Atlas deployments of MongoDB, the authentication database is always the admin database.
```
themantissa: needs same change as data source for database_users.

website/docs/r/database_user.html.markdown
themantissa: That works for me.  Just make sure any changes get updated doc.

mongodbatlas/data_source_mongodbatlas_database_user.go
```diff
@@ -81,21 +83,23 @@ func dataSourceMongoDBAtlasDatabaseUserRead(d *schema.ResourceData, meta interfa
...
 		return fmt.Errorf("error setting `labels` for database user (%s): %s", d.Id(), err)
 	}
 
 	d.SetId(encodeStateID(map[string]string{
-		"project_id":    projectID,
-		"username":      username,
-		"database_name": databaseName,
+		"project_id":         projectID,
+		"username":           username,
+		"auth_database_name": databaseName,
```
themantissa, line 102: I think this is a lot clearer as some found this confusing, thank you!  Is there any way to still support the old database_name as well without getting too messy?  If not then we'll need to be sure to log this as a breaking change in the Changelog, Fyi @marinsalinas

themantissa (changes requested):
Just one small doc change and a few notes.

mongodbatlas/data_source_mongodbatlas_database_user.go
PacoDw: I think that it is a breaking change, what do you think @marinsalinas ?

PacoDw: Also the little change in the client breaked up all :S

themantissa: Oh no, if it's breaking too much we should just leave it as is then, sorry to suggest it as didn't think about how much impact it has and we are already short on time.  Let's just revert this and save the headache.

PacoDw: It's okay it breaks a little bit  but with this PR all the breaking changes are fixed

website/docs/d/database_users.html.markdown
PacoDw: Thanks, @themantissa, good catch! 👍

themantissa (approved):
LGTM - reminder we'll want to note in the changelog that database_name is changing to auth_database_name as of 0.4.0 so users can change their terraform configuration to match.

PR #129, opened 2020-02-10
chore: added database_name as deprecated attribute

PacoDw:
- To avoid `auth_database_name` breaks the user's old configuration, the `database_name` attribute was added and its usage tagged as a deprecated warning.


themantissa (approved):
LGTM!

marinsalinas (approved):
LGTM!

PR #131, opened 2020-02-11
Fix snapshot import with hyphened cluster_name

### LINKED ISSUE #109: Issue with import  mongodbatlas_cloud_provider_snapshot_restore_job
_This issue was originally opened by @sushilvarma2 as hashicorp/terraform#23903. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    

Hello All,

We have an issue where mongo db cluster was created through Terraform. And Snapshot was created manually. I am trying to import snapshot details using terraform import. I have used the command as mentioned here:

Since  my clustername has hyphen in it so terraform is not able to understand this request and throwing an error.
clusterName: projectname-environment-mongo-global-cluster
Can someone suggest please how I can overcome this issue ?

Command:
terraform import mongodbatlas_cloud_provider_snapshot_restore_job.test 5cf5a45a9ccf6400e60981b6-MyCluster-5d1b654ecf09a24b888f4c79

Reference: 
Project_id, cluster_name and snapshot_id (Unique identifier of the snapshot), in the format PROJECTID-CLUSTERNAME-JOBID
https://www.terraform.io/docs/providers/mongodbatlas/r/cloud_provider_snapshot_restore_job.html

marinsalinas:
closes #109

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot.go
+++ b/mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"strings"
+	"regexp"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
@@ -166,10 +166,11 @@ func resourceMongoDBAtlasCloudProviderSnapshotCreate(d *schema.ResourceData, met
 	}
 
 	d.SetId(encodeStateID(map[string]string{
-		"snapshot_id":  snapshot.ID,
 		"project_id":   d.Get("project_id").(string),
 		"cluster_name": d.Get("cluster_name").(string),
+		"snapshot_id":  snapshot.ID,
 	}))
+
 	return resourceMongoDBAtlasCloudProviderSnapshotRead(d, meta)
 }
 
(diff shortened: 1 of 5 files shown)


marinsalinas:
Hello @sushilvarma2, could you please validate if the error is still happening?

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_test.go
```diff
@@ -166,3 +170,25 @@ func testAccMongoDBAtlasCloudProviderSnapshotConfig(projectID, clusterName, desc
 		}
 	`, projectID, clusterName, description, retentionInDays)
 }
+
+func TestResourceMongoDBAtlasCloudProviderSnapshot_snapshotID(t *testing.T) {
```
themantissa, line 174: Just a note: If this works well then we'll probably want to add this to all import statements since if the project name has hyphens it would fail with any import that isn't ignoring in value hyphens.

themantissa (approved):
LGTM w/ a comment we will likely want to add something to all imports if this comes up again.

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot.go
```diff
@@ -234,20 +229,36 @@ func resourceMongoDBAtlasCloudProviderSnapshotImportState(d *schema.ResourceData
...
+	if len(parts) != 4 {
+		return nil, errors.New("import format error: to import a snapshot, use the format {project_id}-{cluster_name}-{snapshot_id}")
+	}
+
+	return &matlas.SnapshotReqPathParameters{
+		GroupID:     parts[1],
+		ClusterName: parts[2],
+		SnapshotID:  parts[3],
+	}, nil
+}
```
PacoDw, line 264: Pretty nice!! Thanks @marinsalinas 👍

PacoDw (approved):

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_test.go
marinsalinas: yes, it would be great to add it to the remaining resources. But I think by the next release because I wouldn't like to break something. 😅

PR #134, opened 2020-02-13
Cluster: added backup_enabled as a deprecated attribute for new all AWS clusters

PacoDw:


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -452,6 +453,26 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 		clusterRequest.NumShards = pointy.Int64(cast.ToInt64(n))
 	}
 
+	if providerName == "AWS" {
+		backup, backupOk := d.GetOk("backup_enabled")
+		providerbackup, providerbackupOk := d.GetOk("provider_backup_enabled")
+		if !backupOk && !providerbackupOk {
+			return errors.New("For AWS cluster one of backup_enabled or provider_backup_enabled must be configured")
```
themantissa, line 460: Not correct, you do not have to setup backup.  both are optional in the API because a customer does not have to have backup.

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -486,21 +484,20 @@ func testAccCheckMongoDBAtlasClusterDestroy(s *terraform.State) error {
...
-			name         = "%[2]s"
-			disk_size_gb = 100
-			num_shards   = 1
+			project_id                   = "%[1]s"
+			name                         = "%[2]s"
+			disk_size_gb                 = 10
+			num_shards                   = 1
 			replication_factor           = 3
 			provider_backup_enabled      = %[3]s
-			pit_enabled 				 = %[3]s
```
themantissa, line 495: why would we remove pit_enabled?  We want to test this as it's one of the most needed things in 4.0 as it's a new flag for those using provider_backup_enabled to have PIT restores turned on.

themantissa (commented):
I do not think we should try to use Terraform's provider deprecation feature with the move from continuous backup (backup_enabled) to cloud provider backup (provider_backup_enabled) as it's currently in flux while it's rolled out and the Atlas API will produce the needed errors as items are deprecated within Atlas.  Remember:
-Both backup_enabled and provider_backup_enabled are optional - users don't have to have backup
-pitEnabled is a new flag we needed for _provider_backup_enabled and we have that so good there.

Once backup_enabled is truly deprecated we can expose this within the provider but I think trying to do so now will potentially result in bugs/issues.

mongodbatlas/resource_mongodbatlas_cluster_test.go
marinsalinas: I agree with @themantissa, is there any reason why you (@PacoDw ) removed?

mongodbatlas/resource_mongodbatlas_cluster.go
marinsalinas: I think it is just a validation to check if both are set but, since conflicts with backup_enabled and provider_backup_enabled are set, this validation is not needed.

Could you please test what happens if you remove the validation and set both arguments.

marinsalinas (changes requested):

themantissa:
Per discussion holding on deprecating anything w/ backup until CPS w/ PIT fully rolled out on Atlas

PR #158, opened 2020-02-26
cluster: removed array of regions due to they could be changed

### LINKED ISSUE #157: Can't create cluster ME_SOUTH_1 region
I am trying to create M10 cluster in AWS me-south-1 region but got an error.

image

Here my request payload in trace mode 
image

And the response from them
image

My terraform file
image
And plan
image

Also, I have tried to create cluster in us-east-1 region and them was successful
image

Seems to terraform provider erase my `regionName` when I set `ME_SOUTH_1` but why? In documentation mentioned that Mongo Atlas supports me-south-1 region
image

Can you help me?
Thanks in advance!

PacoDw:
closes #157

### DIFF
--- a/mongodbatlas/provider.go
+++ b/mongodbatlas/provider.go
@@ -125,95 +125,25 @@ func decodeStateID(stateID string) map[string]string {
 }
 
 func valRegion(reg interface{}, opt ...string) (string, error) {
-
-	regions := []string{
-		"US_EAST_1",
-		"US_EAST_2",
-		"US_WEST_1",
-		"US_WEST_2",
-		"CA_CENTRAL_1",
-		"SA_EAST_1",
-		"AP_NORTHEAST_1",
-		"AP_NORTHEAST_2",
-		"AP_SOUTH_1",
-		"AP_SOUTHEAST_1",
-		"AP_SOUTHEAST_2",
-		"EU_CENTRAL_1",
-		"EU_NORTH_1",
-		"EU_WEST_1",
-		"EU_WEST_2",
-		"EU_WEST_3",
-		"AZURE",
-		"AZURE_CHINA",
-		"AZURE_GERMANY",
-		"US_CENTRAL",
-		"US_EAST",
-		"US_NORTH_CENTRAL",
-		"US_WEST",
-		"US_SOUTH_CENTRAL",
-		"BRAZIL_SOUTH",
-		"CANADA_EAST",
-		"CANADA_CENTRAL",
-		"EUROPE_NORTH",
-		"EUROPE_WEST",
-		"UK_SOUTH",
-		"UK_WEST",
-		"FRANCE_CENTRAL",
-		"ASIA_EAST",
-		"ASIA_SOUTH_EAST",
-		"AUSTRALIA_EAST",
-		"AUSTRALIA_SOUTH_EAST",
-		"INDIA_CENTRAL",
-		"INDIA_SOUTH",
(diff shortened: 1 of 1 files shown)


mongodbatlas/provider.go
```diff
@@ -125,95 +125,25 @@ func decodeStateID(stateID string) map[string]string {
 }
 
 func valRegion(reg interface{}, opt ...string) (string, error) {
-
-	regions := []string{
```
shum, line 129: I would actually prefer if we went back to the hardcoded list of supported regions, perhaps we could move it out to a seperate file like `regions.go` for easier maintenance - this gives us a reference for what regions are supported, and easier troubleshooting, rather than letting everything through.

mongodbatlas/provider.go
```diff
@@ -125,95 +125,25 @@ func decodeStateID(stateID string) map[string]string {
...
-		"EUROPE_WEST_6",
-	}
-
 	region, err := cast.ToStringE(reg)
 	if err != nil {
 		return "", err
 	}
+	if region == "" {
+		return "", fmt.Errorf("region must be set")
+	}
```
shum, line 135: Can we add a simple check that `region` matches a value in the `regions` array, otherwise error?

mongodbatlas/provider.go
shum: We can simply add `ME_SOUTH_1` to the list

shum (changes requested):

mongodbatlas/provider.go
PacoDw: I thought that in the beginning, but if on your side makes a lot of changes removing and adding we need to maintain that file all the time, instead we can let this validation works on MongoDB server-side to show the correct information for the user. I don't have any problem to get back the list, but I think we need another opinion about this, What do you think @marinsalinas?

xmariopereira: Can’t you create function that call all available regions from mongodb api?

PacoDw: That is a really nice idea 👍 , but I don't know if it exists an entry point to retrieve all the available regions, let me take a look to verify it 👀 .

xmariopereira: you might have one problem, some instances types are not available in some regions, you need to do probably some error handling or logic on your side.

kostyazgara: Sorry guys to interrupt your discussion. I don't know every detail about how should be designed providers but if Atlas API validates all requests and doesn't start creation if the payload is invalid why a provider should validate payload on its own side? You can easily send a request and handle the error if it appeared.

Sorry again, but this feature might crucial for my project and should be fixed as soon as possible.

shum (approved):
LGTM. We can defer the error handling to the API endpoint.

PacoDw:
Thank you so much @shum !

marinsalinas (approved):
LGTM!

kostyazgara:
Can you say me, when changes will be applied in terraform provider and I can use them?

PacoDw:
@KostyaZgara I think the new release coming in the next week, but for now, you can use the master branch to use this new change, to use it you can follow these steps:
- Download the repository, make the following command to create the binary: `go build` 
- Once you get the binary you can cut and paste in your project root. (I think you will need to remove the release with you are working)
- Later make: `terraform init` and that's it.
Note: When the new release is done, you need to remove the current binary and make 'terraform init' to download automatically the last plugin version.

Issue #175, opened 2020-03-16
Docs with wrong resource type


hieuln:
Based on the document here: https://www.terraform.io/docs/providers/mongodbatlas/r/private_endpoint_interface_link.html the resource is:

resource "mongodbatlas_private_endpoint_link" "test" {
}

but the correct (and working) one is:

resource "mongodbatlas_private_endpoint_interface_link" "test" {
}

themantissa:
Fixed - doc gets updated on the terraform website when we release so expect this there with 0.4.3.  Thank you!!

PR #176, opened 2020-03-16
fix-#175 - missing word in resource name

themantissa:
Add the missing term interface to the documentation for private link interface.

Note - doc updates go out with new versions so this will merge and show up on Terraform's site when 0.4.3 is released.


PacoDw (approved):
Nice! Thank you so much!

PR #177, opened 2020-03-16
added roles in schema of notifications for alert configurations

coderGo93:
- Added field `roles` for schema `notification`
- Added value for roles in config of tf because now it's required for test


marinsalinas (changes requested):
Could.you add this field into the website documentation?

coderGo93:
> Could.you add this field into the website documentation?

you're right, I forgot that, just updated documentation, thank you @marinsalinas

themantissa (changes requested):
This PR shouldn't need to modify the go files (mod/sum).   Unless there's a reason they shouldn't be merged in.  otherwise lgtm.

marinsalinas:
> This PR shouldn't need to modify the go files (mod/sum). Unless there's a reason they shouldn't be merged in. otherwise lgtm.

Actually it should because this PR needs a new version of the client but, just he should remove the `replace` line in go.mod and update to the latest version of the client.

themantissa:
Thank you @marinsalinas  - I knew something looked a bit wrong.

marinsalinas (changes requested):
Please run `go mod vendor` and commit as `chore: update vendor dependencies", avoiding this causes Travis to fail.

themantissa (approved):
With fixes and checks passing LGTM, @marinsalinas your turn :)

PacoDw (approved):
LGTM!

PR #178, opened 2020-03-17
Typo in `provider_disk_type_name` description

caitlinelfring:
There is no `PP4` disk type, it should be `P4`

See https://docs.atlas.mongodb.com/reference/api/clusters-create-one/#request-body-parameters under `providerSettings.diskTypeName`


themantissa (approved):
Thank you @caitlin615 for the fix.  We'll add this in and it will go out w/ the next release.

PR #180, opened 2020-03-18
Cloud Provider Snapshot Backup Policy

PacoDw:
Added:

- `mongodbatlas_cloud_provider_snapshot_backup_policy` resource.
- `mongodbatlas_cloud_provider_snapshot_backup_policy` resource acceptance testing.
- `mongodbatlas_cloud_provider_snapshot_backup_policy` resource website documentation.
- `mongodbatlas_cloud_provider_snapshot_backup_policy` data source.
- `mongodbatlas_cloud_provider_snapshot_backup_policy` data source acceptance testing.
- `mongodbatlas_cloud_provider_snapshot_backup_policy` data source website documentation.

### Example Usage
```hcl
resource "mongodbatlas_cluster" "my_cluster" {
  project_id   = "<PROJECT-ID>"
  name         = "clusterTest"
  disk_size_gb = 5

  //Provider Settings "block"
  provider_backup_enabled     = true // enable cloud provider snapshots
}

resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
  project_id   = mongodbatlas_cluster.my_cluster.project_id
  cluster_name = mongodbatlas_cluster.my_cluster.name

  reference_hour_of_day    = 3
  reference_minute_of_hour = 45
  restore_window_days      = 4


  policies {
    id = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.id

    policy_item {
      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.0.id
      frequency_interval = 1
      frequency_type     = "hourly"
      retention_unit     = "days"
      retention_value    = 1
    }
    policy_item {
      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.1.id
      frequency_interval = 1
      frequency_type     = "daily"
      retention_unit     = "days"
      retention_value    = 2
    }
    policy_item {
      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.2.id
      frequency_interval = 4
      frequency_type     = "weekly"
      retention_unit     = "weeks"
      retention_value    = 3
    }
    policy_item {
      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.3.id
      frequency_interval = 5
      frequency_type     = "monthly"
      retention_unit     = "months"
      retention_value    = 4
    }
  }
}

data "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
  project_id   = mongodbatlas_cloud_provider_snapshot_backup_policy.test.project_id
  cluster_name = mongodbatlas_cloud_provider_snapshot_backup_policy.test.cluster_name
}
```

Also, we added a new parameter for the cluster resource and data sources`snapshot_backup_policy`, it represents the cloud provider snapshot backup policy for the created cluster. It's a computed attribute so to get all backup policies information the cluster must enable the `provider_backup_enabled` or `backup_enabled` depending on the mongo version.

Note the above configuration on how to get each policy item id to modify them.


mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
```diff
@@ -0,0 +1,406 @@
...
+
+	if err := snapshotScheduleUpdate(d, conn, projectID, clusterName); err != nil {
+		return err
+	}
+
+	return resourceMongoDBAtlasCloudProviderSnapshotBackupPolicyRead(d, meta)
+}
+
+func resourceMongoDBAtlasCloudProviderSnapshotBackupPolicyDelete(d *schema.ResourceData, meta interface{}) error {
+	// there is no way to det the default configuration
```
themantissa: Adjust comment to clarify further:
//There is no resource to delete a backup policy, it can only be updated.

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
```diff
@@ -0,0 +1,406 @@
...
+	return nil
+}
+
+func resourceMongoDBAtlasCloudProviderSnapshotBackupPolicyImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	conn := meta.(*matlas.Client)
+
+	parts := strings.SplitN(d.Id(), "-", 2)
```
themantissa, line 207: We've run into problems with importing with both project (and recently database users) as users often using "-" in their names.  Can we ensure we account for that going forward?

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -895,9 +905,9 @@ func flattenReplicationSpecs(rSpecs []matlas.ReplicationSpec) []map[string]inter
 	specs := make([]map[string]interface{}, 0)
 	for _, rSpec := range rSpecs {
 		spec := map[string]interface{}{
-			"id":             rSpec.ID,
-			"num_shards":     rSpec.NumShards,
-			"zone_name":      rSpec.ZoneName,
+			"id":         rSpec.ID,
+			"num_shards": rSpec.NumShards,
+			// "zone_name":      rSpec.ZoneName,
```
themantissa: why is zone_name commented out?

website/docs/d/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,108 @@
+---
+layout: "mongodbatlas"
+page_title: "MongoDB Atlas: cloud_provider_snapshot_backup_policy"
+sidebar_current: "docs-mongodbatlas-datasource-cloud-provider-snapshot-backup-policy"
+description: |-
+    Provides an Cloud Provider Snapshot Backup Policy Datasource.
```
themantissa: Change An to A:
     Provides **a** Cloud Provider Snapshot Backup Policy Datasource.

website/docs/d/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,108 @@
...
+---
+
+# mongodbatlas_cloud_provider_snapshot_backup_policy
+
+`mongodbatlas_cloud_provider_snapshot_backup_policy` provides an Cloud Provider Snapshot Backup Policy entry datasource. Atlas Cloud Provider Snapshots Backup Policy provides the current snapshot schedule and retention settings for the cluster .
```
themantissa: Just minor adjustments:
provides a Cloud Provider Snapshot Backup Policy entry datasource. An Atlas Cloud Provider Snapshot Backup Policy provides the current snapshot schedule and retention settings for the cluster.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
+---
+layout: "mongodbatlas"
+page_title: "MongoDB Atlas: cloud_provider_snapshot_backup_policy"
+sidebar_current: "docs-mongodbatlas-resource-cloud-provider-snapshot-backup-policy"
+description: |-
+    Provides an Cloud Provider Snapshot Backup Policy resource.
```
themantissa: Replace An with A
     Provides **a** Cloud Provider Snapshot Backup Policy resource.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+---
+
+# mongodbatlas_cloud_provider_snapshot_backup_policy
+
+`mongodbatlas_cloud_provider_snapshot_backup_policy` provides a resource that enables you to view and modify the snapshot scheduling and retention settings for an Atlas cluster with Cloud Provider Snapshots enabled
+
```
themantissa, line 12: slight adjustment:
 `mongodbatlas_cloud_provider_snapshot_backup_policy` provides a resource that enables you to view and modify the snapshot schedule and retention settings for an Atlas cluster with Cloud Provider Snapshots enabled

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.3.id
+      frequency_interval = 5
+      frequency_type     = "monthly"
+      retention_unit     = "months"
+      retention_value    = 4
+    }
+  }
+}
+
+data "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
```
themantissa: data shouldn't be here?

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+resource "mongodbatlas_cluster" "my_cluster" {
+  project_id   = "<PROJECT-ID>"
+  name         = "clusterTest"
+  disk_size_gb = 5
+
+  //Provider Settings "block"
+  provider_name               = "AWS"
+  provider_region_name        = "EU_CENTRAL_1"
+  provider_instance_size_name = "M10"
+  provider_backup_enabled     = true // enable cloud provider snapshots
```
themantissa: Update comment on this one to read  (again they might be able to adjust the actual policy without this being true - need to test - but it's doesn't have any meaning unless provider_backup_enabled is true)
// must be enabled in order to use cloud_provider_snapshot_backup_policy resource

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+data "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
+  project_id   = mongodbatlas_cloud_provider_snapshot_backup_policy.test.project_id
+  cluster_name = mongodbatlas_cloud_provider_snapshot_backup_policy.test.cluster_name
+}
+```
+
+## Argument Reference
+
+* `project_id` - (Required) The unique identifier of the project for the Atlas cluster.
+* `cluster_name` - (Required) The name of the Atlas cluster that contains the snapshots backup policy you want to retrieve.
```
themantissa: remove s from snapshots:
The name of the Atlas cluster that contains the snapshot backup policy you want to retrieve.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -322,6 +322,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 					},
 				},
 			},
+			"snapshot_backup_policy": computedCloudProviderSnapshotBackupPolicySchema(),
```
themantissa, line 325: FYI I _think_ we can compute this even if they user is not using provider_backup_enabled = true but we should probably test it to be sure.  Basically the snapshot backup policy is only applicable if provider_backup_enabled = true.  backup_enabled does not use this policy.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+* `restore_window_days` - (Optional) Number of days back in time you can restore to with point-in-time accuracy. Must be a positive, non-zero integer.
+* `update_snapshots` - (Optional) Specify true to apply the retention changes in the updated backup policy to snapshots that Atlas took previously.
+
+### Policies
+* `policies` - (Required) Contains a document for each backup policy item in the desired updated backup policy.
+* `policies.#.id` - (Required) Unique identifier of the backup policy that you want to update.
+
+#### Policy Item
+* `policies.#.policy_item` - (Required) Array of backup policy items.
+* `policies.#.policy_item.#.id` - (Required) Unique identifier of the backup policy item.
```
themantissa: Similar to above: 

(Required) Unique identifier of the backup policy item.    `policies.#.policy_item.#.id` is a value obtained via the mongodbatlas_cluster resource.  `provider_backup_enabled` of the mongodbatlas_cluster resource must be set to true.   See the example above for how to refer to the mongodbatlas_cluster resource for`policies.#.policy_item.#.id`

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `cluster_id` - Unique identifier of the Atlas cluster.
+* `next_snapshot` - Timestamp in the number of seconds that have elapsed since the UNIX epoch when Atlas takes the next snapshot.
+
+## Import
+
+Cloud Provider Snapshot Backup Policy entries can be imported using project project_id, cluster_name, in the format `PROJECTID-CLUSTERNAME`, e.g.
```
themantissa, line 107: Would someone need to import if the cluster already existed or would they ever need to import since really cluster is always doing that for them?

website/docs/r/cluster.html.markdown
```diff
@@ -361,6 +361,28 @@ In addition to all arguments above, the following attributes are exported:
     - DELETED
     - REPAIRING
 
+### Cloud Provider Snapshot Backup Policy
```
themantissa, line 365: We shouldn't export this unless here unless they are using cloud_provider_snapshots.  I'm increasingly thinking the compute needs a check for provider_backup_enabled being true since this doesn't apply if backup_enabled = true.

themantissa (approved):
This is really cool in how it flows - made some comments in that this policy only applies for cloud provider snapshots so I think we need some logic not compute if provider_backup_enabled isn't true.

themantissa (changes requested):
Ugh hit wrong radio buttons - from the approve comment - This is really cool in how it flows - made some comments in that this policy only applies for cloud provider snapshots so I think we need some logic to not compute if provider_backup_enabled isn't true.

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
PacoDw: good catch! thank you :)

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
PacoDw: Yes don't worry about it, in this resource, we won't have problems because the clusterName is in the last position, so the "strings.SplitN(d.Id(), "-", 2)" function parts it in two string when it catches the first dash coincidence ("-"), in this case, it is in before the clusterName, the problem is when there are three variables and the variable in the middle has dashes.
I thought it and then tested it while I developed the resource 👍

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: Good catch! because the Cluster doesn't return the zone name variable in the response instead of that it returns the name of the zone like an object, e.j. :
```
"replicationSpecs": [
                {
                    "id": "5e72a859c4fb8e098f5b2c11",
                    "numShards": 1,
                    "regionsConfig": {
                        "EU_CENTRAL_1": {  // <---- here
                            "analyticsNodes": 0,
                            "electableNodes": 3,
                            "priority": 7,
                            "readOnlyNodes": 0
                        }
                    },
                    "zoneName": "Zone 1"
                }
            ],
```
Then when a ran a cluster test I got an error in the log, so I catched this, so to avoid errors I commented to check this with @marinsalinas, so @marinsalinas do you know if the response was always like this?

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: thank you :)

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: yes it must even be computed if the user is no using the provider_backup_enabled = true

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: The user could import it to change the policies configuration it the cluster is already exist

website/docs/r/cluster.html.markdown
PacoDw: Internally if the user no uses the backup_enabled = true, that part of the state will have clean fields until the user enables the backup_enabled, I think the best way is to comment this

mongodbatlas/resource_mongodbatlas_cluster.go
marinsalinas: I'm.mot sure but you can remove it and then add validation to set if it's.presemt.

PacoDw: I will check it to be ensured :)

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
themantissa: Awesome, thank you for ensuring and the explanation.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
themantissa: Ah, okay I need to make sure I get the logic here.  The policy will always exist if they have cloud provider snapshots enabled so this would be needed if they did not create the cluster in terraform and wanted to import both the cluster and the policy or if they use terraform before we added this?

website/docs/r/cluster.html.markdown
themantissa: Even if the user a user enables backup_enabled it will (should) have clean fields as that has nothing to do this cloud provider backup, other than they are mutually exclusive.  I will test what is returned from this resource in the API if a user uses the old backup_enabled.  Hopefully I can still test it as we are deprecating it for new clusters!

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: For import, the provider backup enabled must be enabled for the cluster if not the won't get any policy data. If the user wants to use the new resource so they want to enable the provider backup and make `terraform refresh` to get the info in the cluster.

website/docs/r/cluster.html.markdown
PacoDw: it is validated for both provider backup enabled and backup enabled attribute for now.

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: I just added a validation for that attribute, because it was missing

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: Will this not result in an error?    Based on what I tested the API for cloud provider snapshots returns an error "BACKUP_CONFIG_NOT_FOUND" unless provider_backup_enabled is true.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
themantissa: I think this may have gotten resolved prematurely?

website/docs/r/cluster.html.markdown
themantissa: Unresolving as I sent some screen shots over slack and want to be sure it is clear that backup_enabled and snapshot backup policy are mutually exclusive as I'm not sure it is.   Only provider_backup_enabled and snapshot_backup_ policy work together.  Let me know if I can answer any more questions on that.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
themantissa: We should probably note that in the documentation then.  Perhaps in the provider_backup_enabled doc we add (For example: If setting to true for an existing cluster or imported cluster be sure to run `terraform refresh` after applying to enable modification of the Cloud Provider Snapshot Backup Policy going forward.)

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: Don't worry that's covered, I validated that type of error :), if it happens just the snapshot_backup_policy won't be fill, it keeps empty.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: I'm sorry I lost on this topic :S

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: that's right let me add it

website/docs/r/cluster.html.markdown
PacoDw: I will add a note about your last comment, "Cloud Provider Snapshot Backup Policy will add if provider_backup_enabled is enabled, instead, it will keep empty"

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
themantissa: Sorry should have been more specific - One suggestion got added, the other did not so I reopened this comment to ensure both get updated.  I'll double check the latest commit too in case I missed it.

website/docs/r/cluster.html.markdown
themantissa: Peferct, maybe just a bit clearer with:

Cloud Provider Snapshot Backup Policy will be added if provider_backup_enabled is enabled, if not, returned values will be empty.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: okay Thank you so much!

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: Ah, then we should be fine for now then as eventually we'll have all clusters using cps.

website/docs/r/cluster.html.markdown
PacoDw: Thank you I will change it right now

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -73,6 +73,8 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
 }
 ```
 
+~> **IMPORTANT:** If setting to true for an existing cluster or imported cluster be sure to run terraform refresh after applying to enable modification of the Cloud Provider Snapshot Backup Policy going forward.
+
```
themantissa: I'm not sure if this is the best location for this - would this be for if someone is setting provider_backup_enabled = true?  If so I think including this information in the Cluster resource documentation where provider_backup_enabled argument is defined is would make more sense?

themantissa (changes requested):
Responded to a few of the questions.  I think we are getting there!

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: make sense I will change it to the cluster documentation, thanks

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -0,0 +1,113 @@
...
+* `project_id` - (Required) The unique identifier of the project for the Atlas cluster.
+* `cluster_name` - (Required) The name of the Atlas cluster that contains the snapshot backup policy you want to retrieve.
+* `reference_hour_of_day` - (Optional) UTC Hour of day between 0 and 23, inclusive, representing which hour of the day that Atlas takes snapshots for backup policy items.
+* `reference_minute_of_hour` - (Optional) UTC Minutes after referenceHourOfDay that Atlas takes snapshots for backup policy items. Must be between 0 and 59, inclusive.
+* `restore_window_days` - (Optional) Number of days back in time you can restore to with point-in-time accuracy. Must be a positive, non-zero integer.
+* `update_snapshots` - (Optional) Specify true to apply the retention changes in the updated backup policy to snapshots that Atlas took previously.
+
+### Policies
+* `policies` - (Required) Contains a document for each backup policy item in the desired updated backup policy.
+* `policies.#.id` - (Required) Unique identifier of the backup policy that you want to update.
```
themantissa: Last thing -change the policies.$.id to be more descriptive since one has to refer to the cluster resource, change to:

(Required) Unique identifier of the backup policy that you want to update.
`policies.#.id` is a value obtained via the mongodbatlas_cluster resource. provider_backup_enabled of the mongodbatlas_cluster resource must be set to true. See the example above for how to refer to the mongodbatlas_cluster resource for policies.#.id

themantissa (changes requested):
1 thing that got missed a few times.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: it's done 👍

themantissa (approved):
LGTM!! Thank you :D

marinsalinas (approved):
LGTM!

PR #184, opened 2020-03-24
Fix documentation of roles block role_name

fbreckle:
The docs are wrong in this part. It requires `role_name` not `name`.

Using `name` gives: An argument named "name" is not expected here.


themantissa:
@fbreckle nice catch!  Thank you for the PR to fix.

themantissa (approved):
minor doc fix - LGTM

PR #193, opened 2020-04-13
Peering Container documentation fix

themantissa:
Upstream Atlas API docs incorrect described these parameters.  Fixing here and submitted a bug report upstream.  Examples in both, however, are correct.

### DIFF
--- a/website/docs/d/network_container.html.markdown
+++ b/website/docs/d/network_container.html.markdown
@@ -47,8 +47,8 @@ In addition to all arguments above, the following attributes are exported:
 * `id` - The Network Peering Container ID.
 * `atlas_cidr_block` - CIDR block that Atlas uses for your clusters. Atlas uses the specified CIDR block for all other Network Peering connections created in the project. The Atlas CIDR block must be at least a /24 and at most a /21 in one of the following [private networks](https://tools.ietf.org/html/rfc1918.html#section-3).
 * `provider_name`  - Cloud provider for this Network Peering connection. If omitted, Atlas sets this parameter to AWS.
-* `region_name` - AWS region.
-* `region` - Azure region where the container resides.
+* `region_name` - The Atlas AWS region name for where this container will exist.
+* `region` - The Atlas Azure region name for where this container will exist.
 * `azure_subscription_id` - Unique identifer of the Azure subscription in which the VNet resides.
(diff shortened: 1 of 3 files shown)


website/docs/d/network_container.html.markdown
```diff
@@ -47,8 +47,8 @@ In addition to all arguments above, the following attributes are exported:
 * `id` - The Network Peering Container ID.
 * `atlas_cidr_block` - CIDR block that Atlas uses for your clusters. Atlas uses the specified CIDR block for all other Network Peering connections created in the project. The Atlas CIDR block must be at least a /24 and at most a /21 in one of the following [private networks](https://tools.ietf.org/html/rfc1918.html#section-3).
 * `provider_name`  - Cloud provider for this Network Peering connection. If omitted, Atlas sets this parameter to AWS.
-* `region_name` - AWS region.
-* `region` - Azure region where the container resides.
+* `region_name` - The Atlas AWS region name for where this container will exist.
```
shum, line 50: Wondering if the modifier should go be placed on the container rather than the region name, such as: 
```suggestion
* `region_name` - The AWS region name for where this Atlas container will exist.
```
or
```suggestion
* `region_name` - The AWS region name for where Atlas' network peering container will exist.
```
Since I'm not sure what the original confusion was caused by, also fine with "the Atlas AWS/Azure region name" pattern.

shum (approved):
Small alternative wording proposal

website/docs/d/network_container.html.markdown
themantissa: So the actual Atlas doc is wrong, it's the Atlas region name that is needed, not the AWS region name.  The examples are right the main Atlas API docs but not the text (DOCSP-9914 for fix) so was trying to clarify here as we had someone trying to use the AWS region name, which is slightly different, so it failed.  Make more sense now?

shum: Oh got it, then your original suggestion is correct. "Atlas AWS/Azure region name" is a bit of a mouthful but I don't think we have anything better available. 

LGTM.

themantissa: That's why I was open to suggestions - it's awkward to phrase.  If docs had a better way I'll update this then.  Thanks for the feedback/discussion!

PR #202, opened 2020-04-22
Changed roles to computed

### LINKED ISSUE #201: Alert configuration roles array should not be required
Provider version 0.5.0

Issue:
The MongoDB Atlas documentation states that notifications.roles is required, when it is not (see https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/) a request will be submitted to correct the docs bug.   However roles really only applies if the notification type is GROUP, and even then when omitted it simply sends the alert to all users in the project.    Currently requiring roles is likely to break alert configs set to EMAIL, which does not need roles.

PacoDw:
Changed required statement for roles attribute to be computed and added a test case to validate it.

closes #201

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_alert_configuration.go
+++ b/mongodbatlas/resource_mongodbatlas_alert_configuration.go
@@ -246,7 +246,7 @@ func resourceMongoDBAtlasAlertConfiguration() *schema.Resource {
 						},
 						"roles": {
 							Type:     schema.TypeList,
-							Required: true,
+							Optional: true,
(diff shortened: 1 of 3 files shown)


themantissa (changes requested):
wow, thank you for fast turn around.  We should include the documentation for the resource, right now we have required, just need to change to optional for now.

themantissa (approved):
LGTM - just need one DoU reviewer.  Thank you!!

coderGo93 (approved):
LGTM

PR #203, opened 2020-04-22
Fix: improved validation to avoid error 404

### LINKED ISSUE #200: terraform plan and apply fails after upgrading this module to 0.5.0
I have a MongoDB cluster managed with Terraform, with no backup enabled:
```terraform
resource "mongodbatlas_cluster" "this" {
  project_id                   = "<project-id>"
  provider_name                = "AWS"
  provider_region_name         = "US-EAST-1"
  name                         = "MyCluster"
  provider_instance_size_name  = "M10"
  mongo_db_major_version       = "4.2"
}
```
The cluster was created using an earlier version of this module, but when running `plan` or `apply` with the new version, I get:
```
Error: error getting a Cloud Provider Snapshot Backup Policy for the cluster(MyCluster): GET https://cloud.mongodb.com/api/atlas/v1.0/groups/<project-id>/clusters/MyCluster/backup/schedule: 404 (request "Not Found") No backup configuration exists for cluster <cluster-id> in group <project-id>.
```
It appears that relevant [API](https://docs.atlas.mongodb.com/reference/api/cloud-provider-snapshot-schedule-get-all/) returns `HTTP 404` if no cloud provider snapshots are available, which will be also the case when I would decide to add a cloud provider snapshot later.

My current work-around is not to upgrade to the new version.

PacoDw:
closes #200

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
+++ b/mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
@@ -304,9 +305,12 @@ func flattenPolicyItems(items []matlas.PolicyItem) []map[string]interface{} {
 }
 
 func flattenCloudProviderSnapshotBackupPolicy(d *schema.ResourceData, conn *matlas.Client, projectID, clusterName string) ([]map[string]interface{}, error) {
-	backupPolicy, _, err := conn.CloudProviderSnapshotBackupPolicies.Get(context.Background(), projectID, clusterName)
+	backupPolicy, res, err := conn.CloudProviderSnapshotBackupPolicies.Get(context.Background(), projectID, clusterName)
 	if err != nil {
-		if strings.Contains(fmt.Sprint(err), "BACKUP_CONFIG_NOT_FOUND") || strings.Contains(fmt.Sprint(err), "400") {
+		if res.StatusCode == http.StatusNotFound ||
+			strings.Contains(fmt.Sprint(err), "BACKUP_CONFIG_NOT_FOUND") ||
+			strings.Contains(fmt.Sprint(err), "Not Found") ||
+			strings.Contains(fmt.Sprint(err), "404") {
 			return []map[string]interface{}{}, nil
 		}
 		return []map[string]interface{}{}, fmt.Errorf(errorSnapshotBackupPolicyRead, clusterName, err)


mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
```diff
@@ -304,9 +305,12 @@ func flattenPolicyItems(items []matlas.PolicyItem) []map[string]interface{} {
 }
 
 func flattenCloudProviderSnapshotBackupPolicy(d *schema.ResourceData, conn *matlas.Client, projectID, clusterName string) ([]map[string]interface{}, error) {
-	backupPolicy, _, err := conn.CloudProviderSnapshotBackupPolicies.Get(context.Background(), projectID, clusterName)
+	backupPolicy, res, err := conn.CloudProviderSnapshotBackupPolicies.Get(context.Background(), projectID, clusterName)
 	if err != nil {
-		if strings.Contains(fmt.Sprint(err), "BACKUP_CONFIG_NOT_FOUND") || strings.Contains(fmt.Sprint(err), "400") {
+		if res.StatusCode == http.StatusNotFound ||
+			strings.Contains(fmt.Sprint(err), "BACKUP_CONFIG_NOT_FOUND") ||
+			strings.Contains(fmt.Sprint(err), "Not Found") ||
```
themantissa, line 312: should we still have 400 too?  Just in case another case returns a 400 vs 404?

themantissa (commented):
Question in comment - let me know your thoughts @PacoDw .

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_backup_policy.go
PacoDw: I want to be sure that the validation validates only when the backup config is not found I checked the code documentation (https://docs.atlas.mongodb.com/reference/api/api-errors/) to find all related to backup_config to see the codes and I only found the 404 for this case.

themantissa: Good point - let me make sure that doc is updated and we aren't missing something.

PacoDw: Thank you so much :)

themantissa: Per slack discussion resolving this as 404 should get us the error we want (for when no policy exists because CPS not enabled).  TY!

themantissa (commented):
LGTM!!

themantissa (approved):

themantissa (approved):
LGTM - sorry didn't realize there was a new commit!

marinsalinas (approved):

PR #208, opened 2020-04-27
Added field container_id in resource cluster

coderGo93:
- Added field in schema of cluster called `container_id` in resource cluster
- Show `container_id` in cluster state when needed and if is `AZURE` as provider name

Now we can use it as the following configuration when the cluster is created first and let it creates the container automatically to retrieve the `container_id` in the cluster and be used for the peering connection.

## AWS
```hcl
  project_id   = local.project_id
  name         = "terraform-test"
  disk_size_gb = 5

  //Provider Settings "block"
  provider_region_name        = "US_EAST_2"
}

resource "aws_default_vpc" "default" {
  tags = {
    Name = "Default VPC"
  }
}

resource "mongodbatlas_network_peering" "mongo_peer" {
  accepter_region_name   = "us-east-2"
  project_id             = local.project_id
  container_id           = mongodbatlas_cluster.test.container_id
  provider_name          = "AWS"
  route_table_cidr_block = "172.31.0.0/16"
  vpc_id                 = aws_default_vpc.default.id
  aws_account_id         = local.AWS_ACCOUNT_ID
}

resource "aws_vpc_peering_connection_accepter" "aws_peer" {
  vpc_peering_connection_id = mongodbatlas_network_peering.mongo_peer.connection_id
  auto_accept               = true

  tags = {
    Side = "Accepter"
  }
}
```

## AZURE
```hcl
  project_id = local.project_id
  name       = "terraform-test"
  num_shards = 1

  //Provider Settings "block"
  provider_name               = "AZURE"
  provider_region_name        = "US_EAST_2"
}

resource "mongodbatlas_network_peering" "test" {
  project_id            = local.project_id
  atlas_cidr_block      = "192.168.0.0/21"
  container_id          = mongodbatlas_cluster.test.container_id
  provider_name         = "AZURE"
  azure_directory_id    = local.AZURE_DIRECTORY_ID
  azure_subscription_id = local.AZURE_SUBCRIPTION_ID
  resource_group_name   = local.AZURE_RESOURSE_GROUP_NAME
  vnet_name             = local.AZURE_VNET_NAME
}
```
## GCP
```hcl
  project_id   = local.project_id
  name         = "terraform-manually-test"
  num_shards   = 1
  disk_size_gb = 5

  //Provider Settings "block"
  provider_name               = "GCP"
  provider_region_name        = "US_EAST_2"
}

resource "mongodbatlas_network_peering" "test" {
  project_id       = "${local.project_id}"
  atlas_cidr_block = "192.168.0.0/18"

  container_id   = mongodbatlas_cluster.test.container_id
  provider_name  = "GCP"
  gcp_project_id = local.GCP_PROJECT_ID
  network_name   = "default"
}

data "google_compute_network" "default" {
  name = "default"
}

resource "google_compute_network_peering" "peering" {
  name         = "peering-gcp-terraform-test"
  network      = data.google_compute_network.default.self_link
  peer_network = "https://www.googleapis.com/compute/v1/projects/${mongodbatlas_network_peering.test.atlas_gcp_project_id}/global/networks/${mongodbatlas_network_peering.test.atlas_vpc_name}"
}
```

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -502,12 +502,12 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 	}
 
 	var containerId string
-	if providerName == "AZURE"{
+	if providerName == "AZURE" {
 		if v, ok := d.GetOk("container_id"); ok {
 			containerId = v.(string)
 			_, res, err := conn.Containers.Get(context.Background(), projectID, containerId)
 			if err != nil {
-				if res.StatusCode == 404 {//Container doesn't exists
+				if res.StatusCode == 404 { //Container doesn't exists
 					options := &matlas.ContainersListOptions{
 						ProviderName: providerName,
 					}
@@ -527,7 +527,7 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 						return fmt.Errorf(errorContainerDelete, decodeStateID(d.Id())["container_id"], err)
 					}
 					containers := containersResp.([]matlas.Container)
-					if len(containers) > 0{
+					if len(containers) > 0 {
 						containerId = containers[0].ID
 					}
 				}
@@ -654,7 +654,9 @@ func resourceMongoDBAtlasClusterRead(d *schema.ResourceData, meta interface{}) e
 		return fmt.Errorf(errorClusterSetting, "labels", clusterName, err)
 	}
 
-	d.Set("container_id", ids["container_id"])
+	if err := d.Set("container_id", ids["container_id"]); err != nil {
+		return fmt.Errorf(errorClusterSetting, "container_id", clusterName, err)
+	}
 
 	/*
 		Get the advaced configuration options and set up to the terraform state
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -620,7 +620,7 @@ func resourceMongoDBAtlasClusterRead(d *schema.ResourceData, meta interface{}) e
 		return fmt.Errorf(errorClusterSetting, "labels", clusterName, err)
 	}
 
-	if providerName == "AZURE" {
+	if providerName == "AZURE" || providerName == "AWS" {
 		options := &matlas.ContainersListOptions{
 			ProviderName: providerName,
 		}
@@ -631,7 +631,9 @@ func resourceMongoDBAtlasClusterRead(d *schema.ResourceData, meta interface{}) e
 
 		if len(containers) != 0 {
 			for _, container := range containers {
-				if container.ProviderName == "AZURE" && container.Region == cluster.ProviderSettings.RegionName {
+				if container.ProviderName == providerName &&
+					container.Region == cluster.ProviderSettings.RegionName || // For Azure
+					container.RegionName == cluster.ProviderSettings.RegionName { // For AWS
 					if err := d.Set("container_id", container.ID); err != nil {
 						return fmt.Errorf(errorClusterSetting, "container_id", clusterName, err)
 					}
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -359,7 +359,6 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"snapshot_backup_policy": computedCloudProviderSnapshotBackupPolicySchema(),
 			"container_id": {
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 		},
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -357,6 +357,11 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				},
 			},
 			"snapshot_backup_policy": computedCloudProviderSnapshotBackupPolicySchema(),
+			"container_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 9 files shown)


themantissa:
Is this ready for review or still WIP?

coderGo93:
@themantissa Sorry, still WIP because I haven't tested yet.

themantissa:
Ah, can you wait to request a review then @coderGo93 till you are ready?  Thank you.

coderGo93:
Sure, no problem @themantissa

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -502,12 +502,12 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 	}
 
 	var containerId string
-	if providerName == "AZURE"{
+	if providerName == "AZURE" {
 		if v, ok := d.GetOk("container_id"); ok {
 			containerId = v.(string)
```
PacoDw: Why do you have this validation? if the `containerId` attribute is computed (you can't get the value), it never will set by the user, also it's unnecessary to create a variable `containerId` (because it will always be empty)

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -502,12 +502,12 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
...
 			_, res, err := conn.Containers.Get(context.Background(), projectID, containerId)
```
PacoDw: This code doesn't work, containerId is not set because the attribute is computed, you need to validate if a container exists if not, you need to wait until it creates automatically in order to get the containerId and set it to the `container_id` computed attribute.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -527,7 +527,7 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 						return fmt.Errorf(errorContainerDelete, decodeStateID(d.Id())["container_id"], err)
 					}
 					containers := containersResp.([]matlas.Container)
-					if len(containers) > 0{
+					if len(containers) > 0 {
 						containerId = containers[0].ID
 					}
 				}
```
PacoDw: you need to check the logic, due to the above comments.

PacoDw (changes requested):

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: I see, I misunderstood the behavior then, let me change it, thank you paco

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: I misunderstood the behavior, I'll change it, thank you paco

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -502,36 +502,38 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 
 	var containerId string
```
PacoDw: the variable should be `containerID` instead of `containerId`

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -502,36 +502,38 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
...
 	if providerName == "AZURE" {
-		if v, ok := d.GetOk("container_id"); ok {
-			containerId = v.(string)
-			_, res, err := conn.Containers.Get(context.Background(), projectID, containerId)
+		options := &matlas.ContainersListOptions{
+			ProviderName: providerName,
+		}
+		containersOriginal, _, _ := conn.Containers.List(context.Background(), projectID, options)
```
PacoDw: I think we need to validate in case the request/response has an error

PacoDw (changes requested):

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -502,36 +502,38 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
...
+						containerId = container.ID
 					}
 				}
 			}
 		}
+		for _, container := range containersOriginal {
+			if container.ProviderName == "AZURE" && container.Region == cluster.ProviderSettings.RegionName {
+				containerId = container.ID
+			}
+		}
```
PacoDw: all the above logic needs to be in the Read function instead of here(create) because is more ease to handle it, and it's the place where you have to set the computed attributes

PacoDw (changes requested):

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -502,36 +502,38 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
...
+			stateConf := &resource.StateChangeConf{
+				Pending:    []string{"provisioned_container"},
+				Target:     []string{"different"},
+				Refresh:    resourceListNetworkContainerRefreshFunc(d, conn, containersOriginal),
+				Timeout:    1 * time.Hour,
+				MinTimeout: 10 * time.Second,
+				Delay:      2 * time.Minute,
+			}
+			// Wait, catching any errors
+			containersResp, err := stateConf.WaitForState()
```
PacoDw: Also, could you remove the refresh function we don't need, you can use `containerOriginal` to get the `containerID`

PacoDw (changes requested):

PacoDw:
@themantissa it's ready for review 👍

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -614,6 +620,26 @@ func resourceMongoDBAtlasClusterRead(d *schema.ResourceData, meta interface{}) e
 		return fmt.Errorf(errorClusterSetting, "labels", clusterName, err)
 	}
 
+	if providerName == "AZURE" {
```
themantissa: I'm not sure why this is specific to Azure - I know Paco used Azure in his example but we support 3 clouds so all three would need to account for this?  If a cluster is created before the container then the container is created, no matter the provider.  Only one container is created for GCP but for Azure and AWS a container is created per each new region.   However, let me know if I'm missing something obvious please.

website/docs/d/cluster.html.markdown
```diff
@@ -102,7 +102,7 @@ In addition to all arguments above, the following attributes are exported:
 
 * `replication_specs` - Configuration for cluster regions.  See [Replication Spec](#replication-spec) below for more details.
 
-
+* `container_id` - The Network Peering Container ID.
```
themantissa, line 105: I'd add to this: 
 * `container_id` - The Network Peering Container ID. The id of the container either created programmatically by the user before any clusters existed in the project or when the first cluster in the region (AWS/Azure) or project (GCP) was created.  
@shum correct description in your opinion?

We should also update the container_id variable at https://www.terraform.io/docs/providers/mongodbatlas/r/network_peering.html that it can be obtained from the cluster returned values if the cluster has been created before the peering connection.

themantissa (commented):
Question about why Azure only?  Also asking for Chris' review (I'll add him as a reviewer as well, hopefully he has time to look as he knows quite a bit about this as well).

mongodbatlas/resource_mongodbatlas_cluster.go
PacoDw: you're right so let me check with @coderGo93

PacoDw: I checked the ticket and analyze it and yes you're right so I will finish the pending work with this issue

coderGo93 (commented):
LGTM

themantissa (approved):
Think we are covered now in testing!  LGTM :)

PacoDw (approved):
LGM :)

PR #212, opened 2020-04-29
Fix #80 - Update for GCP Encryption at rest

### LINKED ISSUE #80: mongodbatlas_encryption_at_rest key rotation impossible to perform with Azure KeyVault
When using customer keys for encryption at rest it is possible to rotate encryption keys with no downtime through the MongoDB atlas portal by just specifying a different (and accessible) key ID (it even works by changing any credentials, even using a different KeyVault).

The problem is when using this module's `mongodbatlas_encryption_at_rest` resource: having a complete project + cluster + encryption set of resources, like so:

```
resource "mongodbatlas_project" "project" {
  name   = "test-project"
  org_id = "some-org-id"
}

resource "mongodbatlas_cluster" "cluster" {
  project_id = mongodbatlas_project.project.id
  name       = "test-cluster"
  encryption_at_rest_provider = "AZURE"
  provider_name               = "AZURE"
  
  # Other settings...
}

resource "mongodbatlas_encryption_at_rest" "encryption" {
  project_id = mongodbatlas_project.project.id

  # Disabled AWS KMS...

  azure_key_vault = {
    enabled             = true
    #...
    key_identifier      = "SOME-KEY-IDENTIFIER"
    #...
  }

  # Disabled GCE KMS...
}
```

Now, rotating the encryption key should just be changing the `key_identifier` value and applying the change. I.e.:

```
# no changes here...

resource "mongodbatlas_encryption_at_rest" "encryption" {
  project_id = mongodbatlas_project.project.id

  # Disabled AWS KMS...

  azure_key_vault = {
    enabled             = true
    #...
    key_identifier      = "OTHER-KEY-IDENTIFIER"
    #...
  }

  # Disabled GCE KMS...
}
```

However, the plan for this is not an unpdate-in-place, rather a recreation of the whole encryption configuration.

```
$ terraform plan

  # mongodbatlas_encryption_at_rest.encryption must be replaced
-/+ resource "mongodbatlas_encryption_at_rest" "encryption" {
    ...

      ~ azure_key_vault  = { # forces replacement
            "azure_environment"   = "AZURE"
            "client_id"           = "38448f6a-0d05-4100-a883-23fc9bc5d1fa"
            "enabled"             = "true"
          ~ "key_identifier"      = "SOME-KEY-IDENTIFIER" -> "OTHER-KEY-IDENTIFIER"
          ...
      }
      ~ id               = "some-id" -> (known after apply)

...
```

If these encryption settings are used by a cluster, and if attempting to `apply`, you get the following error:

```
Error: error deleting a encryptionAtRest (some-id): PATCH https://cloud.mongodb.com/api/atlas/v1.0/groups/some-id/encryptionAtRest: 409 (request "Conflict") Cannot disable Encryption at Rest on the group because it is still enabled on one or more clusters in the group.
```

This makes it pragmatically impossible to perform an encryption key rotation from this module without first disabling encryption at rest entirely.

coderGo93:
Fix for #80 
- Deleted force new for google cloud kms
- Added a validation for update google cloud kms
- Modified test

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
+++ b/mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
@@ -106,24 +106,20 @@ func resourceMongoDBAtlasEncryptionAtRest() *schema.Resource {
 			},
 			"google_cloud_kms": {
 				Type:     schema.TypeMap,
-				ForceNew: true,
 				Optional: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Type:     schema.TypeBool,
-							ForceNew: true,
 							Required: true,
 						},
 						"service_account_key": {
 							Type:      schema.TypeString,
-							ForceNew:  true,
 							Required:  true,
 							Sensitive: true,
 						},
 						"key_version_resource_id": {
 							Type:      schema.TypeString,
-							ForceNew:  true,
 							Required:  true,
 							Sensitive: true,
 						},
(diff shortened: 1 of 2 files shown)


themantissa (approved):
LGTM as far as I can tell if removing force new for Azure worked than this should work for GCP.   Approving but Paco should give this another review too before proceeding.

PacoDw (approved):
LGTM!

PR #214, opened 2020-04-30
fix: fixed DatabaseUserID to allows names with multiple dashes

### LINKED ISSUE #179: mongodbatlas_database_user can not be imported when they contain dashes "-" in the name
Hi,

we are trying to switch to this provider. While trying out the import for the mongodbatlas_database_user we found out that we can not import database_users which contain "-" in the username. This symbol is allowed in mongoDB and all our users contain dashes. :octopus: 

In https://github.com/terraform-providers/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_database_user.go line number 249 we saw that you are splitting the last information (project_id-username-auth_database_name) block from the import into 3 parts like this:

    parts := strings.SplitN(d.Id(), "-", 3)
	if len(parts) != 3 {
		r eturn nil, errors.New("import format error: to import a database user, use the format {project_id}-{username}-{auth_database_name}")
	}

	projectID := parts[0]
	username := parts[1]
	authDatabaseName := parts[2]

it would be awesome if you could implement it so that we can put for example quotes around the username.

example: "super-cool-database-user"

    func SplitAtDash(s string) []string {
		res := []string{}
		var beg int
		var inString bool

		for i := 0; i < len(s); i++ {
		if s[i] == '-' && !inString {
			res = append(res, s[beg:i])
			beg = i+1
		} else if s[i] == '"' || s[i] == '\'' {
			if !inString {
				inString = true
			} else if i > 0 && s[i-1] != '\\' {
				inString = false
			}
		}
		}
		return append(res, s[beg:])
	}

	func main(){
		strs := []string{
			`12345678901234567890abc-"super-cool-database-user"-admin`,
			`12345678901234567890abc-'super-cool-database-user'-admin`,
			`12345678901234567890abc-super-cool-database-user-admin`,
		}
		for _, str := range strs {
		fmt.Println("input string:", str)
			res := SplitAtDash(str)
			for i := range res {
		fmt.Println(res[i])
			}
		}
	}

Thanks in advance

PacoDw:
Added
- Changes to fix the import function when the username has multiple dashes
- Test case validating the above fix

We can import it using multiple dashes e.g:

```bash
$ terraform import mongodbatlas_database_user.my_user 1112222b3bf99403840e8934-my-user-name-admin
```

closes #179

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_database_user_test.go
+++ b/mongodbatlas/resource_mongodbatlas_database_user_test.go
@@ -167,7 +167,7 @@ func TestAccResourceMongoDBAtlasDatabaseUser_WithLabels(t *testing.T) {
 func TestAccResourceMongoDBAtlasDatabaseUser_importBasic(t *testing.T) {
 	projectID := os.Getenv("MONGODB_ATLAS_PROJECT_ID")
 
-	username := fmt.Sprintf("test_username_%s", acctest.RandString(5))
+	username := fmt.Sprintf("test-username-%s", acctest.RandString(5))
 	resourceName := "mongodbatlas_database_user.test"
 
 	resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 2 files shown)


themantissa (approved):
LGTM (though travis failing again so please ensure tests pass).   Please ensure one other DoU developer reviews as well.

mongodbatlas/resource_mongodbatlas_database_user_test.go
```diff
@@ -167,7 +167,7 @@ func TestAccResourceMongoDBAtlasDatabaseUser_WithLabels(t *testing.T) {
 func TestAccResourceMongoDBAtlasDatabaseUser_importBasic(t *testing.T) {
 	projectID := os.Getenv("MONGODB_ATLAS_PROJECT_ID")
 
-	username := fmt.Sprintf("test_username_%s", acctest.RandString(5))
+	username := fmt.Sprintf("test-username-%s", acctest.RandString(5))
```
coderGo93, line 170: Would be better if you use a function made for this case?, like this
```go
username := acctest.RandomWithPrefix("tf-acc-test-username")
```

PacoDw: yes! thank you so much let me include it

coderGo93 (approved):
LGTM

PR #217, opened 2020-05-05
Update network_peering.html.markdown

themantissa:
remove un-needed parameter from the GCP example - checked at https://docs.atlas.mongodb.com/reference/api/vpc-create-peering-connection/ and https://github.com/terraform-providers/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_network_peering_test.go


themantissa:
@PacoDw can you check this?   I think there was an extra variable not needed in the GCP example so removed.  Thanks!

PacoDw (approved):
That's correct, thank you so much :)

PR #221, opened 2020-05-07
fixes #210: Issues with advanced_configuration section on mongodbatlas_cluster

### LINKED ISSUE #210: Issues with advanced_configuration section on mongodbatlas_cluster
Hi,

There are a few issues with the `advanced_configuration` section of `mongodbatlas_cluster` resource.

First, if I create a new cluster with this configuration:

```
advanced_configuration = {
    minimum_enabled_tls_protocol = "TLS1_2"
}
```

I would expect all settings to have default values expect the one I have defined. That means when calling `/clusters/{name}/processArgs` I should get this as a result:

```
{
    "failIndexKeyTooLong": true,
    "javascriptEnabled": true,
    "minimumEnabledTlsProtocol": "TLS1_2",
    "noTableScan": false,
    "oplogSizeMB": null,
    "sampleRefreshIntervalBIConnector": null,
    "sampleSizeBIConnector": null
}
```

But this is what I get instead:

```
{
    "failIndexKeyTooLong": false,
    "javascriptEnabled": false,
    "minimumEnabledTlsProtocol": "TLS1_2",
    "noTableScan": false,
    "oplogSizeMB": null,
    "sampleRefreshIntervalBIConnector": 0,
    "sampleSizeBIConnector": 0
}
```

The most important part here is that `failIndexKeyTooLong` and `javascriptEnabled` are `true` by default but they are `false`, even my configuration have not stated that I wanted to change those properties. **I believe they should stay with default values.** Also you can see that `sampleRefreshIntervalBIConnector` and `sampleSizeBIConnector` have values as `0` instead of `null`.

The next issue is the fact that even if I go and add the other flags such as:

```
advanced_configuration = {
    fail_index_key_too_long      = true
    javascript_enabled           = true
    minimum_enabled_tls_protocol = "TLS1_2"
}
```

Every time I run `terraform plan` the cluster resource always comes back with 1 change required. This is what appears on the change diff:

```

advanced_configuration       = {
         "fail_index_key_too_long"              = "true"
         "javascript_enabled"                   = "true"
         "minimum_enabled_tls_protocol"         = "TLS1_2"
       - "no_table_scan"                        = "false" -> null
       - "oplog_size_mb"                        = "" -> null
       - "sample_refresh_interval_bi_connector" = "0" -> null
       - "sample_size_bi_connector"             = "0" -> null
}
```

As you can see the properties that were omitted are always coming back as a difference. **I think should not be the case.** The only real option here is to have to define ALL variables such as:

```
advanced_configuration = {
    fail_index_key_too_long              = true
    javascript_enabled                   = true
    minimum_enabled_tls_protocol         = "TLS1_2"
    no_table_scan                        = false
    oplog_size_mb                        = ""
    sample_refresh_interval_bi_connector = "0"
    sample_size_bi_connector             = "0"
  }
```

But is not what the documentation says... besides, what would happen and a new property is created? The issues with `""`/ `0` versus `null` are also worth looking into fixing...

Thanks.

gmlp:
## Description
fixes #210 Issues with advanced_configuration section on mongodbatlas_cluster

Link to any related issue(s): 
- https://github.com/terraform-providers/terraform-provider-mongodbatlas/issues/210

## Type of change:

## Required Checklist:

- [x] I have run make Fmt and formatted my code

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -44,6 +44,14 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: resourceMongoDBAtlasClusterImportState,
 		},
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceMongoDBAtlasClusterResourceV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceMongoDBAtlasClusterStateUpgradeV0,
+				Version: 0,
+			},
+		},
 		Schema: map[string]*schema.Schema{
 			"project_id": {
 				Type:     schema.TypeString,
@@ -287,9 +295,10 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Computed: true,
 			},
 			"advanced_configuration": {
-				Type:     schema.TypeMap,
+				Type:     schema.TypeList,
 				Optional: true,
 				Computed: true,
+				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"fail_index_key_too_long": {
@@ -427,7 +436,7 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 
 	// We need to validate the oplog_size_mb attr of the advanced configuration option to show the error
 	// before that the cluster is created
-	if oplogSizeMB, ok := d.GetOk("advanced_configuration.oplog_size_mb"); ok {
+	if oplogSizeMB, ok := d.GetOk("advanced_configuration.0.oplog_size_mb"); ok {
 		if cast.ToInt64(oplogSizeMB) <= 0 {
 			return fmt.Errorf("`advanced_configuration.oplog_size_mb` cannot be <= 0")
 		}
@@ -505,12 +514,15 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 		the advanced configuration option to attach it
 	*/
 	ac, ok := d.GetOk("advanced_configuration")
-	advancedConfReq := expandProcessArgs(ac.(map[string]interface{}))
-	if ok {
-		_, _, err := conn.Clusters.UpdateProcessArgs(context.Background(), projectID, cluster.Name, advancedConfReq)
-		if err != nil {
-			return fmt.Errorf(errorAdvancedConfUpdate, cluster.Name, err)
+	if aclist, ok1 := ac.([]interface{}); ok1 && len(aclist) > 0 {
+		advancedConfReq := expandProcessArgs(d, aclist[0].(map[string]interface{}))
+		if ok {
+			_, _, err := conn.Clusters.UpdateProcessArgs(context.Background(), projectID, cluster.Name, advancedConfReq)
+			if err != nil {
+				return fmt.Errorf(errorAdvancedConfUpdate, cluster.Name, err)
+			}
 		}
+
 	}
 
 	d.SetId(encodeStateID(map[string]string{
@@ -756,12 +768,14 @@ func resourceMongoDBAtlasClusterUpdate(d *schema.ResourceData, meta interface{})
 		Check if advaced configuration option has a changes to update it
 	*/
 	if d.HasChange("advanced_configuration") {
-		advancedConfReq := expandProcessArgs(d.Get("advanced_configuration").(map[string]interface{}))
-
-		if !reflect.DeepEqual(advancedConfReq, matlas.ProcessArgs{}) {
-			_, _, err := conn.Clusters.UpdateProcessArgs(context.Background(), projectID, clusterName, advancedConfReq)
-			if err != nil {
-				return fmt.Errorf(errorAdvancedConfUpdate, clusterName, err)
+		ac := d.Get("advanced_configuration")
+		if aclist, ok1 := ac.([]interface{}); ok1 && len(aclist) > 0 {
+			advancedConfReq := expandProcessArgs(d, aclist[0].(map[string]interface{}))
+			if !reflect.DeepEqual(advancedConfReq, matlas.ProcessArgs{}) {
+				_, _, err := conn.Clusters.UpdateProcessArgs(context.Background(), projectID, clusterName, advancedConfReq)
+				if err != nil {
+					return fmt.Errorf(errorAdvancedConfUpdate, clusterName, err)
+				}
 			}
 		}
 	}
@@ -1004,32 +1018,47 @@ func flattenRegionsConfig(regionsConfig map[string]matlas.RegionsConfig) []map[s
 	return regions
 }
 
-func expandProcessArgs(p map[string]interface{}) *matlas.ProcessArgs {
-	res := &matlas.ProcessArgs{
-		FailIndexKeyTooLong:              pointy.Bool(cast.ToBool(p["fail_index_key_too_long"])),
-		JavascriptEnabled:                pointy.Bool(cast.ToBool(p["javascript_enabled"])),
-		MinimumEnabledTLSProtocol:        cast.ToString(p["minimum_enabled_tls_protocol"]),
-		NoTableScan:                      pointy.Bool(cast.ToBool(p["no_table_scan"])),
-		SampleSizeBIConnector:            pointy.Int64(cast.ToInt64(p["sample_size_bi_connector"])),
-		SampleRefreshIntervalBIConnector: pointy.Int64(cast.ToInt64(p["sample_refresh_interval_bi_connector"])),
+func expandProcessArgs(d *schema.ResourceData, p map[string]interface{}) *matlas.ProcessArgs {
+	res := &matlas.ProcessArgs{}
+	if _, ok := d.GetOk("advanced_configuration.0.fail_index_key_too_long"); ok {
+		res.FailIndexKeyTooLong = pointy.Bool(cast.ToBool(p["fail_index_key_too_long"]))
 	}
-	if sizeMB := cast.ToInt64(p["oplog_size_mb"]); sizeMB != 0 {
-		res.OplogSizeMB = pointy.Int64(cast.ToInt64(p["oplog_size_mb"]))
-	} else {
-		log.Printf(errorClusterSetting, `oplog_size_mb`, "", cast.ToString(sizeMB))
+	if _, ok := d.GetOk("advanced_configuration.0.javascript_enabled"); ok {
+		res.JavascriptEnabled = pointy.Bool(cast.ToBool(p["javascript_enabled"]))
+	}
+	if _, ok := d.GetOk("advanced_configuration.0.minimum_enabled_tls_protocol"); ok {
+		res.MinimumEnabledTLSProtocol = cast.ToString(p["minimum_enabled_tls_protocol"])
+	}
+	if _, ok := d.GetOk("advanced_configuration.0.no_table_scan"); ok {
+		res.NoTableScan = pointy.Bool(cast.ToBool(p["no_table_scan"]))
+	}
+	if _, ok := d.GetOk("advanced_configuration.0.sample_size_bi_connector"); ok {
+		res.SampleSizeBIConnector = pointy.Int64(cast.ToInt64(p["sample_size_bi_connector"]))
+	}
+	if _, ok := d.GetOk("advanced_configuration.0.sample_refresh_interval_bi_connector"); ok {
+		res.SampleRefreshIntervalBIConnector = pointy.Int64(cast.ToInt64(p["sample_refresh_interval_bi_connector"]))
+	}
+	if _, ok := d.GetOk("advanced_configuration.0.oplog_size_mb"); ok {
+		if sizeMB := cast.ToInt64(p["oplog_size_mb"]); sizeMB != 0 {
+			res.OplogSizeMB = pointy.Int64(cast.ToInt64(p["oplog_size_mb"]))
+		} else {
+			log.Printf(errorClusterSetting, `oplog_size_mb`, "", cast.ToString(sizeMB))
+		}
 	}
 	return res
 }
 
-func flattenProcessArgs(p *matlas.ProcessArgs) map[string]interface{} {
-	return map[string]interface{}{
-		"fail_index_key_too_long":              cast.ToString(*p.FailIndexKeyTooLong),
-		"javascript_enabled":                   cast.ToString(*p.JavascriptEnabled),
-		"minimum_enabled_tls_protocol":         cast.ToString(p.MinimumEnabledTLSProtocol),
-		"no_table_scan":                        cast.ToString(*p.NoTableScan),
-		"oplog_size_mb":                        cast.ToString(p.OplogSizeMB),
-		"sample_size_bi_connector":             cast.ToString(p.SampleSizeBIConnector),
-		"sample_refresh_interval_bi_connector": cast.ToString(p.SampleRefreshIntervalBIConnector),
+func flattenProcessArgs(p *matlas.ProcessArgs) []interface{} {
+	return []interface{}{
+		map[string]interface{}{
+			"fail_index_key_too_long":              *p.FailIndexKeyTooLong,
+			"javascript_enabled":                   *p.JavascriptEnabled,
+			"minimum_enabled_tls_protocol":         p.MinimumEnabledTLSProtocol,
+			"no_table_scan":                        *p.NoTableScan,
+			"oplog_size_mb":                        p.OplogSizeMB,
+			"sample_size_bi_connector":             p.SampleSizeBIConnector,
+			"sample_refresh_interval_bi_connector": p.SampleRefreshIntervalBIConnector,
+		},
 	}
 }
 
--- a/mongodbatlas/resource_mongodbatlas_cluster_migrate.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster_migrate.go
@@ -329,11 +329,16 @@ func resourceMongoDBAtlasClusterResourceV0() *schema.Resource {
 
 func resourceMongoDBAtlasClusterStateUpgradeV0(rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
 	log.Println("[INFO] Found MongoDB Cluser state v0; migrating to v1")
+	rawState = migrateAdvancedConfiguration(rawState)
+
+	return rawState, nil
+}
+
+func migrateAdvancedConfiguration(rawState map[string]interface{}) map[string]interface{} {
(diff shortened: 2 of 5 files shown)


michelzanini:
Let me know when you have this work completed and I can test on my side before merging. Thanks.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
```diff
@@ -0,0 +1,343 @@
...
+		},
+	}
+}
+
+func resourceMongoDBAtlasClusterStateUpgradeV0(rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+
+	log.Println("[INFO] Found MongoDB Cluser state v0; migrating to v1")
+	log.Printf("%v", rawState["advanced_configuration"])
+
+	//rawState["advanced_configuration"] = []interface{}{}
```
PacoDw: could you remove the logs and the comment plz?

PacoDw (changes requested):

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
gmlp: sure, I just leave the first lo message is important

marinsalinas: Info Is OK

PacoDw:
@gmlp we need to add the changes in the documentation and describe the syntax change

michelzanini:
Hi @gmlp ,

I have checked out this branch to test it and, for a new cluster in a new project, it all worked the way I expected. 

This is the code I had before:
```
advanced_configuration = {
    minimum_enabled_tls_protocol = "TLS1_2"
}
```

And now I changed to:

```
advanced_configuration {
    minimum_enabled_tls_protocol = "TLS1_2"
}
```

And it has worked. The `plan` commands started to understand what actually changed and they don't touch the default values of the other configuration parameters. 

So it's all fine. Thanks.

However, if I do a `plan` on a cluster that existed before, where the state has been created by the previous `0.5.1` version, then I get an error every time I run `plan`:

```
Error: a number is required
```

This is what I get. Can you test creating a cluster with `0.5.1` and then switching to your branch and try? 

Thanks.

gmlp:
Hi @michelzanini, 
Thank you to test this out.
I'm working on creating a testing case for state migration and  I'm going to create a test case for that too.

gmlp:
Hi @michelzanini ,

I can't replicate this issue.


**Creatign cluster with provider version v0.5.1** 
main.tf
```
  project_id                   = var.project_id
  name                         = "testAcc-gmlp"
  disk_size_gb                 = 10

  // Provider Settings "block"

  advanced_configuration  = {
                 fail_index_key_too_long              = true
		javascript_enabled                   = true
                minimum_enabled_tls_protocol         = "TLS_1_2"
		no_table_scan                        = false
		oplog_size_mb                        = 1000
		sample_size_bi_connector			       = 110
		sample_refresh_interval_bi_connector = 310
  }
}
```
Advaced_config state:
```
         ...
          "schema_version": 0,
          "attributes": {
            "advanced_configuration": {
              "fail_index_key_too_long": "true",
              "javascript_enabled": "true",
              "minimum_enabled_tls_protocol": "TLS1_2",
              "no_table_scan": "false",
              "oplog_size_mb": "1000",
              "sample_refresh_interval_bi_connector": "310",
              "sample_size_bi_connector": "110"
            },
            "auto_scaling_disk_gb_enabled": false,
          ...
```
**Update cluster using this version with the fix**
main.tf
```
  project_id                   = var.project_id
  name                         = "testAcc-gmlp"
  disk_size_gb                 = 10

  // Provider Settings "block"

  advanced_configuration  {
    fail_index_key_too_long              = true
		javascript_enabled                   = true
    minimum_enabled_tls_protocol         = "TLS_1_1"
		no_table_scan                        = false
		oplog_size_mb                        = 1000
		sample_size_bi_connector			       = 110
		sample_refresh_interval_bi_connector = 310
  }
}
```
change:                                                                                                                                                                                                            
```
     ...
      ~ advanced_configuration {                                                                                                                                                                                 
            fail_index_key_too_long              = true                                                                                                                                                          
            javascript_enabled                   = true                                                                                                                                                          
          ~ minimum_enabled_tls_protocol         = "TLS1_2" -> "TLS_1_1"                                                                                                                                         
            no_table_scan                        = false                                                                                                                                                         
            oplog_size_mb                        = 1000                                                                                                                                                          
            sample_refresh_interval_bi_connector = 310                                                                                                                                                           
            sample_size_bi_connector             = 110                                                                                                                                                           
        }     
       ...
```     
```
...
"schema_version": 1,
          "attributes": {
            "advanced_configuration": [
              {
                "fail_index_key_too_long": true,
                "javascript_enabled": true,
                "minimum_enabled_tls_protocol": "TLS1_1",
                "no_table_scan": false,
                "oplog_size_mb": 1000,
                "sample_refresh_interval_bi_connector": 310,
                "sample_size_bi_connector": 110
              }
            ],
            "auto_scaling_disk_gb_enabled": false,
             ...
```

website/docs/r/cluster.html.markdown
```diff
@@ -307,11 +307,13 @@ Physical location of the region.
 
 -> **NOTE:** Prior to setting these options please ensure you read https://docs.atlas.mongodb.com/cluster-config/additional-options/.
 
+-> **NOTE:** This argument has been changed to type list make sure you have the proper syntax.
```
marinsalinas: Just add a note that it is a list with only one element (MaxItems: 1).

marinsalinas (changes requested):

website/docs/r/cluster.html.markdown
gmlp: done!

themantissa:
@gmlp - I think to reproduce you need to **not** include the advanced_configuration section in your .tf, create with 0.5.1 and then refresh the state.    If you include it then you've given the defaults. The complexity occurs in that there are defaults that are in Atlas and pulled into state even if the users doesn't include them in the .tf Then users may want to edit a default parameter without knowing/editing the others.

michelzanini:
I tried once more and got the same error.

I think you have to start with less config, like this:

```
advanced_configuration = {
    minimum_enabled_tls_protocol = "TLS1_2"
}
```

Then the state will be:

```
advanced_configuration       = {
        "fail_index_key_too_long"              = "false"
        "javascript_enabled"                   = "false"
        "minimum_enabled_tls_protocol"         = "TLS1_2"
        "no_table_scan"                        = "false"
        "oplog_size_mb"                        = ""
        "sample_refresh_interval_bi_connector" = "0"
        "sample_size_bi_connector"             = "0"
}
```

Then when I switch to use the binary from this branch and fixed the config as :

```
advanced_configuration {
    minimum_enabled_tls_protocol = "TLS1_2"
}
```

Then `terraform plan` starts to fail with immediately after on first run:

```
Error: a number is required
```

I tried running with TG_LOG but could NOT see anything useful to add here...

gmlp:
@michelzanini, @themantissa, I replicated this issue with the state migration, should work now. I created a new test for this scenario https://github.com/terraform-providers/terraform-provider-mongodbatlas/pull/221/commits/4259166edaabd005fb60165a7fe2fa17df1dea48
and this is the fix: https://github.com/terraform-providers/terraform-provider-mongodbatlas/pull/221/commits/82708f041cb10930e79c8730b3a42a56f1d013ec

michelzanini:
I tested again and works now :)
Thanks.

PacoDw (approved):
LGTM! thank you so much 👍

marinsalinas (approved):
Just check why Travis is failing otherwise LGTM!

website/docs/r/cluster.html.markdown
```diff
@@ -307,11 +307,13 @@ Physical location of the region.
 
 -> **NOTE:** Prior to setting these options please ensure you read https://docs.atlas.mongodb.com/cluster-config/additional-options/.
 
+-> **NOTE:** This argument has been changed to type list make sure you have the proper syntax. The list can have only one  item maximum.
```
themantissa: Just a small edit to make this a bit clearer for when we release: 
**_EDITED 5/21 again based on feedback below:_**
This argument's syntax has been changed as of version 0.6.0.  Make sure you have the new syntax in your configuration as show below (remove the previously required = sign).  

@gmlp does that make sense to you?  I get that it's a list but here I think that's confusing to mention?

themantissa: Also checked the cluster data source and the advanced config items never got added so no work now but will add this task to the back, fyi.

themantissa (changes requested):
Really nice work - a complicated fix to get this back on track!  Just one small doc change to make it a bit clearer and we should be good.  Also, @shum would you mind giving this a once over since it will require a state migration want to ensure triple check.

website/docs/r/cluster.html.markdown
michelzanini: When I read `The list can have only one  item maximum` I confused with not being able to do this:

```
advanced_configuration  {
    fail_index_key_too_long            = true
    javascript_enabled                   = true
}
```

But actually this is ok and what is not allowed is this:

```
advanced_configuration  {
    fail_index_key_too_long            = true
}
advanced_configuration  {
    javascript_enabled                   = true
}
```

Is this understanding correct?

If yes, then I suggest maybe a more clear sentence, for example, `You should only define a single advanced_configuration block` (or something like that)...

themantissa: Good point - it's literally the different between having an equal sign before:
advanced_configuration  = { 
and having none now
advanced_configuration  {   

So definitely want to change it if not clear.  I will edit my comment to make it clearer.

gmlp: @themantissa, yup, it makes sense.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
```diff
@@ -0,0 +1,368 @@
...
+		if len(advancedConfig) != 0 {
+			if advancedConfig["oplog_size_mb"] == "" {
+				advancedConfig["oplog_size_mb"] = "0"
+			}
+			if advancedConfig["sample_refresh_interval_bi_connector"] == "" {
+				advancedConfig["sample_refresh_interval_bi_connector"] = "0"
+			}
+			if advancedConfig["sample_size_bi_connector"] == "" {
+				advancedConfig["sample_size_bi_connector"] = "0"
+			}
```
shum, line 349: @gmlp Could you explain what this part of the migration is doing? This look potentially dangerous.

themantissa (approved):
LGTM but pending on @shum review due to complexity of state migration.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
gmlp: @shum, This part is for clusters that were created using v0.5.1 without advanced_config argument. The state will be equal to  https://github.com/terraform-providers/terraform-provider-mongodbatlas/commit/4259166edaabd005fb60165a7fe2fa17df1dea48#diff-92bb51ed35b9397957bd08d5cd49ed0bR119-R121 in order to make it compatible with the new schema. we need to parse this empty string to 0.

gmlp: @marinsalinas , @PacoDw any thoughts here guys?

themantissa: @shum is the concern in passing 0 to the Atlas API and potentially changing the default values since all of these are integers?   (https://docs.atlas.mongodb.com/reference/api/clusters-modify-advanced-configuration-options/)

shum: If I'm understanding correctly however, this means that any cluster created without advanced_config args would now be migrated to have oplog size of 0 because it was previously an empty string? I think this is potentially incorrect.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -427,7 +436,7 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 
 	// We need to validate the oplog_size_mb attr of the advanced configuration option to show the error
 	// before that the cluster is created
-	if oplogSizeMB, ok := d.GetOk("advanced_configuration.oplog_size_mb"); ok {
+	if oplogSizeMB, ok := d.GetOk("advanced_configuration.0.oplog_size_mb"); ok {
```
shum, line 439: Is this list always guaranteed to be length 1? Want to make sure we don't get any null pointers or index out of bounds.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
shum: So we're fetching `advanced_configuration` from `rawState`, then if `oplog_size_mb` wasn't explicitly set by the user ie `""`, then we default it to 0? Is there a way where we continue to leave it out and just assume Atlas defaults if it's not provided?

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: @gmlp can you comment on this?

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
themantissa: @shum this is such a tricky one - when one creates a cluster in Atlas you can't send the advanced configs at the same time to the API.  They are only modifiable AFTER the cluster is created (https://docs.atlas.mongodb.com/reference/api/clusters-modify-advanced-configuration-options/).  So if we want users to be able to configure both the cluster and the advanced configuration options at the same time (like they can in the UI) via Terraform it's tricky.   The best way would likely be to create the cluster and then populate the state after the cluster is created by pulling the defaults from Atlas (https://docs.atlas.mongodb.com/reference/api/clusters-get-advanced-configuration-options/) but if the users doesn't have these defined in their TF file it's then something in state that doesn't really match the tf config. We sorta did this for the snapshot backup policy and it's a bit awkward due to that (because it always sets a default policy but again we have a total mismatch from the UI so it's confusing - we have an issue on it actually).   

This is why I wanted your review here.  I'm very worried we will inadvertently cause an issue due to the complexity of not really being having the full CRUD life-cycle available.   Perhaps we should do a test where a user makes a change to a cluster where these are set to zero and see if the terraform plan sees it as a change to apply.  That would signal a big issue.

themantissa (changes requested):
Holding off on issues Chris noted till we ensure we are good on this one due to the potential impact of issues with this.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
gmlp: Hi @themantissa,@shum, I've been doing some research and testing. I figure out that this can improve and I can skip these conditions. I'll open a new PR with the improvement.

mongodbatlas/resource_mongodbatlas_cluster.go
gmlp: we have the Max Items =1 with this we make sure than more that maximum 1 element can be passed. https://github.com/terraform-providers/terraform-provider-mongodbatlas/pull/221/files#diff-104ff45c049e7dbfb15899b74adb243dL292-L293 and with GetOk method, we ensure that the element 0 exists. I can tweak this to validate if the element exists.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
themantissa: @gmlp thank you so much! Will keep an eye out.

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: @shum should we validate in your opinion?  Thanks to you both.

shum: I think GetOk is sufficient as it already checks "whether or not the key has been set to a non-zero value". LGTM.

mongodbatlas/resource_mongodbatlas_cluster_migrate.go
shum: @gmlp Let me know how you want to proceed with this PR and if we're excluding the migration step from this PR.

michelzanini: We can't just exclude the conversion between empty string and `0` otherwise we will get the `Error: a number is required` again. See the messages on this PR where that's described. Unless this code is removed and something else is changed together.

marinsalinas: Can't we change this something like this to avoid unwanted empty values?

*Disclaimer*: I just wrote this on the fly maybe I have some typos, just I want to show my thoughts on hoe to handle the empty strings.
```go
func migrateAdvancedConfiguration(rawState map[string]interface{}) map[string]interface{} {
	if v, ok := rawState["advanced_configuration"]; ok || v != nil {
		advancedConfig := v.(map[string]interface{})
		if len(advancedConfig) != 0 {

			advaceConfigMap := map[string]interface{}{
				"fail_index_key_too_long":              advancedConfig["fail_index_key_too_long"],
				"javascript_enabled":                   advancedConfig["javascript_enabled"],
				"minimum_enabled_tls_protocol":         advancedConfig["minimum_enabled_tls_protocol"],
				"no_table_scan":                        advancedConfig["no_table_scan"],
			},

			if v1, ok := advancedConfig["oplog_size_mb"]; ok, v1!= "" {
				advaceConfigMap["oplog_size_mb"] = v["oplog_size_mb"].(string)
			}
			if v1, ok := advancedConfig["sample_refresh_interval_bi_connector"]; ok, v1!= ""  {
				advaceConfigMap["sample_refresh_interval_bi_connector"] =  v["sample_refresh_interval_bi_connector"].(string)
			}
			if v1, ok := advancedConfig["sample_size_bi_connector"]; ok, v1!= ""  {
				advaceConfigMap["sample_size_bi_connector"] =  v["sample_size_bi_connector"].(string)
			}

			rawState["advanced_configuration"] = []interface{}{
				advaceConfigMap,
			}

		} else {
			rawState["advanced_configuration"] = []interface{}{}
		}
	} else {
		rawState["advanced_configuration"] = []interface{}{}
	}
	return rawState
}
```

shum: Thanks @marinsalinas, agree that something like this should work as well. Let me know how you and @gmlp want to proceed.

gmlp: Hi @themantissa, @shum the new pull request with option 2 is: https://github.com/terraform-providers/terraform-provider-mongodbatlas/pull/238

themantissa:
Should we close this @gmlp since we fixed with another PR?

gmlp:
replaced by https://github.com/terraform-providers/terraform-provider-mongodbatlas/pull/238

PR #233, opened 2020-05-20
Cluster autoscaling

coderGo93:
Added new fields for cluster about autoscaling
  - `auto_scaling_compute_enabled`
  - `auto_scaling_compute_scale_down_enabled`
  - `provider_auto_scaling_compute_min_instance_size`
  - `provider_auto_scaling_compute_max_instance_size`

Added a testacc for cluster using autoscaling

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net/http"
 	"reflect"
+	"regexp"
 	"time"
 
 	"strconv"
@@ -58,6 +59,16 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Type:     schema.TypeBool,
 				Default:  true,
 				Optional: true,
+			},
+			"auto_scaling_compute_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+			},
+			"auto_scaling_compute_scale_down_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
 			},
 			"backup_enabled": {
 				Type:     schema.TypeBool,
@@ -192,6 +203,16 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Computed: true,
 			},
 			"provider_volume_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"provider_auto_scaling_compute_min_instance_size": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"provider_auto_scaling_compute_max_instance_size": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
@@ -375,6 +396,43 @@ func resourceMongoDBAtlasClusterCreate(d *schema.ResourceData, meta interface{})
 		DiskGBEnabled: pointy.Bool(d.Get("auto_scaling_disk_gb_enabled").(bool)),
 	}
 
+	compute := &matlas.Compute{}
+
+	computeEnabled, computeEnabledOk := d.GetOk("auto_scaling_compute_enabled")
+	scaleDownEnabled, scaleDownEnabledOk := d.GetOk("auto_scaling_compute_scale_down_enabled")
+	minInstanceSize, minInstanceSizedOk := d.GetOk("provider_auto_scaling_compute_min_instance_size")
+	maxInstanceSize, maxInstanceSizeOk := d.GetOk("provider_auto_scaling_compute_max_instance_size")
+	if computeEnabledOk && (!maxInstanceSizeOk || maxInstanceSize.(string) == "") {
+		return fmt.Errorf("`provider_auto_scaling_compute_max_instance_size` must be set when `auto_scaling_compute_enabled` is set")
+	}
+	if !computeEnabledOk && (maxInstanceSizeOk || maxInstanceSize.(string) != "") {
+		return fmt.Errorf("`auto_scaling_compute_enabled` must be set when `provider_auto_scaling_compute_max_instance_size` is set")
+	}
+	if scaleDownEnabledOk && (!minInstanceSizedOk || minInstanceSize.(string) == "") {
+		return fmt.Errorf("`provider_auto_scaling_compute_min_instance_size` must be set when `auto_scaling_compute_scale_down_enabled` is set")
+	}
+	if (!computeEnabledOk || !scaleDownEnabledOk) && (minInstanceSizedOk || minInstanceSize.(string) != "") {
+		return fmt.Errorf("`auto_scaling_compute_enabled` and `auto_scaling_compute_scale_down_enabled` must be set when `provider_auto_scaling_compute_min_instance_size` is set")
+	}
+	if computeEnabledOk {
+		compute.Enabled = pointy.Bool(computeEnabled.(bool))
+	}
+	if scaleDownEnabledOk {
+		compute.ScaleDownEnabled = pointy.Bool(scaleDownEnabled.(bool))
+	}
+	autoScaling.Compute = compute
+
+	compute = &matlas.Compute{}
+	providerAutoScaling := &matlas.AutoScaling{}
+	if minInstanceSizedOk {
+		compute.MinInstanceSize = minInstanceSize.(string)
+		providerAutoScaling.Compute = compute
+	}
+	if maxInstanceSizeOk {
+		compute.MaxInstanceSize = maxInstanceSize.(string)
+		providerAutoScaling.Compute = compute
+	}
+
 	//validate cluster_type conditional
 	if _, ok := d.GetOk("replication_specs"); ok {
 		if _, ok1 := d.GetOk("cluster_type"); !ok1 {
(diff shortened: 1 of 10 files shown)


website/docs/d/clusters.html.markdown
```diff
@@ -99,6 +99,8 @@ In addition to all arguments above, the following attributes are exported:
 * `provider_encrypt_ebs_volume` - Indicates whether the Amazon EBS encryption is enabled. This feature encrypts the server’s root volume for both data at rest within the volume and data moving between the volume and the instance.
 * `provider_region_name` - Indicates Physical location of your MongoDB cluster. The region you choose can affect network latency for clients accessing your databases. Requires the Atlas Region name, see the reference list for [AWS](https://docs.atlas.mongodb.com/reference/amazon-aws/), [GCP](https://docs.atlas.mongodb.com/reference/google-gcp/), [Azure](https://docs.atlas.mongodb.com/reference/microsoft-azure/).
 * `provider_volume_type` - Indicates the type of the volume. The possible values are: `STANDARD` and `PROVISIONED`.
+* `provider_auto_scaling_compute_min_instance_size` - (Optional) Minimum instance size to which your cluster can automatically scale.
+* `provider_auto_scaling_compute_max_instance_size` - (Optional) Maximum instance size to which your cluster can automatically scale.
```
themantissa, line 105: missing the other two new values.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -694,6 +771,41 @@ func resourceMongoDBAtlasClusterUpdate(d *schema.ResourceData, meta interface{})
...
+	}
+	if d.HasChange("auto_scaling_compute_scale_down_enabled") {
+		compute.ScaleDownEnabled = pointy.Bool(d.Get("auto_scaling_compute_scale_down_enabled").(bool))
+		autoScaling.Compute = compute
+		cluster.AutoScaling = autoScaling
+	}
+	providerAutoScaling := &matlas.AutoScaling{}
+	compute = &matlas.Compute{}
+	regex := regexp.MustCompile("[0-9]+")
+	if d.HasChange("provider_auto_scaling_compute_min_instance_size") {
```
themantissa, line 792: Two questions,  why are we only checking if the min is > sizeName and sizeName > max on an update?  The same checks would apply if you set these upon cluster creation?   Did you test if submitting these without checking returned an error code/message from Atlas?

website/docs/r/cluster.html.markdown
```diff
@@ -206,6 +206,11 @@ Note free tier (M0) creation is not supported by the Atlas API and hence not sup
 * `auto_scaling_disk_gb_enabled` - (Optional) Specifies whether disk auto-scaling is enabled. The default is true.
     - Set to `true` to enable disk auto-scaling.
     - Set to `false` to disable disk auto-scaling.
+* `auto_scaling_compute_enabled` - (Optional) Specifies whether cluster tier auto-scaling is enabled. The default is false.
+    - Set to `true` to enable cluster tier auto-scaling. If enabled, you must specify a value for `providerSettings.autoScaling.compute.maxInstanceSize`.
+    - Set to `false` to disable cluster tier auto-scaling.
+* `auto_scaling_compute_scale_down_enabled` - (Optional) Set to `true` to enable the cluster tier to scale down. This option is only available if `autoScaling.compute.enabled` is `true`.
+    - If this option is enabled, you must specify a value for `providerSettings.autoScaling.compute.minInstanceSize`
```
themantissa: We need an important note in the doc that if auto_scaling_compute_enabled then Atlas will automatically scale up to the maximum provided and down to the minimum, if provided.  This will cause the value of instanceSizeName returned to potential be different than what is specified in the Terraform config and if one then applies a plan, not noting this, Terraform will scale the cluster back down to the original instanceSizeName value.  To prevent this a lifecycle customization should be used, i.e.:
lifecycle {
  ignore_changes = instanceSizeName
}

We should also test that a lifecycle customization works as expected, thought that will be tough to auto trigger a scaling event so one could test manually by running a terraform config, change the instance size in the UI to between the autoscale values and see if the lifecycle customerization works properly.

themantissa (changes requested):
A question on validation and a few changes/fixes to docs.  Thanks!!

website/docs/d/clusters.html.markdown
coderGo93: You're right, just added, thank you very much, melissa

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: Just added validation in create, thank you!!

website/docs/r/cluster.html.markdown
coderGo93: I pushed with note important about lifecycle and added a similar situation,its test, thank you so much, melissa !

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -1238,10 +1240,13 @@ func testAccMongoDBAtlasClusterConfigAWSWithAutoscaling(projectID, name, autoSca
 			provider_name               = "AWS"
 			provider_disk_iops 			    = 300
 			provider_encrypt_ebs_volume = false
-			provider_instance_size_name = "M30"
+			provider_instance_size_name = "%[8]s"
 			provider_region_name        = "EU_CENTRAL_1"
 			provider_auto_scaling_compute_min_instance_size = "%[5]s"
 			provider_auto_scaling_compute_max_instance_size = "%[6]s"
+			lifecycle{ // To simulate if there a new instance size name to avoid scale cluster down to original value
+				ignore_changes = [provider_instance_size_name]
```
themantissa, line 1248: I don't think this quite tests what would happen in reality with auto scaling but it does test if upgraded via terraform and adds a lifecycle event. If using autoscaling terraform wouldn't likely to be the tool setting the larger provider_instance_size_name, Atlas would when a scale event is triggered.  You would need to test this manually, I believe, to truly test if lifecycle works as we assume, basically:
-create a cluster with at least autoscaling on and a max value with terraform, example: an M10 with a max of a M40
-go into the Atlas UI and change the cluster size to M20 to simulate an auto scale event
-run a plan and see if terraform sees the size changed
-destroy all that and do again but
-create a cluster with at least autoscaling on and a max with terraform, example: an M10 with a max of a M40 with added the lifecycle ignore_changes in the config
-go into the Atlas UI and change the cluster size to M20 to simulate an auto scale event
-run a plan and see if terraform sees the size changed or if it properly ignores it.   

I'm not sure this is an automatable test unless instead of using the UI you triggered the cluster to scale up with the API directly, outside of terraform.    To actually trigger a scaling event you'd have to create a sustained load over a period of time.

website/docs/r/cluster.html.markdown
```diff
@@ -206,6 +206,22 @@ Note free tier (M0) creation is not supported by the Atlas API and hence not sup
...
+    - Set to `false` to disable cluster tier auto-scaling.
+    
+    -> **IMPORTANT:** If `auto_scaling_compute_enabled` is true,  then Atlas will automatically scale up to the maximum provided and down to the minimum, if provided.
+        This will cause the value of `provider_instance_size_name` returned to potential be different than what is specified in the Terraform config and if one then applies a plan, not noting this,
+         Terraform will scale the cluster back down to the original instanceSizeName value. 
+         To prevent this a lifecycle customization should be used, i.e.:
+        ```hcl
+          lifecycle {
+              ignore_changes = [provider_instance_size_name]
+          }
```
themantissa: Just a few more adjustments - move the important block above "* `auto_scaling_compute_enabled`" - line 209.

-> **IMPORTANT:** If `auto_scaling_compute_enabled` is true,  then Atlas will automatically scale up, if needed, to the maximum provided and down to the minimum if `auto_scaling_compute_scale_down_enabled` is also true. This will cause the value of `provider_instance_size_name` returned to potentially be different than what is specified in the Terraform config and if one then applies a plan, not noticing this, Terraform will scale the cluster back down to the original `provider_instance_size_name` value.   To prevent this a lifecycle customization should be used, i.e.:
        ```hcl
          lifecycle {
              ignore_changes = [provider_instance_size_name]
          }

themantissa (changes requested):
Thank you so much for the quick turnaround on the last review!  One more small doc adjustment.  I also provided some information about how to potentially really test this add and the lifecycle changes - I hope it is helpful.

mongodbatlas/resource_mongodbatlas_cluster_test.go
coderGo93: I see, I did try using TF directly and website UI as you suggested and it seems worked for me, I mean, at first it detected changes from website UI using refresh and plan and the second time used the lifecycle and ommited the change using refresh then apply

website/docs/r/cluster.html.markdown
coderGo93: done, thank you melissa

themantissa: Ty!!

mongodbatlas/resource_mongodbatlas_cluster_test.go
themantissa: That makes sense - just wanted to make sure we did a test to best emulate this.  Ty!

themantissa (approved):
LGTM - just needs a DoU review and we should be good!

PacoDw (approved):
LGTM!

Issue #243, opened 2020-06-14
Error when create or import cluster - panic: runtime error: invalid memory address or nil pointer dereference


afonsoaugusto:
Terraform version: 0.12.26  
Go runtime version: go1.12.13
found provider "terraform-provider-mongodbatlas_v0.6.0_x4"

```

Error: rpc error: code = Unavailable desc = transport is closing


panic: runtime error: invalid memory address or nil pointer dereference
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe0a598]
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4: 
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4: goroutine 41 [running]:
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4: github.com/terraform-providers/terraform-provider-mongodbatlas/mongodbatlas.flattenProcessArgs(...)
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4:  /opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-mongodbatlas/mongodbatlas/resource_mongodbatlas_cluster.go:1176
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4: github.com/terraform-providers/terraform-provider-mongodbatlas/mongodbatlas.resourceMongoDBAtlasClusterRead(0xc0002a8850, 0xfbd740, 0xc0004f4a00, 0xc0002a8850, 0x0)
2020-06-13T22:19:49.509-0300 [DEBUG] plugin.terraform-provider-mongodbatlas_v0.6.0_x4:  /opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-mongodbatlas/mongodbatlas/resource_mongodbatlas_cluster.go:732 +0x2f38
[...]

PR #255, opened 2020-07-01
Fix auto scaling attributes

PacoDw:
## Description

Refactored code implementing cleaner code.

Link to any related issue(s): #243 

## Type of change:

## Required Checklist:


go.mod
```diff
@@ -1,22 +1,16 @@
 module github.com/terraform-providers/terraform-provider-mongodbatlas
 
-go 1.13
+go 1.14
```
themantissa: Does the fix require the go version change?  That's a pretty big change to make if not required.  (and we should update the README from go 1.12 to 1.13 (that got missed somewhere)

themantissa (commented):
Question on go version change.

go.mod
themantissa: @PacoDw fyi^.  Let me know on this.

PacoDw: It isn't required, but it is the last golang version and I put it there to consider it what do you think @marinsalinas? and yes I will change the README file in this same PR, good catch 👍

WafflesMcDuff:
Confirmed locally. This fixes #243

go.mod
themantissa: I order to reduce complexity in troubleshooting PRs should be separated in my experience, the bug fix in it's own PR/version upgrades in another.  Otherwise if you have any side issues then one can't be sure what caused it when more than one factor changes.   Just keeps things more simple

PacoDw: Totally agree :)

marinsalinas (approved):

themantissa (approved):
LGTM!

Issue #222, opened 2020-05-08
Issues with resource and API design for Cloud Provider Snapshot Backup Policy


michelzanini:
Hi,

I am sorry but the description of this issue will be long.

I am having some difficulties configuring Cloud Provider Snapshot Backup Policy from Terraform.
The issues are more regarding how the Terraform resource has been designed rather than bugs.

When a cluster is first created, this is the default backup policy it gets:

![Screen Shot 2020-05-08 at 09 35 24](https://user-images.githubusercontent.com/4479787/81406752-8da31180-9110-11ea-9d74-403dbbbc864f.png)

What I really want to enable are `daily` and `weekly` backups. I do not need `hourly` and `monthly`. By using the UI here I could use the `DELETE` button to remove the `hourly` and `monthly` lines. But, I want to do it with automation, not manually.

With the release of Terraform provider `0.5.0` a new resource for backup policy was created and I immediately tried to use it.
This was my first attempt:

```
resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "backup_policy" {
  policies {
    id = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.id

    policy_item {
      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.0.id
      frequency_interval = 1
      frequency_type     = "daily"
      retention_value    = 3
      retention_unit     = "days"
    }
    policy_item {
      id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.1.id
      frequency_interval = 1
      frequency_type     = "weekly"
      retention_value    = 1
      retention_unit     = "weeks"
    }
  }
}
```

I am trying here to define only `daily` and `weekly` backups. The references to `mongodbatlas_cluster.my_cluster` you see here are to follow what the documentation indicates. Let's assume I have declared `mongodbatlas_cluster.my_cluster` somewhere else.

The problem here is that the first item on Atlas is a `hourly` backup (when a cluster is first created) and my first `policy_item` defined in the resource is a `daily` backup. This will cause a error `400 (request "Bad Request") The frequency type of a policy item must not be changed`. This happens because a `hourly` backup cannot be updated to a `daily` via API - the frequency type can't be changed via update.

[...]

PR #259, opened 2020-07-09
Updated Backup Policy documentation

PacoDw:
## Description

Updated backup_policy documentation explaining how we can implement at least one policy item or remove ones.

Link to any related issue(s): #222 

## Type of change:

## Required Checklist:

### DIFF
--- a/website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
+++ b/website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
@@ -12,9 +12,8 @@ description: |-
 
 -> **NOTE:** Groups and projects are synonymous terms. You may find `groupId` in the official documentation.
 
-
-# Examples Modifying Polices
-When Cloud Backup is enabled for a cluster MongoDB Atlas automatically creates a default Cloud Backup schedule for the cluster with four policy items; hourly, daily, weekly, and monthly. Because of this default creation this provider automatically saves the Cloud Backup Snapshot Policy into the Terraform state. If the default works well for you then you do not need to do anything other than create a cluster with Cloud Backup enabled and your Terraform state will have this information if you need it. However,  if you want the policy to be different than the default simply follow the next examples.
+# Examples - Modifying Polices
+When Cloud Backup is enabled for a cluster MongoDB Atlas automatically creates a default Cloud Backup schedule for the cluster with four policy items; hourly, daily, weekly, and monthly. Because of this default creation this provider automatically saves the Cloud Backup Snapshot Policy into the Terraform state when a cluster is created/modified to use Cloud Backup. If the default works well for you then you do not need to do anything other than create a cluster with Cloud Backup enabled and your Terraform state will have this information if you need it. However, if you want the policy to be different than the default we've provided some examples to help below.
 
 ## Example Usage - Create a Cluster and Modify the 4 Default Policies Simultaneously
 
@@ -41,7 +40,8 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
   reference_minute_of_hour = 45
   restore_window_days      = 4
 
-
+  //Keep all 4 default policies but modify the units and values
+  //Could also just reflect the policy defaults here for later management
   policies {
     id = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.id
 
@@ -80,49 +80,7 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
 }
 ```
 
-## Example Usage - Create a Cluster and change that clusters default backup policy in the same Terraform apply run
-
-```hcl
-resource "mongodbatlas_cluster" "my_cluster" {
-  project_id   = "<PROJECT-ID>"
-  name         = "clusterTest"
-  disk_size_gb = 5
-
-  //Provider Settings "block"
-  provider_name               = "AWS"
-  provider_region_name        = "EU_CENTRAL_1"
-  provider_instance_size_name = "M10"
-  provider_backup_enabled     = true // must be enabled in order to use cloud_provider_snapshot_backup_policy resource
-  provider_disk_iops          = 100
-  provider_encrypt_ebs_volume = false
-}
-
-resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
-  project_id   = mongodbatlas_cluster.my_cluster.project_id
-  cluster_name = mongodbatlas_cluster.my_cluster.name
-
-  reference_hour_of_day    = 3
-  reference_minute_of_hour = 45
-  restore_window_days      = 4
-
-
-  policies {
-    id = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.id
-
-    policy_item {
-      id                 = 5f0747cad187d8609a72f546
-      frequency_interval = 4
-      frequency_type     = "weekly"
-      retention_unit     = "days"
-      retention_value    = 3
-    }
-  }
-}
-```
-
--> **NOTE:** This is the id MongoDB Atlas returns for the policy item we want to keep. Here it is hard coded because you need to either use the actual value from the Terraform state or look to map the policy item you want to keep to it's placement in the state file array that was imported in when the cluster was originally created.
-
-~> **IMPORTANT:** For example in the state file we are using the weekly policy is the third policy in the array so it could be referred to with `mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.2.id` instead of the hard coded value, but it's not recommended when the cluster presents changes or make `terraform refresh` due to once this will applied, the cluster state will remove the rest of the items, so the posicion of the array will change to position 0.
+~> **IMPORTANT:**   `policies.#.policy_item.#.id` is obtained when the cluster is created. The example here shows the default order of the default policy when Cloud Backup is enabled (provider_backup_enabled is set to true).  The default policy is viewable in the Terraform State file.
 
 ## Example Usage - Create a Cluster and Modify 3 Default Policies and Remove 1 Default Policy Simultaneously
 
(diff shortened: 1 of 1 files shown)


website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -73,6 +73,47 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
 }
 ```
 
+## Example Usage implementing at least one Policy Item
```
themantissa: So this is taking me a while to review and after testing it all out I think we need 3 examples:

## Example Usage - Create a Cluster and Modify the 4 Default Policies Simultaneously
_This is the existing example we already had but just need to change the header_

## Example Usage - Create a Cluster and Modify 3 Default Policies and Remove 1 Default Policy Simultaneously
_This needs to be a new example, similar to the existing example that just removes one of the default policy blocks_

## Example Usage - Remove 3 Default Policies Items After the Cluster Has Already Been Created
_Same example that's already here._  
 
Each example also needs a note to explain what it is doing:
First example:
-> **NOTE:** When Cloud Backup is enabled for a cluster MongoDB Atlas automatically creates a default Cloud Backup schedule for the cluster with four policy items; hourly, daily, weekly, and monthly.   Because of this default creation this provider automatically saves the Cloud Backup Snapshot Policy into the Terraform state.  If the default works well for you then you do not need to do anything other than create a cluster with Cloud Backup enabled and your Terraform state will have this information if you need it. However,  if you want the policy to be different than the default when creating the cluster simply follow this example to modify the four default policies.

Second example:
-> **NOTE:** (See text in first example for more details on the default.) If you want the Cloud Backup Snapshot Policy to vary in the number of policies from the default when creating the cluster, perhaps you want to remove one policy item and modify the remaining three, simply follow this example here to remove a policy and modify three.

Third example:
-> **NOTE:** (See text in first example for more details on the default.) If you want the Cloud Backup Snapshot Policy to vary in number of policies for a cluster that was already created/imported, perhaps you want to remove three policy items and modify the remaining policy, simply follow the example here.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -73,6 +73,47 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
...
+
+```hcl
+resource "mongodbatlas_cluster" "my_cluster" {
```
themantissa: I think a comment here explaining "this cluster has already been created and is here as an example" would help.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -73,6 +73,47 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
...
+  reference_hour_of_day    = 3
+  reference_minute_of_hour = 45
+  restore_window_days      = 4
+
+
+  policies {
+    id = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.id
+
+    policy_item {
+      id                 = 5f0747cad187d8609a72f546
```
themantissa: This needs some explanation since it's a hard coded value.   Perhaps this would do it:

This is the id MongoDB Atlas returns for the policy item we want to keep.  Here it is hard coded because you need to either use the actual value from the Terraform state or look to map the policy item you want to keep to it's placement in the state file array that was imported in when the cluster was originally created.  For example in the state file we are using the hourly policy is the first policy in the array so it can be referred to with
mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.0.id instead of the hard coded value.  Summarized: to use the state file value for a policy item you need to determine the array placement # of the same frequency_type you want to keep.

themantissa (changes requested):
This helped me a great deal, as did running through the examples!   I've included quite a bit here that I think will further clear this up.  Let me know what you think.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
PacoDw: I like it it's a great idea. I will implement it, thanks. 👍

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
themantissa: This was resolved but it's still a hard coded value that would be hard for someone to understand where it comes from.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -72,6 +75,197 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
   }
 }
 ```
+# Examples Modifying Polices
+When Cloud Backup is enabled for a cluster MongoDB Atlas automatically creates a default Cloud Backup schedule for the cluster with four policy items; hourly, daily, weekly, and monthly. Because of this default creation this provider automatically saves the Cloud Backup Snapshot Policy into the Terraform state. If the default works well for you then you do not need to do anything other than create a cluster with Cloud Backup enabled and your Terraform state will have this information if you need it. However,  if you want the policy to be different than the default simply follow the next examples.
+
+## Example Usage - Create a Cluster and Modify the 4 Default Policies Simultaneously
+This cluster has already been created and is here as an example
```
themantissa: The headings and the examples are a bit mixed up.  Simultaneously means "at the same time" so this heading and example don't match.  The heading as shown here would mean "you create a cluster and change that clusters default backup policy in the same Terraform apply run" - that is the first example on lines 15-77.  So I think the solve here is to use this heading for the first example and then remove this one.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -72,6 +75,197 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
...
+    }
+
+    # Item removed
+    # policy_item {
+    #   id                 = mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.2.id
+    #   frequency_interval = 4
+    #   frequency_type     = "weekly"
+    #   retention_unit     = "weeks"
+    #   retention_value    = 3
+    # }
```
themantissa, line 137: This is a very good way to illustrate it!  Nice!

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -72,6 +75,197 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
   }
 }
 ```
+# Examples Modifying Polices
+When Cloud Backup is enabled for a cluster MongoDB Atlas automatically creates a default Cloud Backup schedule for the cluster with four policy items; hourly, daily, weekly, and monthly. Because of this default creation this provider automatically saves the Cloud Backup Snapshot Policy into the Terraform state. If the default works well for you then you do not need to do anything other than create a cluster with Cloud Backup enabled and your Terraform state will have this information if you need it. However,  if you want the policy to be different than the default simply follow the next examples.
```
themantissa: This should go to the very top as all of the examples have to do w/ modifying as the initial policy is created by default.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -72,6 +75,197 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
...
+      retention_unit     = "days"
+      retention_value    = 3
+    }
+  }
+}
+```
+
+-> **NOTE:** This is the id MongoDB Atlas returns for the policy item we want to keep. Here it is hard coded because you need to either use the actual value from the Terraform state or look to map the policy item you want to keep to it's placement in the state file array that was imported in when the cluster was originally created.
+
+Summarized: to use the state file value for a policy item you need to determine the array placement # of the same frequency_type you want to keep. With this Terraform configuration the anothers policy items will be removed.
```
themantissa: The NOTE and IMPORTANT both refer to line 256, the hard coded policy_item id - we have to explain to the user where the hard coded value comes from and how to figure this stuff out.  It's all probably best put in a comment in the example above the value.

website/docs/r/cloud_provider_snapshot_backup_policy.html.markdown
```diff
@@ -72,6 +75,197 @@ resource "mongodbatlas_cloud_provider_snapshot_backup_policy" "test" {
...
+      retention_unit     = "months"
+      retention_value    = 4
+    }
+  }
+}
+```
+
+-> **NOTE:** (See text in first example for more details on the default.) If you want the Cloud Backup Snapshot Policy to vary in number of policies for a cluster that was already created/imported, perhaps you want to remove three policy items and modify the remaining policy, simply follow the above example here.
+
+~> **IMPORTANT:** Note in this example we decided to remove the first 3 items so we can't use `mongodbatlas_cluster.my_cluster.snapshot_backup_policy.0.policies.0.policy_item.3.id` this sentence to retrieve the monthly id value of the cluster state due to once the cluster being modified or makes a `terraform refresh` will cause that the three items will remove from the state, so we will get an error due to the index 3 doesn't exists any more and our monthly policy item is moved to the first place of the array.(See text in the first example for more details on it)
```
themantissa: this should be under the heading, Example Usage - Remove 3 Default Policies Items After the Cluster Has Already Been Created, as it is explaining what the example is.

The IMPORTANT probably is best to put near 256 in a comment or use the other explanation cut and pasted in here.

themantissa (changes requested):
The examples are good but some of the placement of the provided info is not in the best place.  Let me know if this makes sense and I can perhaps submit a PR with the changes if that will help. 
Just FYI - I don't want this to hold up 0.6.2 but also want to get better examples out - we may need to do 0.6.2 first and then do a special push for just this after.

themantissa:
@PacoDw this is a little non-standard but I made the changes I think we need in the last commit.  Let me know if they make sense and then I can approve so we can move forward on closing these out.  Thanks!

PacoDw:
@themantissa no problem it's okay 👍

themantissa (approved):
LGTM (since made updates in last commit I think we still need).  Let's wrap this up.

coderGo93 (approved):
LGTM

PR #264, opened 2020-07-15
Added a field AwsIAMType for database user

coderGo93:
## Description

- Added a field for database user called `awsIAMType` in resource and datasource(s) and its test
- Convert to url query escape for `username` when it uses `GET` method to avoid special characters issues.

Link to any related issue(s): 

## Type of change:

## Required Checklist:

## Further comments
Client mongo needs to be updated(merged) before this PR so I can update the vendor.

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_database_user.go
+++ b/mongodbatlas/resource_mongodbatlas_database_user.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-	matlas "go.mongodb.org/atlas/mongodbatlas"
 	"net/url"
 	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	matlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
 func resourceMongoDBAtlasDatabaseUser() *schema.Resource {
--- a/mongodbatlas/resource_mongodbatlas_database_user.go
+++ b/mongodbatlas/resource_mongodbatlas_database_user.go
@@ -120,9 +120,9 @@ func resourceMongoDBAtlasDatabaseUserRead(d *schema.ResourceData, meta interface
 			authDatabaseName = d.Get("auth_database_name").(string)
 		}
 	}
-	dbUsername := url.QueryEscape(username)
+	usernameEscaped := url.QueryEscape(username)
 
-	dbUser, _, err := conn.DatabaseUsers.Get(context.Background(), authDatabaseName, projectID, dbUsername)
+	dbUser, _, err := conn.DatabaseUsers.Get(context.Background(), authDatabaseName, projectID, usernameEscaped)
 	if err != nil {
 		return fmt.Errorf("error getting database user information: %s", err)
 	}
@@ -146,7 +146,7 @@ func resourceMongoDBAtlasDatabaseUserRead(d *schema.ResourceData, meta interface
 	}
 
 	if err := d.Set("aws_iam_type", dbUser.AWSIAMType); err != nil {
-		return fmt.Errorf("error setting `x509_type` for database user (%s): %s", d.Id(), err)
+		return fmt.Errorf("error setting `aws_iam_type` for database user (%s): %s", d.Id(), err)
 	}
 
 	if err := d.Set("roles", flattenRoles(dbUser.Roles)); err != nil {
@@ -217,9 +217,9 @@ func resourceMongoDBAtlasDatabaseUserUpdate(d *schema.ResourceData, meta interfa
 	username := ids["username"]
 	authDatabaseName := ids["auth_database_name"]
 
-	dbUsername := url.QueryEscape(username)
+	usernameEscaped := url.QueryEscape(username)
 
-	dbUser, _, err := conn.DatabaseUsers.Get(context.Background(), authDatabaseName, projectID, dbUsername)
+	dbUser, _, err := conn.DatabaseUsers.Get(context.Background(), authDatabaseName, projectID, usernameEscaped)
 	if err != nil {
 		return fmt.Errorf("error getting database user information to update it: %s", err)
 	}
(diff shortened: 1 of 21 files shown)


mongodbatlas/resource_mongodbatlas_database_user.go
```diff
@@ -139,6 +145,10 @@ func resourceMongoDBAtlasDatabaseUserRead(d *schema.ResourceData, meta interface
 		return fmt.Errorf("error setting `x509_type` for database user (%s): %s", d.Id(), err)
 	}
 
+	if err := d.Set("aws_iam_type", dbUser.AWSIAMType); err != nil {
+		return fmt.Errorf("error setting `x509_type` for database user (%s): %s", d.Id(), err)
```
themantissa: Error message is for the wrong type.

mongodbatlas/resource_mongodbatlas_database_user.go
```diff
@@ -115,8 +120,9 @@ func resourceMongoDBAtlasDatabaseUserRead(d *schema.ResourceData, meta interface
 			authDatabaseName = d.Get("auth_database_name").(string)
 		}
 	}
+	dbUsername := url.QueryEscape(username)
```
themantissa: Perhaps a more informative name?  username_escaped or what ever best fits standard go naming?

Also I'm a bit worried about doing this across the board as we've been able to GET and DELETE with special characters in db names before (like a dash).  Let's make sure this doesn't cause any side impacts by being sure we test that thoroughly please.

mongodbatlas/resource_mongodbatlas_database_user_test.go
```diff
@@ -71,7 +73,34 @@ func TestAccResourceMongoDBAtlasDatabaseUser_withX509Type(t *testing.T) {
 					testAccCheckMongoDBAtlasDatabaseUserAttributes(&dbUser, username),
 					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
 					resource.TestCheckResourceAttr(resourceName, "username", username),
-					resource.TestCheckResourceAttr(resourceName, "x509_type", "MANAGED"),
+					resource.TestCheckResourceAttr(resourceName, "x509_type", "CUSTOMER"),
```
themantissa: I would not remove the MANAGED test but add a 2nd test that allows for the username type for CUSTOMER x509 types.  MANAGED is far more common so we should keep the test.

mongodbatlas/resource_mongodbatlas_database_user_test.go
```diff
@@ -342,7 +375,7 @@ func testAccMongoDBAtlasDatabaseUserWithX509TypeConfig(projectID, roleName, user
 	return fmt.Sprintf(`
 		resource "mongodbatlas_database_user" "test" {
 			username           = "%[3]s"
-			x509_type          = "MANAGED"
+			x509_type          = "CUSTOMER"
```
themantissa: Same as above, keep both.

website/docs/d/database_user.html.markdown
```diff
@@ -62,6 +62,7 @@ In addition to all arguments above, the following attributes are exported:
 * `id` - The database user's name.
 * `roles` - List of user’s roles and the databases / collections on which the roles apply. A role allows the user to perform particular actions on the specified database. A role on the admin database can include privileges that apply to the other databases as well. See [Roles](#roles) below for more details.
 * `x509_type` - X.509 method by which the provided username is authenticated.
+* `aws_iam_type` - The new database user authenticates with AWS IAM credentials.
```
themantissa: Edit - include what value may be here:
aws_iam_type - If set the database user authenticates with AWS IAM credentials.  Default is NONE, `USER` means user has AWS IAM user credentials, `ROLE` -  means user has credentials associated with an AWS IAM role.

website/docs/d/database_users.html.markdown
```diff
@@ -68,6 +68,7 @@ In addition to all arguments above, the following attributes are exported:
 * `auth_database_name` - The user’s authentication database. A user must provide both a username and authentication database to log into MongoDB. In Atlas deployments of MongoDB, the authentication database is always the admin database.
 
 * `x509_type` - X.509 method by which the provided username is authenticated.
+* `aws_iam_type` - The new database user authenticates with AWS IAM credentials.
```
themantissa: Same:
aws_iam_type - If set the database user authenticates with AWS IAM credentials.  Default is `NONE`, `USER` means user has AWS IAM user credentials, `ROLE` -  means user has credentials associated with an AWS IAM role.

themantissa (changes requested):
A few edits and request to not remove test but add.

mongodbatlas/resource_mongodbatlas_database_user.go
coderGo93: You're right, thank you so much

mongodbatlas/resource_mongodbatlas_database_user.go
coderGo93: Sure no problem, I have tested many times with testacc and all OK when creating, getting and deleting but will do more testing to be sure of course

mongodbatlas/resource_mongodbatlas_database_user_test.go
coderGo93: No problem, thank you melissa

website/docs/d/database_user.html.markdown
coderGo93: I'll add this part, thank you melissa

themantissa (approved):
Think you addressed all the comments! Thank you - LGTM

mongodbatlas/resource_mongodbatlas_database_user.go
```diff
@@ -53,6 +53,11 @@ func resourceMongoDBAtlasDatabaseUser() *schema.Resource {
 				Optional: true,
 				Default:  "NONE",
 			},
+			"aws_iam_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "NONE",
+			},
```
PacoDw:
```suggestion
			"aws_iam_type": {
				Type:     schema.TypeString,
				Optional: true,
				Default:  "NONE",
				ValidateFunc: validation.StringInSlice([]string{"NONE", "USER", "ROLE"}, false),
			},
```
if only these attribute must to be set in this attribute, you need to consider put this validation to ensure that the user write at least one of them.

PacoDw (changes requested):

mongodbatlas/resource_mongodbatlas_database_user.go
coderGo93: Thanks for the suggestion, paco, I added it and for x509 too

PacoDw (approved):
LGTM!

PR #268, opened 2020-07-21
feat: add a unique user agent

gssbzn:
## Description

We are trying to track usage of the API and right now is hard to distinguish from the TF provider and a vanilla go client usage.

I'm adding a user agent here to help us better track usage.

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/config.go
+++ b/mongodbatlas/config.go
@@ -26,5 +26,8 @@ func (c *Config) NewClient() interface{} {
 	client.Transport = logging.NewTransport("MongoDB Atlas", transport)
 
 	// Initialize the MongoDB Atlas API Client.
-	return matlasClient.NewClient(client)
+	atlasClient := matlasClient.NewClient(client)
+	atlasClient.UserAgent = "terraform-provider-mongodbatlas"
+
+	return atlasClient
 }


mongodbatlas/config.go
```diff
@@ -26,5 +26,8 @@ func (c *Config) NewClient() interface{} {
 	client.Transport = logging.NewTransport("MongoDB Atlas", transport)
 
 	// Initialize the MongoDB Atlas API Client.
-	return matlasClient.NewClient(client)
+	atlasClient := matlasClient.NewClient(client)
+	atlasClient.UserAgent = "terraform-provider-mongodbatlas"
```
MihaiBojin, line 30: it would be great if we could add the provider's version too!

gssbzn: the provider looks to be unversioned (no ldflags), I'll leave the maintainer to decide if we want to add this (in this PR or a follow up) or not

gssbzn:
Linting seems to be failing on travis, I had plans to fix lint installation later but Opening a PR now

gssbzn:
Opened #269 to fix linting

PacoDw (approved):
LGMT! Thanks

mongodbatlas/config.go
MihaiBojin: Updating the UA like so obscures the underlying go client's version.
I think `atlasClient.SetUserAgent(...)` is a better approach, but it's not a hard requirement from my point of view, since we'll version the TF provider soon and we'll be able to infer which go-client it's using.

MihaiBojin (approved):

mongodbatlas/config.go
themantissa: Ah, so this is kinda a weirdness of Terraform and how we ended up not tracking the version before.  Per https://www.terraform.io/docs/extend/how-terraform-works.html#plugin-locations "Provider and provisioner plugins can be installed in the same directories. Provider plugin binaries are named with the scheme terraform-provider-<NAME>_vX.Y.Z, while provisioner plugins use the scheme terraform-provisioner-<NAME>_vX.Y.Z. **Terraform relies on filenames to determine plugin types, names, and versions.**". So they are versioned by filename (I know).   I don't think there would be an issue versioning with ldflags (but leave it to the DoU team to chime in on this) once we move to our repo again and our own releases.  

So that was my long way of saying we can solve this post repo move (hopefully).

themantissa (approved):
LGTM!  Making note to solve versioning within the code base for tracking for later.

PR #293, opened 2020-08-13
GitHub actions tests

coderGo93:
## Description

Added a flag with env `SKIP_TEST_EXTERNAL_CREDENTIALS` where it's function is to skip acceptance tests that requires external credentials such as GCP,azure and aws

Added a github actions and has condition where it executes tests when the branch is master or is a pull request

Link to any related issue(s): 

## Type of change:

## Required Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirments


coderGo93:
Sorry for many commits, I tried many times since I cannot try locally without using third apps. Commits related to travis can be ignored

themantissa:
@coderGo93 just want to check - the tests errors will be fixed in another PR per our slack, correct?

coderGo93:
> @coderGo93 just want to check - the tests errors will be fixed in another PR per our slack, correct?

That's correct, @themantissa . I'll fix the errors that is not API related like unauthorized error or similar

.github/workflows/automated-test-acceptances.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          go-version: 1.14
+      - name: Acceptance Tests
+        env:
+          SKIP_TEST_EXTERNAL_CREDENTIALS: ${{ secrets.SKIP_TEST_EXTERNAL_CREDENTIALS }}
+          MONGODB_ATLAS_PUBLIC_KEY: ${{ secrets.MONGODB_ATLAS_PUBLIC_KEY }}
+          MONGODB_ATLAS_PRIVATE_KEY: ${{ secrets.MONGODB_ATLAS_PRIVATE_KEY }}
+          MONGODB_ATLAS_PROJECT_ID: ${{ secrets.MONGODB_ATLAS_PROJECT_ID }}
+          MONGODB_ATLAS_ORG_ID: ${{ secrets.MONGODB_ATLAS_ORG_ID }}
+          DB_USERNAME: ${{ secrets.DB_USERNAME }}
```
themantissa: This was not an env we created.  It looked like, from the errors this was only for x509?  Also, ideally any MongoDB Atlas  specific variable keeps the MONGODB_ATLAS_ prefix.

.github/workflows/automated-test-acceptances.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
+      - name: Acceptance Tests
+        env:
+          SKIP_TEST_EXTERNAL_CREDENTIALS: ${{ secrets.SKIP_TEST_EXTERNAL_CREDENTIALS }}
```
themantissa, line 21: This is just a work around so if we do not skip EXTERNAL_CREDENTIALS can we go ahead and setup the env to pick up the secrets?   We'll eventually have credentials to do the tests that require a cloud provider too.  Makes sense to prep for that now.  I think we have them all updated in the README.

.github/workflows/automated-test.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
+      - name: Code Lint
+        run: make tools && make lint
+      - name: Website
+        run: make tools && make website-lint
+      - name: Test
+        run: make test
```
themantissa: Linter picking up missing newlines^

mongodbatlas/data_source_mongodbatlas_custom_db_role_test.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 func TestAccDataSourceMongoDBAtlasCustomDBRole_basic(t *testing.T) {
+	SkipTestExtCred(t)
```
themantissa: This is weird - why does custom db role have an external credential?  Looking at the test it is it because the pre check has checkPeeringEnvAWS?  Why is that?

mongodbatlas/data_source_mongodbatlas_custom_db_roles_test.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 func TestAccDataSourceMongoDBAtlasCustomDBRoles_basic(t *testing.T) {
+	SkipTestExtCred(t)
```
themantissa: Same comment as above, this shoulnd't depend on peering to test.

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -305,6 +305,7 @@ func TestAccResourceMongoDBAtlasCluster_basicAzure(t *testing.T) {
 }
 
 func TestAccResourceMongoDBAtlasCluster_basicGCP(t *testing.T) {
+	SkipTestExtCred(t)
```
themantissa: Interesting - this is another one I'm surprised to see peering in the pre check.  If it's a basic GCP cluster I don't see why we'd have peering involved - i.e. TestAccResourceMongoDBAtlasCluster_withGCPNetworkPeering should cover it.

themantissa (changes requested):
Good progress!  A few questions/comments - let me know if you have questions.  Thanks!

.github/workflows/automated-test-acceptances.yml
coderGo93: Ok, the `DB_USERNAME` you meant, sure is for database user rule, you're right I'll add MONGODB_ATLAS_ at first

mongodbatlas/data_source_mongodbatlas_custom_db_role_test.go
PacoDw: Totally agree, @coderGo93 you need to remove the checkPeeringEnvAWS because these tests don't need AWS Credentials, so these tests don't be skipped

.github/workflows/automated-test-acceptances.yml
coderGo93: That's correct, by default the env `SKIP_TEST_EXTERNAL_CREDENTIALS` is false, so if you don't set the env if will run the test that requires external credentials, obviously if you don't set them it will appear error about credentials for aws/gcp/azure

mongodbatlas/data_source_mongodbatlas_custom_db_role_test.go
coderGo93: Understood, I only added the func that requires additional credentials in precheck or similar

.github/workflows/automated-test-acceptances.yml
themantissa: But won't we need a line to import them into the test env as we set them?   e.g.      AWS_VPC_CIDR_BLOCK: ${{ secrets.AWS_VPC_CIDR_BLOCK }}   Can't we pre-seed them all here for the default SKIP_TEST_EXTERNAL_CREDENTIALS=false but skip setting them in the env when SKIP_TEST_EXTERNAL_CREDENTIALS=true?

coderGo93: Oh, you're right, I need to add those env var as well, I misunderstood you at first. Thank you for that, melissa

.travis.yml
```diff
@@ -17,4 +17,6 @@ matrix:
       script:
         #- make website-test #commented until Terraform Provider Program Validation
         - make website-lint
-script: make test
+    - name: "Test"
+      script:
+        - make test
```
themantissa: This still shows no newline at the end of the file, it passed the linter but should probably fix it up.

mongodbatlas/data_source_mongodbatlas_maintenance_window_test.go
```diff
@@ -34,21 +34,19 @@ func TestAccDataSourceMongoDBAtlasMaintenanceWindow_basic(t *testing.T) {
...
-	projectID := os.Getenv("MONGODB_ATLAS_PROJECT_ID")
+	t.Skip()
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
-		Steps: []resource.TestStep{
-			{
+		Steps:     []resource.TestStep{
+			/*{
```
themantissa: Why this so much in this test commented out?

mongodbatlas/resource_mongodbatlas_network_container_test.go
```diff
@@ -245,7 +245,7 @@ func testAccMongoDBAtlasNetworkContainerConfigAzure(projectID, cidrBlock, provid
 			project_id   		 = "%s"
 			atlas_cidr_block = "%s"
 			provider_name		 = "%s"
-			region			     = "US_EAST_2"
+			region			     = "US_EAST"
```
themantissa: Why the change? (both are Azure regions https://docs.atlas.mongodb.com/reference/microsoft-azure/)

themantissa (changes requested):
Just a couple questions and one nit on a missing EoF warning in the review view.

.travis.yml
coderGo93: weird, I added newline, gotta check again

mongodbatlas/data_source_mongodbatlas_maintenance_window_test.go
coderGo93: because I skip it, if you want I can delete the code related with ASAP? and if I don't comment it will give an error about lint unused code

mongodbatlas/resource_mongodbatlas_network_container_test.go
coderGo93: Because it fails with US_EAST_2 and I don't know how to delete it with mongodb ui, it doesn't appear that container the same can be said for GCP, that's why I changed it so it can pass, do you want me to change it back?

.travis.yml
coderGo93: I did not realize that is the travis, I'm not sure if at the end we should keep the .travis.yml if Github actions can do the same work and "faster"

themantissa: We'll probably remove it once we get github actions working well but let's keep it for now.

mongodbatlas/data_source_mongodbatlas_maintenance_window_test.go
themantissa: If we don't need it then removing it is fine - much cleaner than leaving unused code.

mongodbatlas/resource_mongodbatlas_network_container_test.go
themantissa: You can't delete containers in the MongoDB UI because we don't show it there, only via API.  I'm not sure I follow on why it might be failing, what's the actual error?  Might be showing us a real problem.

PacoDw (approved):
If all the above comments were attended is LGMT 👍

mongodbatlas/provider_test.go
```diff
@@ -148,3 +149,22 @@ func TestRemoveLabel(t *testing.T) {
...
+	}
+}
+
+func checkTeamsIds(t *testing.T) {
+	if os.Getenv("MONGODB_ATLAS_TEAMS_IDS") == "" {
+		t.Fatal("`MONGODB_ATLAS_TEAMS_IDS` must be set for Projects acceptance testing")
+	}
+}
+
+// SkipTestImport temporary solution to avoid error about failed verification, will solve later
```
themantissa, line 165: Created an internal ticket to cover - when we move to jira officially we will use ticket INTMDB-4

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_restore_job.go
```diff
@@ -331,7 +331,7 @@ func resourceMongoDBAtlasCloudProviderSnapshotRestoreJobImportState(d *schema.Re
 
 	d.SetId(encodeStateID(map[string]string{
 		"project_id":              *projectID,
-		"cluster_name":            *clusterName,
+		"cluster_name":            requestParameters.ClusterName,
```
themantissa: I asked about similar fixes in the other pull requests.  If we fix a bug while fixing the tests that should be noted in the PR, that way users know what was fixed.  It's just a "nice to do" and makes understanding what happened easier.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -963,7 +963,7 @@ func resourceMongoDBAtlasClusterImportState(d *schema.ResourceData, meta interfa
 	d.SetId(encodeStateID(map[string]string{
 		"cluster_id":    u.ID,
 		"project_id":    *projectID,
-		"cluster_name":  *name,
+		"cluster_name":  u.Name,
```
themantissa, line 966: Same as other comment - need to call out if we had to fix the actual resource for the test to pass (that means it is a fixed bug)

themantissa (commented):
Also looking good - similar to the other PR just would like a call out in the PR description if any bugs in the actual resources were fixed so that the tests would pass.  thanks!

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_restore_job.go
coderGo93: This is an import that sometimes passes or fails, need to see why does happen, not sure if it's because there are many test acceptances running at same time. I changed this to verify it was the issue but it isn't, that's why I added a env var to skip test import if true in all test acceptance that imports because if you skip one another can appear with similar error.

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: Same answer as another comment because is a test acceptance that imports, there's a weird behavior with terraform that sometimes fails sometimes doesn't and the resource that happen can vary.

themantissa: So when you made this change the import stopped failing?   Or you made the change and it didn't fix it?

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_restore_job.go
themantissa: Okay then if the change didn't fix anything and is not needed it shouldn't be included unless it's there for a reason since this isn't just the test but the actual resource code.

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: It didn't make a difference, I mean locally usually works because I run one test acc, that's why I assume it's related with running many testacc at same time but can't be sure, in fact, when I run make testacc locally, it happened the error for import in some resources

mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_restore_job.go
coderGo93: Ok, I'll change it back for now, thank you

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: If it didn't make a difference it's best to not leave it in unless the fix makes sense for another reason.

themantissa: btw we do rate limit the API, just realized ... maybe you are running into that?  You should get a response code: https://docs.atlas.mongodb.com/api/index.html#rate-limiting

coderGo93: It didn't appear that error, I think it's related to ParallelTest but can't be sure but only happens when there are many testacc running at same time, right now it only has one import I think from paco(threshold attribute related) and it doesn't fail because there are not other similar testacc running at same time I think

themantissa (approved):
LGTM

PR #296, opened 2020-08-19
Fixes many testacc

coderGo93:
## Description

Fixes many testacc, still not sure if everything will work until github actions works.

Fixed: 

- retryable error for IP Whitelist resource.
-  OplogTS bug(was string but now i'ts int)

Link to any related issue(s): 

## Type of change:

## Required Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirments

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_alert_configuration_test.go
+++ b/mongodbatlas/resource_mongodbatlas_alert_configuration_test.go
@@ -315,9 +315,7 @@ func testAccCheckMongoDBAtlasAlertConfigurationImportStateIDFunc(resourceName st
 			return "", fmt.Errorf("not found: %s", resourceName)
 		}
 
-		ids := decodeStateID(rs.Primary.ID)
-
-		return fmt.Sprintf("%s-%s", ids["project_id"], ids["id"]), nil
+		return fmt.Sprintf("%s-%s", rs.Primary.Attributes["project_id"], rs.Primary.Attributes["alert_configuration_id"]), nil
 	}
 }
 
--- a/mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_restore_job_test.go
+++ b/mongodbatlas/resource_mongodbatlas_cloud_provider_snapshot_restore_job_test.go
@@ -181,9 +181,7 @@ func testAccCheckMongoDBAtlasCloudProviderSnapshotRestoreJobImportStateIDFunc(re
 			return "", fmt.Errorf("not found:: %s", resourceName)
 		}
 
-		ids := decodeStateID(rs.Primary.ID)
-
-		return fmt.Sprintf("%s-%s-%s", ids["project_id"], ids["cluster_name"], ids["snapshot_restore_job_id"]), nil
+		return fmt.Sprintf("%s-%s-%s", rs.Primary.Attributes["project_id"], rs.Primary.Attributes["cluster_name"], rs.Primary.Attributes["snapshot_restore_job_id"]), nil
 	}
 }
 
--- a/mongodbatlas/resource_mongodbatlas_cluster_test.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster_test.go
@@ -862,9 +862,7 @@ func testAccCheckMongoDBAtlasClusterImportStateIDFunc(resourceName string) resou
 			return "", fmt.Errorf("not found: %s", resourceName)
 		}
 
-		ids := decodeStateID(rs.Primary.ID)
-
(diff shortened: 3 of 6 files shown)


themantissa:
@coderGo93 this needs https://github.com/mongodb/terraform-provider-mongodbatlas/pull/293 first correct?  I'd prefer to wait to review till the tests are all working.

coderGo93:
@themantissa Not really because this testacc needs the GH actions, so I rebase in #293  from this branch and then run again with changes of this branch.

themantissa:
@coderGo93 is this ready for a review now?

coderGo93:
@themantissa Yes, in #293 I rebased from here when I made changes here

PacoDw (approved):
LGTM!

mongodbatlas/data_source_mongodbatlas_cloud_provider_snapshot_restore_job.go
```diff
@@ -76,7 +76,7 @@ func dataSourceMongoDBAtlasCloudProviderSnapshotRestoreJob() *schema.Resource {
 				Computed: true,
 			},
 			"oplog_ts": {
-				Type:     schema.TypeString,
+				Type:     schema.TypeInt,
```
themantissa: Was this a bug found?

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
```diff
@@ -243,15 +243,7 @@ func resourceMongoDBAtlasProjectIPWhitelistDelete(d *schema.ResourceData, meta i
...
-			_, err := conn.ProjectIPWhitelist.Delete(context.Background(), ids["project_id"], ids["entry"])
-			if err != nil {
-				if strings.Contains(fmt.Sprint(err), "500") ||
-					strings.Contains(fmt.Sprint(err), "Unexpected error") ||
-					strings.Contains(fmt.Sprint(err), "UNEXPECTED_ERROR") {
-					return resource.RetryableError(err)
-				}
-				return resource.NonRetryableError(fmt.Errorf(errorWhitelistDelete, err))
-			}
+			return resource.RetryableError(fmt.Errorf(errorWhitelistDelete, "Whitelist still exists"))
```
themantissa: Another bug?  Just want to get an idea of the related fixes to the actual resources.

themantissa (commented):
LGTM but two questions I'd like to understand since it's actually changes to the resources, not just the testes.

mongodbatlas/data_source_mongodbatlas_cloud_provider_snapshot_restore_job.go
coderGo93: It was a typo it seems, it should have been Int since it gave me an error of type so I changed it

mongodbatlas/resource_mongodbatlas_project_ip_whitelist.go
coderGo93: Yes, it happens sometimes the error about an ip whitelist still exists, when an ip exists it should retry not another delete because for some reason it can fail and it should have another get to verify that, that's why I deleted Delete and retry instead.

themantissa: Yeah, still think that's a timing issue.  Just note that in the PR description, for example - "Fixed: retryable error for IP Whitelist resource"

mongodbatlas/data_source_mongodbatlas_cloud_provider_snapshot_restore_job.go
themantissa: In the underlying docs it's a string https://docs.atlas.mongodb.com/reference/api/cloud-backup/restore/create-one-restore-job/ so we may want to ensure it's not a bigger bug or a miss of something else.

coderGo93: mm weird, to me appears is a integer for `oplogTs`

themantissa: Ah, I think the docs are wrong.  In the POST it has it as a string but in the GET it shows an integer, which makes more sense.  I think this fix is right and docs wrong (which test caught!). Let's leave it in but document in the PR description (not the title, just the description is fine)

themantissa: Ignore previous comment - string in docs in both places.  Could Terraform be storing the string as an int?  Seems weird.

PR #297, opened 2020-08-19
Fix #285: Unable to import $external auth users

### LINKED ISSUE #285: Unable to import $external auth users
I'm upgrading provider in my current infrastructure but I cannot import already defined users as they have an "-" in the name, and they are authenticated against $external database.

But the regex used to split the import ID, is using - and $external is not allowed as a database name:
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_database_user.go
``` go
299 func splitDatabaseUserImportID(id string) (projectID, username, authDatabaseName *string, err error) {
300  	var re = regexp.MustCompile(`(?s)^([0-9a-fA-F]{24})-(.*)-([a-z]{1,15})$`)
```

So this regex should allow the "$external" keyword in the third group for allowing user imports.

Some regex tests:
image

image

Current behaviour:
- User is splitted and the rest of the string is used as database name

```
$terraform import  mongodbatlas_database_user.external "0000aaaa0000aaaa0000aaaa-some_test-user-$external"
Error: couldn't import user(some_test) in project(0000aaaa0000aaaa0000aaaa), error: GET https://cloud.mongodb.com/api/atlas/v1.0/groups/0000aaaa0000aaaa0000aaaa/databaseUsers/user-$external/some_test: 404 (request "Not Found") No user with username some_test exists.
```

Proposed fix:
- Adding $external literal to the third grup in an OR statement

``` go
299 func splitDatabaseUserImportID(id string) (projectID, username, authDatabaseName *string, err error) {
300  	var re = regexp.MustCompile(`(?s)^([0-9a-fA-F]{24})-(.*)-([a-z]{1,15}|\$external)$`)
```
image

PacoDw:
## Description

- Improved the regex to import the auth database name with the symbol $
- Refactored tests cases and it was added a new test case to validate when the auth database name has the symbol $

Link to any related issue(s): #285 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_database_user.go
+++ b/mongodbatlas/resource_mongodbatlas_database_user.go
@@ -289,15 +289,15 @@ func resourceMongoDBAtlasDatabaseUserImportState(d *schema.ResourceData, meta in
 
 	d.SetId(encodeStateID(map[string]string{
 		"project_id":         *projectID,
-		"username":           *username,
-		"auth_database_name": *authDatabaseName,
+		"username":           u.Username,
+		"auth_database_name": u.DatabaseName,
 	}))
 
 	return []*schema.ResourceData{d}, nil
 }
 
 func splitDatabaseUserImportID(id string) (projectID, username, authDatabaseName *string, err error) {
-	var re = regexp.MustCompile(`(?s)^([0-9a-fA-F]{24})-(.*)-([a-z]{1,15})$`)
+	var re = regexp.MustCompile(`(?s)^([0-9a-fA-F]{24})-(.*)-([$a-z]{1,15})$`)
 	parts := re.FindStringSubmatch(id)
 
 	if len(parts) != 4 {
--- a/mongodbatlas/resource_mongodbatlas_database_user_test.go
+++ b/mongodbatlas/resource_mongodbatlas_database_user_test.go
@@ -15,11 +15,12 @@ import (
 )
 
 func TestAccResourceMongoDBAtlasDatabaseUser_basic(t *testing.T) {
-	var dbUser matlas.DatabaseUser
-
-	resourceName := "mongodbatlas_database_user.basic_ds"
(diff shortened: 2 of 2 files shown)


mongodbatlas/resource_mongodbatlas_database_user.go
```diff
@@ -289,15 +289,15 @@ func resourceMongoDBAtlasDatabaseUserImportState(d *schema.ResourceData, meta in
...
+		"username":           u.Username,
+		"auth_database_name": u.DatabaseName,
 	}))
 
 	return []*schema.ResourceData{d}, nil
 }
 
 func splitDatabaseUserImportID(id string) (projectID, username, authDatabaseName *string, err error) {
-	var re = regexp.MustCompile(`(?s)^([0-9a-fA-F]{24})-(.*)-([a-z]{1,15})$`)
+	var re = regexp.MustCompile(`(?s)^([0-9a-fA-F]{24})-(.*)-([$a-z]{1,15})$`)
```
themantissa, line 300: Just an FYI the admin database can only be admin or $external so could possibly simplify this, not required but if we run into more issues may make sense.

themantissa (approved):
LGTM, left an informational comment on current state of auth db.  Needs DoU reviewer.

mongodbatlas/resource_mongodbatlas_database_user.go
PacoDw: I made a test validation in both cases (with `admin` and `$external`) and they worked fine. I decided to leave in this way because in the future if `$external` change or add a new for example `$internal` it will work fine.

themantissa: Not a bad idea to future proof it!  Thank you for testing and agree.

coderGo93 (approved):
LGTM

Issue #232, opened 2020-05-20
Can't create alerts with using many event_type 


rafafalves:
Hi,
I'm trying create alerts from many different event_types using version 0.5.1, but just the type OUTSIDE_METRIC_THRESHOLD is working.
If I try any other type from Host, Replica Set or Billing groups, I received an error message like this:
REMOVED BY ADMIN

Creating any alert with OUTSIDE_METRIC_THRESHOLD parameter and trying change the configuration to met with desired other type, terraform return that changes was applied but looking in portal changes was not made.

Already tried configuration like this for all event_types that are returning error:

resource "mongodbatlas_alert_configuration" "test4" {
  project_id = var.id_project
  event_type = "UPGRADED"
  enabled    = true

  notification {
    type_name     = "SLACK"
    interval_min  = 5
    delay_min     = 0
    api_token     = var.slack_token
    channel_name  = var.slack_channel
  }
}

Also tried change the values that I'm passing from variable to "clear value" in code ans the error are the same

rafafalves:
Another point that I forgot to comment in post. paramters from HOST group that appear in documentatino with HOST_, not work even trying make a change. If removed HOST_ and let just desired option like in the configuration example, I can make a change from OUTSIDE_METRIC_THRESHOLD type to new one. (UPGRADED, RESTARTED, NOW_SECONDARY and NOW_PRIMARY).
Anyway, the changes are not completed looking at the portal.

themantissa:
@rafafalves  thank you for the information.  Issue #227  is similar and we are looking into the best course of action.   Will update as more info

themantissa:
@rafafalves apologies for the length of time to resolution here but wanted to assure you and anyone else that we are working on this and will post updates as we have them.

themantissa:
@rafafalves just an end of week update - we are updating the list of what is/isn't available as alerts in Atlas and then will update the provider that should correct the behavior you are seeing.   In the mean time let me know if you have any specific questions.

PacoDw:
Hello @rafafalves, I created a PR implementing the threshold attribute if you want to check more details or play with the new feature see the PR #298.

Let me know if you have another comment or concern, thanks

themantissa:
@rafafalves  we've released a preview version that should allow you to test if this fixes the issue.  If you can confirm we'd appreciate it.
https://github.com/mongodb/terraform-provider-mongodbatlas/releases/tag/0.6.4

[...]

PR #298, opened 2020-08-19
Fix #227 #232: Added a new Threshold attribute

### LINKED ISSUE #227: Can't create a alert for Replication Oplog Window
Hi,

I can't create a alert for `Replication Oplog Window is below 1 hours` using Terraform.

This is what I have:

```
resource "mongodbatlas_alert_configuration" "pager_duty_alert_replication_window" {
  project_id = mongodbatlas_project.project.id
  event_type = "REPLICATION_OPLOG_WINDOW_RUNNING_OUT"
  enabled    = true

  notification {
    type_name    = "PAGER_DUTY"
    service_key  = var.pager_duty_service_key
    delay_min    = 0
    interval_min = local.pager_duty_interval_min
  }
}
```

But this won't work as this alert needs a threshold. See error:

```
400 (request "Bad Request") A threshold must be specified for the specified event type.
```

This threshold has a different JSON format than when using `OUTSIDE_METRIC_THRESHOLD`. For `OUTSIDE_METRIC_THRESHOLD` the JSON looks like this:

```
"metricThreshold": {
   "metricName": "CONNECTIONS_PERCENT",
   "mode": "AVERAGE",
   "operator": "GREATER_THAN",
   "threshold": 80.0,
   "units": "RAW"
},
```

But for this metric, `REPLICATION_OPLOG_WINDOW_RUNNING_OUT`, it looks like this:

```
"threshold": {
   "operator": "LESS_THAN",
   "threshold": 1
}
```

I tried using in Terraform:

```
threshold = {
    operator    = "LESS_THAN"
    threshold   = 1
}
```

But get this error:

```
Error: Unsupported argument
An argument named "threshold" is not expected here.
```

And:

```
metric_threshold = {
    operator    = "LESS_THAN"
    threshold   = 1
    units       = "RAW"
    mode        = "AVERAGE"
  }
```

And get:

```
400 (request "Bad Request") The metric threshold should only be specific for host metric alerts.
```

Is it possible that this metric is not support on Terraform right now, or how do I create it?

Thanks.

PacoDw:
## Description

- Resource
- [x] Threshold attribute was implemented as a new feature
- [x] Added a new test case to validate a configuration with the new attribute
- [x] Added Documentation

- Datasource
- [x] Implemented the threshold in the data source
- [x] Added a test case
- [x] Added Documentation

Link to any related issue(s): #227  #232 

## Type of change:

## Required Checklist:

Example Usage 
```hcl
resource "mongodbatlas_alert_configuration" "test" {
  project_id = "<PROJECT-ID>"
  event_type = "REPLICATION_OPLOG_WINDOW_RUNNING_OUT"
  enabled    = true

  notification {
    type_name     = "GROUP"
    interval_min  = 5
    delay_min     = 0
    sms_enabled   = false
    email_enabled = true
    roles         = ["GROUP_CHARTS_ADMIN", "GROUP_CLUSTER_MANAGER"]
  }

  matcher {
    field_name = "HOSTNAME_AND_PORT"
    operator   = "EQUALS"
    value      = "SECONDARY"
  }

  threshold = {
    operator    = "LESS_THAN"
    threshold   = 1
    units       = "HOURS"
  }
}

data "mongodbatlas_alert_configuration" "test" {
	project_id             = "${mongodbatlas_alert_configuration.test.project_id}"
	alert_configuration_id = "${mongodbatlas_alert_configuration.test.alert_configuration_id}"
}
```


website/docs/d/alert_configuration.html.markdown
```diff
@@ -66,35 +100,35 @@ In addition to all arguments above, the following attributes are exported:
 * `enabled` - If set to true, the alert configuration is enabled. If enabled is not exported it is set to false.
 * `event_type` - The type of event that will trigger an alert.
   Alert type. Possible values:
-	 - Host 	
+	 - Host
```
themantissa: event_type was massively updated in the docs - https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/ so to close both issues we'll need to address this.   The updated values can be found in the "Condition Tab" under eventTypeName.   We don't need the descriptions here but I would perhaps list the legal values and put a link to the underlying API for all the explanations as they are subject to change.

mongodbatlas/resource_mongodbatlas_alert_configuration.go
```diff
@@ -144,6 +143,44 @@ func resourceMongoDBAtlasAlertConfiguration() *schema.Resource {
 					},
 				},
 			},
+			"threshold": {
```
themantissa, line 128: I'm leaving the comment here as quite a few variables between 26-145 needs to be matched to the updated documentation to address #232. It may make sense, where there is a long list of potential values (like 
metricThreshold.metricName which the updated list is now https://docs.atlas.mongodb.com/reference/alert-host-metrics/#measurement-types) to remove the validations and default to Atlas for errors.

themantissa (changes requested):
This is tough to review in that it is a PR for both 227 and 232 but really focused on fixing 227, which is the missing handling for threshold.  This seems to have most of that, as well as the addition of the 
notifications.roles.  However, it's still missing much of what would fix 232, which tends to be around parameter values that were wrong in the doc and are now updated.   These often have many possible values such as, using the Atlas API names, eventTypeName, matchers.fieldName, matchers.value, matchers.operator, metricThreshold.metricName (links to https://docs.atlas.mongodb.com/reference/alert-host-metrics/#measurement-types).  I think it may just make sense to remove the validations for those, allow Atlas to provide the error, and update the documentation.

website/docs/d/alert_configuration.html.markdown
PacoDw: It makes sense, I think this is better to remove them and delegate the error handling to MongoDB server, and update the documentation, thanks

mongodbatlas/resource_mongodbatlas_alert_configuration.go
```diff
@@ -119,23 +100,6 @@ func resourceMongoDBAtlasAlertConfiguration() *schema.Resource {
 						"units": {
 							Type:     schema.TypeString,
 							Optional: true,
-							ValidateFunc: validation.StringInSlice([]string{
```
themantissa: I think we'd be okay to leave this one in as it seems much less likely to change much compared to the others.

website/docs/d/alert_configuration.html.markdown
```diff
@@ -65,68 +99,33 @@ In addition to all arguments above, the following attributes are exported:
...
-    - Billing 	
-      - `CREDIT_CARD_ABOUT_TO_EXPIRE`
-      - `CHARGE_SUCCEEDED`
-      - `INVOICE_CLOSED`
-
-    -> **NOTE:** If this is set to OUTSIDE_METRIC_THRESHOLD, the metricThreshold field must also be set.
+
+  -> **IMPORTANT:** Event Type has many possible values, you can see more details at: https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/ (you can found them in the "Condition Tab" under eventTypeName)
+
+  -> **NOTE:** If this is set to OUTSIDE_METRIC_THRESHOLD, the metricThreshold field must also be set.
```
themantissa: Slight edit but this is a great way to do it!  Change to something like:

 -> **IMPORTANT:** Event Type has many possible values.  All current options at available at https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/  Details for both conditional and metric based alerts can be found by selecting the tabs on the [alert config page](https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/) and checking the latest eventTypeName options. 

  -> **NOTE:** If `event_type` is set to OUTSIDE_METRIC_THRESHOLD, the metricThreshold field must also be set.

website/docs/d/alert_configuration.html.markdown
```diff
@@ -147,7 +146,7 @@ Rules to apply when matching an object against this alert configuration. Only en
 The threshold that causes an alert to be triggered. Required if `event_type_name` : "OUTSIDE_METRIC_THRESHOLD".
 
 * `metric_name` - Name of the metric to check.
```
themantissa: Let's add to line 149:
* `metric_name` - Name of the metric to check. The full list of current options is available [here](https://docs.atlas.mongodb.com/reference/alert-host-metrics/#measurement-types)

website/docs/d/alert_configuration.html.markdown
```diff
@@ -174,6 +173,15 @@ The threshold that causes an alert to be triggered. Required if `event_type_name
...
 * `mode` - This must be set to AVERAGE. Atlas computes the current metric value as an average.
 
+### Threshold
+* `operator` - Operator to apply when checking the current metric value against the threshold value.
+  Accepted values are:
+    - `GREATER_THAN`
+    - `LESS_THAN`
+
+* `threshold` - Threshold value outside of which an alert will be triggered.
+* `units` - The units for the threshold value. Depends on the type of metric.
```
themantissa, line 185: This should be the same list as up top: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/298/files#diff-59fafd448a2d3d95ad00fb5e8c1093b1R155

website/docs/r/alert_configuration.html.markdown
```diff
@@ -45,64 +45,60 @@ resource "mongodbatlas_alert_configuration" "test" {
...
-      - `CREDIT_CARD_ABOUT_TO_EXPIRE`
-      - `CHARGE_SUCCEEDED`
-      - `INVOICE_CLOSED`
-
-    -> **NOTE:** If this is set to OUTSIDE_METRIC_THRESHOLD, the metricThreshold field must also be set.
+
+  -> **IMPORTANT:** Event Type has many possible values, you can see more details at: https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/ (you can found them in the "Condition Tab" under eventTypeName)
+
+  -> **NOTE:** If this is set to OUTSIDE_METRIC_THRESHOLD, the metricThreshold field must also be set.
+
```
themantissa: All the same changes as in data source - won't call all of them out but the same edits will apply.

website/docs/r/alert_configuration.html.markdown
```diff
@@ -45,64 +45,60 @@ resource "mongodbatlas_alert_configuration" "test" {
 }
 ```
 
+```hcl
```
themantissa, line 51: Can you include something like the following above line 13:

-> **NOTE:** In order to allow for a fast pace of change to alert variables some validations have been removed from this resource in order to unblock alert creation.  Impacted areas have links to the MongoDB Atlas API documentation so always check it for the most current information: https://docs.atlas.mongodb.com/reference/api/alert-configurations-create-config/

themantissa (changes requested):
Looking good - just a few docs changes and one areas I think we can leave.

mongodbatlas/resource_mongodbatlas_alert_configuration.go
PacoDw: okay :)

mongodbatlas/resource_mongodbatlas_alert_configuration.go
```diff
@@ -100,6 +100,20 @@ func resourceMongoDBAtlasAlertConfiguration() *schema.Resource {
...
+								"GROUP",
+								"OPS_GENIE",
+								"ORG",
+								"PAGER_DUTY",
+								"SLACK",
+								"SMS",
+								"TEAM",
+								"USER",
+								"VICTOR_OPS",
+								"WEBHOOK"}, false),
```
themantissa: This is the wrong array of strings, check the old code, should be:
"RAW",
								"BITS",
								"BYTES",
								"KILOBITS",
								"KILOBYTES",
								"MEGABITS",
								"MEGABYTES",
								"GIGABITS",
								"GIGABYTES",
								"TERABYTES",
								"PETABYTES",
								"MILLISECONDS",
								"SECONDS",
								"MINUTES",
								"HOURS",
								"DAYS"}, false),

mongodbatlas/resource_mongodbatlas_alert_configuration.go
```diff
@@ -124,6 +138,20 @@ func resourceMongoDBAtlasAlertConfiguration() *schema.Resource {
...
```
themantissa: same as above, wrong data.

themantissa (changes requested):
units is wrong data otherwise rest is good, fix and I can approve.

themantissa (approved):
LGTM - we may have a bit more work once the API docs get a final update but this should be most of it.  Thank you!

coderGo93 (approved):
Really long but it seems good, the only "issue" is the import state, probably would fail when testacc runs for weird reason(hit or miss) but LGTM

PR #321, opened 2020-09-26
Third party integrations 

leofigy:
## Description

This PR includes the schemas for the datasources 

- mongodbatlas_third_party_integration
- mongodbatlas_third_party_integrations

Resources: 

- mongodbatlas_third_party_integration

Test Cases

- TestAccdataSourceMongoDBAtlasThirdPartyIntegration_basic
- TestAccdataSourceMongoDBAtlasThirdPartyIntegrations_basic
- TestAccResourceMongoDBAtlasThirdPartyIntegration_basic
- TestAccResourceMongoDBAtlasThirdPartyIntegration_importBasic
- TestAccResourceMongoDBAtlasThirdPartyIntegration_updateBasic

Including tests and documentation, please take a look the API reference was 
- https://docs.atlas.mongodb.com/reference/api/third-party-integration-settings-get-one/
- https://docs.atlas.mongodb.com/reference/api/third-party-integration-settings-get-all/ 

Please take a look on Monday @coderGo93 , if you want to rename some functions, add/remove attributes properties, 
and comment here. 

Link to any related issue(s): 

## Type of change:

## Required Checklist:


themantissa:
Just gotta say this is a fantastic PR description!  ty!

mongodbatlas/data_source_mongodbatlas_third_party_integration_test.go
```diff
@@ -0,0 +1,292 @@
...
+	`
+
+	PAGERDUTY = `
+	resource "mongodbatlas_third_party_integration" "%[1]s" {
+		project_id = "%[2]s"
+		type = "%[3]s"
+		service_key = "%[4]s"
+	}
+	`
+	SLACK = `
```
gssbzn: FYI slack no longer works via api

website/docs/r/third_party_integration.markdown
```diff
@@ -0,0 +1,91 @@
...
+     Provides a Third-Party Integration Settings resource.
+---
+
+# mongodbatlas_third_party_integration
+
+`mongodbatlas_third_party_integration` Provides a Third-Party Integration Settings for the given type.
+
+-> **NOTE:** Groups and projects are synonymous terms. You may find `groupId` in the official documentation.
+
+-> **NOTE:** Slack integrations now use the OAuth2 verification method and must be initially configured, or updated from a legacy integration, through the Atlas third-party service integrations page. Legacy tokens will soon no longer be supported.
```
themantissa: Can you add the link to the integrations page:
https://docs.atlas.mongodb.com/tutorial/third-party-service-integrations/

Slack integrations now use the OAuth2 verification method and must be initially configured, or updated from a legacy integration, through the [Atlas third-party service integrations page](https://docs.atlas.mongodb.com/tutorial/third-party-service-integrations/). Legacy tokens will soon no longer be supported.

website/docs/r/third_party_integration.markdown
```diff
@@ -0,0 +1,91 @@
...
+
+~> **IMPORTANT** Each project can only have one configuration per {INTEGRATION-TYPE}.
+
+~> **IMPORTANT:** All arguments including the secrets will be stored in the raw state as plain-text. [Read more about sensitive data in state.](https://www.terraform.io/docs/state/sensitive-data.html)
```
themantissa, line 19: This is excellent^. Thank you!

website/docs/r/third_party_integration.markdown
```diff
@@ -0,0 +1,91 @@
...
+   * `flow_name` - Your Flowdock Flow name.
+   * `api_token` - Your API Token.
+   * `org_name` - Your Flowdock organization name.
+* `WEBHOOK`
+   * `url` - Your webhook URL.
+   * `secret` - An optional field for your webhook secret.
+
+## Attributes Reference
+
+* `id` - Unique identifier used for terraform for internal manages and can be used to import.
```
themantissa: Tiny edit:
 * `id` - Unique identifier used by terraform for internal management, which can also be used to import.

themantissa (changes requested):
Really nicely done.   Just a few, very small, doc edits.   Just needs those and a DoU review and done!

mongodbatlas/data_source_mongodbatlas_third_party_integration_test.go
themantissa: Correct but didn't they leave the legacy functionality in the API?   The note about it now needed OAuth is in the "important calls outs" from our docs below.  I guess we could just start w/ this already removed but then we would be different than reality.

leofigy: Hi @gssbzn , @themantissa, I updated the code & docs to deprecate Slack in the plugin.

coderGo93:
Nice, the PR seems good but I have question related to schema resource, most of them are optional and sensitive, but for only optional those parameters is not necessary to appear in state(Computed)?

leofigy:
Hi @coderGo93 

It depends, if you see for "datasources" the schema is the following -
```
			"api_key": {
				Type:      schema.TypeString,
				Sensitive: true,
				Computed:  true,
			},
```
This means is computed, because we won't know it until we read the value from the api and sensitive because it's a secret. 

Now in the other hand, for a "resource" 

```
			"api_key": {
				Type:      schema.TypeString,
				Sensitive: true,
				Optional:  true,
			},
```
The optional makes noise in the sense, that we are reusing the same schema for several "types" of integrations. 
So not all of them are required.

coderGo93 (approved):
LGTM to me, just to be sure, before merge and once you get approvals from two party, please run the make testacc to be sure if all test has been passed.

website/docs/r/third_party_integration.markdown
```diff
@@ -12,8 +12,6 @@ description: |-
 
 -> **NOTE:** Groups and projects are synonymous terms. You may find `groupId` in the official documentation.
 
--> **NOTE:** Slack integrations now use the OAuth2 verification method and must be initially configured, or updated from a legacy integration, through the Atlas third-party service integrations page. Legacy tokens will soon no longer be supported.
```
themantissa: I'd actually leave this note, this explains how to now setup - otherwise people might think we left it off (we did for a reason). And still add the link 
[Atlas third-party service integrations page](https://docs.atlas.mongodb.com/tutorial/third-party-service-integrations/)

themantissa (approved):
Approve but do think we should leave the comment in about how to setup Slack (see note) but not blocking approval on that.  If you can add that before merging it might prove helpful.  Otherwise lgtm, thanks!

website/docs/r/third_party_integration.markdown
leofigy: Hi @themantissa I put the note back , and the link as well :)

themantissa: Thank you and extra bonus points for :) in the commit!

mongodbatlas/data_source_mongodbatlas_third_party_integration_test.go
gssbzn: We were not able to "create" a slack integration, I guess the only use of the API is now for editing existing legacy versions, so I don't know how well that may play with users, and how much support that may bring, just flagging as we (mcli) decided to just remove the slack command to avoid people coming after us with "can't create", also the API fails silently, you'll get a 200 but the integration won't be enabled

gssbzn (commented):
@leofigy don't wait for me on this one, I have no idea of terraform, just wanted to flag the slack thing as it was a gotcha for us in the cli

Issue #329, opened 2020-10-05
Unexpected behaviour from resource `mongodbatlas_teams` when adding username for user not yet part of/Pending to join Organisation


muscovitebob:
# Terraform CLI and Terraform MongoDB Atlas Provider Version

```hcl
Terraform v0.12.20
Mongodbatlas v0.6.1
```

### Terraform Configuration File 

```hcl
resource "mongodbatlas_project" "myproject" {
  name   = "myproject"
  org_id = "orgid"
  teams {
    team_id    = mongodbatlas_teams.myteam.team_id
    role_names = ["GROUP_OWNER"]
  }
}

resource "mongodbatlas_teams" "myteam" {
  org_id    = "orgid"
  name      = "My Team"
  usernames = ["user.that.exists.in.org@organization.com, user.that.does.not.exist.in.org@organization.com"]
}
```

### Steps to Reproduce

Have an organisation in which there is a `user.that.exists.in.org@organization.com`, whom have logged in at least once.
Invite a new user, `user.that.does.not.exist.in.org@organization.com`, to the organisation. They are listed as Pending in the organisation users overview.
Add the `user.that.does.not.exist.in.org@organization.com` to the usernames array in the `mongodbatlas_teams` resource.

### Expected Behavior

User is added to the team successfully, or a helpful error is produced if this is not possible. If this is not possible, the myteam team retains all of its members before we attempted adding a new member.

### Actual Behavior

Terraform plan reports user will be added to array. 

Terraform apply errors out:
```
Error: error getting Atlas User (user.that.does.not.exist.in.org@organization.com) information: GET https://cloud.mongodb.com/api/atlas/v1.0/users/byName/user.that.does.not.exist.in.org@organization.com: 401 (request "Unauthorized") The currently logged in user does not have the user administrator role for any group, team, or organization containing user user.that.does.not.exist.in.org@organization.com.

  on mongodb_extra.tf line 12, in resource "mongodbatlas_teams" "myteam":
  12: resource "mongodbatlas_teams" "myteam" {
```

Repeated terraform plan reports that the `myteam` team needs to have all of the usernames that were previously part of it added again. Navigating to the MongoDB Atlas project GUI confirms that the team has 0 members after this error.

### Additional Context

If this is indeed caused by the `user.that.does.not.exist.in.org@organization.com` being in status Pending for organisation member, the error message should probably be more helpful.

muscovitebob:
[...]

PR #361, opened 2020-12-11
INTMDB-121: Prevents removing existing users

leofigy:
## Description

Hi @themantissa checked the issue, and it's a bit weird. 

**Behaviour**

* The Create API allows you create a team with users that are in pending state when you create the resource for the first time. 
   This has a side effect that it allow you to get the user ID, when you list the users with `conn.Teams.GetTeamUsersAssigned`
  
* However if you add a pending user to an existent team, the addition fails because we use `conn.AtlasUsers.GetByName`
   with the error of 401 (at the time I tested the message was not updated yet). 

**Addressed** 
- The code removed the existing users, and then checked what was in the config file (main.tf) and updated with those users fetching the info. But the code did this before actually trying to get the info. so that's bad. I moved the code after we are sure, we have the info

- The case where the resource is created the first time with pending users, I fixed that using a cached if we get info from that user from `conn.Teams.GetTeamUsersAssigned`.

**_Not addressed_** 

I left the case where we actually cannot figure it out the `user.ID`. and that happens because the API we use in the client requires the user.ID as parameter https://docs.atlas.mongodb.com/reference/api/teams-add-user. 

Link to any related issue(s):

https://github.com/mongodb/terraform-provider-mongodbatlas/issues/329

## Type of change:

- [X] Bug fix (non-breaking change which fixes an issue)

## Required Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirments

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_team.go
+++ b/mongodbatlas/resource_mongodbatlas_team.go
@@ -133,31 +133,69 @@ func resourceMongoDBAtlasTeamUpdate(d *schema.ResourceData, meta interface{}) er
 		// First, we need to remove the current users of the team and later add the new users
 		// Get the current team's users
 		users, _, err := conn.Teams.GetTeamUsersAssigned(context.Background(), orgID, teamID)
+
 		if err != nil {
 			return fmt.Errorf(errorTeamRead, err)
 		}
 
-		// Removing each user
+		// Removing each user - Let's not modify the state before making sure we can continue
+
+		// existig users
+		index := make(map[string]matlas.AtlasUser)
 		for i := range users {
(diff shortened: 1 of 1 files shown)


themantissa (approved):
LGTM - thank you for the explanation of what was done and not.  I think this gets the core issue fixed as reported.  We will have some improvements in the invite process coming that should also help this.   Would be good to see if the issue reporter can try this fix as well.  Thanks!  
Oh @leofigy  and please also include the internal ticket in the PR title - I added the one for this.  Just helps me keep what is what clear.

coderGo93 (approved):
LGTM!

PR #374, opened 2021-01-02
Change db_user password to optional in the document

KeisukeYamashita:
## Description

As you can see here (https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_database_user.go#L48-L53), the `password` argument is `Optional: true` but the document says it's `Required`. It's misleading the developers and I want to fix it.

## Type of change:

## Required Checklist:


themantissa (changes requested):
@KeisukeYamashita this is an interesting one!  It is technically optional but that depends on which kind of authentication it is.  It is required if you are using SCRAM-SHA.  The original Atlas doc we used had it as required but now has a more accurate version: 
https://docs.atlas.mongodb.com/reference/api/database-users-create-a-user
_Conditional | User's password. Required if: x509Type, ldapAuthType, and awsIAMType are NONE, meaning that the database user authenticates with SCRAM-SHA._

If we change the doc we should probably follow the lead of the original doc and call it Conditional with when it is required.

KeisukeYamashita:
Thank you for your review 👍 

Since there are only `Required` or `Optional` in the HCL level, I will want to leave as it is for now. 
Closing this. Feel free to reopen  🙇

PR #380, opened 2021-01-14
INTMDB-160: Resetting an encryption at rest

coderGo93:
## Description

First issue, when it re apply, it re creates(destroy then create) instead of only changes.

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments
It seems the first issue it's fixed, I tested myself and it doesn't destroy again

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
+++ b/mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
@@ -66,6 +66,20 @@ func resourceMongoDBAtlasEncryptionAtRest() *schema.Resource {
 						},
 					},
 				},
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					v := val.(map[string]interface{})
+
+					_, akOk := v["access_key_id"]
+					_, saOk := v["secret_access_key"]
+					_, rOk := v["role_id"]
+
+					if (akOk && saOk && rOk) || (akOk && rOk) || (saOk && rOk) {
+						errs = append(errs, fmt.Errorf("%q For credentials: `access_key_id` and `secret_access_key` are allowed but not `role_id`."+
+							" For roles: `access_key_id` and `secret_access_key` are not allowed but `role_id` is allowed", key))
+					}
+
+					return
+				},
 			},
 			"azure_key_vault": {
 				Type:     schema.TypeMap,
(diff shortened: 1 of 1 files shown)


mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -271,10 +267,8 @@ func flattenAWSKMS(m *matlas.AwsKms) map[string]interface{} {
 		return map[string]interface{}{
 			"enabled":                cast.ToString(m.Enabled),
 			"access_key_id":          m.AccessKeyID,
-			"secret_access_key":      m.SecretAccessKey,
 			"customer_master_key_id": m.CustomerMasterKeyID,
 			"region":                 m.Region,
-			"role_id":                m.RoleID,
```
themantissa, line 277: I'm not sure I understand why these two values are being removed?

themantissa (changes requested):
Asking a question as not sure on the removal.

mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
coderGo93: Those parameters doesn't appear in API response and in docs it doesn't appear either in response, if I set those values, it will always appear "changes" since the result its empty like "" > "values"

themantissa: Ah! That makes sense, I didn't expect role_id to not be returned but if you verified the actual return matches docs then we are good.

themantissa (approved):
LGTM, thank you!

mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
coderGo93: I tested again, it seems with `role_id` doesn't give changes only with secret key. so I put it back the role id

leofigy (approved):
LGTM ! :)

themantissa (approved):
LGTM!

mongodbatlas/resource_mongodbatlas_encryption_at_rest.go
themantissa: Ah, so perhaps a docs mistake since I'm guessing it returned.   TY!

PR #413, opened 2021-02-13
INTMDB-18 : DataSource and Resource support for Online Archive

leofigy:
## Description

This PR adds the support for Online Archive 
original documentation https://docs.atlas.mongodb.com/reference/api/online-archive-get-all-for-cluster/

**Two new datasources**

-    online_archive   (Read)
-    online_archives (Read)

**One new resource**

-   online_archives (Create, Read, Update, Import, Delete)

Link to any related issue(s):

## Type of change:

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation fix/enhancement

## Required Checklist:

- [] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements


mongodbatlas/data_source_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,212 @@
...
+		// argument values
+		"project_id": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"cluster_name": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"atlas_id": {
```
themantissa, line 77: Kind of a nit but I think atlas_id is confusing.  Since this is the returned id of the online archive perhaps archive_id might make more sense for readability?

themantissa: Kind of a nit but I think atlas_id is confusing.  Since this is the returned id of the online archive perhaps archive_id might make more sense for readability?

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/mwielbut/pointy"
+	matlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+const (
+	errorOnlineArchivesCreate = "error creating MongoDB Online Archive: %s"
```
themantissa: Small nit: We almost always want to include Atlas in the error like this that has MongoDB: 	
errorOnlineArchivesCreate = "error creating MongoDB **Atlas** Online Archive: %s"

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+						ValidateFunc: validation.IntAtLeast(0),
+					},
+					"field_type": {
+						Type:     schema.TypeString,
+						Computed: true,
+					},
+				},
+			},
+		},
+		// mongodb_atlas id
```
themantissa: Same as above, I think archive_id makes more sense.

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+
+	outOnlineArchive, _, err := conn.OnlineArchives.Get(context.Background(), projectID, clusterName, atlasID)
+
+	if err != nil {
+		reset := strings.Contains(err.Error(), "404") && !d.IsNewResource()
+		if reset {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("error MongoDBAtlas Online Archive with id %s, read error %s", atlasID, err.Error())
```
themantissa: Small nit - MongoDB Atlas needs a space.

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	newData := fromOnlineArchiveToMapInCreate(outOnlineArchive)
+
+	for key, val := range newData {
+		if err := d.Set(key, val); err != nil {
+			return fmt.Errorf("error MongoDBAtlas Online Archive with id %s, read error %s", atlasID, err.Error())
```
themantissa: Same, MongoDB Atlas needs a space

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+	return nil
+}
+
+func resourceMongoDBAtlasOnlineArchiveImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	conn := meta.(*matlas.Client)
+	parts := strings.Split(d.Id(), "-")
+
+	if len(parts) != 3 {
+		return nil, errors.New("import format error to import a Mongo Online Archive, use the format {project_id}-{cluste_rname}-{atlas_id} ")
```
themantissa: Mongo need the DB and Atlas : 
return nil, errors.New("import format error to import a Mongo**DB** **Atlas** Online Archive, use the format {project_id}-{cluste_rname}-{atlas_id} ")

Also this is where atlas id might get truly confusing for users.  It doesn't make one think archive Id.

themantissa (changes requested):
A few small nits/suggestions but overall looks good. Are you doing the tests and documentation in another PR?

mongodbatlas/data_source_mongodbatlas_online_archive.go
andreaangiolillo: +1 on `archive_id`

mongodbatlas/data_source_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,212 @@
...
+	clusterName := d.Get("cluster_name").(string)
+	atlasID := d.Get("atlas_id").(string)
+
+	outOnlineArchive, _, err := conn.OnlineArchives.Get(context.Background(), projectID, clusterName, atlasID)
+
+	if err != nil {
+		return fmt.Errorf("error reading Online Archive datasource with id %s: %s", atlasID, err.Error())
+	}
+
+	newData := fromOnlineArchiveToMap(outOnlineArchive)
```
andreaangiolillo:
```suggestion
	onlineArchiveMap := fromOnlineArchiveToMap(outOnlineArchive)
```

mongodbatlas/data_source_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,212 @@
...
+	}))
+
+	return nil
+}
+
+func dataSourceMongoDBAtlasOnlineArchivesRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*matlas.Client)
+	projectID := d.Get("project_id").(string)
+	clusterName := d.Get("cluster_name").(string)
+	archives, _, err := conn.OnlineArchives.List(context.Background(), projectID, clusterName, &matlas.ListOptions{})
```
andreaangiolillo, line 182: [q] Could we keep the names of the variables consistent? I am happy with `archives` but then we should use `archive` instead of `outOnlineArchive` in `dataSourceMongoDBAtlasOnlineArchiveRead`

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	// clean up criteria for empty values
+	for key, val := range criteria {
+		if isEmpty(val) {
+			delete(criteria, key)
+		}
+	}
+
+	schemaVals["criteria"] = criteria
```
andreaangiolillo: [q] Is there a reason why we do not set `schemaVals["criteria"] = criteria` as `schemaVals["criteria"] = []interface{}{criteria}` here instead of doing this change in several other places 🤔 ?

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+
+	return schemaVals
+}
+
+func fromOnlineArchiveToMapInCreate(in *matlas.OnlineArchive) map[string]interface{} {
+	localSchema := fromOnlineArchiveToMap(in)
+	criteria := localSchema["criteria"]
+	localSchema["criteria"] = []interface{}{criteria}
+
+	delete(localSchema, "partition_fields")
```
andreaangiolillo, line 372: [q] Why do we delete `partition_fields` from the Map?

mongodbatlas/resource_mongodbatlas_online_archive.go
andreaangiolillo:
```suggestion
		return nil, errors.New("import format error to import a MongoDB Atlas Online Archive, use the format {project_id}-{cluster_name}-{archive_id}")
```

leofigy: agree :) modified to be archive_id

leofigy:
Hi @themantissa  added the documentation, and the PR comments. about the question of the testing yes, it's pending. we need another PR, because we need to think how to inject data into the cluster (or fake data) to allow the call to the API to work :) otherwise it will be pending forever

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+		DBName:   d.Get("db_name").(string),
+		CollName: d.Get("coll_name").(string),
+	}
+
+	requestInput.Criteria = mapCriteria(d)
+
+	if partitions, ok := d.GetOk("partition_fields"); ok {
+		list := partitions.([]interface{})
+
+		if len(list) > 0 {
```
andreaangiolillo, line 249: [nit] I would create a function to return a `partitionList` so that this piece of code is cleaner (too many nested ifs). Something like:



```
func mapToArchivePayload(d *schema.ResourceData) matlas.OnlineArchive {
	// shared input
	requestInput := matlas.OnlineArchive{
		DBName:   d.Get("db_name").(string),
		CollName: d.Get("coll_name").(string),
	}

	requestInput.Criteria = mapCriteria(d)

	if partitions, ok := d.GetOk("partition_fields"); ok {
		list := partitions.([]interface{})

		if len(list) == 0{
			return requestInput
		}

		requestInput.PartitionFields = partionList(list)
		
	}

	return requestInput
}
```

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+
+	atlasID := ids["atlas_id"]
+	projectID := ids["project_id"]
+	clusterName := ids["cluster_name"]
+
+	// if the criteria or the paused is enable then perform an update
+	paused := d.HasChange("paused")
+	criteria := d.HasChange("criteria")
+
+	// nothing to do, let's go
```
andreaangiolillo, line 290: [nit] I think that this comment is not useful and can be understood with the code : )
```suggestion
```

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	if !paused && !criteria {
+		return nil
+	}
+
+	request := matlas.OnlineArchive{}
+
+	// reading current value
```
andreaangiolillo, line 297:
```suggestion
```

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	// computed attribute
+	schemaVals := map[string]interface{}{
+		"cluster_name": in.ClusterName,
+		"atlas_id":     in.ID,
+		"paused":       in.Paused,
+		"state":        in.State,
+		"coll_name":    in.CollName,
+	}
+
+	// criteria
```
andreaangiolillo:
```suggestion
```

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+				"field_type": partition.FieldType,
+				"order":      partition.Order,
+			}
+
+			expected = append(expected, partition)
+		}
+		schemaVals["partition_fields"] = expected
+	}
+
+	return schemaVals
```
andreaangiolillo: [nit]
```suggestion
	if len(in.PartitionFields) == 0 {
		return schemaVals
	}
	
	partitionFieldsMap := make([]map[string]interface{}, 0, len(in.PartitionFields))
	for _, field := range in.PartitionFields {
		if field == nil {
			continue
		}

		fieldMap := map[string]interface{}{
			"field_name": field.FieldName,
			"field_type": field.FieldType,
			"order":      field.Order,
		}

		partitionFieldsMap = append(partitionFieldsMap, fieldMap)
	}
	
	schemaVals["partition_fields"] = partitionFieldsMap
	

	return schemaVals
```

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+		return v == ""
+	case *string:
+		if v == nil {
+			return true
+		}
+		return *v == ""
+	}
+
+	return false
+}
```
andreaangiolillo:
```suggestion
func isEmpty(val interface{}) bool {
	if val == nil {
		return true
	}

	switch v := val.(type) {
	case *bool, *float64, *int64:
		if v == nil {
			return true
		}
	case string:
		return v == ""
	case *string:
		if v == nil {
			return true
		}
		return *v == ""
	}

	return false
}
```

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,414 @@
...
+	localSchema["criteria"] = []interface{}{criteria}
+
+	delete(localSchema, "partition_fields")
+	return localSchema
+}
+
+func mapCriteria(d *schema.ResourceData) *matlas.OnlineArchiveCriteria {
+	criteriaList := d.Get("criteria").([]interface{})
+
+	criteria := criteriaList[0].(map[string]interface{})
```
andreaangiolillo, line 379: [q] Here there is not check that `criteriaList` has at least 1 element. Is this check somewhere else or I am missing something here?

andreaangiolillo: I think this check is done only in `resourceMongoDBAtlasOnlineArchiveUpdate` but not in `mapToArchivePayload`

mongodbatlas/data_source_mongodbatlas_online_archive.go
```diff
@@ -0,0 +1,212 @@
+package mongodbatlas
+
+import (
```
andreaangiolillo, line 3: [q] Why don't we have a test file for `data_source_mongodbatlas_online_archive.go` and `resource_mongodbatlas_private_endpoint.go` ? Is it because they need a cluster up and running? if yes, Don't we have unit tests? Thanks!!

andreaangiolillo (commented):
Great work on this PR 💯 !! The main thing that I think should be changed is `atlas_id` with `archive_id` as Melissa suggested. Besides that, I have only some questions and nits.

mongodbatlas/resource_mongodbatlas_online_archive.go
leofigy: Hello for this one, there is an issue because partition_fields is a list (it's order aware), For example the API returns a third element that is used in the date_field to query the collection and do the archive log. having a change of order sometimes from the API vs what you requested from the config leads to a config conflict.  So I preferred to honor the input from the user :)

andreaangiolillo: Thank you very much! : )

abner-dou (approved):
LGTM

mongodbatlas/resource_mongodbatlas_online_archive.go
leofigy: Hello @andreaangiolillo for this validation of the criteria, it should be in the schema definition, let me add the MinItems

mongodbatlas/resource_mongodbatlas_online_archive.go
themantissa: thank you!

themantissa (approved):
thank you for the changes!   approving - just need @andreaangiolillo 's review with the changes as well.  If you'd like on the test front could see if the online archive team might have some ideas.  Just let me know.

mongodbatlas/data_source_mongodbatlas_online_archive.go
leofigy: Hello @andreaangiolillo, in the provider most of the tests are [Acceptance Tests](https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html) to actually create the infra through the go client (mongodbatlas). For this case the limitation is that I need the cluster with data already preseeded . We could automate it like create the cluster, add another client to create the collection and insert some dummy data, but that would mean that just this test could take like ~30 min and I will need to include extra dependencies not intended for the provider itself. :)

mongodbatlas/resource_mongodbatlas_online_archive.go
leofigy: yea I know it's ugly. yea it's a leftover let me remove code from other parts, the reason behind it was the re-use of functions for both datasource and resource. So I had some values sometimes for datasource and others in resource, so I had to manipulate the data. But I updated the client, so no longer needed :)

andreaangiolillo (approved):
LGTM, thanks for all the changes!

PR #418, opened 2021-03-06
INTMDB-185: Added parameter regions for GCP network container

coderGo93:
## Description

- Added a parameter `regions` for GCP in resource/datasource(s) of network container
- Added test `TestAccResourceMongoDBAtlasNetworkContainer_WithRegionsGCP`
- Added test `TestAccDataSourceMongoDBAtlasNetworkContainer_WithGCPRegions`
- Added test `TestAccDataSourceMongoDBAtlasNetworkContainer_WithGCPRegions`

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments
Before the atlas client should be released first so I can updated the go.mod with official tag instead of auto generated with `go get`


themantissa (approved):
LGTM - needs DOU review and good to merge.  Thank you!

mongodbatlas/data_source_mongodbatlas_network_container.go
```diff
@@ -61,6 +61,13 @@ func dataSourceMongoDBAtlasNetworkContainer() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"regions": {
```
leofigy, line 64: I know it's a data source but question, this will returned by all the types ?

mongodbatlas/data_source_mongodbatlas_network_container.go
```diff
@@ -124,6 +131,10 @@ func dataSourceMongoDBAtlasNetworkContainerRead(d *schema.ResourceData, meta int
 		return fmt.Errorf("error setting `vnet_name` for Network Container (%s): %s", d.Id(), err)
 	}
 
+	if err = d.Set("regions", container.Regions); err != nil {
```
leofigy, line 134: I mean , if it's azure, aws or gcp, the api returns always the regions field? (even empty)

mongodbatlas/data_source_mongodbatlas_network_containers.go
```diff
@@ -70,6 +70,13 @@ func dataSourceMongoDBAtlasNetworkContainers() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
+						"regions": {
```
leofigy, line 73: same here

leofigy (changes requested):
LGTM just confirm the regions field is not affected in the set for azure or gcp. 

And do a negative testing , create an azure and set the regions field.
I know it's late , but all these fields with the same names , 
region, region_name, regions is really confusing 

would not be better to have like an gcp_regions , or subobject for

gcp {
    regions = []
}

mongodbatlas/data_source_mongodbatlas_network_container.go
coderGo93: No, if not using, it will not appear in response neither in state, for now it can be only used for gcp, for azure is the singular `region`

coderGo93:
> LGTM just confirm the regions field is not affected in the set for azure or gcp.
> 
> And do a negative testing , create an azure and set the regions field.
> I know it's late , but all these fields with the same names ,
> region, region_name, regions is really confusing
> 
> would not be better to have like an gcp_regions , or subobject for
> 
> gcp {
> regions = []
> }


You can put regions using azure but it will not send in request because it's validated to only in gcp like azure with region or aws with region_name in create function, also if you do that you could get an update-in-change behavior in the re apply.

mongodbatlas/data_source_mongodbatlas_network_container.go
leofigy: yea, I see the client code, curious do we get any warning, because moving from nil to empty type?
	Regions             []string `json:"regions,omitempty"`    // Regions are available for GCP

leofigy (approved):

PR #434, opened 2021-04-16
INTMDB-199: Fixes the error when updating an replication specs after removed one zone

coderGo93:
## Description
Changed to TypeSet for `replication_specs` schema, added a validation to compare the zone name if it's the same for old and new values and the get the ID generated from original object and set it in request for update to avoid an error like `Updating id of existing replication spec`

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -223,7 +223,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Computed: true,
 			},
 			"replication_specs": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				Computed: true,
 				Elem: &schema.Resource{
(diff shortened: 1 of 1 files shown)


themantissa (approved):
LGTM, @leofigy please review to ensure this covers INTMDB-199.  I think so but want to be sure.  Thanks!

leofigy:
Hello @themantissa , @coderGo93 

After spending a while, I think we must roll back the type to an array, instead of a Set. 
I know the diff is awful, but we're taking care finding the IDs to fix the patch request. 

I played with Set, but the type is even worse because it generates a diff everytime because the computed attributed even if you only change the order and the items remain the same, it seems the hash is not that good. 

I was trying to see if we could skip using DiffSuppressFunc. but it didn't work because does not evaluate the composed attribute at the top :(

leofigy (changes requested):
Hello please rollback to use the TypeList, and just make sure we keep retaining the ID

leofigy (approved):
LGTM thanks

themantissa:
Just ack'ing that I still approve with the 2nd commit with new changes.  Thank you

Issue #189, opened 2020-04-08
Cluster not renamed when the name changes


haidaraM:
Looks like the cluster is not renamed when we change its name. Terraform seems to successfully update the name but nothing changes in Atlas Console. 

The doc says that the `name` can't be changed after cluster creation (https://www.terraform.io/docs/providers/mongodbatlas/r/cluster.html#name). So I guess the cluster should be recreated when its name changes :-) ?

themantissa:
@haidaraM - correct, the UI supports changing the cluster name but the API does not.   Terraform considers the name change a change of cluster and will delete the old cluster and create with the new name so be cautious and aware of this behavior.

haidaraM:
The issue is that currently Terraform doesn't delete the old cluster. The cluster keeps it's old name even though I apply the change. Terraform plan says that the resource will be updated in place.
Is this the expected behavior  ?

themantissa:
Interesting - let me test it out again.

themantissa:
Ah, I skimmed this too quickly.   The behavior I was describing is for Atlas Projects (you can change the name in the UI, not via API and if you try to it will delete the project and recreate - there is a warning we put in the docs) but this is for Clusters!  Cluster names are special and cannot be changed so this is a bit of a bug in the Provider!  It should error when you try to change the name but it looks like it is applying, but since it cannot apply that is shouldn't try.   @haidaraM deleted my incorrect comment and noted this a non-blocking bug.  It doesn't impact anything but may lead to false impressions of what changed.

andreykaipov:
> It should error when you try to change the name but it looks like it is applying, but since it cannot apply that is shouldn't try.

@themantissa Maybe I'm misunderstanding this, but is the expected behavior to have the provider force a new cluster, or for the Atlas API to respond with a 400 such that that the provider errors out?

I tried to change our cluster names yesterday, and Terraform accepted the plan, so I'm guessing the Atlas API incorrectly accepts the request. However, the API doesn't really need to change to improve the provider. I think https://github.com/mongodb/terraform-provider-mongodbatlas/pull/435 accomplishes this.

themantissa:
@andreykaipov based on your PR I tested and the API actually correctly returned an error when a cluster name was submitted via PATCH so we can actually change the provider to capture that error correctly (not sure why before I said it shouldn't try, it can and then return the error.)  I'm putting a ticket in to improve this but will take a bit of time as we have quite a bit on our near term roadmap.

[...]

PR #435, opened 2021-04-16
feature: force new cluster on name change

andreykaipov:
## Description

Currently changing the cluster name shows a plan with in-place changes. The plan can be applied, but no change from the UI is visible. This is misleading since Atlas doesn't let you change cluster names, and in my case even severed the connections to our clusters. I'm guessing this name is used internally by Atlas, and changing it broke some things.

The documentation for the provider says [once the cluster is created, its name cannot be changed](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/cluster#name), so this attribute should definitely be set to improve the user's experience.

Link to any related issue(s): https://github.com/mongodb/terraform-provider-mongodbatlas/issues/189

## Type of change:

- [x] Documentation fix/enhancement (in this case the documentation is already correct! 😄)

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -140,6 +140,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			},
 			"name": {
 				Type:     schema.TypeString,
+				ForceNew: true,
 				Required: true,
 			},
 			"mongo_db_major_version": {


andreykaipov:
I didn't really run all of the acceptance tests, but I tested it out against our clusters and it's the expected behavior.

Upon changing the name...

Before:
image
After:
image

themantissa:
(edit) @andreykaipov thank you for the submission.   Considering that simply changing the name would force a replacement I think it makes more sense to properly error on the API side when trying to change the name vs having Terraform do a replacement, which will likely end up with some inadvertently destroying their clusters.    

Actually just checked and the API returns:
{
  "detail" : "Cannot change name of cluster during an update.",
  "error" : 400,
  "errorCode" : "CLUSTER_CANNOT_CHANGE_NAME",
  "parameters" : [ ],
  "reason" : "Bad Request"
}
Let me check on if we can better surface this up.

andreykaipov:
Regardless of why the provider isn't properly propagating the error, are you saying the preferred behavior in this situation is to have Terraform misleadingly show an in-place change, but error out when actually applying the plan?

Terraform's purpose is to abstract infrastructure with declarative configuration, and the purpose of a plan is to show [what actions are to be performed to achieve the desired state in our configuration](https://www.terraform.io/docs/cli/commands/plan.html). In this situation, the correct action for Terraform to take is to perform a destroy first, since the name within Atlas is immutable, so that's what the plan should show.

If the concern is users inadvertently destroying their clusters, then that's understandable, but that's what reviewing the plan is for! I'd argue those using Terraform understand the risks of not properly reviewing the plans they're applying. However, there's not much merit in doing a review when the plan is intentionally misleading.

leofigy:
Hello @andreykaipov 

The intend is to error out, when we detect a name change in the update request if is an existent resource. I got your point that people should check a plan. and the ForceNew would be the purest way to fix this. But there is always a trade off.

themantissa:
Closing in favor of internal ticket INTMDB-201 to address this issue.

andreykaipov:
I understand. Thanks for considering it.

On the off-chance someone else is considering renaming their cluster names, I recommend updating the relevant docs to something like this instead then:

> `name` - (Required) Name of the cluster as it appears in Atlas. Once the cluster is created, its name cannot be changed. If you'd like to change the name, you'll first need to destroy your cluster.

I can submit a docs PR if y'all are open to that instead.

themantissa:
I understand what you are trying to achieve @andreykaipov but I think this might lead some to the wrong understanding.   Destroying the cluster does not change the name, it allows you to create a new cluster with a different name.  This is slightly different in that if someone has data in the cluster that data is gone when destroyed.    I believe as it is now, along with surfacing the correct error, should be sufficient.    But thank you for being willing to help!

PR #463, opened 2021-05-20
INTMDB-200: Fixes a bug about updating a region name with GCP

coderGo93:
## Description

Fixes the bug of `provider_region_name` are recognized but not applied properly

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -1180,14 +1180,17 @@ func expandReplicationSpecs(d *schema.ResourceData) ([]matlas.ReplicationSpec, e
 			}
 
 			replaceRegion := ""
+			originalRegion := ""
 
 			if okPRName && d.Get("provider_name").(string) == "GCP" && cast.ToString(d.Get("cluster_type")) == "REPLICASET" {
 				if d.HasChange("provider_region_name") {
 					replaceRegion = vPRName.(string)
+					original, _ := d.GetChange("provider_region_name")
+					originalRegion = original.(string)
 				}
 			}
 
-			regionsConfig, err := expandRegionsConfig(spec["regions_config"].(*schema.Set).List(), replaceRegion)
+			regionsConfig, err := expandRegionsConfig(spec["regions_config"].(*schema.Set).List(), originalRegion, replaceRegion)
 			if err != nil {
 				return rSpecs, err
 			}
@@ -1221,7 +1224,7 @@ func flattenReplicationSpecs(rSpecs []matlas.ReplicationSpec) []map[string]inter
 	return specs
 }
 
-func expandRegionsConfig(regions []interface{}, replaceRegion string) (map[string]matlas.RegionsConfig, error) {
+func expandRegionsConfig(regions []interface{}, originalRegion, replaceRegion string) (map[string]matlas.RegionsConfig, error) {
 	regionsConfig := make(map[string]matlas.RegionsConfig)
 
 	for _, r := range regions {
@@ -1232,7 +1235,7 @@ func expandRegionsConfig(regions []interface{}, replaceRegion string) (map[strin
 			return regionsConfig, err
 		}
 
-		if replaceRegion != "" {
+		if replaceRegion != "" && r == originalRegion {
 			r, err = valRegion(replaceRegion)
 		}
 		if err != nil {
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -1181,7 +1181,7 @@ func expandReplicationSpecs(d *schema.ResourceData) ([]matlas.ReplicationSpec, e
 
 			replaceRegion := ""
 
-			if okPRName && d.Get("provider_name").(string) == "GCP" {
+			if okPRName && d.Get("provider_name").(string) == "GCP" && cast.ToString(d.Get("cluster_type")) == "REPLICASET" {
 				if d.HasChange("provider_region_name") {
 					replaceRegion = vPRName.(string)
 				}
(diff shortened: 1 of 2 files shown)


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1175,7 +1178,16 @@ func expandReplicationSpecs(d *schema.ResourceData) ([]matlas.ReplicationSpec, e
 					id = oldID
 				}
 			}
-			regionsConfig, err := expandRegionsConfig(spec["regions_config"].(*schema.Set).List())
+
```
leofigy, line 1181: I just have one doubt in the logic, this is suppose to work with just one region, but of course it works, my only concern is that the change detection is inside the expand , of course it's called by the Update. but no sides effects when this is more than one region right?

leofigy (approved):

mongodbatlas/resource_mongodbatlas_cluster.go
themantissa: @coderGo93 did this concern get addressed?

coderGo93: Hi, I tried to add another region and it didn't work because the configuration was not correct from API error response, but when I tested, it worked in request but somehow after it's over in response for replication specs didn't change but it worked for region_name and with re plan it appears not changes.

themantissa: @coderGo93  not sure I'm following - so when you tested more than one region it worked but the API still errored?

coderGo93: @themantissa Not because it seems it's not possible to do that, the error is like this `400 (request "INVALID_CLUSTER_CONFIGURATION") The specified cluster configuration is not valid.` or Unexpected error , I guess that happens when you set region name along with two replication specs and cluster type or maybe I did something wrong in config, but it works without appearing changes after re plan with the config example from the ticket INTMDB-200 without needing to use the replication specs schema.
BTW I tried the second config example from the ticket with replication specs and still worked for me

leofigy: Hi @coderGo93 

Let's try to clarify all doubts.

**The use case**

if someone creates with only one region, on behalf of the user we populate the replication_specs with just one region.

**The issue** 

If the customer just updates the region name, we didn't update the only region in their replication specs. So that was the fix. 

**The concerns.** 

The fix works, but take in mind we do have different type of clusters , from the documentation https://docs.atlas.mongodb.com/reference/api/clusters-create-one/, and the code fix is inside the expand function so it does not validate the cluster type. So let's say someone by mistake starts playing in GCP with the provider region name, and you detect that change in the expand function you will overwrite the zones for all the items. 

So before merging this two options move your fix outside the expand function or in the other hand, just validate that the API actually complains :) . otherwise we will end up with surprises.

abner-dou (approved):
LGTM

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: Pushed changes by adding validation when cluster type is REPLICASET along if the provider is GCP in update. Also I added to check if the original region is equal to one element of replication specs so it will replace one element instead of all regions.
Steps:
1. Create a cluster with `provider_region_name`
2. After created, update the `provider_region_name`
3. Wait for update to be finished, it can take a while from 10 min to 1hr
4. After updated, with terraform plan shows no changes infrastructure

Config file used for testing
```hcl
resource "mongodbatlas_cluster" "terraform_cluster3" {
  project_id = var.project_id
  name       = "test-cluster1"
  mongo_db_major_version       = "4.4"
  provider_name               = "GCP"
  disk_size_gb                = 10
  num_shards = 1
  provider_region_name        = "US_WEST_2"
}

```

andreaangiolillo: @coderGo93  Could you also test that everything is working for clusterType `SHARDED` and `GEOSHARDED` with more than 1 shard, please?

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1175,7 +1178,19 @@ func expandReplicationSpecs(d *schema.ResourceData) ([]matlas.ReplicationSpec, e
...
+			replaceRegion := ""
+			originalRegion := ""
+
+			if okPRName && d.Get("provider_name").(string) == "GCP" && cast.ToString(d.Get("cluster_type")) == "REPLICASET" {
```
andreaangiolillo, line 1185: [q] I am not sure why we have the check only for `REPLICASET`

andreaangiolillo, line 1185: [q] What would happen when the type is `SHARDED`?

andreaangiolillo (commented):
I have just some questions : )

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: For my understanding, since the user created a basic cluster, it seems in response it creates automatically with cluster type `REPLICASET` along the replication specs , this why I added validation, for this special scenario

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: Yes, I just tested and it worked for me, it only will occur when the provider is GCP and  cluster type is REPLICASET with additional code when it should replace one element instead of all elements in update.

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: if the provider is not GCP and the cluster type is not replicate set(since its automated somehow in api response) with the user config, it will not update the region name in replication specs as it commented in ticket

andreaangiolillo: Thanks!

mongodbatlas/resource_mongodbatlas_cluster.go
andreaangiolillo: Thank you very much!

andreaangiolillo (approved):
LGTM

themantissa (commented):
LGTM - looks ready to merge.

Issue #320, opened 2020-09-25
500 response on /groups/GROUP-ID/peers


igaskin:
```

Error: error creating MongoDB Network Peering Connection: POST https://cloud.mongodb.com/api/atlas/v1.0/groups/GROUP-ID/peers: 500 (request "Internal Server Error") Unexpected error.

  on atlas.tf line 111, in resource "mongodbatlas_network_peering" "eu_west_1":
 111: resource "mongodbatlas_network_peering" "eu_west_1" {
```

### Terraform CLI and Terraform MongoDB Atlas Provider Version

```hcl
terraform version
Terraform v0.13.3
+ provider registry.terraform.io/-/aws v3.8.0
+ provider registry.terraform.io/-/mongodbatlas v0.6.5
+ provider registry.terraform.io/-/random v2.3.0
+ provider registry.terraform.io/hashicorp/aws v3.8.0
+ provider registry.terraform.io/hashicorp/random v2.3.0
+ provider registry.terraform.io/mongodb/mongodbatlas v0.6.5```

### Terraform Configuration File 

```hcl
resource "mongodbatlas_network_peering" "eu_west_1" {
  accepter_region_name   = "eu-west-1"
  project_id             = var.project_id
  provider_name          = var.provider_name
  aws_account_id         = data.aws_caller_identity.default.account_id
  vpc_id                 = var.vpc_ids["eu_west_1"]
  route_table_cidr_block = var.vpc_cidrs["eu_west_1"]
  container_id           = mongodbatlas_network_container.eu_west_1.id
}```

### Steps to Reproduce
1. `terraform init`
2. `terraform apply`

### Expected Behavior
Succeed with 200

### Actual Behavior
Fails with 500

nikhil-mongo:
@igaskin Please refer to the linked [repo](https://github.com/nikhil-mongo/aws-atlas-terraform) for setting up the VPC Peering from Atlas to AWS.
You may need to change some attributes' values. Let us know if this helps you.
Thanks

igaskin:
@nikhil-mongo thanks for sharing the sample repo.  The code in the repo matches the attributes that I've set on my end as well.  If you are unable to reproduce the issue, then feel free to close this issue.  I was never actually able to successfully run this end-to-end.  I had to manually create resources and import them into my terraform state

nikhil-mongo:
[...]

PR #471, opened 2021-06-04
INTMDB-128: Modified design when you can get .id from various resources

coderGo93:
- Refactor when you can get id if you are using encoded state id or only id like `container_id` from various resource

## Description

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments
#320


leofigy:
Okay adding the explanation here.
The idea is to be able to support both the regular resource ID, and decode on-behalf of the user the resource ID that is embedded in the internal ID. because there have been a lot of confusions, on which ID to use.

leofigy (changes requested):
Hi @coderGo93 awesome work, just one request could we create a terratest for all the modified resources, just for the sake of sanity. 

And in addition to this could we add a datasource for project , and re-use the id of that datasource in the other resources.

mongodbatlas/resource_mongodbatlas_privatelink_endpoint_service_test.go
```diff
@@ -80,10 +80,11 @@ func TestAccResourceMongoDBAtlasPrivateLinkEndpointServiceAWS_import(t *testing.
 				),
 			},
 			{
-				ResourceName:      resourceName,
-				ImportStateIdFunc: testAccCheckMongoDBAtlasPrivateLinkEndpointServiceImportStateIDFunc(resourceName),
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            resourceName,
+				ImportStateIdFunc:       testAccCheckMongoDBAtlasPrivateLinkEndpointServiceImportStateIDFunc(resourceName),
```
abner-dou, line 84: I have a question, is it necessary to use so long func names?

I mean, in this method you are in `mongodbatlas_privatelink_endpoint...`method, maybe the name for this func can be `testAccCheckServiceImportStateIDFunc`

having names like this can affect the readability and the maintainability

abner-dou (approved):
LGTM!
(I just left one question, but is not so important)

mongodbatlas/resource_mongodbatlas_privatelink_endpoint_service_test.go
coderGo93: As long it doesn't cross the limit of the line-length from golangci.yml it should be fine, mm it seems readable for me since it's usually testAccCheckMongoDBAtlas[Resource/datasource][Extra(like ImportState, destroy, etc] something like that, but it somehow it crosses the limit, yes it would be nice to shorten it and still readable for the resource it belongs to

threebee (approved):
LGTM!

coderGo93:
> Hi @coderGo93 awesome work, just one request could we create a terratest for all the modified resources, just for the sake of sanity.
> 
> And in addition to this could we add a datasource for project , and re-use the id of that datasource in the other resources.

Done, thank you for the suggestion @leofigy

PR #417, opened 2021-02-25
TeamsUpdate - fixing small bug, again missing update

leofigy:
## Description

Missing refreshing the state when the team is removed from the outside terraform, but still in the configuration file

Link to any related issue(s):

## Type of change:

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [X] I have read the Terraform contribution guidelines
- [X] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [X] I have added any necessary documentation (if appropriate)
- [X] I have run make fmt and formatted my code

PR #494, opened 2021-07-19
INTMDB-188: fixed issue related with read non-existing resource

abner-dou:
## Description

Read methods fixed in resources, there was a bug when refreshing the state when the resource is removed from the outside terraform, but still in the configuration file
Related to [TeamsUpdate - fixing small bug, again missing update](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/417)

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cloud_provider_access.go
+++ b/mongodbatlas/resource_mongodbatlas_cloud_provider_access.go
@@ -3,6 +3,7 @@ package mongodbatlas
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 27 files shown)


themantissa (approved):
LGTM, thank you!

mongodbatlas/resource_mongodbatlas_cloud_provider_access.go
```diff
@@ -107,9 +108,14 @@ func resourceMongoDBAtlasCloudProviderAccessRead(d *schema.ResourceData, meta in
...
 
-	roles, _, err := conn.CloudProviderAccess.ListRoles(context.Background(), projectID)
+	roles, resp, err := conn.CloudProviderAccess.ListRoles(context.Background(), projectID)
 
 	if err != nil {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			d.SetId("")
+			return nil
+		}
+
```
andreaangiolillo, line 118: [nit] These 3 lines are used in other places as well. Maybe we can wrap this piece of code in a function a call it where is needed.

andreaangiolillo (approved):
LGTM

thetonymaster (approved):

mongodbatlas/resource_mongodbatlas_cloud_provider_access.go
abner-dou: I think this validation is so little that even If we create a single function, we will need to verify if return an error and will code exactly other 3 lines
e.g.

```
 func validateResponce(resp  *mongodbatlas.Response) err {
 if resp != nil && resp.StatusCode == http.StatusNotFound {
			d.SetId("")
			return nil
		}
 return fmt.Errorf("is not empty")
 }
 
 
 
 if err:=validateResponse(resp), err==nil{
 return nil
 }
 

```

what do you think @andreaangiolillo ?

abner-dou: I want to avoid overengineering :D

andreaangiolillo: SGTM, mine was more something to think of. Thanks!

PR #489, opened 2021-07-01
INTMDB-183: Migrate to TF SDK 2

coderGo93:
## Description
- Migrated to SDK Version 2
- Because of bad practice and is not supported for sdk v2, using TypeMap as object will show error with `make test`, so the fix was adding another attribute and deprecate the old attribute, and for the new attributes was changed to TypeList with MaxItem 1, resources/datasources affected are `alert_configuration`, `cloud_provider_snapshot_restore_job`, `mongodbatlas_encryption_at_rest`

Link to any related issue(s):

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Documentation fix/enhancement

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have read the Terraform contribution guidelines
- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code

PR #496, opened 2021-07-23
INTMDB-202: Changed to TypeSet for replication specs

coderGo93:
## Description

- Changed to TypeSet for `replication_specs` and use a custom hashing because since `id` its computed, it will always detect diff, so it's using a custom hashing without id
- Added `TestAccResourceMongoDBAtlasCluster_RegionsConfig` 

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments
WIll wait to #489  to be merged first, since this PR it's rebased from there


themantissa:
Just adding a note this is on hold till #489 is fully reviewed and merged so this will not include the 167 files impacted by the SDK upgrade.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -236,8 +235,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 								Schema: map[string]*schema.Schema{
 									"region_name": {
 										Type:     schema.TypeString,
-										Optional: true,
-										Computed: true,
+										Required: true,
```
threebee, line 238: [Q] Earlier this was `Optional` and now it is changed to `Required`. Will this break if clients do not pass `region_name`?

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -1626,3 +1711,27 @@ func testAccMongoDBAtlasClusterConfigGCPRegionName(
...
+	  provider_backup_enabled = true
+	  cluster_type            = "GEOSHARDED"
+	  // Provider Settings "block"
+	  provider_name               = "AWS"
+	  provider_disk_iops          = 1200
+	  provider_instance_size_name = "M30"
+	  %[3]s
+
+		lifecycle {
+		# avoid cluster has been auto-scaled to different instance size
```
threebee, line 1732: [nit] Seem like this comment has a typo? `# avoid cluster has been auto-scaled to different instance size` -> `# avoid cluster has been auto-scaled to different instance size`?

threebee (commented):
Looking good, I have added a question and a nit.

mongodbatlas/resource_mongodbatlas_cluster.go
coderGo93: Because I re-read again the docs and it seems it's required to use the region, so I changed to required, I can roll back to optional if you want. Since the region name it's using as the key name, if you don't put the name the key would be empty along with the field and values, even with optional you got an error saying that the region must be set if you don't set `region_name`

mongodbatlas/resource_mongodbatlas_cluster_test.go
coderGo93: I don't get it, what typo? :thinking:

threebee: I made a typo hahah, I think it should be `# avoid cluster been auto-scaled to different instance size`?

mongodbatlas/resource_mongodbatlas_cluster.go
threebee: This makes sense. I think we should keep it `Required` then.

threebee (approved):
LGTM!

thetonymaster (approved):
LGTM

Issue #482, opened 2021-06-24
Scaling max/min is applied each time with disabled autoscaling


aeimer:
### Terraform CLI and Terraform MongoDB Atlas Provider Version

```hcl
Terraform v0.15.3
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azurerm v2.63.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
+ provider registry.terraform.io/mongodb/mongodbatlas v0.9.1
```

### Terraform Configuration File 

```hcl
resource "mongodbatlas_cluster" "main" {
  for_each = local.mongodb_clusters

  name       = each.value
  project_id = mongodbatlas_project.main.id

  auto_scaling_compute_enabled                    = var.mongodbatlas_config[each.value].mongodbatlas_autoscale # Set to false
  auto_scaling_compute_scale_down_enabled         = var.mongodbatlas_config[each.value].mongodbatlas_auto_down_scale # Set to false
  auto_scaling_disk_gb_enabled                    = true
  cluster_type                                    = "REPLICASET"
  mongo_db_major_version                          = "4.4"
  provider_auto_scaling_compute_max_instance_size = var.mongodbatlas_config[each.value].mongodbatlas_instance_max # Set to M10
  provider_auto_scaling_compute_min_instance_size = var.mongodbatlas_config[each.value].mongodbatlas_instance_min # Set to M10
  provider_backup_enabled                         = true
  provider_instance_size_name                     = var.mongodbatlas_config[each.value].mongodbatlas_instance_min
  provider_name                                   = "AZURE"

  advanced_configuration {
    minimum_enabled_tls_protocol = "TLS1_2"
  }

  replication_specs {
    num_shards = 1
    dynamic "regions_config" {
      for_each = local.mongodb_regions
      content {
        region_name     = regions_config.value["atlasregion"]
        electable_nodes = regions_config.value["electable_nodes"]
        priority        = regions_config.value["priority"]
      }
    }
  }

  labels {
    key   = "environment"
    value = var.environment_name
  }

  lifecycle {
    ignore_changes = [
      provider_instance_size_name
    ]
  }
}

```

### Expected Behavior

The provider should not find any changes to be done.
Especially as the update needs quite a few seconds to minutes, it would be nice to be skipped.

### Actual Behavior

The provider show `provider_auto_scaling_compute_max_instance_size` and `provider_auto_scaling_compute_min_instance_size` to be `not set`?!.

[...]

PR #515, opened 2021-08-16
INTMDB-219: Fixed cluster scaling issue

abner-dou:
## Description

Fix an issue in mongodbatlas_cluster resource, this issue was showing pending changes when the scale was set on `false` but with values in `scaling_size` in configuration file

now, a validation was implemented in order to verify if the scaling is enabled, if don't, terraform will ignore the scaling_size

more information on [issue 482](https://github.com/mongodb/terraform-provider-mongodbatlas/issues/482)

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -1526,9 +1526,14 @@ func clusterConnectionStringsSchema() *schema.Schema {
 }
 
 func isEqualProviderAutoScalingMinInstanceSize(k, old, newStr string, d *schema.ResourceData) bool {
-	canScaleDown, _ := d.GetOk("auto_scaling_compute_scale_down_enabled")
-	canScaleUp, _ := d.GetOk("auto_scaling_compute_enabled")
-	if canScaleDown != nil && canScaleUp != nil && canScaleUp.(bool) && canScaleDown.(bool) {
+	canScaleDown, scaleDownOK := d.GetOk("auto_scaling_compute_scale_down_enabled")
+	canScaleUp, scaleUpOk := d.GetOk("auto_scaling_compute_enabled")
+
+	if !scaleDownOK || !scaleUpOk {
+		return true //if the return is true, it means that both values are the same and there's nothing to do
+	}
+
+	if canScaleUp.(bool) && canScaleDown.(bool) {
 		if old != newStr {
 			return false
 		}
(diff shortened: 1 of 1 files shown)


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1525,3 +1524,24 @@ func clusterConnectionStringsSchema() *schema.Schema {
 		},
 	}
 }
+
+func isEqualProviderAutoScalingMinInstanceSize(k, old, newStr string, d *schema.ResourceData) bool {
+	canScaleDown, _ := d.GetOk("auto_scaling_compute_scale_down_enabled")
+	canScaleUp, _ := d.GetOk("auto_scaling_compute_enabled")
+	if canScaleDown != nil && canScaleUp != nil && canScaleUp.(bool) && canScaleDown.(bool) {
```
andreaangiolillo: [q] Is `canScaleUp.(bool) && canScaleDown.(bool)` really needed? is it not enough to check that is not `nil`?

andreaangiolillo: [q] why aren’t we checking that `canScaleDown == true || canScaleUp== true`?

andreaangiolillo (approved):
LGTM, I have added a few questions.

mongodbatlas/resource_mongodbatlas_cluster.go
abner-dou: I'm checking if there's not nil (the result of the `GetOk` is an interface) before try to evaluate the variables,  and also when we evaluate the var `canScaleDown` I chosse the `AND` operator because in order to use the canScaleDown we need to have the variable `canScaleUp` as true according to the documentation: 

> auto_scaling_compute_scale_down_enabled - (Optional) Set to true to enable the cluster tier to scale down. This option is only available if autoScaling.compute.enabled is true.

thetonymaster: Nitpick: can we simplify this? kinda hard to read and yeah well it's a four conditions if

thetonymaster (approved):

mongodbatlas/resource_mongodbatlas_cluster.go
abner-dou: I splitted into 2 blocks this condition

PR #524, opened 2021-08-23
Support for organization API Keys

vicgp-te:
## Description

This PR adds support to create organization API keys: mongodbatlas_organization_api_key

This creates organization API keys with roles and access list, tests have been added covering all cases and tested with my own Atlas account, import also works.

Note that this does not support the assignments of API keys to projects, i'm working on that at this moment and will be a separate resource.

Link to any related issue(s):

## Type of change:

- [X] New feature (non-breaking change which adds functionality)

## Required Checklist:

## Further comments

Please double check that i'm not missing anything in `website/mongodbatlas.erb`, i don't see all entries there so not sure if i need to add this new one.

### DIFF
--- a/mongodbatlas/data_source_mongodbatlas_organization_api_key.go
+++ b/mongodbatlas/data_source_mongodbatlas_organization_api_key.go
@@ -1,0 +1,96 @@
+package mongodbatlas
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceMongoDBAtlasOrganizationApiKey() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceMongoDBAtlasOrganizationApiKeyRead,
+		Schema: map[string]*schema.Schema{
+			"org_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"api_key_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"roles": {
+				Type:     schema.TypeSet,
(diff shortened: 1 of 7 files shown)


themantissa:
@vgarcia-te we have specifically not supported creation of API key creation in Terraform as we've added support via a secret engine in Vault - a much more secure way to generate API keys: https://www.vaultproject.io/docs/secrets/mongodbatlas   Is the more secure option not the preferred for your use case?

vicgp-te:
@themantissa API key creation using Vault is a good alternative, but for users that don't use Vault, like us, it is not worth to set it up only for this.
The idea of managing the API keys with Terraform is for full automation, we have practically everything in Terraform like i guess many companies also do. If deploying Atlas infrastructure requires dealing with 2 different systems, Terraform for everything and Vault for the keys, this doesn't become practical.

Plus, from what i can see from that link, having to create keys in Vault is manual, which is what we are trying to avoid. Recently, we have found that this Terraform module is missing a few things that make us have to do a set of manual steps every time we deploy a new Atlas cluster, and we want to avoid that to have a full end to end Atlas deployment based on Terraform only. Because of this, soon you will see more contributions from us (we are ThousandEyes/Cisco).

I would really like to see the management of API keys in Terraform, like the rest, and offer 2 alternatives for users: Vault or Terraform.

themantissa:
@vgarcia-te thank you for the feedback.   Vault also has an API and can be automated but the examples, like Terraform, shows the CLI for ease of explanation.    My concern here is that any creation of an API key/pair in Terraform is going to put that secret in the state file which is not encrypted while Vault will store it securely within the software.   Hence it's can be quite a risk depending on the environment.   I'm happy to discuss further with your team to ensure this risk is well understood and how we can best proceed.

It would also be great to know what else is considered missing and we can tell you if it's on our roadmap or not.  Would you be able to share more?

Obviously this may be better handled via a meeting or email - feel free to reach out directly to me at melissa.plunkett@mongodb.com or I can reach out to schedule something if you can share your preferred contact method.

themantissa:
@vgarcia-te any objection to closing this PR out?

PR #544, opened 2021-08-27
test: skip instead of fatal for team ids missing

gssbzn:
## Description

A fix attempt for dependabot, it seems it doesn't have access to the env var so tests fail, I think it should be better to skip rather than fail 

## Type of change:


themantissa (approved):
You are my hero - this has been driving me crazy!  Thank you!

coderGo93 (approved):
Good work, I was going to do something like that haha, indeed, for some reason even if you set secrets in the fork repo won't either. So LGTM! thank you gustavo.

PR #663, opened 2022-01-24
INTMDB-293: Added container_id in advanced cluster

coderGo93:
## Description

Added `container_id` for resource and datasource of advanced cluster

Link to any related issue(s):

## Type of change:

## Required Checklist:


website/docs/r/advanced_cluster.html.markdown
```diff
@@ -246,6 +246,7 @@ replication_specs {
```
themantissa, line 248: In clusters we include the following note at the top of the docs page. We should do the same here.  https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/website/docs/r/cluster.html.markdown?plain=1#L15

website/docs/r/advanced_cluster.html.markdown
```diff
@@ -246,6 +246,7 @@ replication_specs {
...
 * `region_configs` - (Optional) Configuration for the hardware specifications for nodes set for a given regionEach `region_configs` object describes the region's priority in elections and the number and type of MongoDB nodes that Atlas deploys to the region. Each `region_configs` object must have either an `analytics_specs` object, `electable_specs` object, or `read_only_specs` object. See [below](#region_configs)
+* `region_configs_container_ids` - A key-value map for Network Peering Container ID with syntax as `"providerName:regionName" = "containerId"`. Example `AWS:US_EAST_1" = "61e0797dde08fb498ca11a71`.
```
themantissa: Why not call this container_id like we did with clusters (container_ids is fine too but really just keeping the same should be fine)?  It can still be a map but just explained via this doc, as you've done.  Something like: 

```suggestion
* `container_id` - A key-value map of the Network Peering Container ID. The Container ID is the id of the container either created programmatically by the user before any clusters existed in a project or when the first cluster in the region (AWS/Azure) or project (GCP) was created.  The syntax is `"providerName:regionName" = "containerId"`. Example `AWS:US_EAST_1" = "61e0797dde08fb498ca11a71`.

```

themantissa (changes requested):
Looks good but see comment on name.  Thanks!

website/docs/r/advanced_cluster.html.markdown
coderGo93: Because is not inside of regions_config, and is a type map because it can appear more containers ids if there's more region configs in the config file, and it will appear like this 
```
"AWS:US_EAST_1" = "61e0797dde08fb498ca11a71",
"AWS:US_EAST_2" = "61e0797dde08fb498ca11a72",
```
For some reason if I put the computed parameter `container_id` in the `regions_config` it will always appear a diff after plan even if you don't make changes, not sure why happens, maybe it's because is a TypeSet of another TypeSet, but after reading this [solution](https://discuss.hashicorp.com/t/custom-provider-how-to-reference-computed-attribute-of-typemap-list-set-defined-as-nested-block/22898/2) and I applied the same, and now it works without appearing a diff after plan.

themantissa: Okay I get why you would name it that because of how it works but that's still likely better handled in the description of the documentation imo, e.g.:

 `container_id` - A key-value map of the Network Peering Container ID(s) for the configuration specified in `region_configs`. The Container ID is the id of the container either created programmatically by the user before any clusters existed in a project or when the first cluster in the region (AWS/Azure) or project (GCP) was created.  The syntax is `"providerName:regionName" = "containerId"`. Example `AWS:US_EAST_1" = "61e0797dde08fb498ca11a71`.

coderGo93: Understood, will make changes, thank you.

themantissa (approved):
LGTM! Thank you!

mongodbatlas/resource_mongodbatlas_advanced_cluster.go
```diff
@@ -865,13 +889,21 @@ func flattenAdvancedReplicationSpecs(apiObjects []*matlas.AdvancedReplicationSpe
 
 		if len(tfMapObjects) > 0 {
 			tfMapObject := tfMapObjects[i].(map[string]interface{})
-			tfList = append(tfList, flattenAdvancedReplicationSpec(apiObject, tfMapObject))
+			object, err := flattenAdvancedReplicationSpec(ctx, apiObject, tfMapObject, d, conn)
```
thetonymaster: [NIT] Change the `object` variable name to something more descriptive

thetonymaster: `tfMapObject` can be nil right? `flattenAdvancedReplicationSpec` actually returns a nil, so can this be done in a single operation instead of doing this twice?

thetonymaster (changes requested):

thetonymaster (approved):
LGTM

andreaangiolillo (approved):
LGTM

Issue #723, opened 2022-05-03
Typos in documentation for resource/mongodbatlas_search_index


daniel-orlov:
Where: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/search_index#example-usage
Or in the code: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/website/docs/r/search_index.html.markdown

What: both Basic and Advanced examples contain camelCase attribute names like **collectionName** and **searchAnalyzer** instead of expected snake_case ones.
Should be: collection_name, search_analyzer etc

Runs into `Error: Unsupported argument` if executed.

themantissa:
@daniel-orlov  we welcome PRs to fix docs errors - please sign the CLA if you'd like to contribute.

themantissa:
@daniel-orlov  we welcome PRs for docs fixes, just need to sign the CLA first.

PR #725, opened 2022-05-05
Fix attributes case in examples for search_index

daniel-orlov:
## Description

The examples in documentation for the resource/mongodbatlas_search_index contained camelCase terraform attributes. If used as is they result in `Error: Unsupported argument` on execution. This PR changes case to snake_case in documentation for resource attributes and examples.

Link to any related issue(s): https://github.com/mongodb/terraform-provider-mongodbatlas/issues/723

## Type of change:

## Required Checklist:

### DIFF
--- a/website/docs/r/search_index.html.markdown
+++ b/website/docs/r/search_index.html.markdown
@@ -20,11 +20,11 @@ resource "mongodbatlas_search_index" "test" {
   cluster_name = "<CLUSTER_NAME>"
   
   analyzer = "lucene.standard"
-  collectionName = "collection_test"
+  collection_name = "collection_test"
   database = "database_test"
   mappings_dynamic = true
   
-  searchAnalyzer = "lucene.standard"
+  search_analyzer = "lucene.standard"
 }
 ```
 
@@ -34,7 +34,7 @@ resource "mongodbatlas_search_index" "test" {
   project_id = "%[1]s"
   cluster_name = "%[2]s"
   analyzer = "lucene.standard"
-  collectionName = "collection_test"
+  collection_name = "collection_test"
   database = "database_test"
   mappings_dynamic = false
   mappings_fields = <<-EOF
@@ -70,7 +70,7 @@ resource "mongodbatlas_search_index" "test" {
 }
 EOF
   name = "name_test"
-  searchAnalyzer = "lucene.standard"
+  search_analyzer = "lucene.standard"
   analyzers = <<-EOF
   [{
   "name": "index_analyzer_test_name",
@@ -254,7 +254,7 @@ An [Atlas Search analyzer](https://docs.atlas.mongodb.com/reference/atlas-search
         * `type` - Must be `uaxUrlEmail`
         *  `max_token_length` - The maximum number of characters in one token.
 
-* `tokenFilters` - Array containing zero or more token filters. Always require a type field, and some take additional options as well:
+* `token_filters` - Array containing zero or more token filters. Always require a type field, and some take additional options as well:
   ```terraform
   "token_filters":{
     "type": "<FILTER_TYPE>",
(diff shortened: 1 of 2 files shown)


website/docs/r/search_index.html.markdown
```diff
@@ -254,7 +254,7 @@ An [Atlas Search analyzer](https://docs.atlas.mongodb.com/reference/atlas-search
         * `type` - Must be `uaxUrlEmail`
         *  `max_token_length` - The maximum number of characters in one token.
 
-* `tokenFilters` - Array containing zero or more token filters. Always require a type field, and some take additional options as well:
+* `token_filters` - Array containing zero or more token filters. Always require a type field, and some take additional options as well:
```
themantissa, line 257: The other two corrects are correct but I don't think this one is.  tokenFilters is correct (https://www.mongodb.com/docs/atlas/atlas-search/analyzers/custom/) - because it is simply included in the content one would pass into `analyzers`.  This is likely left over documentation from when we had a more formal design of the search resource.  I'll create an internal ticket to fix this part as the tests use the same and I'm not sure it's correct.  Please go ahead and remove this from the PR and we'll investigate further.

themantissa (changes requested):
Please remove the suggestions marked and we'll be able to accept. Thank you for doing this.

website/docs/r/search_index.html.markdown
daniel-orlov: Got you! I think then there are a few more cases of JSON field names vs Terraform attributes confusions on the list, let me quickly fix that.

daniel-orlov:
Looks like there were a few more cases of JSON fields with snake_case. I have carefully studied the official docs [on custom analyzers](https://www.mongodb.com/docs/atlas/atlas-search/analyzers/custom/) and compared field names to the ones in Mongo API.

I suppose there was confusion when copy-pasting original docs parts to the Terraform module documentation, which resulted in some situations in wrong (Terraform snake_case) cases applied onto actual Mongo API JSON camelCase fields.

website/docs/r/search_index.html.markdown
themantissa: https://help-hr.mongodb.com/hc/en-us/requests/53603

themantissa: @daniel-orlov this PR looks really good. The only line still off is 257, the one this comment is on.  All the others are fixed and look great. Could you push a fix for this and we'll merge so this will be in our next release?

daniel-orlov: @themantissa Thank you for your comments! I am afraid I have to disagree with you again - it should be either token_filters in snake_case here AND on line 259 or it will just be misleading. I understand that camelCase is used in the Mongo API and you provided the reference, but here we are documenting terraform objects' fields, it makes no sense to confuse the user by using the wrong attribute name.  

Moreover the docs have been stating It the same [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/website/docs/guides/1.0.1-upgrade-guide.html.markdown?plain=1#L40) and [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/website/docs/guides/1.0.1-upgrade-guide.html.markdown?plain=1#L73) as well.

I once again kindly ask you to reread the references and ask yourself a question - what are we referencing here - a Mongo API field or a Terraform attribute? 

Because if you are right and it should be tokenFilters indeed, there's no way other corrections I made are valid  - because the Mongo API docs state otherwise (e.g. char_filters).

If you have time, I am open to scheduling a Google Meet (or other conferencing tool of your choice) session and we can go over it again.

themantissa: @daniel-orlov I'm glad you disagreed because I was yet again looking at the tests and not the actual resource code.  You are correct - this is the right casing/variable!  Glad to have some help on this one :) I'll resolve this comment and merge the PR as is (and get that test fixed too by our devs!)

themantissa (approved):
Thank you so much for your work here!

PR #833, opened 2022-09-08
INTMDB-387: Enable Azure NVME for Atlas Dedicated clusters

martinstibbe:
## Description

INTMDB-387: Enable Azure NVME for Atlas Dedicated clusters

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -1148,7 +1148,7 @@ func expandProviderSetting(d *schema.ResourceData) (*matlas.ProviderSettings, er
 		providerSettings.AutoScaling = &matlas.AutoScaling{Compute: compute}
 	}
 
-	if d.Get("provider_name") == "AWS" {
+	if d.Get("provider_name") == "AWS" || d.Get("provider_name") == "AZURE" {
 		// Check if the Provider Disk IOS sets in the Terraform configuration and if the instance size name is not NVME.
 		// If it didn't, the MongoDB Atlas server would set it to the default for the amount of storage.
 		if v, ok := d.GetOk("provider_disk_iops"); ok && !strings.Contains(providerSettings.InstanceSizeName, "NVME") {
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -1148,13 +1148,20 @@ func expandProviderSetting(d *schema.ResourceData) (*matlas.ProviderSettings, er
 		providerSettings.AutoScaling = &matlas.AutoScaling{Compute: compute}
 	}
 
-	if d.Get("provider_name") == "AWS" || d.Get("provider_name") == "AZURE" {
+	if d.Get("provider_name") == "AWS" {
 		// Check if the Provider Disk IOS sets in the Terraform configuration and if the instance size name is not NVME.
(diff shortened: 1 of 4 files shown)


themantissa (approved):
LGTM but fyi for @shum we didn't actually include the same in https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_advanced_cluster.go so either should work, and advanced is preferred, for those just starting out (we are going to have to move everyone from one to the other).   Still good to do till we deprecate clusters.

andreaangiolillo (approved):
LGTM

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1148,7 +1148,7 @@ func expandProviderSetting(d *schema.ResourceData) (*matlas.ProviderSettings, er
 		providerSettings.AutoScaling = &matlas.AutoScaling{Compute: compute}
 	}
 
-	if d.Get("provider_name") == "AWS" {
+	if d.Get("provider_name") == "AWS" || d.Get("provider_name") == "AZURE" {
```
shum: I just realized that `providerSettings.diskIOPS` is the Disk IOPS setting for AWS storage only. 

https://www.mongodb.com/docs/atlas/reference/api/clusters-modify-one/

Instead, we probably need to do the same with `providerSettings.diskTypeName` https://www.mongodb.com/docs/atlas/reference/api/clusters-modify-one/#std-label-modify-cluster-providerSettings-diskTypeName where the user cannot provide diskTypeName if the isntance size contains NVME.

something like:
```
if d.Get("provider_name") == "AZURE" {
		if v, ok := d.GetOk("provider_disk_type_name"); ok && !strings.Contains(providerSettings.InstanceSizeName, "NVME") {
			providerSettings.DiskTypeName = pointy.Int64(cast.ToInt64(v))
		}
		providerSettings.EncryptEBSVolume = pointy.Bool(true)
	}
	return providerSettings, nil
}
```

shum (changes requested):

shum:
> LGTM but fyi for @shum we didn't actually include the same in https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_advanced_cluster.go so either should work, and advanced is preferred, for those just starting out (we are going to have to move everyone from one to the other). Still good to do till we deprecate clusters.

Good to know. I guess this validation is just to prevent folks who don't read the manual from shooting themselves in the foot.

mongodbatlas/resource_mongodbatlas_cluster.go
martinstibbe: @shum I assume you wanted d.Get("provider_disk_type_name") vs pointy.Int64(cast.ToInt64(v)) for the name when it is permitted to set a value ?  i.e.```
 if d.Get("provider_name") == "AZURE" {
		if v, ok := d.GetOk("provider_disk_type_name"); ok && !strings.Contains(providerSettings.InstanceSizeName, "NVME") {
			providerSettings.DiskTypeName = cast.ToString(v)
		}
		providerSettings.EncryptEBSVolume = pointy.Bool(true)
	}
```

shum (approved):

mongodbatlas/resource_mongodbatlas_cluster.go
shum: yea, sorry about that, copy pasted and didn't read everything

PR #943, opened 2022-11-29
INTMDB-465 - PrivateLink Endpoint Service & Connection Strings

evertsd:
## Description

Added a secondary RefreshFunc to _mongodbatlas_privatelink_endpoint_service_ to wait for all clusters to be idle. This allows us to retrieve proper connection_string information following the creation or deletion of endpoint services.

Closes INTMDB-465
Addresses HELP-39042

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_advanced_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_advanced_cluster.go
@@ -942,18 +942,15 @@ func flattenAdvancedReplicationSpecs(ctx context.Context, apiObjects []*matlas.A
 
 	var tfList []map[string]interface{}
 
-	for _, apiObject := range apiObjects {
+	for i, apiObject := range apiObjects {
 		if apiObject == nil {
 			continue
 		}
 
 		var tfMapObject map[string]interface{}
 
-		for _, v := range tfMapObjects {
-			if v.(map[string]interface{})["id"] == apiObject.ID {
-				tfMapObject = v.(map[string]interface{})
-				break
-			}
+		if len(tfMapObjects) > i {
+			tfMapObject = tfMapObjects[i].(map[string]interface{})
 		}
 
 		advancedReplicationSpec, err := flattenAdvancedReplicationSpec(ctx, apiObject, tfMapObject, d, conn)
--- a/mongodbatlas/resource_mongodbatlas_advanced_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_advanced_cluster.go
@@ -942,15 +942,18 @@ func flattenAdvancedReplicationSpecs(ctx context.Context, apiObjects []*matlas.A
 
 	var tfList []map[string]interface{}
 
-	for i, apiObject := range apiObjects {
+	for _, apiObject := range apiObjects {
 		if apiObject == nil {
 			continue
 		}
 
 		var tfMapObject map[string]interface{}
 
-		if len(tfMapObjects) > 0 {
-			tfMapObject = tfMapObjects[i].(map[string]interface{})
+		for _, v := range tfMapObjects {
+			if v.(map[string]interface{})["id"] == apiObject.ID {
+				tfMapObject = v.(map[string]interface{})
+				break
+			}
 		}
 
 		advancedReplicationSpec, err := flattenAdvancedReplicationSpec(ctx, apiObject, tfMapObject, d, conn)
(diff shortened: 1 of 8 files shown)


Zuhairahmed (commented):
Thanks, @evertsd can you double check if we need to also apply these fixes to Cluster and/or PrivateLink Endpoint Serverless?

evertsd:
I'll apply these fixes to serverless private endpoints as well. In the case of cluster vs. advanced cluster, it's just a matter of reading clusters to tell when they're idle, and I'm opting to rely on the newer AdvancedCluster API.

evertsd:
Ok @Zuhairahmed these changes have been applied to serverless instance create as well now

Zuhairahmed (approved):
LGTM! Not blocking @evertsd, but just for my sanity...should there be any doc updates here? I'm assuming since currently Terraform does not wait for cluster update when creating private endpoints we are just correcting intended behavior, but let me know if you disagree.

mongodbatlas/resource_mongodbatlas_advanced_cluster.go
```diff
@@ -1124,6 +1125,36 @@ func resourceClusterAdvancedRefreshFunc(ctx context.Context, name, projectID str
 	}
 }
 
+func resourceClusterListAdvancedRefreshFunc(ctx context.Context, projectID string, client *matlas.Client) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		clusters, resp, err := client.AdvancedClusters.List(ctx, projectID, nil)
+
+		if err != nil && strings.Contains(err.Error(), "reset by peer") {
```
andreaangiolillo, line 1134: [q] Can you use `errors.Is` here?

mongodbatlas/resource_mongodbatlas_privatelink_endpoint_service_serverless.go
```diff
@@ -123,6 +123,22 @@ func resourceMongoDBAtlasPrivateLinkEndpointServiceServerlessCreate(ctx context.
...
+		MinTimeout: 5 * time.Second,
+		Delay:      5 * time.Minute,
+	}
+
+	_, err = clusterConf.WaitForStateContext(ctx)
+
+	if err != nil {
+		// error awaiting serverless instances to IDLE should not result in failure to apply changes to this resource
+		log.Printf(errorServerlessInstanceListStatus, err)
+	}
```
andreaangiolillo:
```suggestion
	if _, err = clusterConf.WaitForStateContext(ctx); err != nil {
		// error awaiting serverless instances to IDLE should not result in failure to apply changes to this resource
		log.Printf(errorServerlessInstanceListStatus, err)
	}
```

mongodbatlas/resource_mongodbatlas_advanced_cluster.go
```diff
@@ -1124,6 +1125,36 @@ func resourceClusterAdvancedRefreshFunc(ctx context.Context, name, projectID str
...
+			return nil, "REPEATING", nil
+		}
+
+		if err != nil && clusters == nil && resp == nil {
+			return nil, "", err
+		} else if err != nil {
```
andreaangiolillo:
```suggestion
		} 
	 if err != nil {
```

andreaangiolillo (approved):
comments are not blocking. LGTM

mongodbatlas/resource_mongodbatlas_advanced_cluster.go
evertsd: I'm not sure if I can use `errors.Is` out of the box, is `err.Error() == "reset by peer"`? Or does it just contain it? 

Even if it's just containment, I could theoretically implement something like the following, would that be preferable?
```
type ErrorContains {
    Message string
}

func (e *ErrorContains) Error() string {
    return e.Message
}

func (e *ErrorContains) Is(tgt error) bool {
    return strings.Contains(tgt.Error(), e.Message)
}

var ResetByPeerError = &ErrorContains{ Message = "reset by peer" }
```

mongodbatlas/resource_mongodbatlas_advanced_cluster.go
evertsd: I applied these suggestions in another commit, alongside similar changes. Thanks for the feedback!

mongodbatlas/resource_mongodbatlas_advanced_cluster.go
evertsd: I'm resolving this for now, but am open to improving this down the road depending on your response.

Issue #1057, opened 2023-02-09
Terraform plan fails if API key created by `mongodbatlas_api_key` resource is deleted outside of Terraform


itspngu:
### Terraform CLI and Terraform MongoDB Atlas Provider Version

```hcl
Terraform v1.3.7
on linux_amd64
+ provider registry.terraform.io/mongodb/mongodbatlas v1.8.0
```

### Terraform Configuration File 

```hcl
terraform {
  required_version = ">=1.3"
  required_providers {
    mongodbatlas = {
      source  = "mongodb/mongodbatlas"
      version = "~>1.8"
    }
    time = {
      source  = "hashicorp/time"
      version = "~>0.9"
    }
  }
}

variable "name" {
  description = "The name of the MongoDB Atlas project to create"
  type        = string
}
variable "vault_egress_ip_cidr_ranges" {
  description = "The IP ranges in CIDR notation that Vault traffic originates from"
  type        = list(string)
}

data "mongodbatlas_roles_org_id" "this" {}

resource "time_rotating" "mongodbatlas_vault_creds" {
  rotation_days = 7
}
resource "mongodbatlas_api_key" "vault" {
  org_id = data.mongodbatlas_roles_org_id.this.org_id

  description = "vault-${var.name}-root-${time_rotating.mongodbatlas_vault_creds.unix}"
  role_names = [
    "ORG_OWNER",
  ]

  depends_on = [
    time_rotating.mongodbatlas_vault_creds,
  ]
}
resource "mongodbatlas_access_list_api_key" "vault" {
  for_each = toset(var.vault_egress_ip_cidr_ranges)

  org_id     = data.mongodbatlas_roles_org_id.this.org_id
  api_key_id = mongodbatlas_api_key.vault.api_key_id

  cidr_block = each.value

  depends_on = [
    mongodbatlas_api_key.vault
  ]
}
```

### Steps to Reproduce

1. `terraform init`
2. `terraform plan`
3. `terraform apply`
4. Delete API key (e.g. via Atlas web UI)
5. `terraform plan`
6. Receive error

### Expected Behavior

Terraform should create a new API key because the API key it created previously is no longer present.

### Actual Behavior

Terraform fails completing the `plan` stage, forcing you to manually remove the resource from tfstate before planning again:

```
[...]

PR #1097, opened 2023-03-13
INTMDB-581: plan fails if API key created by mongodbatlas_api_key resource is deleted outside of Terraform

martinstibbe:
## Description

plan fails if API key created by mongodbatlas_api_key resource is deleted outside of Terraform

Link to any related issue(s):
https://github.com/mongodb/terraform-provider-mongodbatlas/issues/1057

## Type of change:

## Required Checklist:


mongodbatlas/resource_mongodbatlas_api_key.go
```diff
@@ -92,8 +92,12 @@ func resourceMongoDBAtlasAPIKeyRead(ctx context.Context, d *schema.ResourceData,
 	orgID := ids["org_id"]
 	apiKeyID := ids["api_key_id"]
 
-	apiKey, _, err := conn.APIKeys.Get(ctx, orgID, apiKeyID)
+	apiKey, resp, err := conn.APIKeys.Get(ctx, orgID, apiKeyID)
 	if err != nil {
+		if resp != nil && resp.StatusCode == http.StatusNotFound || resp.StatusCode == http.StatusBadRequest {
+			d.SetId("")
+			return nil
```
andreaangiolillo, line 101: why do we also consider `StatusBadRequest`? Shouldn't we only consider StatusNotFound?

mongodbatlas/resource_mongodbatlas_api_key.go
```diff
@@ -92,8 +92,12 @@ func resourceMongoDBAtlasAPIKeyRead(ctx context.Context, d *schema.ResourceData,
...
+		}
```
andreaangiolillo, line 102: Could we add a warning saying that we could not find the key and we are silently removing the API key?

mongodbatlas/resource_mongodbatlas_api_key.go
martinstibbe: If they deleted API key and it tries to find it comes back with bad request so had to catch that scenario

mongodbatlas/resource_mongodbatlas_api_key.go
martinstibbe: If API key missing it will offer to create it during plan result so it would be removed from state after apply and replaced with new one could add message to catch that edge case

Zuhairahmed (approved):
LGTM

mongodbatlas/resource_mongodbatlas_api_key.go
martinstibbe: Added a log message provider also shows this 
```
Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply":

  # mongodbatlas_api_key.test has been deleted
```

andreaangiolillo (approved):
LGTM

PR #1217, opened 2023-06-07
INTMDB-858: Move mongodbatlas_cluster_outage_simulation acceptance tests to use cloud-dev

andreaangiolillo:
## Description

This PR defines a new workflow to run concurrent TF acceptance tests against cloud-dev. **Only the cluster outage resources/data source have been added for now.** 

Screenshot 2023-06-07 at 17 33 32

- Ticket: [INTMDB-858](https://jira.mongodb.org/browse/INTMDB-858)
- Action: [5201802898](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/5201802898)
- New MongoDB Atlas Organization - Cloud-dev: Terraform Test GitHub

#### How to run the workflow

- The workflow can be started manually: In this case all the jobs in the workflow will run
- PR: the workflow runs automatically after the code_health tests are successful. The specific acceptance tests run only if the resource/data sources files have changed in the PR
- Push: When merging against the master branch, the workflow runs after code_health tests are successful and it runs all run all the acceptance tests (not checking file changes).

 

## Type of change:

## Required Checklist:

### The cluster outage tests running concurrently against cloud-dev
Screenshot 2023-06-07 at 16 23 37

Screenshot 2023-06-07 at 15 57 48


Zuhairahmed (approved):
LGTM

GNUmakefile
```diff
@@ -37,7 +37,7 @@ test: fmtcheck
 .PHONY: testacc
 testacc: fmtcheck
 	@$(eval VERSION=acc)
-	TF_ACC=1 go test $(TEST) -run '$(TEST_REGEX)' -v -parallel 5 $(TESTARGS) -timeout $(ACCTEST_TIMEOUT) -cover -ldflags="$(LINKER_FLAGS)"
+	TF_ACC=1 go test $(TEST) -run '$(TEST_REGEX)' -v -parallel 20 $(TESTARGS) -timeout $(ACCTEST_TIMEOUT) -cover -ldflags="$(LINKER_FLAGS)"
```
gssbzn, line 40: `-parallel` should not exceed the number of cores, do you know how many cores you have in GH?

andreaangiolillo: My understanding is that `go test` uses `GOMAXPROCS` (which is set to the number of CPUs) as the default value when the `-parallel` flag is not specified. I'm not currently aware of any recommendation advocating for using this default value. If you happen to have come across an article or resource that suggests this, could you share it with me?

gssbzn: I mean yes, parallel defaults to `GOMAXPROCS` but the flag is more to lower it than increase it, there's no benefit to bump it above the number of available cores since each parallel test will be in contention for resources and not really running in parallel


TL;DR: what's the point of using 20 here if you have 4 cores and realistically only running 4 tests in parallel

andreaangiolillo: The ubuntu-latest seems to have only 2 cores [supported-runners-and-hardware-resources](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). I will try to remove the parallel flag and see if the test has the same behaviour. I will add an update here. Thanks!

andreaangiolillo: I tested here [9401709790](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/5210481259/jobs/9401709790). When the `-parallel` flag is not provided, it runs only 2 acceptance tests in parallel (2 projects available in cloud-dev)
Screenshot 2023-06-08 at 12 12 35


When the flag is provided, it was running all of them in parallel (4 acceptance tests -> 4 projects in cloud-dev)
Screenshot 2023-06-07 at 16 23 37

Since these are not unit tests and there is I/0, what may be happening is that once a thread is waiting for a response from the cloud, a new thread is started to run another test.

gssbzn: there's a difference between parallelism and concurrency, I can't see the logs for [9401709790](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/5210481259/jobs/9401709790). so can't confirm yet but considering concurrent vs parallel what go is doing is spinning `--parallel` processes if the number of processes exceed the number of CPUs what will happen is go will have to pause some tests to run other tests, you will see this on the logs as PAUSE/CONT statements so they run concurrently, if the times are better with  `-parallel 20` than with 4 or other number feel free to keep but 20 is not really 20 parallel test but 10 concurrent test (assuming 2 CPU)

andreaangiolillo: Thanks for the clarification! Yes, the time is better with `--parallel 20` as the job took 24 minutes [5201802898](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/5201802898) vs 47 minutes [5210481259](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/5210481259?pr=1219)

gssbzn (approved):
LGTM thanks for the conversations on --parallel

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,60 @@
+name: 'Acceptance Tests'
+
+on:
+  workflow_dispatch: {}
+  workflow_call: 
+    inputs:
+      parent-event-name: # we want to run all the acceptance tests when the worflow is run in master (push)
+        required: true
+        type: string
```
colm-quinn: [q] Is master the full 12 hour run on every merge? I wonder instead do we only want to run the concurrent tests on master automatically, with an on-demand run for acceptance tests/on feature branches?

colm-quinn: Looking at the filters, we only run the cluster sim tests here? IF that is true, can you update the comment to indicate it's a partial run?

colm-quinn (approved):
LGTM, small comment on incorrect comment text (I think) Just want to make sure we don't suddenly start running a lot of slow tests on Prod if we merge this change and start working on master

.github/workflows/acceptance-tests.yml
andreaangiolillo: > Looking at the filters, we only run the cluster sim tests here? 

yes, the workflow runs only 1 TF resource which takes ~24 minutes to complete. 

> IF that is true, can you update the comment to indicate it's a partial run?

sure

andreaangiolillo: updated

maastha (approved):
LGTM

martinstibbe (approved):
LGTM

PR #1262, opened 2023-06-21
INTMDB-873 Improve the Atlas region parameter in terraform configuration

maastha:
## Description

Jira Ticket: [INTMDB-873](https://jira.mongodb.org/browse/INTMDB-873)

This PR adds a mapping for all cloud provider regions vs. Atlas regions for mongodbatlas_cluster resource as we want to allow users to use cloud provider region names and Atlas region names interchangeably.

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Testing
Following test cases have been tested manually in addition to the acceptance tests:
1. Create and update cluster of type "REPLICASET" with multiple region_configs in replication_specs (provider_name="AWS")
2. Create and update cluster of type "REPLICASET" with single r region_configs in replication_specs (provider_name="AWS")
3. Create cluster with provider TENANT and backing_provider_name="AWS"
4. Create and update cluster of type "GEOSHARDED" with multiple region_configs in replication_specs (provider_name="AWS")

Partial `terraform.tfstate` from testing [1]:
```
 "provider_name": "AWS",
            "provider_region_name": "US-EAST-1",
            "provider_volume_type": "STANDARD",
            "replication_factor": 0,
            "replication_specs": [
              {
                "id": "......",
                "num_shards": 1,
                "regions_config": [
                  {
                    "analytics_nodes": 0,
                    "electable_nodes": 2,
                    "priority": 5,
                    "read_only_nodes": 0,
                    "region_name": "US_WEST_1"
                  },
                  {
                    "analytics_nodes": 0,
                    "electable_nodes": 2,
                    "priority": 6,
                    "read_only_nodes": 2,
                    "region_name": "US-EAST-2"
                  },
                  {
                    "analytics_nodes": 0,
                    "electable_nodes": 3,
                    "priority": 7,
                    "read_only_nodes": 0,
                    "region_name": "US_EAST_1"
                  }
                ],
                "zone_name": "ZoneName managed by Terraform"
              }
            ],
```

Partial `terraform.tfstate` from testing [2] (GEOSHARDED):
```
            "cluster_type": "GEOSHARDED",
            "provider_name": "AWS",
            "provider_region_name": "",
            "provider_volume_type": "STANDARD",
            "replication_factor": 0,
            "replication_specs": [
              {
                "id": ".....",
                "num_shards": 2,
                "regions_config": [
                  {
                    "analytics_nodes": 0,
                    "electable_nodes": 3,
                    "priority": 7,
                    "read_only_nodes": 0,
                    "region_name": "US_EAST_1"
                  }
                ],
                "zone_name": "Zone 1"
              },
              {
                "id": "6493089500bc3c3e29f00b40",
                "num_shards": 2,
                "regions_config": [
                  {
                    "analytics_nodes": 0,
                    "electable_nodes": 3,
                    "priority": 7,
                    "read_only_nodes": 0,
                    "region_name": "eu-central-1"
                  }
                ],
                "zone_name": "Zone 2"
              }
            ],
```


mongodbatlas/data_source_mongodbatlas_cluster.go
```diff
@@ -497,6 +498,22 @@ func dataSourceMongoDBAtlasClusterRead(ctx context.Context, d *schema.ResourceDa
 	return nil
 }
 
+func flattenReplicationSpecs(rSpecs []matlas.ReplicationSpec) []map[string]interface{} {
```
maastha, line 501: Moved `flattenReplicationSpecs ` to data source cluster as is so that region mapping logic is isolated to to cluster resource.

mongodbatlas/atlas_region_helper.go
```diff
@@ -0,0 +1,104 @@
+package mongodbatlas
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+type AtlasRegion string
```
maastha, line 9: PENDING: add remaining Atlas and cloud regions

maastha:
Cluster acceptance tests currently having transient failures, looking into that.
Tests for this change passing locally:
Screenshot 2023-06-22 at 09 09 11

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1361,22 +1376,188 @@ func flattenReplicationSpecs(rSpecs []matlas.ReplicationSpec) []map[string]inter
 		specs = append(specs, spec)
 	}
 
-	return specs
+	return updateRegionNames(d, specs)
```
maastha, line 1383: main change

mongodbatlas/atlas_region_helper.go
```diff
@@ -0,0 +1,303 @@
...
+
+const (
+	// AWS:
+	// Americas
+	UsEast1    AtlasRegion = "US_EAST_1"
+	UsWest2    AtlasRegion = "US_WEST_2"
+	CaCentral1 AtlasRegion = "CA_CENTRAL_1"
```
andreaangiolillo, line 16: I really don't like the idea of maintaining this list 😿  One major drawback I see with having a hardcoded list is that whenever Atlas introduces a new region, users won't be able to utilize it until we manually update the region list in the Terraform code and release the provider.


Could we investigate if we can update [listCloudProviderRegions](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Clusters/operation/listCloudProviderRegions) (or create a new endpoint in Atlas) to return the information that we need? The idea is that we already have an enum with all the regions in the MMS codebase so that when the Atlas team adds a new region, we can have this info in TF for free without the need for this hardcoded list.

Once we have the endpoint, the TF cluster resource will just need to make an additional API call to get the correct region name.

I understand this means dedicating more time to this ticket. Tagging @colm-quinn to see if he agrees on dedicating more time to this ticket to investigate this other solution.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1273,8 +1275,14 @@ func flattenProviderSettings(d *schema.ResourceData, settings *matlas.ProviderSe
 		log.Printf(errorClusterSetting, "provider_name", clusterName, err)
 	}
 
-	if err := d.Set("provider_region_name", settings.RegionName); err != nil {
-		log.Printf(errorClusterSetting, "provider_region_name", clusterName, err)
+	if val, ok := d.GetOk("provider_region_name"); ok {
+		if err := d.Set("provider_region_name", val); err != nil {
+			log.Printf(errorClusterSetting, "provider_region_name", clusterName, err)
+		}
```
andreaangiolillo, line 1285: shouldn't we return an error here?

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1361,22 +1376,188 @@ func flattenReplicationSpecs(rSpecs []matlas.ReplicationSpec) []map[string]inter
...
+}
+
+// This method creates a map of all the replication_specs present in Terraform state for the cluster
+// keyed on a hash value generated from property values of a replication_spec(s).
+//
+// The value of each entry is another map keyed on a hash value generated from property values of each region_config(s) in the replication_spec
+//
+// A "region_name_by_user" property is also added to each region_config to keep track if the user is
+// using an Atlas region name or a Cloud Provider's region name.
+func getStateFileSpecsMap(stateSpecs interface{}, providerName, backingProviderName string) map[uint64]map[uint64]interface{} {
```
andreaangiolillo, line 1431: [nit] golang comment for a function should start with the function name https://tip.golang.org/doc/comment
```suggestion
// getStateFileSpecsMap creates a map of all the replication_specs present in Terraform state for the cluster
// keyed on a hash value generated from property values of a replication_spec(s).
//
// The value of each entry is another map keyed on a hash value generated from property values of each region_config(s) in the replication_spec
//
// A "region_name_by_user" property is also added to each region_config to keep track if the user is
// using an Atlas region name or a Cloud Provider's region name.
func getStateFileSpecsMap(stateSpecs interface{}, providerName, backingProviderName string) map[uint64]map[uint64]interface{} {
```

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1361,22 +1376,188 @@ func flattenReplicationSpecs(rSpecs []matlas.ReplicationSpec) []map[string]inter
...
+		}
+		repSpecsHashKey := getReplicationSpecHash(stateSpec, true)
+		stateReplicationSpecsMap[repSpecsHashKey] = stateRegionConfigsMap
+	}
+
+	return stateReplicationSpecsMap
+}
+
+// excluding 'id' property when generating hash of a ReplicationSpec
+func getReplicationSpecHash(repSpec map[string]interface{}, stateSpec bool) uint64 {
```
andreaangiolillo, line 1459: cluster is not the only resource that uses provider regions. Should we update other resources as well to use these functions?

mongodbatlas/atlas_region_helper.go
maastha: While I don't see the cost of maintaining this mapping to be a lot as we don't expect new regions to be added very often, I'm okay with adding a new endpoint for this mapping so it can be available to a broader audience and move the maintenance to MMS if it already has region enums.

For my knowledge, how would the ownership of that API be worked out? (since you mentioned about an Atlas team)

mongodbatlas/resource_mongodbatlas_cluster.go
maastha: I would address other resources separately. For this ticket (HELP) I think we should focus on only mongodbatlas_cluster resource, given the complexity in handling replication specs/region configs. 

We can look into other resources such as mongodbatlas_advanced_cluster and others after this.

cc @Zuhairahmed

mongodbatlas/atlas_region_helper.go
maastha: @andreaangiolillo Came across [this comment](https://jira.mongodb.org/browse/CLOUDP-123180?focusedCommentId=4841378&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4841378) in a similar JIRA ticket for Atlas CLI. Do you remember more about this? Can this existing endpoint already return all region mappings?

andreaangiolillo: The ticket is referring to this endpoint https://github.com/10gen/mms/pull/38225, which is the same one I shared with you via Slack. However, it seems that we didn't actually implement the desired functionality for that endpoint. Instead, we only created the related AtlasCLI command. If you need more information about the ticket, feel free to contact Bianca

maastha: Synced with @blva offline, as per her knowledge, the endpoint will return only list of regions by instance size and not mapping with Atlas regions.

colm-quinn: If we can leverage the enums/list defined in Atlas to do this via an extension to the existing API, then I think it's a good approach and example of leverage, where we solve once for all integrations.

When we originally spoke about the mapping conversion here, we were hoping to generate the names by transformation, rather than by maintaining a manual list. Def can see a maintenance issue & source of bugs going forward with the current implementation

colm-quinn: To add, I can see the mapping being in the generated SDK have a lot of user value, again a reason to extend the API

maastha: > When we originally spoke about the mapping conversion here, we were hoping to generate the names by transformation, rather than by maintaining a manual list. 

Aah I see, @colm-quinn generating by transformation was never an option [as noted here](https://jira.mongodb.org/browse/CLOUDP-123180?focusedCommentId=4841378&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4841378) 

I have started looking into the Atlas endpoint for this mapping.

mongodbatlas/resource_mongodbatlas_cluster.go
Zuhairahmed: while agree @maastha that we can we have separate tickets for future unrelated resources, given how we are actively directing users to leverage mongodbatlas_advanced_clusters resource, suggest this PR should address support in both mongodbatlas_cluster and mongodbatlas_advanced_clusters.

maastha:
Closing PR as we no longer want to introduce this functionality.

PR #1365, opened 2023-08-02
refactor: Migrate Resource: mongodbatlas_project to Terraform Plugin Framework and remove api_keys attribute

maastha:
## Description

Migrate Resource: mongodbatlas_project to Terraform Plugin Framework.
- This PR also removes api_keys attribute from project resource.

Link to any related issue(s): [INTMDB-908](https://jira.mongodb.org/browse/INTMDB-908)

## Type of change:

## Required Checklist:


mongodbatlas/provider_test.go
```diff
@@ -30,6 +35,14 @@ func init() {
...
 		},
 	}
+
+	config := Config{
+		PublicKey:    os.Getenv("MONGODB_ATLAS_PUBLIC_KEY"),
+		PrivateKey:   os.Getenv("MONGODB_ATLAS_PRIVATE_KEY"),
+		BaseURL:      os.Getenv("MONGODB_ATLAS_BASE_URL"),
+		RealmBaseURL: os.Getenv("MONGODB_REALM_BASE_URL"),
+	}
+	testMongoDBClient, _ = config.NewClient(context.Background())
```
AgustinBettati, line 44: I think here we can leave this testMongoDBClient and remove the testAccProviderSdkV2 which is being used for the same purpose

mongodbatlas/fw_data_source_mongodbatlas_projects.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return fmt.Errorf("error while getting project properties for project %s: %v", project.ID, err.Error())
+		}
+		projectModel := toTFProjectDataSourceModel(ctx, project, atlasTeams, atlasProjectSettings, atlasLimits)
+		results[i] = projectModel
+	}
+
+	stateModel.Results = results
+	stateModel.TotalCount = types.Int64Value(int64(projectsRes.TotalCount))
+	stateModel.ID = types.StringValue(fmt.Sprint(projectsRes.TotalCount))
```
maastha: This is a dummy value pretty much because acceptance tests require every data source to have an 'id' field. TF will get rid of it at some point but for now they recommend to just add it and include a deprecation note to ensure users don't use it https://github.com/hashicorp/terraform-plugin-testing/issues/84#issuecomment-1480006432

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	}
+	if !projectPlan.IsPerformanceAdvisorEnabled.IsUnknown() {
+		projectSettings.IsPerformanceAdvisorEnabled = projectPlan.IsPerformanceAdvisorEnabled.ValueBoolPointer()
+	}
+	if !projectPlan.IsRealtimePerformancePanelEnabled.IsUnknown() {
+		projectSettings.IsRealtimePerformancePanelEnabled = projectPlan.IsRealtimePerformancePanelEnabled.ValueBoolPointer()
+	}
+	if !projectPlan.IsSchemaAdvisorEnabled.IsUnknown() {
+		projectSettings.IsSchemaAdvisorEnabled = projectPlan.IsSchemaAdvisorEnabled.ValueBoolPointer()
+	}
```
AgustinBettati, line 350: For these optional fields, would .IsNull() be the more appropriate option? Using as reference: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#null-and-unknown-values

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+
+	// adding updated limits into the project
+	if len(planLimits) > 0 {
+		err := setProjectLimits(ctx, connV2, projectID, planLimits)
+		if err != nil {
+			return fmt.Errorf("error adding limits into the project during update: %v", err.Error())
+		}
+	}
+	return nil
```
AgustinBettati: would give a shot at implementing the previous getChangesInSet, to avoid having to remove and create all elements of the set for a single change

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+
+func NewProjectDS() datasource.DataSource {
+	return &ProjectDS{}
+}
+
+type ProjectDS struct {
+	client *MongoDBClient
+}
+
+type tfProjectDSModel struct {
```
andreaangiolillo, line 29: Any reason why this can not be projectModel?
```suggestion
type projectModel struct {
```

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+	RegionUsageRestrictions                     types.String    `tfsdk:"region_usage_restrictions"`
+	ProjectID                                   types.String    `tfsdk:"project_id"`
+	Name                                        types.String    `tfsdk:"name"`
+	OrgID                                       types.String    `tfsdk:"org_id"`
+	Created                                     types.String    `tfsdk:"created"`
+	ID                                          types.String    `tfsdk:"id"`
+	Limits                                      []tfLimitModel  `tfsdk:"limits"`
```
andreaangiolillo:
```suggestion
	Limits                                      []*tfLimitModel  `tfsdk:"limits"`
```

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+	Teams                                       []tfDSTeamModel `tfsdk:"teams"`
```
andreaangiolillo:
```suggestion
	Teams                                       []*tfDSTeamModel `tfsdk:"teams"`
```

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+	ClusterCount                                types.Int64     `tfsdk:"cluster_count"`
+	IsCollectDatabaseSpecificsStatisticsEnabled types.Bool      `tfsdk:"is_collect_database_specifics_statistics_enabled"`
+	IsRealtimePerformancePanelEnabled           types.Bool      `tfsdk:"is_realtime_performance_panel_enabled"`
+	IsSchemaAdvisorEnabled                      types.Bool      `tfsdk:"is_schema_advisor_enabled"`
+	IsPerformanceAdvisorEnabled                 types.Bool      `tfsdk:"is_performance_advisor_enabled"`
+	IsExtendedStorageSizesEnabled               types.Bool      `tfsdk:"is_extended_storage_sizes_enabled"`
+	IsDataExplorerEnabled                       types.Bool      `tfsdk:"is_data_explorer_enabled"`
+}
+
+type tfDSTeamModel struct {
```
andreaangiolillo:
```suggestion
type teamModel struct {
```

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+	TeamID    types.String `tfsdk:"team_id"`
+	RoleNames types.List   `tfsdk:"role_names"`
+}
+
+func (d *ProjectDS) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_project"
+}
```
andreaangiolillo: why do we need this?

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (d *ProjectDS) Configure(ctx context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
```
andreaangiolillo, line 56: Are we forced to call this function Configure? In generate a better go patter would be to call it newClient and return the client to set for the struct. Feel free to skip this suggestion if Configure is part of an interface

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+	} else {
+		name := projectState.Name.ValueString()
+		project, _, err = conn.Projects.GetOneProjectByName(ctx, name)
+		if err != nil {
+			resp.Diagnostics.AddError("error when getting project from Atlas", fmt.Sprintf(errorProjectRead, name, err.Error()))
+			return
+		}
+	}
+
+	// GET PROJECT PROPS
```
andreaangiolillo:
```suggestion
```

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+		project *matlas.Project
+	)
+
+	if !projectState.ProjectID.IsNull() {
+		projectID := projectState.ProjectID.ValueString()
+		project, _, err = conn.Projects.GetOneProject(ctx, projectID)
+		if err != nil {
+			resp.Diagnostics.AddError("error when getting project from Atlas", fmt.Sprintf(errorProjectRead, projectID, err.Error()))
+			return
+		}
```
andreaangiolillo, line 180: why the projectID is null? Shouldn't always be set as part of the create?

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+		resp.Diagnostics.AddError("error when getting project properties", fmt.Sprintf(errorProjectRead, project.ID, err.Error()))
+		return
+	}
+
+	projectState = toTFProjectDataSourceModel(ctx, project, atlasTeams, atlasProjectSettings, atlasLimits)
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &projectState)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
andreaangiolillo, line 201: is this a patter of the framework? I see that we don't return error anymore 🤔

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func toTFProjectDataSourceModel(ctx context.Context, project *matlas.Project,
+	teams *matlas.TeamsAssigned, projectSettings *matlas.ProjectSettings, limits []admin.DataFederationLimit) tfProjectDSModel {
+	projectStateModel := tfProjectDSModel{
+		ID:           types.StringValue(project.ID),
+		ProjectID:    types.StringValue(project.ID),
```
andreaangiolillo, line 208: any reason why we have both id and projectID for the same value?

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+	projectState = toTFProjectDataSourceModel(ctx, project, atlasTeams, atlasProjectSettings, atlasLimits)
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &projectState)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+func toTFProjectDataSourceModel(ctx context.Context, project *matlas.Project,
+	teams *matlas.TeamsAssigned, projectSettings *matlas.ProjectSettings, limits []admin.DataFederationLimit) tfProjectDSModel {
```
andreaangiolillo, line 205: Can the projectSettings be null? if yes, we are going to get null pointers errors in line 218

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,261 @@
...
+		IsDataExplorerEnabled:                       types.BoolValue(*projectSettings.IsDataExplorerEnabled),
+		IsExtendedStorageSizesEnabled:               types.BoolValue(*projectSettings.IsExtendedStorageSizesEnabled),
+		IsPerformanceAdvisorEnabled:                 types.BoolValue(*projectSettings.IsPerformanceAdvisorEnabled),
+		IsRealtimePerformancePanelEnabled:           types.BoolValue(*projectSettings.IsRealtimePerformancePanelEnabled),
+		IsSchemaAdvisorEnabled:                      types.BoolValue(*projectSettings.IsSchemaAdvisorEnabled),
+		Teams:                                       toTFTeamsDataSourceModel(ctx, teams),
+		Limits:                                      toTFLimitsDataSourceModel(ctx, limits),
+	}
+
+	return projectStateModel
```
andreaangiolillo:
```suggestion
	return tfProjectDSModel{
		ID:           types.StringValue(project.ID),
		ProjectID:    types.StringValue(project.ID),
		Name:         types.StringValue(project.Name),
		OrgID:        types.StringValue(project.OrgID),
		ClusterCount: types.Int64Value(int64(project.ClusterCount)),
		Created:      types.StringValue(project.Created),
		IsCollectDatabaseSpecificsStatisticsEnabled: types.BoolValue(*projectSettings.IsCollectDatabaseSpecificsStatisticsEnabled),
		IsDataExplorerEnabled:                       types.BoolValue(*projectSettings.IsDataExplorerEnabled),
		IsExtendedStorageSizesEnabled:               types.BoolValue(*projectSettings.IsExtendedStorageSizesEnabled),
		IsPerformanceAdvisorEnabled:                 types.BoolValue(*projectSettings.IsPerformanceAdvisorEnabled),
		IsRealtimePerformancePanelEnabled:           types.BoolValue(*projectSettings.IsRealtimePerformancePanelEnabled),
		IsSchemaAdvisorEnabled:                      types.BoolValue(*projectSettings.IsSchemaAdvisorEnabled),
		Teams:                                       toTFTeamsDataSourceModel(ctx, teams),
		Limits:                                      toTFLimitsDataSourceModel(ctx, limits),
	}
```

mongodbatlas/fw_data_source_mongodbatlas_projects.go
```diff
@@ -0,0 +1,202 @@
...
+import (
+	"context"
+	"fmt"
+
+	"go.mongodb.org/atlas-sdk/v20230201002/admin"
+	matlas "go.mongodb.org/atlas/mongodbatlas"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
```
andreaangiolillo:
```suggestion
	"go.mongodb.org/atlas-sdk/v20230201002/admin"
	matlas "go.mongodb.org/atlas/mongodbatlas"
	"github.com/hashicorp/terraform-plugin-framework/datasource"
	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
	"github.com/hashicorp/terraform-plugin-framework/types"
```

mongodbatlas/fw_data_source_mongodbatlas_projects.go
```diff
@@ -0,0 +1,202 @@
...
+	return &ProjectsDS{}
+}
+
+type ProjectsDS struct {
+	client *MongoDBClient
+}
+
+type tfProjectsDSModel struct {
+	ID           types.String       `tfsdk:"id"`
+	Results      []tfProjectDSModel `tfsdk:"results"`
```
andreaangiolillo:
```suggestion
	Results      []*tfProjectDSModel `tfsdk:"results"`
```

mongodbatlas/fw_data_source_mongodbatlas_projects.go
andreaangiolillo: Can we create a ticket to track this work (remove the ID once the issue is solved)? Also, instead of using totalCount can we use `id.UniqueId()`?

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: no, I used .IsNull() initially but that didn't work as expected because we also have planmodifier for Unknown state for these fields (so these are basically considered unknown and not null)

mongodbatlas/fw_data_source_mongodbatlas_projects_test.go
```diff
@@ -6,22 +6,19 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-	matlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
 func TestAccProjectDSProjects_basic(t *testing.T) {
 	projectName := fmt.Sprintf("test-datasource-project-%s", acctest.RandString(10))
 	orgID := os.Getenv("MONGODB_ATLAS_ORG_ID")
 	teamsIds := strings.Split(os.Getenv("MONGODB_ATLAS_TEAMS_IDS"), ",")
-	apiKeysIds := strings.Split(os.Getenv("MONGODB_ATLAS_API_KEYS_IDS"), ",")
```
andreaangiolillo, line 18: why did we remove this?

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/setplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+
+	expanders "github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/expanders"
```
andreaangiolillo, line 30: this seems to need a go imports

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+
+func NewProjectRS() resource.Resource {
+	return &ProjectRS{}
+}
+
+type ProjectRS struct {
+	client *MongoDBClient
+}
+
+type tfProjectRSModel struct {
```
andreaangiolillo, line 56:
```suggestion
type projectRSModel struct {
```

mongodbatlas/provider_test.go
maastha: That is still used in existing un-migrated resource tests right?

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	ID                                          types.String `tfsdk:"id"`
+	ClusterCount                                types.Int64  `tfsdk:"cluster_count"`
+	IsDataExplorerEnabled                       types.Bool   `tfsdk:"is_data_explorer_enabled"`
+	IsPerformanceAdvisorEnabled                 types.Bool   `tfsdk:"is_performance_advisor_enabled"`
+	IsRealtimePerformancePanelEnabled           types.Bool   `tfsdk:"is_realtime_performance_panel_enabled"`
+	IsSchemaAdvisorEnabled                      types.Bool   `tfsdk:"is_schema_advisor_enabled"`
+	IsExtendedStorageSizesEnabled               types.Bool   `tfsdk:"is_extended_storage_sizes_enabled"`
+	IsCollectDatabaseSpecificsStatisticsEnabled types.Bool   `tfsdk:"is_collect_database_specifics_statistics_enabled"`
+	WithDefaultAlertsSettings                   types.Bool   `tfsdk:"with_default_alerts_settings"`
+}
```
andreaangiolillo, line 73: Can't we do this? Or the opposite, the DS using the resource model
```suggestion
type tfProjectRSModel struct {
tfProjectDSModel
}
```

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: yeah let me give that a try

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+		return
+	}
+	client, ok := req.ProviderData.(*MongoDBClient)
+
+	if !ok {
+		resp.Diagnostics.AddError(errorConfigureSummary, fmt.Sprintf(errorConfigure, req.ProviderData))
+		return
+	}
+	r.client = client
+}
```
andreaangiolillo: this seems used also by other ds and resources. Can we extract the implementation in another function and reuse it inside the Configure?

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	_, _, err = conn.Projects.UpdateProjectSettings(ctx, project.ID, projectSettings)
+	if err != nil {
+		errd := deleteProject(ctx, r.client.Atlas, project.ID)
+		if errd != nil {
+			resp.Diagnostics.AddError("error during project deletion when updating project settings", fmt.Sprintf(errorProjectDelete, project.ID, err.Error()))
+			return
+		}
+		resp.Diagnostics.AddError(fmt.Sprintf("error updating project's settings assigned (%s):", project.ID), err.Error())
+		return
+	}
```
andreaangiolillo:
```suggestion
	if _, _, err = conn.Projects.UpdateProjectSettings(ctx, project.ID, projectSettings); err != nil {
		errd := deleteProject(ctx, r.client.Atlas, project.ID)
		if errd != nil {
			resp.Diagnostics.AddError("error during project deletion when updating project settings", fmt.Sprintf(errorProjectDelete, project.ID, err.Error()))
			return
		}
		resp.Diagnostics.AddError(fmt.Sprintf("error updating project's settings assigned (%s):", project.ID), err.Error())
		return
	}
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+
+	// READ
+	// get project
```
andreaangiolillo:
```suggestion

```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	}
+
+	err = updateProjectSettings(ctx, conn, &projectState, &projectPlan)
+	if err != nil {
+		resp.Diagnostics.AddError("error in project settings update", fmt.Sprintf(errorProjectUpdate, projectID, err.Error()))
+		return
+	}
+
+	// READ
+	// get project
```
andreaangiolillo:
```suggestion
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	err := deleteProject(ctx, r.client.Atlas, projectID)
+
+	if err != nil {
+		resp.Diagnostics.AddError("error when destroying resource", fmt.Sprintf(errorProjectDelete, projectID, err.Error()))
+		return
+	}
+}
+
+func (r *ProjectRS) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```
andreaangiolillo, line 525:
```suggestion
	resource.ImportStatePassThroughID(ctx, path.Root("id"), req, resp)
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	// remove all current teams
+	for _, team := range stateTeams {
+		_, err := conn.Teams.RemoveTeamFromProject(ctx, projectID, team.TeamID.ValueString())
+		if err != nil {
+			return fmt.Errorf("error removing team from the project: %v", err.Error())
+		}
+	}
+
+	// adding updated teams into the project
+	if len(planTeams) > 0 {
```
andreaangiolillo:
```suggestion
	if len(planTeams) == 0 {
	    return nil
	}
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+		_, _, err := conn.Projects.AddTeamsToProject(ctx, projectID, toAtlasProjectTeams(ctx, planTeams))
+		if err != nil {
+			return fmt.Errorf("error adding teams to the project: %v", err.Error())
+		}
```
andreaangiolillo:
```suggestion
		if _, _, err := conn.Projects.AddTeamsToProject(ctx, projectID, toAtlasProjectTeams(ctx, planTeams)); err != nil {
			return fmt.Errorf("error adding teams to the project: %v", err.Error())
		}
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+
+/*
+This assumes the project CRUD outcome will be the same for any non-zero number of dependents
+
+If all dependents are deleting, wait to try and delete
+Else consider the aggregate dependents idle.
+
+If we get a defined error response, return that right away
+Else retry
+*/
```
andreaangiolillo:
```suggestion
resourceProjectDependentsDeletingRefreshFunc assumes the project CRUD outcome will be the same for any non-zero number of dependents

If all dependents are deleting, wait to try and delete
Else consider the aggregate dependents idle.

If we get a defined error response, return that right away
Else retry
*/
```
```suggestion
This assumes the project CRUD outcome will be the same for any non-zero number of dependents

If all dependents are deleting, wait to try and delete
Else consider the aggregate dependents idle.

If we get a defined error response, return that right away
Else retry
*/
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+	return func() (interface{}, string, error) {
+		var target *matlas.ErrorResponse
+		clusters, _, err := client.AdvancedClusters.List(ctx, projectID, nil)
+		dependents := AtlastProjectDependents{AdvancedClusters: clusters}
+
+		if errors.As(err, &target) {
+			return nil, "", err
+		} else if err != nil {
+			return nil, "RETRY", nil
+		}
```
andreaangiolillo:
```suggestion
		if errors.As(err, &target) {
			return nil, "", err
		}
		
		if err != nil {
			return nil, "RETRY", nil
		}
```

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+If we get a defined error response, return that right away
+Else retry
+*/
+func resourceProjectDependentsDeletingRefreshFunc(ctx context.Context, projectID string, client *matlas.Client) retry.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		var target *matlas.ErrorResponse
+		clusters, _, err := client.AdvancedClusters.List(ctx, projectID, nil)
+		dependents := AtlastProjectDependents{AdvancedClusters: clusters}
+
+		if errors.As(err, &target) {
```
andreaangiolillo, line 855: what is target? can we give it a better name?

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+			return nil, "", err
+		} else if err != nil {
+			return nil, "RETRY", nil
```
andreaangiolillo: retry should be a constant. Also why are we returning it?

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+		}
+
+		if dependents.AdvancedClusters.TotalCount == 0 {
+			return dependents, "IDLE", nil
```
andreaangiolillo: IDLE should be a constant

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,857 @@
...
+		}
+
+		for _, v := range dependents.AdvancedClusters.Results {
+			if v.StateName != "DELETING" {
```
andreaangiolillo: same

mongodbatlas/fw_resource_mongodbatlas_project_migration_test.go
```diff
@@ -0,0 +1,229 @@
...
+				Config: fmt.Sprintf(`resource "mongodbatlas_project" "test" {
+					name   = "%s"
+					org_id = "%s"
+				  }`, projectName, orgID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "org_id", orgID),
+				),
+			},
+			{
+				ProtoV6ProviderFactories: testAccProviderV6Factories,
```
andreaangiolillo, line 62: how this step will work? is the first step using the old provider and the second step the framework?

mongodbatlas/provider_test.go
```diff
@@ -22,6 +24,9 @@ var testAccProviderV6Factories map[string]func() (tfprotov6.ProviderServer, erro
 // this provider instance has to be passed into mux server factory for its configure method to be invoked
 var testAccProviderSdkV2 *schema.Provider
 
+// Manually configured client client required for Framework-based acceptance tests
```
andreaangiolillo:
```suggestion
// testMongoDBClient is used to configure client required for Framework-based acceptance tests
```

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: That was to avoid confusion with atlas models, any models/types that might be created only for tests or temporarily otherwise for any computation. I initally did have `projectModel` but overall I found `tf` prefix to be much more readable especially when dealing with conversion functions, etc, that this is specific to the actual terraform plan/state/config model in question.

These models contain attributes that are Terraforms types, so these are not normal Go models.

Also, since right now we have everything in one package we can end up with naming conflicts when we start adding models for all resources as well as their corresponding data sources. So I think we should definitely keep these separate.

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: The Framework uses Metadata() method to define resource name. In this case resp.TypeName will be `mongodbatlas_project` where `mongodbatlas` is the req.ProviderTypeName defined in the provider.

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: Yes. Configure is part of an interface.

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: This is the data source, so the user can set either name or project_id, both of which are optional and we query accordingly based on what the user has provided. This logic has been migrated as is from the previous data source.
https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/data-sources/project#argument-reference

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: Right. We don't have to return the error anymore, but only add them to `resp.Diagnostics` and return

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: Yes. // https://github.com/hashicorp/terraform-plugin-testing/issues/84#issuecomment-1480006432
Every data source needs to have an 'id' value. As per comment here https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1365#discussion_r1284372784

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: projectSettings is always retrieved from `getProjectPropsFromAPI()` where we do check if the API ever returns an error so this cannot be null.

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: I do have that setup. This is how it auto-formats all the imports for me. We ran into this last time too. Let's sync our goimports settings because this is auto-formatted :smile:

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: No. Unfortunately there were differences in all models. For example, data source models don't have `WithDefaultAlertsSettings` or `ProjectOwnerID` fields and these models need to be exactly same as the schemas

andreaangiolillo: Thanks!

mongodbatlas/fw_resource_mongodbatlas_project.go
andreaangiolillo: 👍

mongodbatlas/fw_data_source_mongodbatlas_project.go
andreaangiolillo: Thanks for confirming this

mongodbatlas/fw_data_source_mongodbatlas_project.go
andreaangiolillo: Thanks

mongodbatlas/fw_data_source_mongodbatlas_project.go
andreaangiolillo: Can we update this to make project a constant and use format? `("%s_%s, req.ProviderTypeName, projectResourceName)`

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: Can change it like this, didn't seem to add a lot because of the error handling that's why didn't do it initially. But let me know.

```suggestion
func (r *ProjectRS) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
	client, err := configure(req.ProviderData)
	if err != nil {
		resp.Diagnostics.AddError(errorConfigureSummary, err.Error())
		return
	}
	r.client = client
}

func configure(providerData any) (*MongoDBClient, error) {
	if providerData == nil {
		return nil, fmt.Errorf("ProviderData is null")
	}
	client, ok := providerData.(*MongoDBClient)
	if !ok {
		return nil, fmt.Errorf(errorConfigure, providerData)
	}
	return client, nil
}
```

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: This is an in-built terraform function

mongodbatlas/fw_resource_mongodbatlas_project_migration_test.go
maastha: Correct. First step will use version 1.11.0 of our provider that was published today.
Second step will create a plan using the Framework-based implementation of this resource/provider and check that no changes should be discovered as part of the plan.

mongodbatlas/provider_test.go
AgustinBettati: Correct. Essentially we can leave either `testMongoDBClient`, or the usage of `testAccProviderSdkV2.Meta().(*MongoDBClient)`. Between both I prefer testMongoDBClient.

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: This is called when deleting project. RETRY is returned so it can be checked as a status for a refresh function. See https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1365/files#diff-c2af32c41f07d77e630b76e890c8a1dfee1e053fdeac86f8fd31a0bb8e5c9663R801

This method migrated as is from SDKv2 resource

mongodbatlas/fw_data_source_mongodbatlas_projects_test.go
maastha: Because we are also removing api_keys as part of this. Do you mean something else?

mongodbatlas/provider_test.go
maastha: I don't think we should change that in all the other resource and data source tests unless required as we'd be running all the tests then. At least not as part of this PR. We can address that separately.

mongodbatlas/fw_data_source_mongodbatlas_projects.go
maastha: I'm not sure how id.UniqueId() works, but I will give that a try

mongodbatlas/provider_test.go
AgustinBettati: Thats okay, in that case I would however remove the created `testMongoDBClient` in favor of the existing `testAccProviderSdkV2.Meta().(*MongoDBClient)` that is being used in acceptance tests, mainly for consistency.

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,862 @@
...
+const (
+	errorProjectCreate  = "error creating Project: %s"
+	errorProjectRead    = "error getting project(%s): %s"
+	errorProjectDelete  = "error deleting project (%s): %s"
+	errorProjectSetting = "error setting `%s` for project (%s): %s"
+	errorProjectUpdate  = "error updating project (%s): %s"
+)
+
+var _ resource.Resource = &ProjectRS{}
+var _ resource.ResourceWithImportState = &ProjectRS{}
```
AgustinBettati, line 46: not sure if I see value in these statements, checking of correct interface implementation will be done as soon as resource is configured in the provider

mongodbatlas/provider_test.go
maastha: @AgustinBettati While that would be consistent in the sense that we use only `testAccProviderSdkV2.Meta().(*MongoDBClient)` throughout our acceptance tests to get the client, it would be inconsistent and incorrect in context of a **Framework-based migrated** resource and it's tests to continue using an SDKv2 based entity. This would also require us to update all the tests once we are ready to get rid of everything that SDKv2 in our provider.

We can instead rename `testMongoDBClient` to  something like `testMongoDBClient_Fw` (open to suggestions for the name) to clarify that this is specific to the Framework migrated resources.
Let me know what you think.

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: Could you please elaborate `as soon as resource is configured in the provider`?

AgustinBettati: When the resource is added into the provider, it will have to comply with the resource.Resource interface: 
`return []func() resource.Resource{
		NewProjectRS,
}`


I do however see that for the ImportState interface there would be no compilation errors if something is incorrectly defined, so line 42 does have value.
To add, hashicorp example uses this convention: https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/internal/provider/example_resource.go#L22 so we can follow with it.

mongodbatlas/fw_data_source_mongodbatlas_project.go
AgustinBettati: As mentioned [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1366#discussion_r1284562820), I have found the `tf` prefix helpful when reading and developing.

mongodbatlas/provider_test.go
AgustinBettati: If we will progressively transition from testAccProviderSdkV2.Meta().(*MongoDBClient) to using the new testMongoDBClient then I think the existing comments are sufficient

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: Discussed this with @andreaangiolillo offline as well. We will continue using `tf` an `fw` naming conventions. 
This will also be captured in this migration wiki as well.

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: Resolving as we will be using `tfResourceNameRSModel` as convention as discussed.

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: Resolving as we will be using this naming convention as discussed.

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: updated

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: updated

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: updated to constant.

mongodbatlas/fw_data_source_mongodbatlas_projects.go
maastha: updated

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: updated

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: updated

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: updated

mongodbatlas/fw_data_source_mongodbatlas_projects.go
maastha: updated to use  id.UniqueId()

mongodbatlas/fw_data_source_mongodbatlas_projects_test.go
andreaangiolillo: I forgot that we are deprecating apiKeysIds. Resolving this

andreaangiolillo (approved):
LGTM. Thanks for addressing all the comments. (Make sure that the failing tests is not related to your changes)

mongodbatlas/fw_resource_mongodbatlas_project.go
maastha: implemented getChangesInSet for teams and limits

maastha:
> LGTM. Thanks for addressing all the comments. (Make sure that the failing tests is not related to your changes)

All project* acceptance tests passing.

AgustinBettati (approved):

mongodbatlas/fw_data_source_mongodbatlas_project.go
```diff
@@ -0,0 +1,255 @@
+package mongodbatlas
+
```
Zuhairahmed, line 2: does "fw_" ahead of each of the file names stand for "framework"? if so suggest we call out somewhere in repo README file to let users know which resources / data sources have been migrated to new Terraform Plugin Framework

Zuhairahmed (approved):
minor suggestion comment, otherwise LGTM

mongodbatlas/fw_data_source_mongodbatlas_project.go
maastha: @Zuhairahmed added to CONTRIBUTING.md

mongodbatlas/fw_resource_mongodbatlas_project.go
```diff
@@ -0,0 +1,941 @@
...
+	_ = projectPlan.Limits.ElementsAs(ctx, &planLimits, false)
+	atlasLimits = filterUserDefinedLimits(atlasLimits, planLimits)
+	projectPlanNew := toTFProjectResourceModel(ctx, projectRes, atlasTeams, atlasProjectSettings, atlasLimits)
+	updatePlanFromConfig(projectPlanNew, &projectPlan)
+
+	// save updated data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &projectPlanNew)...)
+}
+
+func (r *ProjectRS) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
```
andreaangiolillo, line 506: .

PR #1366, opened 2023-08-03
feat: new framework provider, main and acceptance tests to use mux server with existing sdk v2 provider

AgustinBettati:
## Description

Ticket: [INTMDB-971](https://jira.mongodb.org/browse/INTMDB-971)

Add new Terraform Plugin Framework and enable mux server so existing SDK v2 provider can be combined together. All acceptance tests and main file are now using the muxed server provider.

## Type of change:

## Required Checklist:

## Further comments

- Manual validation was carried out to force use of framework provider for parsing provider configuration and generate same schema validations.
- Regarding documentation: readme and all examples were updated with new terraform version requirement. Website documentation has not been updated yet (need to define correct place to write statement).


AgustinBettati:
As we are updating to terraform protocol v6, moving forward users will have to use terraform CLI version 1.0+ (previous requirement was 10.13+). I adjusted all examples and README.md to state this new requirement, however I do not see a clear place where we can mention this in the homepage of our terraform registry. Looked into other providers and have not seen this requirement mentioned there either. @Zuhairahmed do you have inputs on this?

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+	RoleARN           types.String `tfsdk:"role_arn"`
+	SessionName       types.String `tfsdk:"session_name"`
+	SourceIdentity    types.String `tfsdk:"source_identity"`
+	Tags              types.Map    `tfsdk:"tags"`
+	TransitiveTagKeys types.Set    `tfsdk:"transitive_tag_keys"`
+}
+
+func (p *MongodbtlasProvider) Metadata(ctx context.Context, req provider.MetadataRequest, resp *provider.MetadataResponse) {
+	resp.TypeName = "mongodbatlas"
+	// resp.Version = "" we can look into defining this value automatically for release process
```
andreaangiolillo: is this a left over? if not, could you add addition info about when this code will be removed? A jira ticket is enough

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+	AssumeRole           types.List   `tfsdk:"assume_role"`
+	SecretName           types.String `tfsdk:"secret_name"`
+	Region               types.String `tfsdk:"region"`
+	StsEndpoint          types.String `tfsdk:"sts_endpoint"`
+	AwsAccessKeyID       types.String `tfsdk:"aws_access_key_id"`
+	AwsSecretAccessKeyID types.String `tfsdk:"aws_secret_access_key"`
+	AwsSessionToken      types.String `tfsdk:"aws_session_token"`
+}
+
+type tfAssumeRoleModel struct { //nolint:govet //prefer to maintain order of attributes over small byte optimization
```
andreaangiolillo: I don't like the `tf` |`fw`. Can we use framework instead?

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-mux/tf5to6server"
+	"github.com/hashicorp/terraform-plugin-mux/tf6muxserver"
+	sdkv2schema "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	cstmvalidator "github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/validator"
+)
+
+type MongodbtlasProvider struct{}
+
+type tfMongodbAtlasProviderModel struct { //nolint:govet //prefer to maintain order of attributes over small byte optimization
```
andreaangiolillo:
```suggestion
type tfMongodbAtlasProviderModel struct { //nolint:govet //prefer to maintain order of attributes over small byte optimization
```

this is not what we decided to do for TF. happy to open again this discussion if you want (feel free to open a discussion on team slack channel) but for now, you should run `make fix-lint` and remove `nolint`

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+			},
+			"aws_access_key_id": schema.StringAttribute{
+				Optional: true,
+			},
+			"aws_secret_access_key": schema.StringAttribute{
+				Optional: true,
+			},
+			"aws_session_token": schema.StringAttribute{
+				Optional: true,
+			},
```
andreaangiolillo: let's add a description for all the fields

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+				Description: "Assume role session tag keys to pass to any subsequent sessions.",
+			},
+		},
+	},
+}
+
+func (p *MongodbtlasProvider) Configure(ctx context.Context, req provider.ConfigureRequest, resp *provider.ConfigureResponse) {
+	var (
+		data tfMongodbAtlasProviderModel
+	)
```
andreaangiolillo:
```suggestion
	var data tfMongodbAtlasProviderModel

```

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
andreaangiolillo, line 204: let's follow golanf standards and return the error

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+
+	defaultValuesAndValidations(&data, resp)
```
andreaangiolillo: this is really hard to read. instead of using a pointer, let's return the object as part of the function

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+		PublicKey:    data.PublicKey.ValueString(),
+		PrivateKey:   data.PrivateKey.ValueString(),
+		BaseURL:      data.BaseURL.ValueString(),
+		RealmBaseURL: data.RealmBaseURL.ValueString(),
+	}
+
+	var assumeRoles []tfAssumeRoleModel
+	data.AssumeRole.ElementsAs(ctx, &assumeRoles, true)
+
+	if len(assumeRoles) > 0 {
```
andreaangiolillo, line 221:
```suggestion
	if len(assumeRoles) == 0 {
	return nil
	}
```

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+		config.AssumeRole = parseTfModel(ctx, &assumeRoles[0])
+		secret := data.SecretName.ValueString()
+		region := data.Region.ValueString()
+		awsAccessKeyID := data.AwsAccessKeyID.ValueString()
+		awsSecretAccessKey := data.AwsSecretAccessKeyID.ValueString()
+		awsSessionToken := data.AwsSessionToken.ValueString()
+		endpoint := data.StsEndpoint.ValueString()
+		var err error
+		config, err = configureCredentialsSTS(&config, secret, region, awsAccessKeyID, awsSecretAccessKey, awsSessionToken, endpoint)
```
andreaangiolillo: passing a pointer to a function and then changing the value of the object of the pointer is not a good pattern as decreases readability

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+			err.Error(),
+		)
+		return
+	}
+
+	resp.DataSourceData = client
+	resp.ResourceData = client
+}
+
+func parseTfModel(ctx context.Context, tfAssumeRoleModel *tfAssumeRoleModel) *AssumeRole {
```
andreaangiolillo, line 252:
```suggestion
func parseModel(ctx context.Context, tfAssumeRoleModel *tfAssumeRoleModel) *AssumeRole {
```

andreaangiolillo, line 252: Could you add a comment to explain what is the goal of the function?

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+		data.AwsSessionToken = types.StringValue(MultiEnvDefaultFunc([]string{
+			"AWS_SESSION_TOKEN",
+			"TF_VAR_AWS_SESSION_TOKEN",
+		}, "").(string))
+	}
+}
+
+func (p *MongodbtlasProvider) DataSources(context.Context) []func() datasource.DataSource {
+	return []func() datasource.DataSource{
+		// NewExampleDataSource,
```
andreaangiolillo, line 363:
```suggestion
```

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+}
+
+func (p *MongodbtlasProvider) Resources(context.Context) []func() resource.Resource {
+	return []func() resource.Resource{
+		// NewExampleResource,
```
andreaangiolillo, line 369:
```suggestion
```

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,394 @@
...
+
+func NewFrameworkProvider() provider.Provider {
+	return &MongodbtlasProvider{}
+}
+
+func MuxProviderFactory() func() tfprotov6.ProviderServer {
+	return muxProviderFactoryUsingExistingSdkV2(NewSdkV2Provider())
+}
+
+// creates mux server using existing sdk v2 provider passed as parameter, and new framework provider
```
andreaangiolillo:
```suggestion
// muxProviderFactoryUsingExistingSdkV2 creates mux server using existing sdk v2 provider passed as parameter, and new framework provider
```

mongodbatlas/provider.go
```diff
@@ -337,7 +302,106 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
...
+	if err != nil {
+		return nil, diag.FromErr(err)
+	}
+	return client, nil
+}
+
+func setDefaultsAndValidations(d *schema.ResourceData) error {
+	mongodbgovCloud := pointy.Bool(d.Get("is_mongodbgov_cloud").(bool))
+	if *mongodbgovCloud {
+		if err := d.Set("base_url", "https://cloud.mongodbgov.com"); err != nil {
```
andreaangiolillo: https://cloud.mongodbgov.com should be a constant

mongodbatlas/provider.go
```diff
@@ -337,7 +302,106 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
...
+			return err
+		}
+	} else {
+		if err := setValueFromConfigOrEnv(d, "base_url", []string{
+			"MONGODB_ATLAS_BASE_URL",
+			"MCLI_OPS_MANAGER_URL",
+		}); err != nil {
+			return err
+		}
+	}
```
andreaangiolillo: base_url should work with both gov and not gov
```suggestion
	if *mongodbgovCloud {
		if err := d.Set("base_url", "https://cloud.mongodbgov.com"); err != nil {
			return err
		}
}
		if err := setValueFromConfigOrEnv(d, "base_url", []string{
			"MONGODB_ATLAS_BASE_URL",
			"MCLI_OPS_MANAGER_URL",
		}); err != nil {
			return err
		}
	}
```

AgustinBettati: For this case I replicated the same behaviour we have in our existing sdk v2 provider https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/provider.go#L317.
Not sure if we would want to incorporate this change within the changes of the migration.

mongodbatlas/provider.go
AgustinBettati: will adjust

mongodbatlas/fw_provider.go
AgustinBettati: adjusted

mongodbatlas/fw_provider.go
AgustinBettati: Is being deleted in project migration PR (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1365/files#diff-50a1299c66e854062730103c8144024ebf3a8ecc6f1b620231d15ed6f5997424L354)

mongodbatlas/fw_provider.go
AgustinBettati: same as above: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1365/files#diff-50a1299c66e854062730103c8144024ebf3a8ecc6f1b620231d15ed6f5997424L360

mongodbatlas/fw_provider.go
AgustinBettati: adjusted

mongodbatlas/fw_provider.go
AgustinBettati: Adjusted. No strong feeling here, just found it less error prone to preserve a certain order during development.

mongodbatlas/fw_provider.go
AgustinBettati: yes, was left over. The Version attribute is currently not being used for any functionality (https://github.com/hashicorp/terraform-plugin-framework/blob/55f0d6bc4be2a57adb06a5c25c28640712c88129/provider/metadata.go#L21C2-L21C2).
However, looking more into this found the version is easily accessible so will include now.

mongodbatlas/fw_provider.go
AgustinBettati: I did like the tf convention as it is concise and quickly lets you know that the model is used for handling data to/from terraform. We can carry the conversation over here https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1365#discussion_r1284391764 as I believe it is the same.

mongodbatlas/fw_provider.go
maastha: as discussed in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1365, this is how Framework handles errors

mongodbatlas/fw_provider.go
```diff
@@ -0,0 +1,395 @@
...
+func NewFrameworkProvider() provider.Provider {
+	return &MongodbtlasProvider{}
+}
+
+func MuxProviderFactory() func() tfprotov6.ProviderServer {
+	return muxProviderFactoryUsingExistingSdkV2(NewSdkV2Provider())
+}
+
+// muxProviderFactoryUsingExistingSdkV2 creates mux server using existing sdk v2 provider passed as parameter, and new framework provider
+func muxProviderFactoryUsingExistingSdkV2(sdkV2Provider *sdkv2schema.Provider) func() tfprotov6.ProviderServer {
```
maastha: I'd just rename this to `muxedProviderFactory` or `muxedProviderServer`

mongodbatlas/fw_provider.go
AgustinBettati: Added descriptions based off of the following documentation: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs#aws-secrets-manager

mongodbatlas/fw_provider.go
AgustinBettati: adjusted to muxedProviderFactory, also added comment with context of why this function is defined.

Zuhairahmed:
> As we are updating to terraform protocol v6, moving forward users will have to use terraform CLI version 1.0+ (previous requirement was 10.13+). I adjusted all examples and README.md to state this new requirement, however I do not see a clear place where we can mention this in the homepage of our terraform registry. Looked into other providers and have not seen this requirement mentioned there either. @Zuhairahmed do you have inputs on this?

https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/website/docs/index.html.markdown

mongodbatlas/fw_provider.go
andreaangiolillo: This is not addressed

mongodbatlas/fw_provider.go
andreaangiolillo: This is not addressed

mongodbatlas/fw_provider.go
andreaangiolillo: This is not addressed

mongodbatlas/fw_provider.go
andreaangiolillo: This is not addressed

mongodbatlas/provider.go
andreaangiolillo: This change just improves readability so I would add it

AgustinBettati: adjusted

mongodbatlas/fw_provider.go
AgustinBettati: Not sure if I understand this suggestion, if assume roles is not defined we will still want to create a client right? Having the return would not allow this.

andreaangiolillo: yes, you are right. Resolving this comment

mongodbatlas/fw_provider.go
AgustinBettati: adjusted

mongodbatlas/fw_provider.go
AgustinBettati: just to be clear, to avoid any mutations to the parameter that is being passed, the config would be passed as a value and all changes would be present in the returned config correct?

mongodbatlas/fw_provider.go
AgustinBettati: gocritic does suggest passing data parameter by pointer due to size (290 bytes), we could still return the modified by value to make the function signature more clear.

andreaangiolillo: it is okay to pass a pointer as input. IMO what decreases readability is to change the value of the object of the pointer inside the function. A better approach is to get the pointer as input and then return a new object 
Example`data := defaultValuesAndValidations(&data, resp)`

mongodbatlas/fw_provider.go
andreaangiolillo: no, you can pass config as a reference. My point here is to make sure to not change the object pointed by `config` inside the function

AgustinBettati: From what I understand we have 2 options to address the concern:

- receive the config as value, so it is clear that all changes will only be present in the returned config.
- receive by pointer, and within the function create a new instance of the struct before modifying.

I do feel the first one is more clear, but maybe I am not fully understanding the second option.

mongodbatlas/fw_provider.go
AgustinBettati: correct, this is aligned with my most recent commit

mongodbatlas/fw_provider.go
andreaangiolillo: I am okay with both options. [Keep in mind that option n1 may make linter sad if the struct size is above 80 bytes](https://go-critic.com/overview.html#hugeparam)

andreaangiolillo (commented):
LGTM. Thanks for addressing all the comments. Remaining comments are not blocking and can be addressed in a follow-up PR

andreaangiolillo (approved):

maastha (approved):
LGTM, if tests are not blocking.

AgustinBettati:
All checks have passed, merging PR.

PR #1383, opened 2023-08-11
refactor: Migrate mongodbatlas_encryption_at_rest to Terraform Plugin Framework

maastha:
## Description

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments
Migration and acceptance tests run locally:
Screenshot 2023-08-23 at 12 34 16

Import:
Screenshot 2023-08-22 at 23 38 13
Resource in state after import:
```
  {
      "mode": "managed",
      "type": "mongodbatlas_encryption_at_rest",
      "name": "example",
      "provider": "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "aws_kms_config": [
              {
                "access_key_id": null,
                "customer_master_key_id": "arn:aws:kms:us-east-1:....................xxxxxxxxxxxxxxxxxxxxxxxxxx",
                "enabled": true,
                "region": "US_EAST_1",
                "role_id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
                "secret_access_key": null
              }
            ],
            "azure_key_vault_config": [
              {
                "azure_environment": "",
                "client_id": "",
                "enabled": false,
                "key_identifier": "",
                "key_vault_name": "",
                "resource_group_name": "",
                "secret": null,
                "subscription_id": "",
                "tenant_id": ""
              }
            ],
            "google_cloud_kms_config": [
              {
                "enabled": false,
                "key_version_resource_id": "",
                "service_account_key": null
              }
            ],
            "id": "63bec58b014da65b8f73c06c",
            "project_id": "63bec58b014da65b8f73c06c"
          },
          "sensitive_attributes": []
        }
      ]
    },
```


mongodbatlas/framework/validator/aws_kms_config_validator.go
```diff
@@ -0,0 +1,51 @@
...
+		response.Diagnostics.Append(diag.Errors()...)
+		return
+	}
+
+	attrMap := obj.Attributes()
+	ak, akOk := attrMap["access_key_id"]
+	sa, saOk := attrMap["secret_access_key"]
+	r, rOk := attrMap["role_id"]
+
+	if ((akOk && !ak.IsNull()) && (saOk && !sa.IsNull()) && (rOk && !r.IsNull())) ||
```
AgustinBettati:
```suggestion
	if (accessKeyDefined && secretAccessKeyDefined && roleIdDefined) ||
```
Would extract variables just to make it easier to read

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -0,0 +1,576 @@
...
+		}
+		if err != nil {
+			resp.Diagnostics.AddError(fmt.Sprintf(errorCreateEncryptionAtRest, projectID), err.Error())
+			return
+		}
+		break
+	}
+
+	// read
+	encryptionResp, response, err := conn.EncryptionsAtRest.Get(context.Background(), projectID)
```
AgustinBettati: Is there a particular reason why we are reading after the create? I am seeing for migrated resource we rely directly of the response from the create endpoints to populate the state.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -0,0 +1,576 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	projectID := encryptionAtRestPlan.ProjectID.ValueString()
+	encryptionAtRestReq := &matlas.EncryptionAtRest{
+		GroupID:        projectID,
+		AwsKms:         *newAtlasAwsKms(encryptionAtRestPlan.AwsKmsConfig),
+		AzureKeyVault:  *newAtlasAzureKeyVault(encryptionAtRestPlan.AzureKeyVaultConfig),
+		GoogleCloudKms: *newAtlasGcpKms(encryptionAtRestPlan.GoogleCloudKmsConfig),
```
AgustinBettati: to keep in mind, I can see that in previous implementation these fields were defined in request only if defined in configuration
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/af8d75bc7f81d2073931e776cba3fa9059cf2d5d/mongodbatlas/resource_mongodbatlas_encryption_at_rest.go#L231
Now even if not defined we are still defining the field with empty objects.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -0,0 +1,576 @@
...
+	// get current state
+	resp.Diagnostics.Append(req.State.Get(ctx, &encryptionAtRestState)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	projectID := encryptionAtRestState.ProjectID.ValueString()
+
+	// Use the ID only with the IMPORT operation
+	if encryptionAtRestState.ID.ValueString() != "" && (projectID == "") {
+		projectID = encryptionAtRestState.ID.ValueString()
```
AgustinBettati, line 282: Could we define the ProjectID during the import function? This is an example:
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-189585-plugin-framework-migration/mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go#L552

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -0,0 +1,576 @@
...
+
+func hasAzureKeyVaultConfigChanged(azureKeyVaultConfigPlan, azureKeyVaultConfigState []tfAzureKeyVaultConfigModel) bool {
+	return !reflect.DeepEqual(azureKeyVaultConfigPlan, azureKeyVaultConfigState)
+}
+
+func hasAwsKmsConfigChanged(awsKmsConfigPlan, awsKmsConfigState []tfAwsKmsConfigModel) bool {
+	return !reflect.DeepEqual(awsKmsConfigPlan, awsKmsConfigState)
+}
+
+func resetDefaultsFromConfig(ctx context.Context, encryptionAtRestRSCurrent, encryptionAtRestRSNew, encryptionAtRestRSConfig *tfEncryptionAtRestRSModel) {
```
AgustinBettati: finding it hard to understand why method is receiving 3 terraform models. For resetting defaults, I would imagine you need to receive the state to be mutated, and the state from which to obtain values. 
Would start by clarifying which one is being mutated, or possibly adding as a return value.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest_test.go
```diff
@@ -116,10 +117,11 @@ func TestAccAdvRSEncryptionAtRest_basicAWS(t *testing.T) {
 			SecretAccessKey:     os.Getenv("AWS_SECRET_ACCESS_KEY_UPDATED"),
 			CustomerMasterKeyID: os.Getenv("AWS_CUSTOMER_MASTER_KEY_ID_UPDATED"),
 			Region:              os.Getenv("AWS_REGION_UPDATED"),
+			RoleID:              os.Getenv("AWS_ROLE_ID"),
 		}
 	)
 
-	resource.ParallelTest(t, resource.TestCase{
```
AgustinBettati, line 122: curious about this change, was it for a particular reason?

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: I was basically trying to replicate how this works today, in SDKv2 implementations, you'll notice we always call the read handler at the end of Create/Update handlers for the final response. I'll update this if the Create response is sufficient, let me confirm.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: You're right, nice catch, I had that logic in place initially but for some reason got removed during a refactoring, I'll add that back. Although TBH I don't think that really makes a difference from what I've seen.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: This was intentional, by using this condition I'm also setting variable `isImport` which is used to handle import functionality in Read()

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest_test.go
maastha: Yes, these tests currently do not run in CI and are only run locally. When running all tests locally together if we try to run all tests at once that results in inconsistencies because all of them currently use the same projectID to set/update/delete the same resource.

Let me try to update the configs for this resource to create separate projects and see if that works

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: I understand, I've added comments inside the handlers to make it more clear, I'm happy to add additional comments in this method as well.
 I faced some issues in this area (resetting values) with this resource because certain values are not returned by the API on Create/Update which need to be taken up from the config to keep it consistent with the state, otherwise the Framework throws an error for inconsistency.
Another scenario is that the API returns all blocks whether or not the user set it, in that scenario we need to unset the unused blocks based on what is in the config otherwise we run into the same inconsistency issue.

The idea is that during Read() (also called during Import where there's no user-defined config) we don't have the Config object available so we use the state/plan instead for this, but for Create/Update we have access to the Config object which we then use for this purpose.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest_test.go
maastha: Tried this and we can't create separate projects for each test in the corresponding TF config because that messes with the checkDestroy functionality in test.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: added additional comments to make it more clear

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: updated

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: updated

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest_test.go
AgustinBettati: Thanks for looking into this and providing context.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
AgustinBettati: Thanks for the comments in the code, this helps a lot. 
For reference, in alert configuration I faced a similar issue were I had certain attributes that where not returned by the API and had to be defined from existing state/config to avoid errors. The following function https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-189585-plugin-framework-migration/mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go#L641 was defined which receives only a single tfModel and receives the terraform state or plan depending on the context it is invoked:
> Create -> pass the Plan // planned state for the resource.
> Read -> pass the State // current state for the resource.
> Update -> pass the Plan // planned state for the resource.
> Import -> pass empty state as no existing state or plan is available.

You might have additional edge cases that justify receiving 2 models, but thought it would be worth sharing.

maastha: @AgustinBettati yes, thank you for sharing this! You're right, for encryption at rest there are additional edge cases as listed above. The initial approach here was to use only a single other model but due to above cases, ended up having to use 2. Also, was trying to keep any resetting logic in a single method for all CRUD handlers.

AgustinBettati: Okay, thanks for the follow up!

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -0,0 +1,566 @@
...
+	for i := 0; i < 5; i++ {
+		encryptionResp, _, err = conn.EncryptionsAtRest.Create(ctx, encryptionAtRestReq)
+		if err != nil {
+			if strings.Contains(err.Error(), "CANNOT_ASSUME_ROLE") || strings.Contains(err.Error(), "INVALID_AWS_CREDENTIALS") ||
+				strings.Contains(err.Error(), "CLOUD_PROVIDER_ACCESS_ROLE_NOT_AUTHORIZED") {
+				log.Printf("warning issue performing authorize EncryptionsAtRest not done try again: %s \n", err.Error())
+				log.Println("retrying ")
+				time.Sleep(10 * time.Second)
+				encryptionAtRestReq.GroupID = projectID
+				continue
```
andreaangiolillo, line 248: Was this logic already here? Can we use `retry.StateChangeConf` instead? you can file a ticket to improve the logic to use `retry.StateChangeConf`

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -0,0 +1,566 @@
...
+
+func newTFEncryptionAtRestRSModel(ctx context.Context, projectID string, encryptionResp *matlas.EncryptionAtRest, plan *tfEncryptionAtRestRSModel) *tfEncryptionAtRestRSModel {
+	encryptionAtRest := tfEncryptionAtRestRSModel{
+		ID:                   types.StringValue(projectID),
+		ProjectID:            types.StringValue(projectID),
+		AwsKmsConfig:         newTFAwsKmsConfig(ctx, &encryptionResp.AwsKms, plan.AwsKmsConfig),
+		AzureKeyVaultConfig:  newTFAzureKeyVaultConfig(ctx, &encryptionResp.AzureKeyVault, plan.AzureKeyVaultConfig),
+		GoogleCloudKmsConfig: newTFGcpKmsConfig(ctx, &encryptionResp.GoogleCloudKms, plan.GoogleCloudKmsConfig),
+	}
+	return &encryptionAtRest
```
andreaangiolillo:
```suggestion
	return &tfEncryptionAtRestRSModel{
		ID:                   types.StringValue(projectID),
		ProjectID:            types.StringValue(projectID),
		AwsKmsConfig:         newTFAwsKmsConfig(ctx, &encryptionResp.AwsKms, plan.AwsKmsConfig),
		AzureKeyVaultConfig:  newTFAzureKeyVaultConfig(ctx, &encryptionResp.AzureKeyVault, plan.AzureKeyVaultConfig),
		GoogleCloudKmsConfig: newTFGcpKmsConfig(ctx, &encryptionResp.GoogleCloudKms, plan.GoogleCloudKmsConfig),
	} 
```

andreaangiolillo (approved):
LGTM

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
maastha: yes, this has been migrated as is. Created https://jira.mongodb.org/browse/INTMDB-1039

mongodbatlas/framework/validator/aws_kms_config_validator.go
maastha: uodated

PR #1385, opened 2023-08-14
refactor: migrate alert_configuration resource and removing deprecated fields

AgustinBettati:
## Description

Ticket: [INTMDB-996](https://jira.mongodb.org/browse/INTMDB-996)

## Type of change:

## Required Checklist:

## Further comments

- All acceptance tests are running, including the ones that require external credentials:
Screenshot 2023-08-16 at 10 35 33
- I encountered some scenarios of failing non-empty plans in migrations tests, more details here: https://discuss.hashicorp.com/t/terraform-framework-migration-tests-with-non-empty-plan-due-to-optional-attributes/57305. (update) **Solution**:  the optional attributes that could not be defined to null properly were set as computed, given that the api always responds with default values for them.


mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -0,0 +1,811 @@
...
+		return
+	}
+
+	ids := decodeStateID(alertConfigState.ID.ValueString())
+
+	alert, getResp, err := conn.AlertConfigurations.GetAnAlertConfig(context.Background(), ids[encodedIDKeyProjectID], ids[encodedIDKeyAlertID])
+	if err != nil {
+		// deleted in the backend case
+		if getResp != nil && getResp.StatusCode == http.StatusNotFound {
+			alertConfigState.ID = types.StringValue("") // TODO verify this works
```
AgustinBettati: need to verify this works the same why it does in sdk v2 for marking resource as destroyed

AgustinBettati: fixed to correct usage of remove resource method.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
```diff
@@ -60,13 +59,8 @@ func TestAccConfigRSAlertConfiguration_EmptyMetricThresholdConfig(t *testing.T)
...
 				Config: testAccMongoDBAtlasAlertConfigurationConfigEmptyMetricThresholdConfig(orgID, projectName, true),
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckMongoDBAtlasAlertConfigurationExists(resourceName, alert),
-					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
-					resource.TestCheckResourceAttr(resourceName, "notification.#", "1"),
-					resource.TestCheckResourceAttrSet(resourceName, "threshold_config.#"),
-				),
-				ExpectNonEmptyPlan: true,
+				// metric threshold is not created so state is inconsistent with plan
+				ExpectError: regexp.MustCompile("Error: Provider produced inconsistent result after apply"),
```
AgustinBettati, line 63: When an empty metric threshold config block is defined, sdk v2 would provide a non empty plan showing differences after the apply, while framework now directly throws an error display inconsistent state after apply.

mongodbatlas/data_source_mongodbatlas_alert_configuration.go
```diff
@@ -504,3 +507,178 @@ func convertNotificationToCtyValues(notification *matlas.Notification) map[strin
 
 	return values
 }
+
```
AgustinBettati, line 510: these are methods that I moved from old resource implementation as they are used in the data source, will be removed when data source is migrated as well.

mongodbatlas/data_source_mongodbatlas_alert_configuration.go
```diff
@@ -504,3 +507,178 @@ func convertNotificationToCtyValues(notification *matlas.Notification) map[strin
...
+			"metric_name": m.MetricName,
+			"operator":    m.Operator,
+			"threshold":   cast.ToString(m.Threshold),
+			"units":       m.Units,
+			"mode":        m.Mode,
+		}
+	}
+
+	return map[string]interface{}{}
+}
```
andreaangiolillo, line 523: why are you still using `map[string]interface` instead of go struct?

mongodbatlas/data_source_mongodbatlas_alert_configuration.go
```diff
@@ -504,3 +507,178 @@ func convertNotificationToCtyValues(notification *matlas.Notification) map[strin
...
+			"operator":  m.Operator,
+			"units":     m.Units,
+			"threshold": m.Threshold,
+		}}
+	}
+
+	return []interface{}{}
+}
+
+func expandAlertConfigurationNotification(d *schema.ResourceData) ([]matlas.Notification, error) {
```
andreaangiolillo, line 563: instead of using `expand**` let's use go convention `newMongoDBAtlasNotification` (`new[NameOfStructThatIsReturned`)

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -0,0 +1,811 @@
...
+	alertConfigurationID := parts[1]
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), encodeStateID(map[string]string{
+		"id":         alertConfigurationID,
+		"project_id": projectID,
+	}))...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("project_id"), projectID)...)
+}
+
+func expandTFAlertConfigurationNotification(tfNotificationSlice []tfNotificationModel) ([]matlas.Notification, error) {
```
andreaangiolillo: same as above. I would like to move away from expand - flat format and use `new[NameOfResource]` which is widely used in golang

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -0,0 +1,811 @@
...
+				OpsGenieRegion: conversion.StringValueToFramework(value.OpsGenieRegion),
+				TeamID:         conversion.StringValueToFramework(value.TeamID),
+				TypeName:       conversion.StringValueToFramework(value.TypeName),
+				Username:       conversion.StringValueToFramework(value.Username),
+				EmailEnabled:   types.BoolValue(value.EmailEnabled != nil && *value.EmailEnabled),
+				SMSEnabled:     types.BoolValue(value.SMSEnabled != nil && *value.SMSEnabled),
+			}
+		}
+	} else {
+		for i := range matlasSlice {
```
andreaangiolillo:
```suggestion
	if len(matlasSlice) != len(currStateNotifications) { // notifications were modified elsewhere from terraform, or import statement is being called
		for i := range matlasSlice {
			value := matlasSlice[i]
			notifications[i] = tfNotificationModel{
				TeamName:       types.StringValue(value.TeamName),
				Roles:          value.Roles,
				ChannelName:    conversion.StringValueToFramework(value.ChannelName),
				DatadogRegion:  conversion.StringValueToFramework(value.DatadogRegion),
				DelayMin:       types.Int64Value(int64(*value.DelayMin)),
				EmailAddress:   conversion.StringValueToFramework(value.EmailAddress),
				IntervalMin:    types.Int64Value(int64(value.IntervalMin)),
				MobileNumber:   conversion.StringValueToFramework(value.MobileNumber),
				OpsGenieRegion: conversion.StringValueToFramework(value.OpsGenieRegion),
				TeamID:         conversion.StringValueToFramework(value.TeamID),
				TypeName:       conversion.StringValueToFramework(value.TypeName),
				Username:       conversion.StringValueToFramework(value.Username),
				EmailEnabled:   types.BoolValue(value.EmailEnabled != nil && *value.EmailEnabled),
				SMSEnabled:     types.BoolValue(value.SMSEnabled != nil && *value.SMSEnabled),
			}
			return notifications
		}
		for i := range matlasSlice {
```

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -0,0 +1,811 @@
...
+	matchers := make([]tfMatcherModel, len(matlasSlice))
+	if len(matlasSlice) != len(currStateSlice) { // matchers were modified elsewhere from terraform, or import statement is being called
+		for i, value := range matlasSlice {
+			matchers[i] = tfMatcherModel{
+				FieldName: conversion.StringValueToFramework(value.FieldName),
+				Operator:  conversion.StringValueToFramework(value.Operator),
+				Value:     conversion.StringValueToFramework(value.Value),
+			}
+		}
+	} else {
```
andreaangiolillo:
```suggestion
	if len(matlasSlice) != len(currStateSlice) { // matchers were modified elsewhere from terraform, or import statement is being called
		for i, value := range matlasSlice {
			matchers[i] = tfMatcherModel{
				FieldName: conversion.StringValueToFramework(value.FieldName),
				Operator:  conversion.StringValueToFramework(value.Operator),
				Value:     conversion.StringValueToFramework(value.Value),
			}
		}
		return matches
```

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_migration_test.go
```diff
@@ -0,0 +1,92 @@
...
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"mongodbatlas": {
+						VersionConstraint: "1.11.0",
+						Source:            "mongodb/mongodbatlas",
+					},
+				},
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckMongoDBAtlasAlertConfigurationExists(resourceName, alert),
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
```
andreaangiolillo, line 78: can we check for additional fields?

mongodbatlas/framework/conversion/string.go
```diff
@@ -17,3 +17,12 @@ func TypesListToString(ctx context.Context, set types.List) []string {
 	_ = set.ElementsAs(ctx, &results, false)
 	return results
 }
+
+// StringValueToFramework converts a string value to a Framework String value.
+// An empty string is converted to a null String. Useful for optional attributes.
+func StringValueToFramework(v string) types.String {
+	if v == "" {
+		return types.StringNull()
```
andreaangiolillo, line 25: why we cannot use types.StringValue(v) when v is `""`?

maastha: I believe that's because in SDKv2 there was no concept of null values for string, ints etc so they always stored "" (for strings) or 0 (for integers) whether or not the user provided value for that attribute in the config. 
But now the Framework expects the state to be same as config and if the user has not specified an attribute in the config it is expected to be null in the state (sometimes we get "" for some values from the API, so Terraform throws inconsistent values error)
@AgustinBettati correct me if I'm wrong.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -0,0 +1,811 @@
...
+						"operator": schema.StringAttribute{
+							Optional: true,
+						},
+						"threshold": schema.Float64Attribute{
+							Optional: true,
+						},
+						"units": schema.StringAttribute{
+							Optional: true,
+							Validators: []validator.String{
+								stringvalidator.OneOf(
```
maastha, line 231: Does this validator ignoreCase as in validation.StringInSlice()? See previous [implementation](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_alert_configuration.go#L155)
Applicable to all other uses here

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
maastha: +1

mongodbatlas/framework/conversion/string.go
```diff
@@ -17,3 +17,12 @@ func TypesListToString(ctx context.Context, set types.List) []string {
 	_ = set.ElementsAs(ctx, &results, false)
 	return results
 }
+
+// StringValueToFramework converts a string value to a Framework String value.
+// An empty string is converted to a null String. Useful for optional attributes.
+func StringValueToFramework(v string) types.String {
```
maastha:
```suggestion
func StringNullIfEmpty(v string) types.String {
```

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_migration_test.go
```diff
@@ -0,0 +1,92 @@
+package mongodbatlas
```
maastha, line 1: Can we add test cases for each collection in the resource (empty/null if optional, multiple elements wherever possible)
I have had to add logic for null collections, hence we should make sure

mongodbatlas/framework/conversion/string.go
AgustinBettati: Correct, this is relevant primarily for migration tests. If the empty string is defined directly you end up with non-empty plan diffs (related issue: https://github.com/hashicorp/terraform/issues/31887).

mongodbatlas/data_source_mongodbatlas_alert_configuration.go
AgustinBettati: you can ignore additions to the data_source_mongodbatlas_alert_configuration, these are legacy methods I moved from resource file to data source as they were being used, but when migrating the data sources all of these functions will be removed/replaced.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: agree to moving forward with this convention. Just to be clear, for the case of converting from api models to terraform models I see that the toTF* prefix is being used (in project resource and my changes). Are we in agreement to rename these to newTF*  as well?

mongodbatlas/data_source_mongodbatlas_alert_configuration.go
AgustinBettati: more detail in comment below

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: Previous implementation defines `ignoreCase` as false for all StringInSlice calls, meaning it is case sensitive. I have verified that stringvalidator.OneOf() is also caes sensitive, so we should be good here.

mongodbatlas/data_source_mongodbatlas_alert_configuration.go
andreaangiolillo: ah, I see. Thanks for the clarification here

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
andreaangiolillo: my vote is for `newTF*` 👍

mongodbatlas/framework/conversion/string.go
andreaangiolillo: Thanks for the clarifications here

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_migration_test.go
AgustinBettati: Added additional migration tests to verify all scenarios:
- empty/multiple matchers
- empty/defined metric threshold config
- empty/defined threshold config
- multiple notifications

mongodbatlas/framework/conversion/string.go
AgustinBettati: agree, adjusted

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
maastha: I find `toTF*` more readable, but no strong opinion, whatever works for the team.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
maastha: Are we planning to handle this error? Or are we waiting for a response from Hashicorp on this? 
Migration tests are basically to ensure backward compatibility so this would be a breaking change.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
maastha: Great, thank you!

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: adjusted.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
AgustinBettati: This acceptance test defines an an empty block `metric_threshold_config {}` which does not have an initial validation as as all attributes are optional. 
- In 1.11.0 the apply runs successfully, but in the terraform state `metric_threshold_config` is an empty array. That is why when running terraform plan afterwards you receive a non empty plan as terraform mentions that this block needs to be created.
- With the new framework, after running terraform apply the output state is the same, with `metric_threshold_config` as an empty array. However the terraform apply responds with an error because the framework now validates that the state is inconsistent with the defined configuration.

I am not that concerned about this case as it is essentially an incorrectly defined configuration that users would fix when using 1.11.0

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_migration_test.go
AgustinBettati: added additional checks for all migrations tests 👍

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
maastha: Got it, makes sense now, so this essentially is an invalid test case. I thought this was about a null list but I see it is actually defined as empty.
In my case, there were optional lists which even if not defined in the config at all were being saved as empty arrays in the state.
Thanks for the clarification, Agustin!

andreaangiolillo (approved):
LGTM. Thanks for addressing all the comments

maastha (approved):
LGTM

PR #1406, opened 2023-08-23
fix: add instructions for updates of the Atlas SDK

wtrocki:
## Description

Adding support for automatic updates of major releases of the sdk. 
In the long run, we can consider having manually triggered github action that will create an update PR.
 
 ## Command in action
 
 All imports have been updated to the latest after running command.
 
Screenshot 2023-08-23 at 10 40 32

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -134,4 +134,5 @@ link-git-hooks: ## Install git hooks
 .PHONY: update-atlas-sdk
 update-atlas-sdk: ## Update the atlas-sdk dependency
 	go install github.com/icholy/gomajor@v0.9.5
+	## Fetch the latest major version and update imports.
 	gomajor get go.mongodb.org/atlas-sdk/v20230201001@latest
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -133,5 +133,5 @@ link-git-hooks: ## Install git hooks
 
 .PHONY: update-atlas-sdk
 update-atlas-sdk: ## Update the atlas-sdk dependency
-	go install github.com/icholy/gomajor@latest
+	go install github.com/icholy/gomajor@v0.9.5
 	gomajor get go.mongodb.org/atlas-sdk/v20230201001@latest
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -130,3 +130,8 @@ link-git-hooks: ## Install git hooks
 	@echo "==> Installing all git hooks..."
 	find .git/hooks -type l -exec rm {} \;
 	find .githooks -type f -exec ln -sf ../../{} .git/hooks/ \;
+
+.PHONY: update-atlas-sdk
+update-atlas-sdk: ## Update the atlas-sdk dependency
+	go install github.com/icholy/gomajor@v0.9.5
+	gomajor get go.mongodb.org/atlas-sdk/v20230201001@latest
```
AgustinBettati, line 138: question: would the version `v20230201001` have to be modified manually or will it always fetch the latest major version?

andreaangiolillo: My understanding is that it should change for every new release.
If I am correct, could we update the command to use an env variable for instance `ATLAS_SDK_VERSION=v20230201001` and update  CONTRIBUTING.md to include an explanation that this var must contain the version of the SDK that we want to add to TF

wtrocki: > would the version v20230201001 have to be modified manually 

No. 

>  or will it always fetch the latest major version?

yes (including latest minor and patch)

TL;DR:

Someone back in 2009 make call that:

`go get go.mongodb.org/atlas-sdk` always refers to 0.x.x and 1.x.x versions.

any major release above 1  needs:

`go get go.mongodb.org/atlas-sdk/someversion` - the version here is irrelevant but needs to be present to fetch the latest.

GNUmakefile
```diff
@@ -130,3 +130,8 @@ link-git-hooks: ## Install git hooks
 	@echo "==> Installing all git hooks..."
 	find .git/hooks -type l -exec rm {} \;
 	find .githooks -type f -exec ln -sf ../../{} .git/hooks/ \;
+
+.PHONY: update-atlas-sdk
+update-atlas-sdk: ## Update the atlas-sdk dependency
+	go install github.com/icholy/gomajor@v0.9.5
```
wtrocki, line 136:
```suggestion
	go install github.com/icholy/gomajor@v0.9.5
	## Fetch the latest major version and update imports.
```

andreaangiolillo (approved):
LGTM

AgustinBettati (approved):
LGTM, thank you!

PR #1412, opened 2023-08-24
fix: avoid diff in state after import for undefined optional attribute in alert config notification

AgustinBettati:
## Description

**Related ticket**: case found when testing [INTMDB-981](https://jira.mongodb.org/browse/INTMDB-981)

**Summary**: adjust optional attribute `interval_min` to handle default values the same way as `delay_min` to avoid non empty plan after using import operation.

For the following valid resource that does not define `interval_min` attribute:
```
resource "mongodbatlas_alert_configuration" "test" {
  project_id = "64c0f3f5ce752426ab9f506b"
  event_type = "NO_PRIMARY"
  enabled    = true

  notification {
    type_name    = "PAGER_DUTY"
    delay_min    = 4
    service_key = "<service key>"
  }
}
```

When import statement and HCL code is generated using the following data source for the created resource above:
```
data "mongodbatlas_alert_configuration" "test" {
  project_id             = "64c0f3f5ce752426ab9f506b"
  alert_configuration_id = "aWQ=:NjRlNzBmNDJjM2MyOTM2OGFiNWY4Mjk5-cHJvamVjdF9pZA==:NjRjMGYzZjVjZTc1MjQyNmFiOWY1MDZi"

  output {
    type = "resource_import"
  }
  output {
    type = "resource_hcl"
  }
}
```

[**Previous**] Terraform plan output on the generated resource code results in a non empty plan:
Screenshot 2023-08-24 at 17 48 23

[**New**] Terraform plan output:
Screenshot 2023-08-24 at 17 48 36

## Type of change:

## Required Checklist:

## Further comments
 Test that was defined needs credentials, so ran all resource and data source tests:
Screenshot 2023-08-24 at 17 31 55
Screenshot 2023-08-24 at 17 32 45
Screenshot 2023-08-24 at 17 33 11

### DIFF
--- a/mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
+++ b/mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
@@ -291,6 +291,10 @@ func (r *AlertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						},
 						"interval_min": schema.Int64Attribute{
 							Optional: true,
+							Computed: true,
+							PlanModifiers: []planmodifier.Int64{
+								int64planmodifier.UseStateForUnknown(),
+							},
 						},
 						"mobile_number": schema.StringAttribute{
 							Optional: true,
@@ -709,9 +713,6 @@ func newTFNotificationModelList(matlasSlice []matlas.Notification, currStateNoti
 		if !currState.EmailAddress.IsNull() {
 			newState.EmailAddress = conversion.StringNullIfEmpty(value.EmailAddress)
 		}
-		if !currState.IntervalMin.IsNull() {
-			newState.IntervalMin = types.Int64Value(int64(value.IntervalMin))
-		}
 		if !currState.MobileNumber.IsNull() {
 			newState.MobileNumber = conversion.StringNullIfEmpty(value.MobileNumber)
 		}
@@ -728,6 +729,7 @@ func newTFNotificationModelList(matlasSlice []matlas.Notification, currStateNoti
 			newState.Username = conversion.StringNullIfEmpty(value.Username)
 		}
 
+		newState.IntervalMin = types.Int64Value(int64(value.IntervalMin))
 		newState.DelayMin = types.Int64Value(int64(*value.DelayMin))
 		newState.EmailEnabled = types.BoolValue(value.EmailEnabled != nil && *value.EmailEnabled)
 		newState.SMSEnabled = types.BoolValue(value.SMSEnabled != nil && *value.SMSEnabled)
--- a/mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
+++ b/mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
@@ -317,6 +317,39 @@ func TestAccConfigRSAlertConfiguration_importConfigNotifications(t *testing.T) {
 	})
 }
 
+// used for testing notification that does not define interval_min attribute
+func TestAccConfigRSAlertConfiguration_importPagerDuty(t *testing.T) {
+	SkipTestExtCred(t) // Will skip because requires external credentials aka api key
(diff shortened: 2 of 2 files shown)


mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -291,6 +291,10 @@ func (r *AlertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						},
 						"interval_min": schema.Int64Attribute{
 							Optional: true,
+							Computed: true,
+							PlanModifiers: []planmodifier.Int64{
+								int64planmodifier.UseStateForUnknown(),
+							},
```
andreaangiolillo, line 297: > [UseStateForUnknown()](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown): Copies the prior state value, if not null. This is useful for reducing (known after apply) plan outputs for computed attributes which are known to not change over time.

Why is this modifier needed?

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -709,9 +713,6 @@ func newTFNotificationModelList(matlasSlice []matlas.Notification, currStateNoti
 		if !currState.EmailAddress.IsNull() {
 			newState.EmailAddress = conversion.StringNullIfEmpty(value.EmailAddress)
 		}
-		if !currState.IntervalMin.IsNull() {
-			newState.IntervalMin = types.Int64Value(int64(value.IntervalMin))
-		}
```
andreaangiolillo, line 714: I was reading this [plan-modification#plan-modification-process](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#plan-modification-process)
> Plan Modification Process
> When the provider receives a request to generate the plan for a resource change via the framework, the following occurs:
> 
> Set any attributes with a null configuration value to the [default value](https://developer.hashicorp.com/terraform/plugin/framework/resources/default).
> If the plan differs from the current resource state, the framework marks computed attributes that are null in the configuration as unknown in the plan. This is intended to prevent unexpected Terraform errors. Providers can later enter any values that may be known.
> Run attribute plan modifiers.
> Run resource plan modifiers.

I am wondering if `currState.IntervalMin.isNill` is `false`  but `currState.IntervalMin.IsUnknown()` is `true` when calling `terraform plan`. Could you verify? if that is the case we should add `currState.IntervalMin.IsUnknown()` in all of this ifs

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: Essentially to simplify the terraform plan output. The `interval_min` attribute will only change its value when modified by the user.
As a counter example, alert configuration has an `updated` attribute that is computed and is not set with this modifier. In this case is seems appropriate to have the message `(known after apply)` shown on the terraform plan because on every apply it will be modified.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: Yes your statement is correct, I could verify that the attribute in `currState` comes as unknown as it is an undefined computed attribute.

In this case, since it is a value that is always defined in the api request/response (defined as an int in manual sdk), we can simply define the value in the model. This is what solves the inconsistency issue when doing the import operation, because when importing you have no state/config to verify if field was defined by the user, so value coming from the api is set directly.

andreaangiolillo (approved):
LGTM. Thanks for the discussion

AgustinBettati:
@maastha small change here just in case you want to take a look

maastha (approved):
LGTM

PR #1369, opened 2023-08-04
fix: How do I create Organization API Key with Organization Billing Admin permission and Project Read Only for projects

andreaangiolillo:
## Description

JIRA ticket: [INTMDB-898](https://jira.mongodb.org/browse/INTMDB-898)

Link to any related issue(s): https://github.com/mongodb/terraform-provider-mongodbatlas/issues/1278

## Type of change:

- [x] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code

## Further comments

I locally tested the provider with the following configuration files and checked that the resource was successfully created, the `terraform plan` did not return any changes and the `terraform destroy` worked as expected.

```terraform
resource "mongodbatlas_project" "atlas-project" {
  name   = "Issue-1278"
  org_id = "60ddf55c27a5a20955a707d7"
}

resource "mongodbatlas_project_api_key" "api_1" {
  description = "test api_key multi"
  project_id  = mongodbatlas_project.atlas-project.id

  project_assignment {
    project_id = mongodbatlas_project.atlas-project.id
    role_names = ["ORG_BILLING_ADMIN", "GROUP_READ_ONLY"]
  }

  project_assignment {
    project_id = "63dcfc256af00a5934e60924"
    role_names = ["GROUP_READ_ONLY"]
  }

  project_assignment {
    project_id = "64c23af6f133166c39176cbf"
    role_names = ["GROUP_OWNER"]
  }
}
```

```terraform
resource "mongodbatlas_project" "atlas-project" {
  name   = "Issue-1278"
  org_id = "60ddf55c27a5a20955a707d7"
}

resource "mongodbatlas_project_api_key" "api_1" {
  description = "test api_key multi"
[...]

PR #1416, opened 2023-08-25
revert: "fix: How do I create Organization API Key with Organization Billing Admin permission and Project Read Only for projects"

andreaangiolillo:
Reverts mongodb/terraform-provider-mongodbatlas#1369

Team Triage Decision: reverting #1369 and wait for the fix in atlas before providing a fix in terraform


marcosuma (approved):
LGTM

Issue #1396, opened 2023-08-17
Error: The terraform-provider-mongodbatlas_v1.11.0 plugin crashed


Kikivsantos:
HI, there.

I'm trying to create an alert on MongoDB Atlas with terraform, but I´m getting the error below:

```

mongodbatlas_alert_configuration.default: Creating...
╷
│ Error: Plugin did not respond
│ 
│   with mongodbatlas_alert_configuration.default,
│   on main.tf line 2, in resource "mongodbatlas_alert_configuration" "default":
│    2: resource "mongodbatlas_alert_configuration" "default" {
│
│ The plugin encountered an error, and failed to respond to the
│ plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may
│ contain more details.
╵

Stack trace from the terraform-provider-mongodbatlas_v1.11.0 plugin:

panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 10 [running]:
github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas.expandAlertConfigurationThresholdConfig(0x4143e5?)
        github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/resource_mongodbatlas_alert_configuration.go:613 +0x314
github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas.resourceMongoDBAtlasAlertConfigurationCreate({0x1758948, 0xc00033e1e0}, 0x0?, {0x121f160?, 0xc000927830?})
        github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/resource_mongodbatlas_alert_configuration.go:307 +0x185
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0xc0006fc8c0, {0x1758980, 0xc00058ef30}, 0xd?, {0x121f160, 0xc000927830})    
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.27.0/helper/schema/resource.go:733 +0x12e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0xc0006fc8c0, {0x1758980, 0xc00058ef30}, 0xc000a80820, 0xc00077eb00, {0x121f160, 0xc000927830})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.27.0/helper/schema/resource.go:864 +0xa7e
[...]

PR #1419, opened 2023-08-28
fix: HELP: Error: The terraform-provider-mongodbatlas_v1.11.0 plugin crashed

andreaangiolillo:
## Description

This PR fixes an NPE when the matcher & metrics threshold are provided with null fields

Link to any related issue(s): #1396

## Type of change:

## Required Checklist:

## Further comments

Next Steps:

- Reproduce the issue with the resource migrated to the new framework
- Backport the fix in 1.12.0 if needed

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_alert_configuration.go
+++ b/mongodbatlas/resource_mongodbatlas_alert_configuration.go
@@ -9,11 +9,10 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"github.com/mwielbut/pointy"
 	"github.com/spf13/cast"
-
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	matlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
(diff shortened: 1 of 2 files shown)


mongodbatlas/resource_mongodbatlas_alert_configuration.go
```diff
@@ -515,8 +514,11 @@ func expandAlertConfigurationMatchers(d *schema.ResourceData) []matlas.Matcher {
 
 	if m, ok := d.GetOk("matcher"); ok {
 		for _, value := range m.([]interface{}) {
-			v := value.(map[string]interface{})
+			if value == nil {
```
marcosuma, line 517: could we do something similar to 

```
			if ok, v := value.(map[string]interface{}); !ok {
				break
			}
```

mongodbatlas/resource_mongodbatlas_alert_configuration.go
```diff
@@ -610,6 +612,10 @@ func expandAlertConfigurationThresholdConfig(d *schema.ResourceData) *matlas.Thr
 		vL := value.([]interface{})
 
 		if len(vL) > 0 {
+			if vL[0] == nil {
+				return nil
+			}
+
 			v := vL[0].(map[string]interface{})
```
marcosuma, line 619: same as here. I have not fully tried myself, but was wondering if something like `ok, v := value.(map[string]interface{}); !ok` would catch the impossibility to cast due to nil.

I also don't fully understand this one: in which cases would there be an array of length > 0 whose first value is nil? And why do we assume that all other values in the array will also be nil?

gssbzn (approved):
LGTM

AgustinBettati (approved):
LGTM, agree with the next steps 👍

mongodbatlas/resource_mongodbatlas_alert_configuration.go
andreaangiolillo: I think I tried to do it but the `ok` was still true but I am not 100% sure now 🤦 . I will add an update in a few minutes to see if this would work. Thanks for flagging this

andreaangiolillo: Yes, We can use the `ok,v `pattern here. I opened https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1420 to address this comment. Thanks

PR #1424, opened 2023-08-29
refactor: extract Configure and Metadata framework functions into single implementation

AgustinBettati:
## Description

Small refactor that removes boilerplate definition of `Metadata` and `Configure` functions in all our plugin-framework resources and data sources into common structs.

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/fw_common.go
+++ b/mongodbatlas/fw_common.go
@@ -1,0 +1,68 @@
+package mongodbatlas
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+// common struct for all framework resources. Implements the following plugin-framework defined functions:
+// - Metadata
+// - Configure
+// client is left empty and populated by the framework when envoking Configure method.
+// resourceName must be defined when creating an instance of a resource.
+type RSCommon struct {
+	client       *MongoDBClient
+	resourceName string
+}
+
+func (r *RSCommon) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = fmt.Sprintf("%s_%s", req.ProviderTypeName, r.resourceName)
+}
+
+func (r *RSCommon) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	client, err := ConfigureClient(req.ProviderData)
+	if err != nil {
+		resp.Diagnostics.AddError(errorConfigureSummary, err.Error())
+		return
+	}
+	r.client = client
+}
+
+// common struct for all framework data sources. Implements the following plugin-framework defined functions:
+// - Metadata
+// - Configure
+// client is left empty and populated by the framework when envoking Configure method.
+// dataSourceName must be defined when creating an instance of a data source.
+type DSCommon struct {
+	client         *MongoDBClient
+	dataSourceName string
+}
+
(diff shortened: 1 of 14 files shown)


mongodbatlas/fw_common.go
```diff
@@ -0,0 +1,68 @@
...
+func (d *DSCommon) Configure(ctx context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	client, err := ConfigureClient(req.ProviderData)
+	if err != nil {
+		resp.Diagnostics.AddError(errorConfigureSummary, err.Error())
+		return
+	}
+	d.client = client
+}
+
+func ConfigureClient(providerData any) (*MongoDBClient, error) {
```
marcosuma: Not sure if I am wrong here, but would you consider making this private if it's only being called in this package?

mongodbatlas/fw_common.go
```diff
@@ -0,0 +1,68 @@
...
+	if providerData == nil {
+		return nil, nil
+	}
+
+	client, ok := providerData.(*MongoDBClient)
+	if !ok {
+		return nil, fmt.Errorf(errorConfigure, providerData)
+	}
+
+	return client, nil
```
marcosuma: nit:
```
if client, ok := providerData.(*MongoDBClient); ok {
    return client, nil
}

return nil, fmt.Errorf(...)
```

mongodbatlas/fw_common.go
marcosuma: any chance we could avoid using `any`?

mongodbatlas/fw_common.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+// common struct for all framework resources. Implements the following plugin-framework defined functions:
+// - Metadata
+// - Configure
+// client is left empty and populated by the framework when envoking Configure method.
+// resourceName must be defined when creating an instance of a resource.
```
andreaangiolillo: comments should start with the name of the Struct or function

`// RSCommon is used ...`

mongodbatlas/fw_common.go
andreaangiolillo: +1, ConfigureClient should be called only here

mongodbatlas/fw_common.go
```diff
@@ -0,0 +1,68 @@
+package mongodbatlas
```
andreaangiolillo, line 1: Does it make sense to move this struct in a subdir to make it clear that does not define a resource or ds?

andreaangiolillo (approved):
LGTM. My comments are not blocking the PR

mongodbatlas/fw_common.go
AgustinBettati: for sure

mongodbatlas/fw_common.go
AgustinBettati: will adjust

mongodbatlas/fw_common.go
AgustinBettati: Not in this case. The ConfigureRequest struct defined by terraform plugin framework defines the `ProviderData` field as `any` becuase the data is provider-specifc.

mongodbatlas/fw_common.go
AgustinBettati: Had looked into this but reached a limitation. When defining `fw_common.go` within framework package, I need to import mongodbatlas package to reference `MongoDBClient`. This leads to an `import cycle not allowed` error.

maastha (approved):
LGTM, thank you for this one!

PR #1433, opened 2023-08-30
chore: draft breaking changes notice

wtrocki:
## Description

Draft breaking changes notice for the team to consider.

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,0 +1,29 @@
+## Breaking Changes in Terraform provider
+
+## Version Compatibilities
+
+Our releases follow the Semantic Versioning (SemVer) convention to ensure predictability and stability. Each version is represented as MAJOR.MINOR.PATCH, where:
+
+- MAJOR version updates may contain breaking changes.
+- MINOR version updates introduce backward-compatible functionality and can contain breaking changes 
+- PATCH version updates encompass backward-compatible bug fixes.
+
+## Release Support Period
+
+Every release of the Atlas Terraform Provider is officially supported for approximately one year from its initial release date.
+Each version of the provider is guaranteed to function properly for that time period.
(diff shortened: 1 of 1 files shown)


BREAKING_CHANGES.md
```diff
@@ -0,0 +1,25 @@
...
+
+Our releases follow the Semantic Versioning (SemVer) convention to ensure predictability and stability. Each version is represented as MAJOR.MINOR.PATCH, where:
+
+- MAJOR version updates may contain breaking changes.
+- MINOR version updates introduce backward-compatible functionality and can contain breaking changes 
+- PATCH version updates encompass backward-compatible bug fixes.
+
+## Release Support Period
+
+Every release of the Atlas Terraform Provider is officially supported for approximately one year from its initial release date.
```
colm-quinn, line 13: I don't think this is true - we support the latest release of the provider. For issues, we encourage upgrading, as we don't release updates for older versions.

In line with how we support the AtlasCLI & other tools where we don't use a magor release number

BREAKING_CHANGES.md
```diff
@@ -0,0 +1,25 @@
+## Breaking Changes in Terraform provider
+
+## Version Compatibilities
+
+Our releases follow the Semantic Versioning (SemVer) convention to ensure predictability and stability. Each version is represented as MAJOR.MINOR.PATCH, where:
+
+- MAJOR version updates may contain breaking changes.
+- MINOR version updates introduce backward-compatible functionality and can contain breaking changes 
+- PATCH version updates encompass backward-compatible bug fixes.
```
colm-quinn, line 9: I think for both versioning & changelog, it would be more accurate to refer to the Terraform guidance/policy, which is what we aim to follow: https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#versioning-specification

PR #1435, opened 2023-08-30
refactor:  moving acceptance test to acceptance-tests folder

andreaangiolillo:
## Description
Jira Ticket: [INTMDB-798](https://jira.mongodb.org/browse/INTMDB-798)

This PR moves all the acceptance tests under the acceptance test folders. The main reason is that after https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1436, the`_test.go` files under the dir `mongodbatlas` will contain **only unit tests**.

This PR is a pre-requisite for https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1436 where I am adding the unit tests for the database user resource

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code

## Further comments

- **cloud-dev is currently broken. I merge the PR only after all the acceptance tests are successful**
- **This PR will be merged into master once 1.11.1 is released**

PR #1436, opened 2023-08-30
poc: Add unit tests for db user resource

andreaangiolillo:
## Description

Jira Ticket: [INTMDB-798](https://jira.mongodb.org/browse/INTMDB-798)

This PR adds unit tests for the Database User resource.
This PR must be merged after https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1435.

```bash
make test                                                                                                                          1.20.3  3.1.1  16.18.0
==> Checking that code complies with gofmt requirements...
go test $(go list ./... | grep -v /integrationtesting) -timeout=30s -parallel=4
?   	github.com/mongodb/terraform-provider-mongodbatlas	[no test files]
?   	github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/conversion	[no test files]
?   	github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/testutils	[no test files]
?   	github.com/mongodb/terraform-provider-mongodbatlas/version	[no test files]
ok  	github.com/mongodb/terraform-provider-mongodbatlas/acceptance-tests	0.825s
ok  	github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas	0.123s
ok  	github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/validator	0.451s
```

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
+++ b/mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
@@ -1,0 +1,281 @@
+package mongodbatlas
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	fwresource "github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	matlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+func TestMongoDBAtlasDatabaseUserResource_Schema(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	schemaRequest := fwresource.SchemaRequest{}
+	schemaResponse := &fwresource.SchemaResponse{}
+
+	// Instantiate the resource.Resource and call its Schema method
+	NewDatabaseUserRS().Schema(ctx, schemaRequest, schemaResponse)
+
+	if schemaResponse.Diagnostics.HasError() {
+		t.Fatalf("Schema method diagnostics: %+v", schemaResponse.Diagnostics)
+	}
+
+	// Validate the schema
+	diagnostics := schemaResponse.Schema.ValidateImplementation(ctx)
+
+	if diagnostics.HasError() {
+		t.Fatalf("Schema validation diagnostics: %+v", diagnostics)
+	}
+}
+
+func TestMongoDBAtlasDatabaseUserResourceSchema_Read(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	resource, schemaResponse, err := newDatabaseUserRSMock(ctx)
+	if err != nil {
+		t.Fatalf("error during the resource creation: %v", err)
+	}
+
+	readRequest, readResponse := newReadRequestResponse(schemaResponse)
+	requestModel := newRequestModel(ctx)
+	readRequest.State.Set(ctx, requestModel)
+
+	resource.Read(ctx, *readRequest, readResponse)
+
+	var responseModel *tfDatabaseUserModel
+	diag := readResponse.State.Get(ctx, &responseModel)
+	if diag.HasError() {
+		t.Fatalf("The READ operation failed: %v", diag)
+	}
+
+	validateResponse(t, requestModel, responseModel)
+}
+
+func TestMongoDBAtlasDatabaseUserResourceSchema_Create(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	resource, schemaResponse, err := newDatabaseUserRSMock(ctx)
+	if err != nil {
+		t.Fatalf("error during the resource creation: %v", err)
+	}
+
+	createRequest, createResponse := newCreateRequestResponse(schemaResponse)
+	requestModel := newRequestModel(ctx)
+	createRequest.Plan.Set(ctx, requestModel)
+
+	resource.Create(ctx, *createRequest, createResponse)
+
+	var responseModel *tfDatabaseUserModel
+	diag := createResponse.State.Get(ctx, &responseModel)
+	if diag.HasError() {
+		t.Fatalf("The CREATE operation failed: %v", diag)
+	}
+
+	validateResponse(t, requestModel, responseModel)
+}
+
+func TestMongoDBAtlasDatabaseUserResourceSchema_Update(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	resource, schemaResponse, err := newDatabaseUserRSMock(ctx)
+	if err != nil {
+		t.Fatalf("error during the resource creation: %v", err)
+	}
+
+	updateRequest, updateResponse := newUpdateRequestResponse(schemaResponse)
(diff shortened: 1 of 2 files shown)


mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
```diff
@@ -0,0 +1,281 @@
...
+		t.Fatalf("Schema method diagnostics: %+v", schemaResponse.Diagnostics)
+	}
+
+	// Validate the schema
+	diagnostics := schemaResponse.Schema.ValidateImplementation(ctx)
+
+	if diagnostics.HasError() {
+		t.Fatalf("Schema validation diagnostics: %+v", diagnostics)
+	}
+}
```
andreaangiolillo, line 34: `TestMongoDBAtlasDatabaseUserResource_Schema` validates the resource schema as explained in the [Hashicorp doc](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/schemas#unit-testing)

mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
```diff
@@ -0,0 +1,281 @@
...
+	return &matlas.DatabaseUser{
+		GroupID:      groupID,
+		Username:     updateRequest.Username,
+		DatabaseName: updateRequest.DatabaseName,
+	}, nil, nil
+}
+
+func (s *MockDatabaseUsersServiceOp) Delete(ctx context.Context, databaseName, groupID, username string) (*matlas.Response, error) {
+	return nil, nil
+}
```
andreaangiolillo, line 281: Mocking the SDK

mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	if requestModel.AuthDatabaseName != responseModel.AuthDatabaseName {
+		t.Fatalf("expected %s, got %s", requestModel.AuthDatabaseName, responseModel.AuthDatabaseName)
+	}
+
+	if requestModel.ProjectID != responseModel.ProjectID {
+		t.Fatalf("expected %s, got %s", requestModel.AuthDatabaseName, responseModel.AuthDatabaseName)
+	}
+}
```
andreaangiolillo, line 147: Validating that the response model has the params in the request Model

mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+func newDatabaseUserRSMock(ctx context.Context) (fwresource.Resource, *fwresource.SchemaResponse, error) {
```
andreaangiolillo, line 149: Mocking the SDK

mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
```diff
@@ -0,0 +1,281 @@
...
+	schemaRequest := fwresource.SchemaRequest{}
+	schemaResponse := &fwresource.SchemaResponse{}
+
+	// Instantiate the resource.Resource and call its Schema method
+	resource := NewDatabaseUserRSWithClient(&MongoDBClient{
+		Atlas: &matlas.Client{
+			DatabaseUsers: &MockDatabaseUsersServiceOp{},
+		},
+	})
```
andreaangiolillo, line 158: Adding the Mock SDK to the resource

wtrocki (approved):

marcosuma:
Exciting to see this! I guess it's a broader conversation with the whole team but since we're starting unit tests it'd be great if we agreed on a naming convention. I really like the "Google one": `UnitOfWork_StateUnderTest_ExpectedBehavior` e.g. instead of `TestMongoDBAtlasDatabaseUserResourceSchema_Create` it would be `MongoDBAtlasDatabaseUserResourceSchema_Create_Succeeds` assuming `MongoDBAtlasDatabaseUserResourceSchema` is the unit of work.

I'd also say that MongoDBAtlas is quite verbose and we could do something like MDBAtlas

Thoughts @andreaangiolillo @maastha @AgustinBettati @colm-quinn?

I wonder if we do something on a broader level? @gssbzn

AgustinBettati:
From a maintenance point of view I do see that each unit test does require several manual steps to set up the requests and validations of the response.

One alternative which potentially could reduce some of this boilerplate is to mock request/responses at an http level, which would allow us to continue using the known `terraform-plugin-testing` and possibly reuse existing configs/checks. Simple [example](https://github.com/hashicorp/terraform-provider-http/blob/main/internal/provider/data_source_http_test.go#L26) mentioned in [hashicorp thread](https://discuss.hashicorp.com/t/mock-api-calls-in-plugin-tests/57322/2?u=agustinbettati). 

Might be worth contrasting both options before moving forward towards other resources.

**Edit**: Looking at this alternative: https://github.com/hashicorp/terraform-provider-google/blob/bee71217ef0c9a29b85ed69fefbce92d4af85405/google/acctest/vcr_utils.go#L392-L422, I see we could define a testing provider which can be used in regular acceptance tests, but this testing provider would use the mocked sdk client as handled in this PR.

gssbzn:
> From a maintenance point of view I do see that each unit test does require several manual steps to set up the requests and validations of the response.

This, and I agree this could use with some research on mocking generation as there's a lot of boiler plate code here

On naming of tests I like golang approach which is similar to what MArco describes but the expected result is a nested test as it favors matrix testing

```go
func TestMethod(t *testing.T){
  tc := []testcase{ }
  for _, case := range tc {

    t.Run(tc.name, ...)
  }
}
```

andreaangiolillo:
> One alternative which potentially could reduce some of this boilerplate is to mock request/responses at an http level, which would allow us to continue using the known terraform-plugin-testing and possibly reuse existing configs/checks. Simple [example](https://github.com/hashicorp/terraform-provider-http/blob/main/internal/provider/data_source_http_test.go#L26) mentioned in [hashicorp thread](https://discuss.hashicorp.com/t/mock-api-calls-in-plugin-tests/57322/2?u=agustinbettati).

@AgustinBettati Looking at this example, the data source takes a URL as input and performs the HTTP requests inside the READ operation. This is not what we are currently doing in our provider where this logic is hidden for us: the SDK takes care of performing the HTTP requests for us. To make this work, we still have to inject the client to use for the acceptance test and mock the API calls 🤔 

Looking at the Google provider example.

AgustinBettati:
Yes I agree that the first example is not relevant for our case, we could redefine the base url and have a mock server but that seems like more work.

> Looking at the Google provider example.

The google example seems more applicable as it defines a new provider that overrides the configure method. We can define a new client that uses the mocked sdk, and then use this test provider directly in acceptance tests. I believe we would be limited to using this mechanism only for rs/ds that are using the plugin framework.

mongodbatlas/fw_resource_mongodbatlas_database_user_test.go
andreaangiolillo, line 281: mock

PR #1437, opened 2023-08-31
fix: add debug support for sdk

wtrocki:
## Description

Enable debug support for the terraform provider and the sdk-go.
Just an idea for the team to consider. PR  is missing docs updates.

The debug flag is part of the Atlas CLI/SDK support plan. Having terraform support for the same would make things much easier to handle. 

## Relationship to TF_Logs

TF_Logs are very detailed terraform logs (large volume). For most of the cases when the issue happens in the API we are mostly interested in the Request/Response pair to be able to reproduce the issue locally.

### DIFF
--- a/mongodbatlas/config.go
+++ b/mongodbatlas/config.go
@@ -28,6 +28,7 @@ type Config struct {
 	PrivateKey   string
 	BaseURL      string
 	RealmBaseURL string
+	Debug        bool
 }
 
 // MongoDBClient contains the mongodbatlas clients and configurations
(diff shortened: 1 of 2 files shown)


wtrocki:
@marcosuma Using SDK debug mode will print all the relevant logs and help with triaging http errors.
Debug mode prints requests and response and all trace information to the logs. 

Investigaton is needed if terraform would be able to consume those logs (Atlas CLI have them out of the box)

I see that Debug is currently hardcoded: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/df074dee671ca863bbd79a8e5df7dac05257bb5a/internal/config/client.go#L120

marcosuma:
thank you @wtrocki , will bring it up to the team

PR #1448, opened 2023-09-05
PoC: define new TestCheckFunc to validate resulting terraform state using golden files

AgustinBettati:
## Description

This is a draft PR that showcases the possibility of using golden files to verify resulting Terraform state in our acceptance tests. 

**Main benefit**: avoids the need to manually verify each attribute using `TestCheckResourceAttr` and `TestCheckResourceAttrSet`. Instead, all of the state is serialized into a json file, and each time a test is executed, the current state is compared against the expected state.

### To add an acceptance test with this check

1. In the checks section of your new acceptance test, use testutils.CheckStateMatchesGoldenFile and reference a json file (which can initially be defined as empty). 
2. Run the test, which will fail and show a message that contains a `cp` command to update the content of the file. Once you have verified that the values present in the json are correct, commit the json file together with your test.

### What happens if a test fails?

A json diff is shown in the error message to quickly identify what changes have been detected in the terraform state.
<details>
<summary>Diff message example</summary>

```
Step 1/1 error: Check failed: Check 1/1 error: serialized state doesn't match the golden file.
        Differences:   map[string]any{
          	"depends_on": nil,
          	"deposed":    nil,
          	"primary": map[string]any{
          		... // 2 identical entries
          		"RawPlan":  map[string]any{},
          		"RawState": map[string]any{},
          		"attributes": map[string]any{
          			... // 2 ignored and 8 identical entries
          			"is_data_explorer_enabled":              string("true"),
          			"is_extended_storage_sizes_enabled":     string("false"),
        + 			"is_performance_advisor_enabled":        string("true"),
          			"is_realtime_performance_panel_enabled": string("true"),
          			"is_schema_advisor_enabled":             string("true"),
          			... // 2 ignored and 51 identical entries
          		},
          		... // 1 ignored and 2 identical entries
          	},
          	"provider": string("registry.terraform.io/hashicorp/mongodbatlas"),
          	"type":     string("mongodbatlas_project"),
          }
        .
        To update the golden file, run:
        cp /var/folders/d6/zqxkvmcs15gb5xtn30rnxz780000gp/T/current-state.json ./mongodbatlas/../test/data/state-golden-files/project-ds/project-ds-limits.json
```
</details>

### Additional considerations:
- For values that change on each execution, a mechanism needs to be put in place to ignore certain attributes. This was partially implemented in the PoC.

## Type of change:

## Required Checklist:


test/data/state-golden-files/project-ds/project-ds-limits.json
```diff
@@ -0,0 +1,87 @@
+{
```
colm-quinn, line 1: Doesn't apply here, but is it possible for secrets to leak as a result here? For instance, how will the golden file handle exporting terraform state for sensitive fields?

mongodbatlas/data_source_mongodbatlas_project_test.go
```diff
@@ -174,9 +175,10 @@ func TestAccProjectDSProject_limits(t *testing.T) {
 			{
 				Config: testAccMongoDBAtlasProjectDSUsingRS(testAccMongoDBAtlasProjectConfigWithLimits(projectName, orgID, []*projectLimit{})),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttrSet("data.mongodbatlas_project.test", "name"),
-					resource.TestCheckResourceAttrSet("data.mongodbatlas_project.test", "org_id"),
-					resource.TestCheckResourceAttrSet("data.mongodbatlas_project.test", "limits.0.name"),
+					testutils.CheckStateMatchesGoldenFile(
+						"data.mongodbatlas_project.test",
+						"../test/data/state-golden-files/project-ds/project-ds-limits.json",
+						[]string{".primary.attributes.org_id", ".primary.attributes.created", ".primary.attributes.name"}),
```
colm-quinn, line 181: Q: Why checking `.primary.attributes.created` now - time/dates tend to be a brittle test.

Also are you dropping the tests on `"limits.0.name"` ?

AgustinBettati: This is actually just the opposite, the `CheckStateMatchesGoldenFile` allows the possibility of ignoring certain fields in the json. This relevant for values that change on each execution, like the `created` attribute that will always be different in value. All other fields will be used for comparison.

test/data/state-golden-files/project-ds/project-ds-limits.json
AgustinBettati: This a good point. Within the context of acceptance tests, I see sensitive values are handled 2 ways:

- Dummy values are defined (example: [data_source_mongodbatlas_database_user_test.go#L52](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/data_source_mongodbatlas_database_user_test.go#L52)) which would not be an issue
- Make use of env variables (example: [resource_mongodbatlas_alert_configuration_test.go#L333](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/resource_mongodbatlas_alert_configuration_test.go#L333)). This case is more concerning because the person creating the test could serialize in the json the value used for testing locally. Will look into this to see if they can be filtered automatically.

marcosuma:
_Sorry, I am gonna leave a message here because I don't think this is a comment related to any piece of the code_

Love the proposal @AgustinBettati! this reminds me screenshot tests that can be defined for UIs (same exact principle). 
I think this proposal goes well with the overall attempt of improving our test infrastructure. 

This looks like a great addition to reduce the volume of dependency to cloud-dev by maintaining a good coverage. We need to work together with @andreaangiolillo on how to define a plan and include these type of tests too.

One challenge is: how do you know at first (the very first time you create the test) that the output is good? I can see how that is subject to human error since they have to review it.

AgustinBettati:
> One challenge is: how do you know at first (the very first time you create the test) that the output is good? I can see how that is subject to human error since they have to review it.

When you generate the json state file for the first time you will check that the relevant attributes are present and contain the correct values. This needed manual check is somewhat similar, but much faster and exhaustive, to how we currently have to write checks to verify certain attributes are present or have a certain value.

PR #1474, opened 2023-09-15
fix: improved error handling for cloud_backup_schedule resource.

marcosuma:
## Description

Link to any related issue(s): https://jira.mongodb.org/browse/HELP-50083

Originally we discovered this issue because client was creating a cloud_backup_schedule concurrently with the creation of the cluster but without specifying a "dependency" (either by using depends_on or by using cluster variable as input param values for the cloud_backup_schedule resource).

Behind the scenes though the plugin was crashing during this workflow because of a bug occurring when trying to check any existing policy and trying to read the response. In fact, we were trying to read the response of the request even when error was returned. This made the plugin to fully crash.

With this fix, we actually achieve what I believe is the correct state:
- if customer **does not specify a dependency** between the two resource it will get an error of this type:
```
│ Error: error creating a Cloud Backup Schedule: PATCH https://cloud-dev.mongodb.com/api/atlas/v1.0/groups/6504522396b90345de085a13/clusters/clusterTest4/backup/schedule: 404 (request "CLUSTER_NAME_NOT_FOUND") No cluster with name "clusterTest4" exists in group 6504522396b90345de085a13.
│ 
│   with mongodbatlas_cloud_backup_schedule.test,
│   on main.tf line 29, in resource "mongodbatlas_cloud_backup_schedule" "test":
│   29: resource "mongodbatlas_cloud_backup_schedule" "test" {
```

which should be clear to understand. Eventually, if they retry it will actually work. But no matter what we are updating the documentation to make sure this is clear.

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cloud_backup_schedule.go
+++ b/mongodbatlas/resource_mongodbatlas_cloud_backup_schedule.go
@@ -454,14 +454,16 @@ func resourceMongoDBAtlasCloudBackupScheduleImportState(ctx context.Context, d *
 }
 
 func cloudBackupScheduleCreateOrUpdate(ctx context.Context, conn *matlas.Client, d *schema.ResourceData, projectID, clusterName string) error {
+	policy := matlas.Policy{}
 	// Get policies items
 	resp, _, err := conn.CloudProviderSnapshotBackupPolicies.Get(ctx, projectID, clusterName)
 	if err != nil {
 		log.Printf("error getting MongoDB Cloud Backup Schedule (%s): %s", clusterName, err)
+	} else if len(resp.Policies) == 1 {
+		policy.ID = resp.Policies[0].ID
 	}
 
 	req := &matlas.CloudProviderSnapshotBackupPolicy{}
-	policy := matlas.Policy{}
 	policyItem := matlas.PolicyItem{}
 	var policiesItem []matlas.PolicyItem
 	export := matlas.Export{}
@@ -533,10 +535,6 @@ func cloudBackupScheduleCreateOrUpdate(ctx context.Context, conn *matlas.Client,
 
 	if d.HasChange("use_org_and_group_names_in_export_prefix") {
 		req.UseOrgAndGroupNamesInExportPrefix = pointy.Bool(d.Get("use_org_and_group_names_in_export_prefix").(bool))
-	}
-
-	if len(resp.Policies) == 1 {
-		policy.ID = resp.Policies[0].ID
 	}
 
 	policy.PolicyItems = policiesItem
--- a/website/docs/r/cloud_backup_schedule.html.markdown
+++ b/website/docs/r/cloud_backup_schedule.html.markdown
@@ -16,6 +16,8 @@ description: |-
 
 -> **NOTE:** If Backup Compliance Policy is enabled for the project for which this backup schedule is defined, you cannot modify the backup schedule for an individual cluster below the minimum requirements set in the Backup Compliance Policy.  See [Backup Compliance Policy Prohibited Actions and Considerations](https://www.mongodb.com/docs/atlas/backup/cloud-backup/backup-compliance-policy/#configure-a-backup-compliance-policy).
 
+-> **NOTE:** When creating a backup schedule you **must either** use the `depends_on` clause to indicate the cluster to which it refers **or** specify the values of `project_id` and `cluster_name` as reference of the cluster resource (e.g. `cluster_name = mongodbatlas_cluster.my_cluster.name` - see the example below). Failure in doing so will result in an error when executing the plan.
+
 In the Terraform MongoDB Atlas Provider 1.0.0 we have re-architected the way in which Cloud Backup Policies are manged with Terraform to significantly reduce the complexity. Due to this change we've provided multiple examples below to help express how this new resource functions.
 
 


mongodbatlas/resource_mongodbatlas_cloud_backup_schedule.go
```diff
@@ -454,14 +454,16 @@ func resourceMongoDBAtlasCloudBackupScheduleImportState(ctx context.Context, d *
 }
 
 func cloudBackupScheduleCreateOrUpdate(ctx context.Context, conn *matlas.Client, d *schema.ResourceData, projectID, clusterName string) error {
+	policy := matlas.Policy{}
 	// Get policies items
 	resp, _, err := conn.CloudProviderSnapshotBackupPolicies.Get(ctx, projectID, clusterName)
 	if err != nil {
 		log.Printf("error getting MongoDB Cloud Backup Schedule (%s): %s", clusterName, err)
+	} else if len(resp.Policies) == 1 {
+		policy.ID = resp.Policies[0].ID
```
lantoli, line 463: Don't we want to do anything with id if we get more than 1 policy in the response?

lantoli (approved):

website/docs/r/cloud_backup_schedule.html.markdown
```diff
@@ -16,6 +16,8 @@ description: |-
 
 -> **NOTE:** If Backup Compliance Policy is enabled for the project for which this backup schedule is defined, you cannot modify the backup schedule for an individual cluster below the minimum requirements set in the Backup Compliance Policy.  See [Backup Compliance Policy Prohibited Actions and Considerations](https://www.mongodb.com/docs/atlas/backup/cloud-backup/backup-compliance-policy/#configure-a-backup-compliance-policy).
 
+-> **NOTE:** When creating a backup schedule you **must either** use the `depends_on` clause to indicate the cluster to which it refers **or** specify the values of `project_id` and `cluster_name` as reference of the cluster resource (e.g. `cluster_name = mongodbatlas_cluster.my_cluster.name` - see the example below). Failure in doing so will result in an error when executing the plan.
+
```
andreaangiolillo, line 20: What about the feedback that the TSE shared with us today? (see https://mongodb.slack.com/archives/CLB2CJ1JB/p1694774725425429?thread_ts=1694106650.522989&cid=CLB2CJ1JB)

> But shouldn’t this resource see if there is a cluster named to what is supplied in the project and then execute, rather than starting execution and failing?

@Zuhairahmed Could you clarify which is the preferred option here?

andreaangiolillo: Actually, nvm my previous message. It makes sense to check for the cluster's status only if the `mongodbatlas_cloud_backup_schedule` fails when the cluster is in the 'IN_PROGRESS' state. If we are confident that the resource won't fail in that situation, displaying the error message is the best approach.

andreaangiolillo (approved):

mongodbatlas/resource_mongodbatlas_cloud_backup_schedule.go
marcosuma: I just moved the existing logic that was on the bottom. I am not really sure why we only re-use the ID if there is only one policy.

@andreaangiolillo  do you have context here?

website/docs/r/cloud_backup_schedule.html.markdown
marcosuma: >It makes sense to check for the cluster's status only if the mongodbatlas_cloud_backup_schedule fails when the cluster is in the 'IN_PROGRESS' state. If we are confident that the resource won't fail in that situation, displaying the error message is the best approach.

Sorry but I don't understand you here.

andreaangiolillo: The cluster has a status field which can have 4 values: `IN_PROGRESS`, `IDLE`, `DELETING` `UPDATING`. I was wondering what would happen if the cluster is not `IDLE` (the cluster is not "ready") and we use `mongodbatlas_cloud_backup_schedule`. I believe the resource won't fail as it will still find a cluster name

mongodbatlas/resource_mongodbatlas_cloud_backup_schedule.go
andreaangiolillo: policies can have only 1 element. Since this function is used also for updates, we need to reuse the policy id to update it as part of the Update operation

marcosuma: @andreaangiolillo I figured that was gonna be the answer but funny enough I was not able to find where is the definition that says it only has 1 element. This is [the closest](https://github.com/mongodb/go-client-mongodb-atlas/blob/d023c4dda54a37f2d7b02242067e9398898e8843/mongodbatlas/cloud_provider_snapshot_backup_policies.go#L44) I found but I couldn't find where it says "only 1 elem".

website/docs/r/cloud_backup_schedule.html.markdown
marcosuma: I think if it doesn't fail it's the best case scenario. If it fails it will return an error (not a crash) that mentions the cluster was not found. For sure an alternative option would be to do what the TSE proposed but that also mean that we are adding one more call to the backend to do that. I would say this proposal should be fine and resolving customer's concerns.

andreaangiolillo: agree. Thanks for the discussion here

PR #1476, opened 2023-09-15
fix: populate total_count in mongodbatlas_alert_configurations data source 

lantoli:
## Description

Ticket: [INTMDB-1024](https://jira.mongodb.org/browse/INTMDB-1024)

- mongodbatlas_alert_configurations data source populates total_count
- updated to use the new SDK
- use RFC3339Nano instead of RFC3339 to format times

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments

Manual tested that the examples in atlas-alert-configuration work exactly the same before and after the changes in this PR, and terraform apply doesn't detect any change in the plan.


mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
```diff
@@ -321,124 +322,111 @@ func outputAlertConfiguration(alert *matlas.AlertConfiguration, outputType, reso
...
+		values["interval_min"] = cty.NumberIntVal(int64(*notification.IntervalMin))
 	}
 
-	if notification.OpsGenieRegion != "" {
-		values["ops_genie_region"] = cty.StringVal(notification.OpsGenieRegion)
+	if conversion.IsStringPresent(notification.MobileNumber) {
+		values["mobile_number"] = cty.StringVal(*notification.MobileNumber)
 	}
 
-	if notification.OrgName != "" {
```
marcosuma, line 424: same here for OrgName, I am assuming the new SDK does not require it?

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
```diff
@@ -321,124 +322,111 @@ func outputAlertConfiguration(alert *matlas.AlertConfiguration, outputType, reso
...
-		values["datadog_region"] = cty.StringVal(notification.DatadogRegion)
-	}
-
-	if notification.EmailAddress != "" {
-		values["email_address"] = cty.StringVal(notification.EmailAddress)
+	if conversion.IsStringPresent(notification.ChannelName) {
+		values["channel_name"] = cty.StringVal(*notification.ChannelName)
 	}
 
-	if notification.FlowName != "" {
```
marcosuma, line 408: has this field been removed? I don't see FlowName in the new code

marcosuma (approved):

marcosuma:
Minor nit: this is a good example where since the PR is comprised of two important pieces (update to new SDK and populate total_count) I would have probably split it into two PRs.

I would say right now the title of the PR mentions we are fixing `total_count` but that is just few lines of code; most of it consists of updating it to the new SDK.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
lantoli: correct, they are for a notification type that doesn't exist any more

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
lantoli: correct, they are for a notification type that doesn't exist any more

lantoli:
> Minor nit: this is a good example where since the PR is comprised of two important pieces (update to new SDK and populate total_count) I would have probably split it into two PRs.
> 
> I would say right now the title of the PR mentions we are fixing `total_count` but that is just few lines of code; most of it consists of updating it to the new SDK.

Agreed, that change is also mentioned in the PR description but it would be better to have a different PR for that

mongodbatlas/framework/conversion/string.go
```diff
@@ -18,11 +18,21 @@ func TypesListToString(ctx context.Context, set types.List) []string {
 	return results
 }
 
+// IsStringPresent returns true if the string is non-empty.
+func IsStringPresent(strPtr *string) bool {
```
AgustinBettati: As this conversion function is not related to any terraform plugin framework types, could be worth placing in util/type_conversion instead.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
```diff
@@ -291,26 +292,26 @@ func computeAlertConfigurationOutput(alert *matlas.AlertConfiguration, definedOu
...
 		})),
 		ProjectID:             types.StringValue(projectID),
-		AlertConfigurationID:  types.StringValue(apiRespConfig.ID),
-		EventType:             types.StringValue(apiRespConfig.EventTypeName),
-		Created:               types.StringValue(apiRespConfig.Created),
-		Updated:               types.StringValue(apiRespConfig.Updated),
+		AlertConfigurationID:  types.StringValue(*apiRespConfig.Id),
+		EventType:             types.StringValue(*apiRespConfig.EventTypeName),
+		Created:               types.StringValue(util.TimeToString(*apiRespConfig.Created)),
+		Updated:               types.StringValue(util.TimeToString(*apiRespConfig.Updated)),
```
AgustinBettati:
```suggestion
		Created:               types.StringPointerValue(util.TimePtrToStringPtr(apiRespConfig.Created)),
		Updated:              types.StringPointerValue(util.TimePtrToStringPtr(apiRespConfig.Updated)),
```
Would favor using `util.TimePtrToStringPtr` just as a safer option to avoid potential nil pointer dereference.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
```diff
@@ -321,124 +322,111 @@ func outputAlertConfiguration(alert *matlas.AlertConfiguration, outputType, reso
...
-		"metric_name": cty.StringVal(metric.MetricName),
-		"operator":    cty.StringVal(metric.Operator),
-		"threshold":   cty.NumberFloatVal(metric.Threshold),
-		"units":       cty.StringVal(metric.Units),
-		"mode":        cty.StringVal(metric.Mode),
+		"metric_name": cty.StringVal(*metric.MetricName),
+		"operator":    cty.StringVal(*metric.Operator),
+		"threshold":   cty.NumberFloatVal(*metric.Threshold),
+		"units":       cty.StringVal(*metric.Units),
+		"mode":        cty.StringVal(*metric.Mode),
```
AgustinBettati: Do we know if these values are always returned? Would check for nils to avoid a potential crash here, and we can check in upstream code if these values are in fact required.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
```diff
@@ -321,124 +322,111 @@ func outputAlertConfiguration(alert *matlas.AlertConfiguration, outputType, reso
...
 
-func convertThresholdToCtyValues(threshold matlas.Threshold) map[string]cty.Value {
+func convertThresholdToCtyValues(threshold *admin.GreaterThanRawThreshold) map[string]cty.Value {
 	return map[string]cty.Value{
-		"operator":  cty.StringVal(threshold.Operator),
-		"units":     cty.StringVal(threshold.Units),
-		"threshold": cty.NumberFloatVal(threshold.Threshold),
+		"operator":  cty.StringVal(*threshold.Operator),
+		"units":     cty.StringVal(*threshold.Units),
+		"threshold": cty.NumberFloatVal(float64(*threshold.Threshold)), // int in new SDK but keeping float64 for backward compatibility
```
AgustinBettati: same as comment above.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -820,3 +963,38 @@ func newTFMatcherModelList(matlasSlice []matlas.Matcher, currStateSlice []tfMatc
 	}
 	return matchers
 }
+
+func newTFMatcherModelListV2(m []map[string]interface{}, currStateSlice []tfMatcherModel) []tfMatcherModel {
```
AgustinBettati, line 967: Is there a reason why SDK v2 returns a map instead of struct for the `matchers` field?

AgustinBettati (changes requested):

mongodbatlas/framework/conversion/string.go
gssbzn: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

lantoli: thanks, moving it in a new commit

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
lantoli: I don't know, I guess it's to support additional fields in the future

mongodbatlas/framework/conversion/string.go
AgustinBettati: taking into account what Gus referenced, we could remove the `mongodbatlas/util` package altogether and place the type_conversion file within the mongodbatlas package (where it is being used).

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
lantoli: thanks, good idea, adding it in a new commit

mongodbatlas/framework/conversion/string.go
lantoli: @gssbzn thanks, you're right about that but at the moment we have very few functions so it would also be a premature optimization to create a package for only one function. as we progress we'll probably extract those from util to a more meaningful package.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
lantoli: there is not problem in this case but I'll add a commit to check this in case it changes in the future

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration.go
lantoli: there is not problem in this case but I'll add a commit to check this in case it changes in the future

mongodbatlas/framework/conversion/string.go
gssbzn: agree on premature abstraction but the idea is more on preventing a util package, once it is there is really hard to get rid of it so up to you folks on how to resolve

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: I see that depending on the alert configuration type the `matchers` field can be defined (with a specific schema) or needs to be left empty. Reference to Atlas API docs: https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Alert-Configurations/operation/getAlertConfiguration.
The Go SDK seems to be using a particular alert config type which does not define the schema. @wtrocki What do you think if we create a ticket to see if the proper schema can be defined?

mongodbatlas/framework/conversion/string.go
lantoli: thanks @gssbzn , following your advice we see that it doesn't make sense to have a util package only used for the TF repo so we're integrating it, no util package needed.   We'll do it in a new PR as this is getting big and it's not related to this PR specifically

AgustinBettati (approved):

mongodbatlas/fw_data_source_mongodbatlas_alert_configurations.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-	matlas "go.mongodb.org/atlas/mongodbatlas"
+	"go.mongodb.org/atlas-sdk/v20230201006/admin"
```
andreaangiolillo, line 13:
```suggestion
	"go.mongodb.org/atlas-sdk/v20230201008/admin"
```

Can we use the latest version? https://github.com/mongodb/atlas-sdk-go/releases/tag/v20230201008.0.0

This can be addressed in a followup PR

andreaangiolillo (approved):
LGTM

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -765,6 +848,39 @@ func newTFMetricThresholdConfigModel(matlasMetricThreshold *matlas.MetricThresho
 	return []tfMetricThresholdConfigModel{newState}
 }
 
+func newTFMetricThresholdConfigModelV2(t *admin.ServerlessMetricThreshold, currStateSlice []tfMetricThresholdConfigModel) []tfMetricThresholdConfigModel {
```
andreaangiolillo, line 851:
```suggestion
func newTFMetricThresholdConfigModel(t *admin.ServerlessMetricThreshold, currStateSlice []tfMetricThresholdConfigModel) []tfMetricThresholdConfigModel {
```
Why using v2 in the name?

lantoli: I upgraded data sources for alert_notification and alert_notifications but not for the resource.  These are 3 functions in the resource file which are also being used by the data sources, so I kept the original functions for the resource and these v2 methods can be used when we migrate the alert configuration resource to the new sdk

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
wtrocki: Matchers are [dynamic](https://github.com/10gen/mms/blob/00f5a0ad4db46a1b012658484b91a87be17f7e36/server/src/main/com/xgen/cloud/alerts/configs/runtime/res/ApiAlertConfigsMatchersResource.java#L29) resource and also exist only on couple subtypes. +1 for jira but we are limited in what we can do here (might be not worth).

`[]map[string]interface{},` usually means dynamic value/object without any structure. 

> NOTE: For terraform we should avoid making those part of the resource state as it would likely trigger unwanted behaviour.

PR #1463, opened 2023-09-11
refactor: use retry.StateChangeConf for encryption-at-rest resource.

marcosuma:
## Description

Use `retry.StateChangeConf` instead of the `for loop` to remove technical debt and improve retry strategy.

Link to any related issue(s): https://jira.mongodb.org/browse/INTMDB-1039

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [X] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [X] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [X] I have added any necessary documentation (if appropriate)
- [X] I have run make fmt and formatted my code

PR #1477, opened 2023-09-19
refactor: use retry.StateChangeConf for encryption-at-rest resource.

marcosuma:
## Description

Link to any related issue(s): https://jira.mongodb.org/browse/INTMDB-1039

In this change we are removing the `for-loop` used as retry mechanism for the `encryption_at_rest` resource creation and using the `retry strategy` provided by the terraform plugin sdk.

### Test

**Note**: I think we should unit test this the method used in the retry logic itself. But for now I've tested it manually with the following 

```
main.tf:

resource "mongodbatlas_encryption_at_rest" "test" {
  project_id = "64fecfd3da30680cf9043172"

  aws_kms_config {
    enabled                 = true
    customer_master_key_id  = "..."
    region                  = "US_EAST_1"
    role_id                 = "64fecfd591da66678694d4b6"

  }
}

```

and succeeded:

```
Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

mongodbatlas_encryption_at_rest.test: Creating...
mongodbatlas_encryption_at_rest.test: Creation complete after 5s [id=64fecfd3da30680cf9043172]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

I've also tried to "fake" it and return "RETRY" instead of "COMPLETED"

```
<<<< return encryptionResp, "COMPLETED", nil

>>>> return encryptionResp, "RETRY", nil
```

and got (unrelated) error which proves it indeed successfuly created the resource but kept retrying.

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
+++ b/mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
@@ -32,6 +32,9 @@ const (
 	errorReadEncryptionAtRest    = "error getting Encryption At Rest: %s"
 	errorDeleteEncryptionAtRest  = "error deleting Encryption At Rest: (%s): %s"
 	errorUpdateEncryptionAtRest  = "error updating Encryption At Rest: %s"
+	retryStrategyPendingState    = "PENDING"
+	retryStrategyCompletedState  = "COMPLETED"
+	retryStrategyErrorState      = "ERROR"
 )
 
 var _ resource.ResourceWithConfigure = &EncryptionAtRestRS{}
@@ -227,16 +230,17 @@ func (r *EncryptionAtRestRS) Create(ctx context.Context, req resource.CreateRequ
 	}
 
 	stateConf := &retry.StateChangeConf{
-		Pending:    []string{"RETRY"},
-		Target:     []string{"COMPLETED", "ERROR"},
+		Pending:    []string{retryStrategyPendingState},
+		Target:     []string{retryStrategyCompletedState, retryStrategyErrorState},
 		Refresh:    resourceMongoDBAtlasEncryptionAtRestCreateRefreshFunc(ctx, projectID, conn, encryptionAtRestReq),
 		Timeout:    1 * time.Minute,
 		MinTimeout: 1 * time.Second,
 		Delay:      0,
 	}
 
-	encryptionResp, err := stateConf.WaitForStateContext(ctx)
-	if err != nil {
+	var encryptionResp interface{}
+	var err error
+	if encryptionResp, err = stateConf.WaitForStateContext(ctx); err != nil {
 		resp.Diagnostics.AddError(fmt.Sprintf(errorCreateEncryptionAtRest, projectID), err.Error())
 		return
 	}
@@ -262,11 +266,11 @@ func resourceMongoDBAtlasEncryptionAtRestCreateRefreshFunc(ctx context.Context, 
 				log.Println("retrying ")
 
 				encryptionAtRestReq.GroupID = projectID
-				return encryptionResp, "RETRY", nil
+				return encryptionResp, retryStrategyPendingState, nil
 			}
-			return encryptionResp, "ERROR", err
+			return encryptionResp, retryStrategyErrorState, err
 		}
-		return encryptionResp, "COMPLETED", nil
+		return encryptionResp, retryStrategyCompletedState, nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


marcosuma:
I am also going to make sure I apply these [proposed changes ](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1463) by Andrea.

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -257,6 +256,24 @@ func (r *EncryptionAtRestRS) Create(ctx context.Context, req resource.CreateRequ
 	}
 }
 
+func resourceMongoDBAtlasEncryptionAtRestCreateRefreshFunc(ctx context.Context, projectID string, conn *matlas.Client, encryptionAtRestReq *matlas.EncryptionAtRest) retry.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		encryptionResp, _, err := conn.EncryptionsAtRest.Create(ctx, encryptionAtRestReq)
+		if err != nil {
+			if strings.Contains(err.Error(), "CANNOT_ASSUME_ROLE") || strings.Contains(err.Error(), "INVALID_AWS_CREDENTIALS") ||
```
marcosuma: This code has just been moved from its previous place. As per one [previous comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1463) I am not sure we can use here `errors.Is(...)` statement because it would not be the same as checking if `string contains`

andreaangiolillo: You can define two errors

```
var cannotAssumeRole := errors.new("CANNOT_ASSUME_ROLE") 
var invalidAWSCredentials "= errors.new("INVALID_AWS_CREDENTIALS")
```
and the use `errors.Is(err, cannotAssumeRole)`

marcosuma: @andreaangiolillo I know we can do that, but my point here is that I don't know if `err.Error()` **only** contains that exact string or it contains more. Does it make sense or am I missing something?

Based on the `errors.Is()` documentation:
```
// Is reports whether any error in err's tree **matches** target.
```

andreaangiolillo: ah, thanks for the clarification. You can find out in mms 

> CANNOT_ASSUME_ROLE(
>       "Atlas cannot assume the specified role.", "Atlas cannot assume the specified role (%s)."),

>   INVALID_AWS_CREDENTIALS(HttpServletResponse.SC_BAD_REQUEST, "Invalid AWS credentials."),


https://github.com/10gen/mms/blob/74bd3430d12c4c446b8f214222b2c3ea9e91a452/server/src/main/com/xgen/svc/mms/api/res/atlas/ApiAtlasEncryptionAtRestResource.java#L273

https://github.com/10gen/mms/blob/74bd3430d12c4c446b8f214222b2c3ea9e91a452/server/src/main/com/xgen/svc/mms/api/res/atlas/ApiAtlasEncryptionAtRestResource.java#L406

maastha (approved):
LGTM

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
```diff
@@ -225,28 +227,23 @@ func (r *EncryptionAtRestRS) Create(ctx context.Context, req resource.CreateRequ
...
 	}
 
-	var encryptionResp *matlas.EncryptionAtRest
+	stateConf := &retry.StateChangeConf{
+		Pending:    []string{retrystrategy.RetryStrategyPendingState},
+		Target:     []string{retrystrategy.RetryStrategyCompletedState, retrystrategy.RetryStrategyErrorState},
+		Refresh:    resourceMongoDBAtlasEncryptionAtRestCreateRefreshFunc(ctx, projectID, conn, encryptionAtRestReq),
+		Timeout:    1 * time.Minute,
+		MinTimeout: 1 * time.Second,
+		Delay:      0,
```
andreaangiolillo, line 236: I believe it makes sense to add a delay here to wait before retrying the operation.

andreaangiolillo (approved):
LGTM

mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
marcosuma: Delay here means:

```
Delay          time.Duration    // Wait this time before starting checks
```

for reference:

```
// StateChangeConf is the configuration struct used for `WaitForState`.
type StateChangeConf struct {
	Delay          time.Duration    // Wait this time before starting checks
	Pending        []string         // States that are "allowed" and will continue trying
	Refresh        StateRefreshFunc // Refreshes the current state
	Target         []string         // Target state
	Timeout        time.Duration    // The amount of time to wait before timeout
	MinTimeout     time.Duration    // Smallest time to wait before refreshes
	PollInterval   time.Duration    // Override MinTimeout/backoff and only poll this often
	NotFoundChecks int              // Number of times to allow not found (nil result from Refresh)

	// This is to work around inconsistent APIs
	ContinuousTargetOccurence int // Number of times the Target state has to occur continuously
}
```

PR #1503, opened 2023-10-03
deprecate: Deletes resources and data sources that were announced to be removed in 1.12.0.

marcosuma:
## Description

There were some resources and data_sources that were declared as deprecated and to be removed in `1.12.0` - we are doing this as part of our next release `1.12.2`.

Link to any related issue(s): 
- https://jira.mongodb.org/browse/INTMDB-1147 
- https://jira.mongodb.org/browse/INTMDB-806

## Type of change:

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.

## Required Checklist:


mongodbatlas/data_source_mongodbatlas_data_lake_pipeline_test.go
```diff
@@ -1,12 +1,14 @@
 package mongodbatlas
```
marcosuma, line 1: @Zuhairahmed @andreaangiolillo @AgustinBettati @lantoli @maastha 

We are deprecating data_lake[s] but the data_lake_pipeline[s] were not deprecated so they are actually staying. 

Is this correct?

mongodbatlas/provider.go
```diff
@@ -219,7 +215,6 @@ func getResourcesMap() map[string]*schema.Resource {
 		"mongodbatlas_privatelink_endpoint":                                        resourceMongoDBAtlasPrivateLinkEndpoint(),
 		"mongodbatlas_privatelink_endpoint_serverless":                             resourceMongoDBAtlasPrivateLinkEndpointServerless(),
 		"mongodbatlas_privatelink_endpoint_service":                                resourceMongoDBAtlasPrivateEndpointServiceLink(),
-		"mongodbatlas_privatelink_endpoint_service_adl":                            resourceMongoDBAtlasPrivateLinkEndpointServiceADL(),
```
AgustinBettati, line 222: Something to consider: Even though we have the deprecation message specifying removal in 1.12.0 version, I am concerned in including the removal of complete resources within a minor version `1.12.2`. As a user updating to a minor version, I would expect a guarantee of no breaking changes.
If we have no urgency with these removals, we could schedule for `1.13.0` as a safer bet.

marcosuma: I think that's something we discussed already, I believe in the risk you are mentioning but I thought we justified it by saying that in any case this was declared to be removed in 1.12.0, so no matter what those customers would have been already impacted and they should have already updated it.

of course these are cases where a little bit of telemetry (e.g. how many are using this resource?) would have been great.

mongodbatlas/data_source_mongodbatlas_data_lake_pipeline_test.go
AgustinBettati: Looking at this issue https://jira.mongodb.org/browse/INTMDB-806 its seems correct.

mongodbatlas/data_source_mongodbatlas_data_lake_pipeline_test.go
```diff
@@ -36,6 +38,25 @@ func TestAccDataSourceClusterRSDataLakePipeline_basic(t *testing.T) {
 	})
 }
 
+func testAccCheckMongoDBAtlasDataLakeDestroy(s *terraform.State) error {
+	conn := testAccProviderSdkV2.Meta().(*MongoDBClient).Atlas
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "mongodbatlas_data_lake" {
+			continue
+		}
```
AgustinBettati: Not sure if there is value in preserving this function given that it looks for the resource mongodbatlas_data_lake which is removed

mongodbatlas/data_source_mongodbatlas_data_lake_pipeline_test.go
Zuhairahmed: Correct, you can see Data Lake Pipelines is a seperate endpoint in latest Atlas Admin API and was not deprecated there either: https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Data-Lake-Pipelines

mongodbatlas/data_source_mongodbatlas_data_lake_pipeline_test.go
marcosuma: good point!

AgustinBettati: Sorry for not stating this before, we could maintain the checkDestroy method but calling `conn.DataLakePipeline.Get(ctx, projectID, name)` to check that the data lake pipeline resource is destroyed, which is relevant for this test

marcosuma: Gotcha, I was assuming it wasn't needed because I think we weren't doing it, but let me look into it

.github/workflows/acceptance-tests.yml
```diff
@@ -47,7 +47,6 @@ jobs:
           generic:
             - 'mongodbatlas/data_source_mongodbatlas_backup_compliance_policy*.go'
             - 'mongodbatlas/resource_mongodbatlas_backup_compliance_policy*.go'
-            - 'mongodbatlas/data_source_mongodbatlas_data_lakes_test*.go'
```
andreaangiolillo, line 50: I see you removed also private endpoint adl. Do we also have to remove it from here?

.github/workflows/acceptance-tests.yml
```diff
@@ -47,7 +47,6 @@ jobs:
...
             - 'mongodbatlas/resource_mongodbatlas_x509_authentication_database_user*.go'
             - 'mongodbatlas/resource_mongodbatlas_auditing*.go'
```
andreaangiolillo, line 51: Could you also make sure to delete the documentation and examples related to these resources?

mongodbatlas/data_source_mongodbatlas_data_lake_pipeline_test.go
marcosuma: @AgustinBettati changes done.

.github/workflows/acceptance-tests.yml
marcosuma: In this file we have:

```
          network:
            - 'mongodbatlas/**_network**.go'  
            - 'mongodbatlas/**_privatelink**.go' 
            - 'mongodbatlas/**_private_endpoint**.go'
```

which cannot be removed because we still have:
```
terraform-provider-mongodbatlas/mongodbatlas/resource_mongodbatlas_privatelink_endpoint_serverless.go
```

I think there is no other change needed for this file @andreaangiolillo

andreaangiolillo: Thanks for checking!

.github/workflows/acceptance-tests.yml
marcosuma: Good point! Let me do that.

marcosuma: @andreaangiolillo done

AgustinBettati (approved):

andreaangiolillo (approved):
LGTM

PR #1513, opened 2023-10-09
fix: alert configuration data source nil pointer with third party notifications

AgustinBettati:
## Description

Jira ticket: [INTMDB-1182](https://jira.mongodb.org/browse/INTMDB-1182)
Bug found when validating fix for [INTMDB-981](https://jira.mongodb.org/browse/INTMDB-981).

This nil pointer error is encountered when using the alert configuration data source for a third party alert configuration notification that does not define the `interval_min` attribute.
An acceptance test reproducing this issue was defined but skipped in CI as it relies on external credentials. Was able to verify that dummy credentials can be used to successfully create third party alert notifications so adjusted the corresponding tests so they all run in our CI moving forward.

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/fw_data_source_mongodbatlas_alert_configuration_test.go
+++ b/mongodbatlas/fw_data_source_mongodbatlas_alert_configuration_test.go
@@ -105,17 +105,18 @@ func TestAccConfigDSAlertConfiguration_withPagerDuty(t *testing.T) {
 	var (
 		alert          = &matlas.AlertConfiguration{}
 		dataSourceName = "data.mongodbatlas_alert_configuration.test"
-		projectID      = os.Getenv("MONGODB_ATLAS_PROJECT_ID")
-		serviceKey     = "dummykey111111111111111111111111"
+		orgID          = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName    = acctest.RandomWithPrefix("test-acc")
+		serviceKey     = dummy32CharKey
 	)
 
 	resource.Test(t, resource.TestCase{
-		PreCheck:                 func() { testAccPreCheck(t) },
+		PreCheck:                 func() { testAccPreCheckBasic(t) },
 		ProtoV6ProviderFactories: testAccProviderV6Factories,
 		CheckDestroy:             testAccCheckMongoDBAtlasAlertConfigurationDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccDSMongoDBAtlasAlertConfigurationConfigWithPagerDuty(projectID, serviceKey, true),
+				Config: testAccDSMongoDBAtlasAlertConfigurationConfigWithPagerDuty(orgID, projectName, serviceKey, true),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckMongoDBAtlasAlertConfigurationExists(dataSourceName, alert),
 					resource.TestCheckResourceAttrSet(dataSourceName, "project_id"),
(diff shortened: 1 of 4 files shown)


mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
```diff
@@ -744,9 +745,9 @@ func newTFNotificationModelListV2(n []admin.AlertsNotificationRootForGroup, curr
 				Roles:          value.Roles,
 				ChannelName:    conversion.StringPtrNullIfEmpty(value.ChannelName),
 				DatadogRegion:  conversion.StringPtrNullIfEmpty(value.DatadogRegion),
-				DelayMin:       types.Int64Value(int64(*value.DelayMin)),
+				DelayMin:       types.Int64PointerValue(util.IntPtrToInt64Ptr(value.DelayMin)),
```
andreaangiolillo, line 748: I am not sure why you need `Int64PointerValue` and `util.IntPtrToInt64Ptr`.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
```diff
@@ -398,12 +397,10 @@ func TestAccConfigRSAlertConfiguration_importPagerDuty(t *testing.T) {
 }
 
 func TestAccConfigRSAlertConfiguration_DataDog(t *testing.T) {
-	SkipTestExtCred(t) // Will skip because requires external credentials aka api key
-	SkipTest(t)        // Will force skip if enabled
 	var (
 		resourceName = "mongodbatlas_alert_configuration.test"
 		projectID    = os.Getenv("MONGODB_ATLAS_PROJECT_ID")
-		ddAPIKey     = os.Getenv("DD_API_KEY")
+		ddAPIKey     = "11111111111111111111111111111111"
```
andreaangiolillo: instead of using 11. Could you just use the value `dummykey11` everywhere? Does the endpoint run validation on the key?

mongodbatlas/fw_resource_mongodbatlas_alert_configuration.go
AgustinBettati: `IntervalMin` and `DelayMin` are of type *int, as in some case they may not be defined. `Int64PointerValue` and `util.IntPtrToInt64Ptr` are used to safely store the optional value into the model, without insecurely accessing the underlying value as done in `*value.DelayMin`.

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
AgustinBettati: It does run a format validation, which is why some keys have a length of 32 while other 36 with a particular format. Extracted to common variables to make this clear.

AgustinBettati:
cloud-dev seems unstable with project creations, will wait some time and run CI again.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration_test.go
```diff
@@ -102,21 +102,21 @@ func TestAccConfigDSAlertConfiguration_withOutput(t *testing.T) {
 }
 
 func TestAccConfigDSAlertConfiguration_withPagerDuty(t *testing.T) {
```
lantoli, line 104: is this still testing PagerDuty or we should change test name?

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
```diff
@@ -398,23 +402,22 @@ func TestAccConfigRSAlertConfiguration_importPagerDuty(t *testing.T) {
 }
 
 func TestAccConfigRSAlertConfiguration_DataDog(t *testing.T) {
```
lantoli, line 404: similar to before, are we still testing Datadog integration?

mongodbatlas/util/type_conversion.go
```diff
@@ -28,6 +28,15 @@ func Int64PtrToIntPtr(i64 *int64) *int {
 	return &i
 }
 
+func IntPtrToInt64Ptr(i *int) *int64 {
```
lantoli, line 31: this seems the inverse to admin.Int64PtrToIntPtr. , don't know if it makes sense to have it in the SDK

lantoli (approved):

andreaangiolillo (approved):

mongodbatlas/fw_resource_mongodbatlas_alert_configuration_test.go
AgustinBettati: yes, terraform configurations being tested are the same. Change here is that api key is now a hardcoded dummy value, when testing in cloud-dev found that there is no need to pass a real api key, as long as the format is correct the alert notifications are created successfully.

mongodbatlas/fw_data_source_mongodbatlas_alert_configuration_test.go
AgustinBettati: responded in other comment

mongodbatlas/util/type_conversion.go
AgustinBettati: this conversion only seems needed when using old sdk (that uses int type), so eventually will be removed.

PR #1529, opened 2023-10-16
fix: Encryption at rest return inconsistent plan when setting secret access key

andreaangiolillo:
## Description
Ticket: [INTMDB-1202](https://jira.mongodb.org/browse/INTMDB-1202)
Help Ticket: [HELP-50897](https://jira.mongodb.org/browse/HELP-50897)

The encryption at rest resource returns inconsistent results after apply when the secret access key is provided

```hcl
--
2023-10-16T16:05:09.424+0100 [ERROR] vertex "mongodbatlas_encryption_at_rest.mongoencrypt" error: Provider produced inconsistent result after apply
╷
│
│ When applying changes to mongodbatlas_encryption_at_rest.mongoencrypt, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new value: .aws_kms_config[0].secret_access_key: inconsistent
│ values for sensitive attribute.
│
```

This PR updates the logic to use the `secret_access_key` from the state file instead from the API response since the field is not returned.

### NOTE
Credentials based authentication is no longer supported by the API, if users try to pass accessKeyID and secretAccessKey for the encryption at rest resource for a new project, they will get the following error

```bash
{
    "detail": "Credentials based authentication is no longer supported for AWS KMS.",
    "error": 400,
    "errorCode": "AWS_KMS_CREDENTIALS_AUTH_DEPRECATED",
    "parameters": [],
    "reason": "Bad Request"
}
```
The Atlas team made the decision to allow customers to keep providing these credentials if the project was already using them to allow them to migrate to the new IAM role-based. As a result, if the user was already providing those credentials, they will still be able to do so. New projects will get the error instead. This is why there are no acceptance tests defined and the terraform documentation does not include a test about it. (code [mms/NDSUISvc.java](https://github.com/10gen/mms/blob/9275618efa3caa23e133b76d5eed4df20b69bf02/server/src/main/com/xgen/svc/nds/svc/NDSUISvc.java#L3383)
```java
    // Prevent credentials in a request when creds are not currently
    // in use.  Allow users with creds to keep using them until they switch
    if (!hasExistingCreds && hasCredentialsInRequest) {
      throw new SvcException(NDSErrorCode.AWS_KMS_CREDENTIALS_AUTH_DEPRECATED);
    }
```

Unfortunately, we cannot deprecate the fields as long as they are supported in the [API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Encryption-at-Rest-using-Customer-Key-Management/operation/updateEncryptionAtRest).

### How did I reproduce the issue
I ran Atlas locally and removed the validation in  [mms/NDSUISvc.java](https://github.com/10gen/mms/blob/9275618efa3caa23e133b76d5eed4df20b69bf02/server/src/main/com/xgen/svc/nds/svc/NDSUISvc.java#L3383)

## Type of change:

## Required Checklist:

## Next Steps
- I am reaching out to the team owning the API to see if they plan to deprecate the fields for the API

### DIFF
--- a/mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
+++ b/mongodbatlas/fw_resource_mongodbatlas_encryption_at_rest.go
@@ -9,8 +9,6 @@ import (
 	"reflect"
 	"time"
 
-	matlas "go.mongodb.org/atlas/mongodbatlas"
-
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
@@ -21,10 +19,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
-
 	"github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/conversion"
 	retrystrategy "github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/retry"
 	validators "github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/framework/validator"
+	matlas "go.mongodb.org/atlas/mongodbatlas"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


AgustinBettati (approved):
Great investigation, LGTM. Would update our terraform migration issues document with this case.

PR #1548, opened 2023-10-24
fix!: fixes global num_shards adding it as computed and removing the default.

marcosuma:
## Description

Based on [documentation](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v1/#tag/Clusters/operation/createLegacyCluster) it looks like this `numShards` field **(the one outside the `replication_specs`)** is not required, it has its own default and it can also be computed:

```
Number of shards up to 50 to deploy for a sharded cluster. The resource returns 1 to indicate a replica set and values of 2 and higher to indicate a sharded cluster. The returned value equals the number of shards in the cluster.
```

### Backward compatibility

| *old version* | *new version* | Result |
|--------|--------|--------| 
| Creates cluster **without** specifying num_shards... | ...and then runs terraform apply without changing anything | _works_ ✅ |
| Creates cluster **without** specifying num_shards... | ...and then runs terraform apply changing _num_shards_ | _works_ ✅ |
| Creates cluster specifying num_shards... | ...and then runs terraform apply without changing anything | _works_ ✅ | 
| Creates cluster specifying num_shards... | ...and then runs terraform apply changing the _num_shards_ | _works_ ✅ | 

Link to any related issue(s): https://jira.mongodb.org/browse/HELP-51008

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -137,7 +137,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"num_shards": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  1,
+				Computed: true,
 			},
 			"cloud_backup": {
 				Type:          schema.TypeBool,
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -389,10 +389,6 @@ func resourceMongoDBAtlasClusterCreate(ctx context.Context, d *schema.ResourceDa
 		if _, ok1 := d.GetOk("cluster_type"); !ok1 {
 			return diag.FromErr(fmt.Errorf("`cluster_type` should be set when `replication_specs` is set"))
 		}
-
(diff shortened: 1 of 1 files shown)


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -137,7 +137,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"num_shards": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  1,
```
maastha, line 140: Should we mention this as a breaking change in the Changelog? Although I understand that this is backward compatible but a change in behavior so might be worth to add to the release guide at least.

https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#example-major-number-increments

maastha (approved):

mongodbatlas/resource_mongodbatlas_cluster.go
marcosuma: @maastha ok, I think we can put that the label on this PR

lantoli (approved):

AgustinBettati (approved):

PR #1564, opened 2023-10-25
fix: 'computed' and 'default' usage based on the documentation

marcosuma:
## Description

Customer is complaining that using TF 1.0.8 certain fields are returning error because the provider (Atlas) is returning values but fields are not marked as "Computed". In other cases fields are marked as "Computed" but as per documentation these fields don't return values defined from the backend itself.

Documentation: https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v1/#tag/Clusters/operation/updateClusterAdvancedConfiguration

Definition of "Computed": `Computed indicates whether the provider may return its own value for this attribute or not. Computed cannot be used with Required. If Required and Optional are both false, the attribute will be considered "read only" for the practitioner, with only the provider able to set its value.`

This needs to go after https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1548

Link to any related issue(s): https://jira.mongodb.org/browse/HELP-51509

**Customer cannot upgrade TF version**

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -79,7 +79,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"backup_enabled": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     false,
+				Computed:    true,
 				Description: "Clusters running MongoDB FCV 4.2 or later and any new Atlas clusters of any type do not support this parameter",
 			},
 			"retain_backups_enabled": {
@@ -90,19 +90,16 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"bi_connector_config": {
 				Type:     schema.TypeList,
 				Optional: true,
-				Computed: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							Computed: true,
 						},
 						"read_preference": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: true,
 						},
 					},
 				},
@@ -293,7 +290,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"paused": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Computed: true,
 			},
 			"srv_address": {
 				Type:     schema.TypeString,
@@ -308,19 +305,16 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Type:       schema.TypeSet,
 				Optional:   true,
 				Set:        HashFunctionForKeyValuePair,
-				Computed:   true,
 				Deprecated: fmt.Sprintf(DeprecationByDateWithReplacement, "September 2024", "tags"),
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"key": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: true,
 						},
 						"value": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Computed: true,
 						},
 					},
 				},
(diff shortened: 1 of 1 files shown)


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -339,7 +336,7 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 			"version_release_system": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "LTS",
+				Computed:     true,
```
AgustinBettati, line 339: For the cases where the default is removed and replaced for computed, are we certain that the API is responding with the same value? Just want to make sure we are not incorporating a change for users that don't have this field defined.

mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -308,19 +308,16 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Type:       schema.TypeSet,
 				Optional:   true,
 				Set:        HashFunctionForKeyValuePair,
-				Computed:   true,
```
AgustinBettati, line 311: Is this a fix for the error "attribute representing nested block must not be unknown itself"? Or simply an improvement seeing Atlas API does not respond with labels.

marcosuma: it's both basically - as [per documentation](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v1/#tag/Clusters/operation/createLegacyCluster) nothing is returned and as you said it's deprecated. 

I am really not sure why we used **Computed** here originally.

mongodbatlas/resource_mongodbatlas_cluster.go
marcosuma: As per [documentation](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v1/#tag/Clusters/operation/createLegacyCluster):

```
Default: "LTS"
Enum: "LTS" "CONTINUOUS"
Method by which the cluster maintains the MongoDB versions. If value is CONTINUOUS, you must not specify mongoDBMajorVersion.
```

marcosuma: Correct me if I am wrong: to me when there is a default from the backend I interpret that as a way (not the only one) for the attribute to be `computed`.

lantoli (approved):

mongodbatlas/resource_mongodbatlas_cluster.go
AgustinBettati: yes that makes sense, thanks for clarifying

mongodbatlas/resource_mongodbatlas_cluster.go
AgustinBettati: Yes agree that computed is not accurate in this case. 
I am curious if the same scenario is happening with `bi_connector_config` and `advanced_configuration` which were bringing a similar issue, no need to adjust in this PR but more for my understanding in case you looked into it.

AgustinBettati (approved):
LGTM

mongodbatlas/resource_mongodbatlas_cluster.go
marcosuma: @AgustinBettati that's what I thought, but then something more complex happens there.
**Why? What is the difference?** The difference is that for labels we don't return anything, while for `advanced_configuration` the backend does return default value like:
```
  # advanced_configuration {
  #   # default_read_concern = ""
  #   # default_write_concern = ""
  #   fail_index_key_too_long = false
  #   javascript_enabled = true
  #   minimum_enabled_tls_protocol = "TLS1_2"
  #   no_table_scan = false
  #   oplog_min_retention_hours = 0
  #   # oplog_size_mb = 0
  #   sample_refresh_interval_bi_connector = 0
  #   sample_size_bi_connector = 0
  #   transaction_lifetime_limit_seconds = 0
  # }
```
**hence** when I do the `terraform apply` and it does a refresh and comparison it returns the error saying that `it was expecting advanced_configuration count to be 0 but the read returned count 1` (which basically mean that when we read the backend we load that configuration because its values are default).

What I've just said should apply also for `bi_connector_config` but I haven't tested that.

PR #1565, opened 2023-10-25
feat: Updates migration tests to run separately and use last released version of provider for plan checks

maastha:
## Description

This PR adds a new Github action workflow 'Migration Tests' and separates out existing migration acceptance tests into this workflow only i.e. migration tests will no longer run as part of existing Acceptance tests workflows in PRs.

_What are 'Migration tests'?_
_Migration tests are similar to Terraform acceptance tests but focus on ensuring that empty plans are generated when upgrading from a previous released version to the current version of the Terraform provider. See [documentation](https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#testing-migration)._ 

**Releases**: This workflow will run during release and use the **last released version** of our provider to run migration tests to ensure no unexpected plans are generated when our users upgrade.

**Manual**: This workflow can be manually run if required, using labels or via UI.

We can eventually disable these tests to run for every PR as migration tests usually cover scenarios already covered by acceptance tests, and require to only run these during releases or manually.

NOTE: This workflow was not tested for a run during release.
NOTE: This PR also introduces convention to use prefix `TestAccMigration*` for all migration tests.

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/code-health.yml
+++ b/.github/workflows/code-health.yml
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run ShellCheck
         uses: bewuethr/shellcheck-action@v2
-  call-accettance-tests-workflow:
+  call-acceptance-tests-workflow:
     needs: [build, lint, shellcheck, unit-test, website-lint]
     secrets: inherit
     permissions:
--- a/.github/workflows/code-health.yml
+++ b/.github/workflows/code-health.yml
@@ -98,4 +98,14 @@ jobs:
     uses: ./.github/workflows/acceptance-tests.yml
     with:
       parent-event-name: ${{ github.event_name }}
+  call-migration-tests-workflow:
+    needs: [build, lint, shellcheck, unit-test, website-lint]
+    secrets: inherit
+    permissions:
+      contents: write
+      pull-requests: read
+      repository-projects: read
+    uses: ./.github/workflows/migration-tests.yml
+    with:
+      parent-event-name: ${{ github.event_name }}
 
--- a/.github/workflows/migration-tests.yml
+++ b/.github/workflows/migration-tests.yml
@@ -2,8 +2,6 @@ name: 'Migration Tests'
 
 on:
   workflow_dispatch: {} # workflow can be run manually
-  schedule:
-    - cron: "0 4 * * *" # workflow runs every day at 04:00 AM
   workflow_call: # workflow runs after code-health
     inputs:
       parent-event-name:
@@ -40,7 +38,7 @@ jobs:
 
   project: 
     needs: [ change-detection ]
-    if: ${{ needs.change-detection.outputs.project == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-project' || inputs.parent-event-name == 'release' }}
+    if: ${{ needs.change-detection.outputs.project == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc-migration' || github.event.label.name == 'run-testacc-project-migration' || inputs.parent-event-name == 'release' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -76,7 +74,7 @@ jobs:
   
   config:
     needs: [ change-detection ]
-    if: ${{ needs.change-detection.outputs.config == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-config' || inputs.parent-event-name == 'release' }}
+    if: ${{ needs.change-detection.outputs.config == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc-migration' || github.event.label.name == 'run-testacc-config-migration' || inputs.parent-event-name == 'release' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/migration-tests.yml
+++ b/.github/workflows/migration-tests.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Get Last Release
         id: get_last_release
         run: |
-         LAST_RELEASE=$(git describe --abbrev=0 --tags)
+         LAST_RELEASE=$(curl -sSfL -X GET https://api.github.com/repos/mongodb/terraform-provider-mongodbatlas/releases/latest | jq -r '.tag_name | ltrimstr("v")')
          echo "Last release: $LAST_RELEASE"
          echo "MONGODB_ATLAS_LAST_VERSION=$LAST_RELEASE" >> $GITHUB_ENV
         shell: bash
--- a/.github/workflows/migration-tests.yml
+++ b/.github/workflows/migration-tests.yml
@@ -13,13 +13,14 @@ on:
     types: [ labeled ]
 
 jobs:  
-  mchange-detection:
+  change-detection:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: read
       repository-projects: read
     outputs:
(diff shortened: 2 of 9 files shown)


maastha:
unit-test step failing due to issue with `Adds coverage report to PRs` step

.github/workflows/code-health.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run ShellCheck
         uses: bewuethr/shellcheck-action@v2
-  call-accettance-tests-workflow:
+  call-acceptance-tests-workflow:
```
lantoli, line 91: finally, thanks! :-)

.github/workflows/migration-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+          config:
+            - 'mongodbatlas/fw_data_source_mongodbatlas_alert_configuration*.go'
+            - 'mongodbatlas/fw_data_source_mongodbatlas_database_user*.go'
+            - 'mongodbatlas/fw_data_source_mongodbatlas_atlas_user*.go'
+            - 'mongodbatlas/fw_resource_mongodbatlas_alert_configuration*.go'
+            - 'mongodbatlas/fw_resource_mongodbatlas_database_user*.go'   
+
+  project: 
+    needs: [ change-detection ]
+    if: ${{ needs.change-detection.outputs.project == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc-migration' || github.event.label.name == 'run-testacc-project-migration' || inputs.parent-event-name == 'release' }}
```
lantoli: i know we have this in other gh actions but I wonder if we could make this easier or have some functions or something

.github/workflows/migration-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+          TF_LOG: ${{ vars.LOG_LEVEL }}
+          TF_ACC: 1
+          PARALLEL_GO_TEST: 20
+          CI: true
+          TEST_REGEX: "^TestAccMigrationProject"
+        run: make testacc
+  
+  config:
+    needs: [ change-detection ]
+    if: ${{ needs.change-detection.outputs.config == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc-migration' || github.event.label.name == 'run-testacc-config-migration' || inputs.parent-event-name == 'release' }}
```
lantoli: same

mongodbatlas/fw_resource_mongodbatlas_database_user_migration_test.go
```diff
@@ -309,14 +316,15 @@ func TestAccConfigRSDatabaseUser_Migration_WithRoles(t *testing.T) {
...
-		orgID        = os.Getenv("MONGODB_ATLAS_ORG_ID")
-		projectName  = acctest.RandomWithPrefix("test-acc")
-		clusterName  = acctest.RandomWithPrefix("test-acc-cluster")
+		resourceName          = "mongodbatlas_database_user.test"
+		username              = acctest.RandomWithPrefix("test-acc-user-")
+		password              = acctest.RandomWithPrefix("test-acc-pass-")
+		orgID                 = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName           = acctest.RandomWithPrefix("test-acc")
+		clusterName           = acctest.RandomWithPrefix("test-acc-cluster")
+		lastVersionConstraint = os.Getenv("MONGODB_ATLAS_LAST_VERSION")
```
lantoli, line 327: do we need to check that this env var exists?

lantoli (approved):

mongodbatlas/fw_resource_mongodbatlas_database_user_migration_test.go
maastha: nice catch, will update

.github/workflows/migration-tests.yml
```diff
@@ -0,0 +1,111 @@
+name: 'Migration Tests'
+
+on:
+  workflow_dispatch: {} # workflow can be run manually
+  workflow_call: # workflow runs after code-health
+    inputs:
+      parent-event-name:
+        required: true
+        type: string
+  pull_request: # you can run a specic job in your PR using GitHub labels
```
marcosuma: is this comment needed?

.github/workflows/migration-tests.yml
marcosuma: perhaps we could start fixing this file and then a follow-up to change the others?

marcosuma (approved):
minor comments, but LGTM!

.github/workflows/migration-tests.yml
maastha: @lantoli I couldn't find a way to make this reusable or use a function, is there any example or reference you could share?

lantoli: I'm not sure about the best way, but for instance what about creating a variable in the workflow and use the variable? something like:

on: ...

env:
   hasToTrigger: ${{ needs.change-detection.outputs.config == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc-migration' || github.event.label.name == 'run-testacc-config-migration' || inputs.parent-event-name == 'release' }}

job...

and use:

if: ${{ env.hasToTrigger }}

lantoli: or use it for the common part and then:
if: ${{ env.hasToTrigger }} || specific conditions here...

maastha: The thing is we need to pass arguments here, because github.event.label.name will be different, so we might be able to do this partially but not for entire condition

AgustinBettati (approved):
great idea having the version update automatically, gives proper value to the migration tests

.github/workflows/migration-tests.yml
maastha: since this doesn't hold a static value, rather holds a condition, i don't believe this is possible to do via a variable as well

mongodbatlas/fw_resource_mongodbatlas_database_user_migration_test.go
maastha: updated

.github/workflows/migration-tests.yml
lantoli: i suppose it's possible while you use variables that are know before starting execution

.github/workflows/migration-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+  workflow_call: # workflow runs after code-health
+    inputs:
+      parent-event-name:
+        required: true
+        type: string
+  pull_request:
+    types: [ labeled ]
+
+jobs: 
+  set-output:
```
maastha, line 14: @lantoli tried with another step: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1565/files#diff-9e63d3ed50a83cef3c77bb7447f2858374bc51e6dd94ac7f65b3420205c70ea6
This is an example PR running this workflow: https://github.com/maastha/terraform-provider-mongodbatlas/pull/1 

Let me know if this approach makes sense, not sure if this is possible with a variable

.github/workflows/migration-tests.yml
maastha: Addressed, see comment https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1565#discussion_r1372020123

.github/workflows/migration-tests.yml
maastha: removed

.github/workflows/migration-tests.yml
lantoli: that's great, thanks!

marcosuma:
I just realised something now: why do we call them `migration`? shouldn't it be `backwardCompatibiltyTest`?

lantoli:
> I just realised something now: why do we call them `migration`? shouldn't it be `backwardCompatibiltyTest`?

@marcosuma I think it's Hashicorp naming

PR #1572, opened 2023-10-30
fix: overrides to attribute behavior for resource elems.

marcosuma:
## Description

In order to preserve backward compatibility to old TF versions (< 1.1.5), I am adding `ConfigMode: schema.SchemaConfigModeAttr` to `advanced_configuration` and `bi_connector_config` that fundamentally enables us to use `Computed: true` to nested blocks. Read below for the official explanation of `ConfigMode`.

I'll be short here but you can read the whole context in the help below.
`advanced_configuration` and `bi_connector_config` are two **nested block attributes** that have the characteristics of being optional but at the same time Atlas API returns a default value for them if not specified. This creates errors in our provider for TF versions <1.1.5 since our v1.12.0 provider version.
In particular the error says that `computed:true` cannot be used for nested blocks. If you remove the `computed:true` at the end of `tf apply` you'll get another error saying that `block count changed from 0 to 1` _(it is referring to the fact that the read function calling the admin API is returning a default value hence it's not matching the initial plan...)_.

To fix that, I am simply declaring those nested block as attributes which means we can use `Computed:true` for them.

**I've tested it with v1.1.4 and v1.5.6** and it works.

Now basically our customer can just use our latest provider version with their original TF version and their `terraform apply` should succeed (they don't need to change anything).
 

From official description of `schema.go`:

```
	// ConfigMode allows for overriding the default behaviors for mapping
	// schema entries onto configuration constructs.
	//
	// By default, the Elem field is used to choose whether a particular
	// schema is represented in configuration as an attribute or as a nested
	// block; if Elem is a *schema.Resource then it's a block and it's an
	// attribute otherwise.
	//
	// If Elem is *schema.Resource then setting ConfigMode to
	// SchemaConfigModeAttr will force it to be represented in configuration
	// as an attribute, which means that the Computed flag can be used to
	// provide default elements when the argument isn't set at all, while still
	// allowing the user to force zero elements by explicitly assigning an
	// empty list.
	//
	// When Computed is set without Optional, the attribute is not settable
	// in configuration at all and so SchemaConfigModeAttr is the automatic
	// behavior, and SchemaConfigModeBlock is not permitted.
```

Note: this is also [extensively used in other repos](https://github.com/search?q=schema.SchemaConfigModeAttr%2C&type=code)

Link to any related issue(s): https://jira.mongodb.org/browse/HELP-51509

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_mongodbatlas_cluster.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster.go
@@ -88,10 +88,11 @@ func resourceMongoDBAtlasCluster() *schema.Resource {
 				Description: "Flag that indicates whether to retain backup snapshots for the deleted dedicated cluster",
 			},
 			"bi_connector_config": {
-				Type:     schema.TypeList,
-				Optional: true,
-				Computed: true,
-				MaxItems: 1,
+				Type:       schema.TypeList,
+				Optional:   true,
+				ConfigMode: schema.SchemaConfigModeAttr,
+				Computed:   true,
+				MaxItems:   1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"enabled": {
@@ -1606,10 +1607,11 @@ func isEqualProviderAutoScalingMaxInstanceSize(k, old, newStr string, d *schema.
 
 func clusterAdvancedConfigurationSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:     schema.TypeList,
-		Optional: true,
-		Computed: true,
-		MaxItems: 1,
+		Type:       schema.TypeList,
+		Optional:   true,
+		Computed:   true,
+		ConfigMode: schema.SchemaConfigModeAttr,
+		MaxItems:   1,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"default_read_concern": {
--- a/mongodbatlas/resource_mongodbatlas_cluster_test.go
+++ b/mongodbatlas/resource_mongodbatlas_cluster_test.go
@@ -10,7 +10,9 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/mongodb/terraform-provider-mongodbatlas/mongodbatlas/testutils"
 	"github.com/mwielbut/pointy"
 	matlas "go.mongodb.org/atlas/mongodbatlas"
 )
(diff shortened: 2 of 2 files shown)


mongodbatlas/resource_mongodbatlas_cluster.go
```diff
@@ -1606,10 +1607,11 @@ func isEqualProviderAutoScalingMaxInstanceSize(k, old, newStr string, d *schema.
...
 func clusterAdvancedConfigurationSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:     schema.TypeList,
-		Optional: true,
-		Computed: true,
-		MaxItems: 1,
+		Type:       schema.TypeList,
+		Optional:   true,
+		Computed:   true,
+		ConfigMode: schema.SchemaConfigModeAttr,
```
AgustinBettati, line 1613: very interesting change, nice finding!
Did you notice any change in the structure of the terraform state when having this change from blocks to attributes? To make sure this change is not breaking, would it be worth adding a migration test case that creates the terraform state with previous version and reads it with the latest one?

marcosuma: good point, let me work on that - thanks for the feedback

lantoli: nice, isn't this happening in other places?

marcosuma: > this happening

what is "this"?

lantoli: other nested attributes that are computed or that might cause similar issues

marcosuma: oh I get it: yes, it might be possible there are other nested blocks that have "computed: true". It might be worth it a follow-up action once we see the success of this.

marcosuma: @AgustinBettati my newly created test is passing `--- PASS: TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_maintainsBackwardCompatibility (1033.39s)`

Now just waiting for the overall `cluster` test to succeed.

mongodbatlas/resource_mongodbatlas_cluster_test.go
```diff
@@ -1410,6 +1412,49 @@ func TestAccClusterRSCluster_basicAWS_PausedToUnpaused(t *testing.T) {
...
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheckBasic(t) },
+		CheckDestroy: testAccCheckMongoDBAtlasClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"mongodbatlas": {
+						VersionConstraint: "1.11.0",
```
AgustinBettati, line 1432: wondering if we want to respect the ```TestAccMigration*``` prefix convention, and add a comment specifying why we leave a fixed version here vs using `lastVersionConstraint` like in our other migration tests (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1565). I understand we want to test specifically the upgrade from 1.11.x due to terraform framework changes in the 1.12.0 release and ensure this schema change has no impact.

marcosuma: I would probably defer that to @maastha since her change hasn't still been deployed. @maastha  you can take a note about this test and include it as a migration test. I left a comment in your PR though that I am not sure why this is called "migration.

@AgustinBettati if you are ok let's leave it like that, I think the name of the test should be enough self-explanatory

AgustinBettati: Yes no need to block this PR, the test as is serving its purpose.

AgustinBettati (approved):

PR #1596, opened 2023-11-05
fix: Fixes Update SDK GitHub action

lantoli:
## Description

Jira ticket: [INTMDB-1252](https://jira.mongodb.org/browse/INTMDB-1252)

Fixes Update SDK GitHub action. 

- Uses a token to create the PR so CI checks are run
- Removes file filters to include all changed files.

Example of PR created: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1599

## Type of change:

## Required Checklist:


.github/workflows/update-sdk.yml
```diff
@@ -13,23 +13,18 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
-      - name: update
-        run: 	| 
-          go install github.com/icholy/gomajor@latest 
-          make update-atlas-sdk
+      - name: Update files
+        run:  make tools update-atlas-sdk
```
wtrocki, line 17: [Nit] Change adds to the build time. Ok to leave it.

.github/workflows/update-sdk.yml
```diff
@@ -13,23 +13,18 @@ jobs:
...
       - name: Verify Changed files
-        uses: tj-actions/verify-changed-files@6d688963a73d28584e163b6f62cf927a282c4d11
+        uses: tj-actions/verify-changed-files@v16
```
wtrocki: We do pin actions to commit. More info:
https://medium.com/@wtr/best-patterns-building-github-actions-workflows-6b03655223ca

```suggestion
        uses: tj-actions/verify-changed-files@6d688963a73d28584e163b6f62cf927a282c4d11
```

wtrocki (approved):

.github/workflows/update-sdk.yml
lantoli: i know, but just a few seconds as a tradeoff for consistency and remove duplication

.github/workflows/update-sdk.yml
lantoli: I see

wtrocki:
We need to verify if action changes work. Best would be to trigger it [from branch](https://medium.com/@wtr/best-patterns-building-github-actions-workflows-6b03655223ca)

lantoli:
> We need to verify if action changes work. Best would be to trigger it [from branch](https://medium.com/@wtr/best-patterns-building-github-actions-workflows-6b03655223ca)

@wtrocki I did a few tests from this branch

scripts/update-sdk-preview.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
wtrocki:
```suggestion
set -euo pipefail

echo "Script will update to dev preview. Note: you need to have latest major version in order to be able to use it"
```

wtrocki:
Verified it myself. All good. Added extra note as script might not work for all the cases

lantoli:
> Verified it myself. All good. Added extra note as script might not work for all the cases

@wtrocki added your note. that's also why I added this note: "If no files were changed you may need first to update to the latest SDK using the 'Update SDK' GitHub Action"

andreaangiolillo (changes requested):
Why do we want to use SDK preview? Doesn't SDK preview use cloud-dev?

GNUmakefile
```diff
@@ -138,3 +139,7 @@ link-git-hooks: ## Install git hooks
 .PHONY: update-atlas-sdk
 update-atlas-sdk: ## Update the atlas-sdk dependency
 	./scripts/update-sdk.sh
+
+.PHONY: try-atlas-sdk-preview
+try-atlas-sdk-preview: ## Use SDK Preview, DONT MERGE
```
marcosuma: If we really have to push this, this needs some sort of automated safety mechanism that would impede a push in master where we're using the sdk preview version. I can see how with a distraction we can incur in unexpected merges.

marcosuma: in the meantime if your purpose is to test search index you can unblock yourself by testing this locally and syncing with the owning team on when is the plan of releasing these APIs in prod. I don't think we should tweak things and build workarounds to project timeline obstacles.

wtrocki:
> Why do we want to use SDK preview? Doesn't SDK preview use cloud-dev?

The idea is to do development of the feature using a preview before the change is available. 

EDIT: closed by mistake.

GNUmakefile
lantoli: @andreaangiolillo @marcosuma as @wtrocki is saying, this is to make it easier for people what wants/need to use the SDK Preview.   
In this way people don't need to do it manually. what problems do you see keeping this make target in case somebody wants to use it?  (there can always be a distraction and merge sdk preview changes even if this make target doesn't exist)

andreaangiolillo: > what problems do you see keeping this make target in case somebody wants to use it?

We are adding additional code to maintain for something that I think we should not use. If we need to use the SDK-preview, a better way which does not involve updating all the repo files is to use `replace =>` with the local SDK pointing to the SDK preview PR.

lantoli: only go.mod and go.sum are modified when using the SDK Preview

andreaangiolillo: Yes, it is true. One additional question: What would happen if the preview PR is not available on the SDK channel when running the script? Would the script fail, would the script run the github action on the atlas SDK?

I think you can move the preview logic to a new  PR so that we can keep discussing it there and unblock this PR

andreaangiolillo (approved):
LGTM. Thanks for the discussions

PR #1605, opened 2023-11-07
feat: Adds type and fields attributes in resource and data sources for search_index

lantoli:
## Description

Jira ticket: [INTMDB-1260](https://jira.mongodb.org/browse/INTMDB-1260)

Adds type and fields attributes in resource and data sources for search_index.

## Type of change:

- [] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.

## Required Checklist:


mongodbatlas/resource_mongodbatlas_search_index.go
```diff
@@ -347,15 +385,24 @@ func resourceMongoDBAtlasSearchIndexCreate(ctx context.Context, d *schema.Resour
 	projectID := d.Get("project_id").(string)
 	clusterName := d.Get("cluster_name").(string)
 	dynamic := d.Get("mappings_dynamic").(bool)
+	var mappings *admin.ApiAtlasFTSMappings
+	indexType := d.Get("type").(string)
```
ciprian-tibulca, line 399: looking only at this PR, I'm not sure where the default value is set ("search")
should we add this?
```suggestion
	indexType := d.Get("type").(string)
	if indexType == nil {
		indexType = "search" // move this to a const like vectorSearch
	}
```

lantoli: in TF if not stated explicitly the default values is the zero value, "" (empty string) in this case.

we're keeping "" to avoid changes in  Terraform plans for current resources, so both "" and "search" are correct.

If we change this for all resources they'll see change plans in search-indexe resources they didn't change:   "" -> "search"

website/docs/r/search_index.html.markdown
```diff
@@ -28,7 +28,25 @@ resource "mongodbatlas_search_index" "test-basic-search-index" {
 }
 ```
 
-### Advanced (with custom analyzers)
+### Basic vector index
+```terraform
+resource "mongodbatlas_search_index" "test-basic-search-vector" {
```
lantoli, line 33: @Zuhairahmed do we want a different example?

mongodbatlas/resource_mongodbatlas_search_index.go
ciprian-tibulca: `If we change this for all resources they'll see change plans in search-index resources they didn't change: "" -> "search"`

Just to double-check, the already created indexes will always have type=""? Asking because, if the API starts returning type="search" for these old search indexes, TF will keep trying to reconcile the state and try to make type="" (as its old config says), or am I wrong?

e.g.:
* 2023-01-01 create search index without a type (not available at that moment)
* 2023-01-01 the GET API returns the search index without a type field
...
* 2023-11-15 the GET API starts returning the same search index with type="search" (instead of "") `// don't know if this will happen but it's important to check`
* 2023-11-16 TF detects a change (type in config is not set, so equals "", while the API says it's "search") even if you apply the plan (where the type is still "")

mongodbatlas/resource_mongodbatlas_search_index.go
```diff
@@ -272,6 +293,10 @@ func resourceMongoDBAtlasSearchIndexRead(ctx context.Context, d *schema.Resource
 		return diag.Errorf("error setting `index_id` for search index (%s): %s", d.Id(), err)
 	}
 
+	if err := d.Set("type", searchIndex.Type); err != nil {
```
AgustinBettati, line 308: Does the API always responds with a value for the type field? Would consider marking as computed or defining a default value in the schema to avoid inconsistencies in the state.

kyuan-mongodb (approved):
LGTM!

mongodbatlas/resource_mongodbatlas_search_index.go
lantoli: the field doesn't exist in the state (terraform.tfstate) so when state is refreshed (e.g. before plan or apply) the state is updated to have:  type = ""    .   But this doesn't raise a plan change to the user.

if the SDK will start to return "search" then a plan change will be shown to the user.

mongodbatlas/resource_mongodbatlas_search_index.go
lantoli: if the index was created withouth type, it won't return type.   API responds same as in creation so I won't use computed.
as it's defined as optional, it has default value = "" (zero value for strings), this is what we want.  When refreshing the state the "type" attribute is added to the state with "" but no plan change is shown to the users.

mongodbatlas/data_source_mongodbatlas_search_index.go
```diff
@@ -140,21 +153,34 @@ func dataSourceMongoDBAtlasSearchIndexRead(ctx context.Context, d *schema.Resour
...
-		return diag.Errorf("error setting `mappings_dynamic` for search index (%s): %s", d.Id(), err)
-	}
-
 	if err := d.Set("synonyms", flattenSearchIndexSynonyms(searchIndex.Synonyms)); err != nil {
 		return diag.Errorf("error setting `synonyms` for search index (%s): %s", d.Id(), err)
 	}
 
-	if len(searchIndex.Mappings.Fields) > 0 {
-		searchIndexMappingFields, err := marshalSearchIndex(searchIndex.Mappings.Fields)
+	if searchIndex.Mappings != nil {
```
marcosuma, line 160: we are changing this if statement - do you foresee any potential breaking change?

marcosuma (commented):
it would be good to add migration and acceptance tests to this change

mongodbatlas/data_source_mongodbatlas_search_indexes.go
```diff
@@ -100,25 +100,29 @@ func flattenSearchIndexes(searchIndexes []admin.ClusterSearchIndex, projectID, c
 
 	for i := range searchIndexes {
 		searchIndexesMap[i] = map[string]any{
-			"project_id":       projectID,
-			"cluster_name":     clusterName,
-			"analyzer":         searchIndexes[i].Analyzer,
-			"collection_name":  searchIndexes[i].CollectionName,
-			"database":         searchIndexes[i].Database,
-			"index_id":         searchIndexes[i].IndexID,
-			"mappings_dynamic": searchIndexes[i].Mappings.Dynamic,
```
marcosuma, line 109: why are we removing mapping_dynamics here and setting it separately below? (just thinking on behaviour changes...)

lantoli: because Dynamics is inside Mappings and Mappings doesn't exist for vector, it's the same behaviour but making sure Mappings exists

mongodbatlas/data_source_mongodbatlas_search_index.go
lantoli: I tried different tests (automatic and manual) and couldn't find any problem. In current indexes searchInde.Mapping is not nil, this statement is false only for (new) vector indexes

lantoli:
> it would be good to add migration and acceptance tests to this change

@marcosuma I added an acceptance test and planning to add more. About migration (and unit) I think the idea is doing it in the new FW resources, this is not migrated yet.

I did manual tests that showed a change in TF state (from unexisting Type and Fields to Type="", Fields=nil) but didn't trigger a plan change to the user.

mongodbatlas/resource_mongodbatlas_search_index_test.go
```diff
@@ -156,6 +175,45 @@ func TestAccSearchIndexRS_importBasic(t *testing.T) {
 	})
 }
 
+func TestAccSearchIndexRS_withVector(t *testing.T) {
```
lantoli, line 229: @marcosuma FYI basic acceptance test, planning to add more

website/docs/r/search_index.html.markdown
Zuhairahmed: @lantoli  suggest for time being we leverage example that Atlas Search team has provided here for `fields` parameter: https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/#std-label-fts-data-types-knn-vector

Zuhairahmed: "fields": {
      "plot_embedding": {
        "type": "knnVector",
        "dimensions": 1536,
        "similarity": "euclidean"
      }
    }

Zuhairahmed: also believe you are missing index `name` parameter which is required: 
image

to prevent confusion with `cluster_name` and `collection_name`, suggest we call this `index_name`

lantoli: thanks. your example won't work directly but adapting the idea.  current "name" attribute is the index name, that doesn't change.

Zuhairahmed: SGTM

mongodbatlas/resource_mongodbatlas_search_index.go
marcosuma: suggest to have this question/answer represented with a test, not necessarily blocking this PR (unless it's trivial creating it).

marcosuma (approved):

mongodbatlas/resource_mongodbatlas_search_index_test.go
```diff
@@ -2,53 +2,98 @@ package mongodbatlas
...
+					resource.TestCheckResourceAttr(datasourceName, "name", indexName),
+					resource.TestCheckResourceAttr(datasourceName, "type", ""),
+					resource.TestCheckResourceAttrSet(datasourceName, "index_id"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccSearchIndexRS_withSearchType(t *testing.T) {
```
lantoli, line 63: @marcosuma FYI this tests checks that you can explicitly pass "search" as index type.

mongodbatlas/resource_mongodbatlas_search_index_test.go
```diff
@@ -2,53 +2,98 @@ package mongodbatlas
...
+
+					resource.TestCheckResourceAttr(datasourceName, "name", indexName),
+					resource.TestCheckResourceAttr(datasourceName, "project_id", projectID),
+					resource.TestCheckResourceAttr(datasourceName, "cluster_name", clusterName),
+					resource.TestCheckResourceAttr(datasourceName, "database", databaseName),
+					resource.TestCheckResourceAttr(datasourceName, "collection_name", collectionName),
+					resource.TestCheckResourceAttr(datasourceName, "mappings_dynamic", "true"),
+					resource.TestCheckResourceAttr(datasourceName, "search_analyzer", searchAnalyzer),
+					resource.TestCheckResourceAttr(datasourceName, "name", indexName),
+					resource.TestCheckResourceAttr(datasourceName, "type", ""),
```
lantoli, line 55: @marcosuma FYI this tests checks that type is empty when created in that way

mongodbatlas/resource_mongodbatlas_search_index.go
```diff
@@ -115,6 +119,15 @@ func returnSearchIndexSchema() map[string]*schema.Schema {
 			Type:     schema.TypeBool,
 			Optional: true,
 		},
+		"type": {
+			Type:     schema.TypeString,
+			Optional: true,
+		},
+		"fields": {
```
AgustinBettati, line 126: have we considered adding a conflictsWith validation to ensure vectorSearch attributes are not defined together with regular search attributes?

AgustinBettati (approved):
LGTM, just a small detail

mongodbatlas/resource_mongodbatlas_search_index.go
lantoli: at the end I preferred that all those checks are done in the server. If we do them in client-side there is some risk that both logics differ now or in the future.

PR #1610, opened 2023-11-08
fix: Fixes search index acceptance tests flow

lantoli:
## Description

Jira ticket: [INTMDB-1271](https://jira.mongodb.org/browse/INTMDB-1271)

Fixes search index acceptance tests flow.

## Type of change:

## Required Checklist:


.github/workflows/acceptance-tests.yml
```diff
@@ -467,4 +469,31 @@ jobs:
           PARALLEL_GO_TEST: 20
           CI: true
           TEST_REGEX: "^TestAccSTSAssumeRole"
-        run: make testacc
\ No newline at end of file
+        run: make testacc
+
+  search_index:
+    needs: [ change-detection ]
+    if: ${{ needs.change-detection.outputs.search_index == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-cluster' || inputs.parent-event-name == 'release' }}
```
andreaangiolillo: `github.event.label.name` allows you to run only this job if you add the specified label to the PR

```suggestion
    if: ${{ needs.change-detection.outputs.search_index == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-search-index' || inputs.parent-event-name == 'release' }}
```

mongodbatlas/resource_mongodbatlas_search_index_test.go
```diff
@@ -21,14 +22,14 @@ func TestAccClusterRSSearchIndex_basic(t *testing.T) {
...
 	)
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:                 func() { testAccPreCheckBasic(t) },
+		PreCheck:                 func() { testAccPreCheckSearchIndex(t) },
 		ProtoV6ProviderFactories: testAccProviderV6Factories,
 		CheckDestroy:             testAccCheckMongoDBAtlasSearchIndexDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccMongoDBAtlasSearchIndexConfig(projectID, clusterName),
+				Config: testAccSearchIndexConfig(t, projectID, clusterName),
```
andreaangiolillo: Can we create the project as part of the test? One problem with this approach is that it may fail if another PR try to run it at the same time (they are using the same project)

andreaangiolillo (commented):
Left a few questions

mongodbatlas/resource_mongodbatlas_search_index_test.go
lantoli: no, because the beta feature we were talking, we have to use an specific project

andreaangiolillo: We could decide to not run it as part of the acceptance test until the feature is released in prod

.github/workflows/acceptance-tests.yml
lantoli: oh, thanks, will change

mongodbatlas/resource_mongodbatlas_search_index_test.go
lantoli: that's an option but to be honest I prefer to leave it as it is so we can have more confidence on the vector search changes, later we can change it to create the project in the test.

andreaangiolillo: Leaving this thread open to see what other people think about it. Personally, I don't like having tests that will likely be red, this adds additional cost as the on-call person, who may not have the context about the feature flag, has to put time into understanding why they are failing

lantoli: i don't see why they should be red frequently because of this, all resources have randomized names so it should be fine

andreaangiolillo: my understanding is that the test will fail if there are two PRs running it because they are using the same Atlas project. Is my understanding correct?

lantoli: they won't fail because they use random cluster names.

anyway I haven't changed this, current tests are using an existing project. If we want to change the test to create a project it's a change to the existing tests and I think that can belong to a different PR.

here I've fixed the flow so test are executed, right not search index tests are not being executed in GH.

andreaangiolillo: Thanks for the clarification!

andreaangiolillo: > If we want to change the test to create a project it's a change to the existing tests and I think that can belong to a different PR.

The fact that the tests are not running in GH makes me think that they may fail if run concurrently. We can always remove them if I am correct

andreaangiolillo (approved):
LGTM. Thanks for the discussion

PR #1618, opened 2023-11-12
chore: Refactors Acceptance Tests flow logic

lantoli:
## Description

Refactors Acceptance Tests flow logic so info about a resource test keeps together.

### Pros
- Information about a test keeps together like the test regular expression and files to detect changes. No need to change/add in 3 different places for testing a resource.
- Reduce duplicated code

### Cons
- It's harder to see what tests were skipped
- All jobs are run and their containers created even if not needed, although the time-consuming steps are skipped if not needed.

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/actions/acceptance-test-action/action.yml
+++ b/.github/actions/acceptance-test-action/action.yml
@@ -27,9 +27,5 @@ runs:
       if: ${{ inputs.forceTestAcc || steps.filter.outputs.src == 'true'}}
       env:
         TEST_REGEX: ${{ inputs.testRegex }}
-        TF_LOG: ${{ vars.LOG_LEVEL }}
-        SKIP_TEST_EXTERNAL_CREDENTIALS: ${{ vars.SKIP_TEST_EXTERNAL_CREDENTIALS }}
-        ACCTEST_TIMEOUT: ${{ vars.ACCTEST_TIMEOUT }}
-        PARALLEL_GO_TEST: "20"
       shell: bash
       run: make testacc
--- a/.github/actions/acceptance-test-action/action.yml
+++ b/.github/actions/acceptance-test-action/action.yml
@@ -14,9 +14,7 @@ runs:
     - uses: dorny/paths-filter@v2
       id: filter
       with:
-        filters: |
-          src: 
-            ${{ inputs.filters }}
+        filters: ${{ inputs.filters }}
     - name: Print hello
       shell: bash
       run: |
--- a/.github/actions/acceptance-test-action/action.yml
+++ b/.github/actions/acceptance-test-action/action.yml
@@ -15,7 +15,8 @@ runs:
       id: filter
       with:
         filters: |
-          src: ${{ inputs.filters }}
+          src: 
+            ${{ inputs.filters }}
     - name: Print hello
       shell: bash
       run: |
--- a/.github/actions/acceptance-test-action/action.yml
+++ b/.github/actions/acceptance-test-action/action.yml
@@ -19,7 +19,6 @@ runs:
         go-version-file: "go.mod"
     - name: Acceptance Tests
       shell: bash
-      working-directory: mongodbatlas
       run: make testacc
     - name: Print bye
       shell: bash
--- a/.github/actions/acceptance-test-action/action.yml
+++ b/.github/actions/acceptance-test-action/action.yml
@@ -8,6 +8,9 @@ inputs:
   filters:
     description: "files to detect if tests should be run"
     required: true
+  testRegex:
+    description: "regular expression for test methods"
+    required: true
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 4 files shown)


mongodbatlas/resource_mongodbatlas_search_index.go
```diff
@@ -485,7 +485,7 @@ func unmarshalSearchIndexAnalyzersFields(mappingString string) []admin.ApiAtlasF
 	}
 	var fields []admin.ApiAtlasFTSAnalyzers
 	if err := json.Unmarshal([]byte(mappingString), &fields); err != nil {
-		log.Printf("[ERROR] cannot unmarshal search index mapping fields: %v", err)
+		log.Printf("[ERROR] can not unmarshal search index mapping fields: %v", err)
```
marcosuma: how important is this change? should we just leave it like that? because eventually I don't think this makes it "better english" 😄

.github/workflows/acceptance-tests.yml
```diff
@@ -12,7 +12,53 @@ on:
   pull_request: # you can run a specic job in your PR using GitHub labels
     types: [ labeled ]
 
+env:
+    FORCE_TEST_ACC: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || inputs.parent-event-name == 'release' }}
+    TF_LOG: ${{ vars.LOG_LEVEL }}
+    SKIP_TEST_EXTERNAL_CREDENTIALS: ${{ vars.SKIP_TEST_EXTERNAL_CREDENTIALS }}
+    ACCTEST_TIMEOUT: ${{ vars.ACCTEST_TIMEOUT }}
+    PARALLEL_GO_TEST: "20"
```
marcosuma: are these "global" env?

.github/workflows/acceptance-tests.yml
```diff
@@ -12,7 +12,53 @@ on:
...
 jobs:  
+  search_index:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
andreaangiolillo, line 26: Could we move `actions/checkout@v4` to `github/actions/acceptance-test-action/action.yml`??

.github/actions/acceptance-test-action/action.yml
```diff
@@ -0,0 +1,31 @@
...
+  testRegex:
+    description: "regular expression for test methods"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: ${{ inputs.filters }}
```
andreaangiolillo: Something to keep in mind with this approach is that we call `dorny/paths-filter@v2` for every job. Before we were calling `dorny/paths-filter@v2` once, and then use the result of that step to run the correct jobs
Screenshot 2023-11-13 at 12 17 12.

One cons of this approach is that we will always create a run job to run every acceptance test, the job will then terminate once `dorny/paths-filter@v2` is run. Right now the job is skipped instead. (Could you confirm this?)

I would suggest you to add pros/cons of this approach in the description of your PR to help the discussion


(not strong opinion here I just want to highlight this new behaviour)

andreaangiolillo (commented):
I would add pros/cons of the proposed approach in the description of the PR to help the discussion

mongodbatlas/resource_mongodbatlas_search_index.go
lantoli: this was to test that file change detection was working fine :-) i'll revert it

.github/workflows/acceptance-tests.yml
lantoli: they are passed to all jobs

.github/workflows/acceptance-tests.yml
lantoli: no, that was my first try but it can't find the action file

.github/actions/acceptance-test-action/action.yml
lantoli: you're right the jobs will be created, but the tests steps will be skipped if not need to run so at the end I think it'll be equivalent.   Also it's true that we call path-filters multiple times but I suppose this action internally has to do the code for each filter group so at then end it does a similar job.

one pro of current version is that it's easier to see what acceptance tests were skipped.

i can change the description to add pros and cons

mongodbatlas/resource_mongodbatlas_search_index.go
lantoli: here I changed a file, for generic I added PR label [run-testacc-generic](https://github.com/mongodb/terraform-provider-mongodbatlas/labels/run-testacc-generic) so i test different ways to trigger the tests

.github/workflows/acceptance-tests.yml
marcosuma: are we sure this is a good idea? Here is where I would like to see all environment variables as close as possible to the job/step so I don't have to go up and down in the file to see the context.

lantoli: we can decide what we want to have closer and which one we want for all. 
for instances in the beginning I was going to delete PARALLEL_GO_TEST and ACCTEST_TIMEOUT because they have their default values, but in this way you can change the var instead of having to change code.

about FORCE_TEST_ACC I think it's a good idea to have it here as this doesn't depend on a specific test. For example we want to run tests always if it's triggered manually (workflow_dispatch) or if it's a release

lantoli: i don't see a big problem in having common vars here. We can move them later if we realize that we want different values for different tests

PR #1621, opened 2023-11-14
feat: `mongodbatlas_search_deployment` data source with acceptance tests and docs

AgustinBettati:
## Description

Jira ticket: [INTMDB-1273](https://jira.mongodb.org/browse/INTMDB-1273)

Currently open as draft pull request as we have to make use of a dev preview of the Atlas SDK.

This PR includes:

- Implementation of `mongodbatlas_search_deployment` data source
- acceptance tests
- documentation

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/fw_data_source_mongodbatlas_search_deployment.go
+++ b/mongodbatlas/fw_data_source_mongodbatlas_search_deployment.go
@@ -1,0 +1,99 @@
+package mongodbatlas
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var _ datasource.DataSource = &SearchDeploymentDS{}
+var _ datasource.DataSourceWithConfigure = &SearchDeploymentDS{}
+
+func NewSearchDeploymentDS() datasource.DataSource {
+	return &SearchDeploymentDS{
+		DSCommon: DSCommon{
+			dataSourceName: searchDeploymentName,
+		},
+	}
+}
+
+type tfSearchDeploymentDSModel struct {
+	ID          types.String `tfsdk:"id"`
+	ClusterName types.String `tfsdk:"cluster_name"`
+	ProjectID   types.String `tfsdk:"project_id"`
+	Specs       types.List   `tfsdk:"specs"`
+	StateName   types.String `tfsdk:"state_name"`
+}
+
+type SearchDeploymentDS struct {
+	DSCommon
+}
+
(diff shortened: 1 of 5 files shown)


mongodbatlas/fw_data_source_mongodbatlas_search_deployment.go
```diff
@@ -0,0 +1,99 @@
...
+	ProjectID   types.String `tfsdk:"project_id"`
+	Specs       types.List   `tfsdk:"specs"`
+	StateName   types.String `tfsdk:"state_name"`
+}
+
+type SearchDeploymentDS struct {
+	DSCommon
+}
+
+func (d *SearchDeploymentDS) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
marcosuma, line 34: this is more of a generic question: in the context of a data source I don't understand the concept of `Computed`: aren't they all computed?

On the other side I'd imagine `Required` means that if we don't fill it up with the backend response terraform would raise an error, but why do we really need to set it as required? Couldn't we just leave it as Optional?

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -31,12 +31,13 @@ var _ resource.ResourceWithImportState = &SearchDeploymentRS{}
 
 const (
 	searchDeploymentDoesNotExistsError = "ATLAS_FTS_DEPLOYMENT_DOES_NOT_EXIST"
+	searchDeploymentName               = "search_deployment"
```
marcosuma, line 34: nit: this looks like a separate PR? no need to do it now, but consider it for the next time

mongodbatlas/fw_data_source_mongodbatlas_search_deployment.go
AgustinBettati: > in the context of a data source, aren't they all computed?

Not exactly. When using a data source there are some attributes that need to be provided by the user in order to identify the resource that will be fetched. In this case the resource is identified through the `project_id` and `cluster_name`, which is why they are marked as required.
You will see this is also clarified in the documentation of the data source, separating the concept of arguments (attributes that have to be defined), and then an attributes sections for values that are exported (or as we call computed).

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
AgustinBettati: Yes true. Tried to keep this PR as clean as possible with respect to the resource implementation, but you usually some pending details always get adjusted when implementing the data source.

AgustinBettati:
merged in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1633

PR #1631, opened 2023-11-15
chore: Reduces duplication in GH action for acceptance tests

lantoli:
## Description

Jira ticket: [INTMDB-1295](https://jira.mongodb.org/browse/INTMDB-1295)

Reduces duplication in GH action for acceptance tests.

## Type of change:

- [] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.

## Required Checklist:

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -122,10 +122,8 @@ jobs:
     if: ${{ needs.change-detection.outputs.cluster_outage_simulation == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-cluster-outage-simulation' || inputs.parent-event-name == 'release'  }}
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Set up Go
-        uses: actions/setup-go@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
       - name: Acceptance Tests
@@ -142,10 +140,8 @@ jobs:
     if: ${{ needs.change-detection.outputs.advanced_cluster == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-advanced-cluster' || inputs.parent-event-name == 'release' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Set up Go
-        uses: actions/setup-go@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
       - name: Acceptance Tests
@@ -162,10 +158,8 @@ jobs:
     if: ${{ needs.change-detection.outputs.cluster == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-cluster' || inputs.parent-event-name == 'release' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Set up Go
-        uses: actions/setup-go@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
       - name: Acceptance Tests
@@ -182,10 +176,8 @@ jobs:
     if: ${{ needs.change-detection.outputs.generic == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-generic' || inputs.parent-event-name == 'release' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Set up Go
-        uses: actions/setup-go@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
       - name: Acceptance Tests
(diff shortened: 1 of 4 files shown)


.github/workflows/migration-tests.yml
```diff
@@ -21,10 +17,10 @@ jobs:
       config: ${{ steps.filter.outputs.config }}
       advanced_cluster: ${{ steps.filter.outputs.advanced_cluster }}
       backup_online_archive: ${{ steps.filter.outputs.backup_online_archive }}
-      shouldTriggerResourceTest: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || inputs.parent-event-name == 'release' }}
+      shouldTriggerResourceTest: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' }}
```
lantoli, line 20: leftovers

.github/workflows/acceptance-tests.yml
```diff
@@ -5,38 +5,41 @@ on:
...
   pull_request: # you can run a specic job in your PR using GitHub labels
     types: [ labeled ]
 
+env:
+  ALWAYS_RUN: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' }}
+  CI: true
+  PARALLEL_GO_TEST: 20
+  ACCTEST_TIMEOUT: ${{ vars.ACCTEST_TIMEOUT }}
+  SKIP_TEST_EXTERNAL_CREDENTIALS: ${{ vars.SKIP_TEST_EXTERNAL_CREDENTIALS }}
+  TF_LOG: ${{ vars.LOG_LEVEL }}
```
marcosuma, line 17: I left this comment in the other PR but I am sorry I missed your eventual response. I have a preference in not having global env variables because it goes in the opposite direction of what you want to do here, which IIUC is simplify and "isolate" the context of individual jobs.

I suggest we leave these at each job level, they will be repeated, yes, but verbosity is not necessarily bad

.github/workflows/acceptance-tests.yml
```diff
@@ -5,38 +5,41 @@ on:
...
-      federation: ${{ steps.filter.outputs.federation }}
-      backup_schedule: ${{ steps.filter.outputs.backup_schedule }}
-      project: ${{ steps.filter.outputs.project }}
-      serverless: ${{ steps.filter.outputs.serverless }}
-      network: ${{ steps.filter.outputs.network }}
-      config: ${{ steps.filter.outputs.config }}
-      assume_role: ${{ steps.filter.outputs.assume_role }}
-      event_trigger: ${{ steps.filter.outputs.event_trigger }}
-      search_index: ${{ steps.filter.outputs.search_index }}
+      cluster_outage_simulation: ${{ steps.filter.outputs.cluster_outage_simulation == 'true' || github.event.label.name == 'run-testacc-cluster-outage-simulation' || env.ALWAYS_RUN == 'true' }}
```
marcosuma, line 26: this is probably making it even less readable :)

lantoli: I agree it's less readable this line but I think it improves the other parts and overall it's better :-)

.github/workflows/acceptance-tests.yml
lantoli: the other PR is still open. Basically my argument is that it's ok to have common variables for things that are common and we don't want to change in each test. 
As an example we want to say CI: true for all tests run from here.   I don't see a benefit in having in each single tests, and there is the problem of forgetting it in some tests as you copy/paste.

also for a value that is used in all tests but one or a few you can decide to override in that test (just declaring the env var. again inside the test job) or remove entirely from the common part and define in each one.

.github/workflows/acceptance-tests.yml
lantoli: for example:
 if: ${{ needs.change-detection.outputs.cluster_outage_simulation == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || github.event.label.name == 'run-testacc-cluster-outage-simulation' || inputs.parent-event-name == 'release'  }}

becames:
    if: ${{ needs.check.outputs.cluster_outage_simulation == 'true' }}

lantoli: also for a value that is used in all tests but one or a few you can decide to override in that test (some declaring the env var. again) or remove entirely from the common part and define in each one.

.github/workflows/acceptance-tests.yml
AgustinBettati: I do see the value in extracting the common env variables, these will apply to all acceptance test jobs and remove noise from more specific configurations each one has. This can also help when adding new common env variables such as the terraform version used to run acceptance tests.
On the other hand having the criteria `ALWAYS_RUN` here is a little confusing, having it inlined should be okay.

lantoli: this is the best way I found to avoid having to repeat the expression in all places, but happy to do it in a different way

lantoli: @AgustinBettati very good point about TF version variables

andreaangiolillo (commented):
I am not sure if this PR is really simplifying the logic of our tests. 
Maybe you could update the description of your PR to highlight which logic these proposed changes are simplifying and why you think they are simpler now. 

Ideally, a person who is not familiar with github actions should be able to understand the logic and I am not sure these changes are a step in that direction.

.github/workflows/acceptance-tests.yml
marcosuma: this reminds me a lot of the best practices for testing.
I remember in Google there was a best practice where "variables, methods, context etc.etc" should be as close as possible (preferably inside) the test itself, so the reader of that (functional/unit/integ) test does not have to move in the file to understand what's going on. As a consequence you would see lots of duplication in the code because same variables, same init logic would be repeated instead of being separated. But this was preferred and recommended. And to me it made lots of sense.

This reminds me of that, we are trying to make this config file more organised at the price of "intuitiveness" and "time-to-understand". 

Said that, I won't push back too much on this - if you both like this approach then I am fine! :)

.github/workflows/acceptance-tests.yml
marcosuma: ok sounds good.

.github/workflows/migration-tests.yml
```diff
@@ -21,10 +17,10 @@ jobs:
       config: ${{ steps.filter.outputs.config }}
       advanced_cluster: ${{ steps.filter.outputs.advanced_cluster }}
       backup_online_archive: ${{ steps.filter.outputs.backup_online_archive }}
-      shouldTriggerResourceTest: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.event.label.name == 'run-testacc' || inputs.parent-event-name == 'release' }}
```
marcosuma, line 24: what was the reson why we had this?

lantoli: the release part is a leftover that was used before in relation with the release process, but not used any more, @andreaangiolillo told me this

.github/workflows/acceptance-tests.yml
lantoli: and we're removing 100 lines ;-)

i see your point, let's discuss offline

.github/workflows/migration-tests.yml
andreaangiolillo: This is the part that can be removed  `inputs.parent-event-name == 'release'` as it is no longer used

PR #1609, opened 2023-11-08
feat: search node management with new `mongodbatlas_search_deployment` resource

AgustinBettati:
## Description

Jira ticket: [INTMDB-1257](https://jira.mongodb.org/browse/INTMDB-1257)

Currently open as draft pull request as we have to make use of a dev preview of the Atlas SDK.

This PR includes:

- Implementation of `mongodbatlas_search_deployment` resource
- acceptance tests
- documentation and examples for new resource

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code
- [x] If changes include deprecations or removals, I defined an isolated PR with a relevant title as it will be used in the auto-generated changelog.

PR #1633, opened 2023-11-16
feat: search node management with `mongodbatlas_search_deployment` resource and data source

AgustinBettati:
## Description

Jira tickets: [INTMDB-1273](https://jira.mongodb.org/browse/INTMDB-1273), [INTMDB-1257](https://jira.mongodb.org/browse/INTMDB-1257)

Includes change from https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1621 and https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1609.

This PR is using the latest SDK release, and adjusting documentation links.

## Type of change:

## Required Checklist:


.github/workflows/acceptance-tests.yml
```diff
@@ -185,6 +188,30 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: 'go.mod'
+      - name: Acceptance Tests
+        env:
+          MONGODB_ATLAS_PUBLIC_KEY: ${{ secrets.MONGODB_ATLAS_PUBLIC_KEY_CLOUD_DEV }}
```
lantoli, line 204: CI: true is missing ;-).   cc @marcosuma

lantoli (approved):

.github/workflows/acceptance-tests.yml
AgustinBettati: true, however that env variable is not used in this job as no tests need to be skipped (several jobs dont define it). Can add just for consistency.

lantoli: yeah, i know it's only used when calling SkipTestForCI , it was more to advocate for a common env zone :-)

marcosuma: will try to answer to your PR :) but I think the global env variables would make the opposite effect here when you create a new job and you don't immediately expect there are global variables that might affect this job's behaviour

examples/atlas-search-deployment/README.md
```diff
@@ -0,0 +1,11 @@
+# MongoDB Atlas Provider - Atlas Cluster with dedicated Search Nodes Deployment
+
+This example shows how you can use Atlas Dedicated Search Nodes in Terraform. As a prerequisite, you need to create a project and cluster resource.
```
marcosuma: this feels like the prerequisite has to be done separately instead it's part of the example
```suggestion
This example shows how you can use Atlas Dedicated Search Nodes in Terraform. As part of it, a project and cluster resource are created as a prerequisite.
```

.github/workflows/acceptance-tests.yml
lantoli: or it could be the opposite, that you try to use SkipTestForCI and don't know why it's not skipping it.

I think it depends on the expectations set on the tests run from that GH action. for me it's clear in this case that all tests "want" to have CI:true even if they don't use/need it right now.

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -0,0 +1,368 @@
...
+	if _, _, err := connV2.AtlasSearchApi.UpdateAtlasSearchDeployment(ctx, projectID, clusterName, &searchDeploymentReq).Execute(); err != nil {
+		resp.Diagnostics.AddError("error during search deployment update", err.Error())
+		return
+	}
+
+	updateTimeout, diags := searchDeploymentPlan.Timeouts.Update(ctx, defaultSearchNodeTimeout)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
marcosuma, line 195: I see this is more or less the same that is in `create` any chance we could make this method leaner and move this inside the `waitSearch...`? or just make it a separate method with a one-liner call here..

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -0,0 +1,368 @@
...
+				return "", retrystrategy.RetryStrategyPendingState, nil
+			}
+			return nil, "", err
+		}
+
+		if util.IsStringPresent(deploymentResp.StateName) {
+			tflog.Debug(ctx, fmt.Sprintf("search deployment status: %s", *deploymentResp.StateName))
+			return deploymentResp, *deploymentResp.StateName, nil
+		}
+		return deploymentResp, "", nil
```
marcosuma, line 316: ah, this could have been so perfectly unit tested... can't wait to see them 😛

marcosuma (approved):
few comments, LGMT overall!

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
AgustinBettati: So there are some details here to take into account:
- for the simple approach of extracting to separate methods and having a one-liner: the part checking if there are errors and finish the update function (L193-L195) can not be extracted. L192 could potentially be extracted into separate function, but not in favour of passing resp and mutating within another function (harder to follow).
- for having the logic inside waitSearch function, an additional parameter would be added to receive the function `searchDeploymentPlan.Timeouts.Create` and `searchDeploymentPlan.Timeouts.Update` which add some complexity to the method signature. This is why I end up passing `timeout time.Duration` directly with some common code in each method.

AgustinBettati: moving forward with the merge, if any further action comes from this comment we can address in a followup PR.

PR #1634, opened 2023-11-16
fix: adds data_process_region field to online archive resource

marcosuma:
## Description

Link to any related issue(s): https://jira.mongodb.org/browse/INTMDB-1265

All description should be clear in the issue.
 
## Type of change:

## Required Checklist:


website/docs/r/online_archive.html.markdown
```diff
@@ -113,6 +114,10 @@ The only field required for criteria type `CUSTOM`
 
 * `expire_after_days` - Number of days used in the date criteria for nominating documents for deletion. Value must be between 7 and 9215.
 
+### Data Process Region
+* `cloud_provider` - Human-readable label that identifies the Cloud service provider where you wish to store your archived data.
+* `region` - Human-readable label that identifies the geographic location of the region where you wish to store your archived data. (`US_EAST_1`, `US_WEST_2`, `SA_EAST_1`, `EU_WEST_1`, `EU_WEST_2`, `EU_CENTRAL_1`, `AP_SOUTH_1`, `AP_SOUTHEAST_2`)
```
AgustinBettati: nit: instead of hardcoding valid values for region we can reference the API docs.

mongodbatlas/resource_mongodbatlas_online_archive.go
```diff
@@ -525,6 +558,15 @@ func fromOnlineArchiveToMap(in *admin.BackupOnlineArchive) map[string]any {
 		schemaVals["data_expiration_rule"] = []any{dataExpirationRule}
 	}
 
+	var dataProcessRegion map[string]any
+	if in.DataProcessRegion != nil && in.DataProcessRegion.CloudProvider != nil && in.DataProcessRegion.Region != nil {
```
AgustinBettati: since both field are optional, would it be possible only one can be defined?
Even though docs shows both fields as optional, it would seem logical that both cloud provider and regions must be defined.

mongodbatlas/resource_mongodbatlas_online_archive_test.go
```diff
@@ -165,6 +166,52 @@ func TestAccBackupRSOnlineArchiveBasic(t *testing.T) {
...
+				// The online archive won't work if the cluster does not have data
+				Config: testAccBackupRSOnlineArchiveConfigFirstStep(orgID, projectName, name),
+				Check: resource.ComposeTestCheckFunc(
+					populateWithSampleData(resourceName, &cluster),
+				),
+			},
+			{
+				Config: testAccBackupRSOnlineArchiveConfigWithProcessRegion(orgID, projectName, name, cloudProvider, "SA_EAST_1"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(onlineArchiveResourceName, "data_process_region.0.cloud_provider", "AWS"),
```
AgustinBettati, line 197: we can add checks to the data source here

mongodbatlas/resource_mongodbatlas_online_archive_test.go
```diff
@@ -331,6 +378,59 @@ func testAccBackupRSOnlineArchiveConfigWithoutSchedule(orgID, projectName, clust
...
+		sync_creation = true
+	}
+
+	data "mongodbatlas_online_archive" "read_archive" {
+		project_id =  mongodbatlas_online_archive.users_archive.project_id
+		cluster_name = mongodbatlas_online_archive.users_archive.cluster_name
+		archive_id = mongodbatlas_online_archive.users_archive.archive_id
+	}
+
+	data "mongodbatlas_online_archives" "all" {
```
lantoli, line 430: @marcosuma @AgustinBettati this is an example where we can have flaky tests because using the plural data source, it doesn't necessarily run after the resources are created as there's no dependency with them and can be executed at any time.
It's referring project_id and cluster_name in the resource but it will work only if some of them is computed. (and I don't think they are)

mongodbatlas/resource_mongodbatlas_online_archive.go
marcosuma: I would leave it like that. `Optional` gives flexibility for future changes as it leaves the schema "free". At the same time it's obvious that these two fields go together: if you specify the region you'll want to specify the provider. If you specify the provider, you _might_ want to specify the region.

mongodbatlas/resource_mongodbatlas_online_archive_test.go
marcosuma: Are you sure? I remember that as long as you specify a variable instead of a value it'll wait until the resource to which the variable refers to is ready.

website/docs/r/online_archive.html.markdown
marcosuma: let me know if new changes are ok

mongodbatlas/resource_mongodbatlas_online_archive.go
AgustinBettati: I usually like going with a more restrictive schema for new attributes for the following reason: In the future you will always be able to change from required to optional if new attributes are defined or a new use case appears, however going from optional to required would be a breaking change. Not blocking the PR but just to keep in mind.

mongodbatlas/resource_mongodbatlas_online_archive_test.go
AgustinBettati: I have the same idea as Marco, as long as you reference an attribute from the resource it will wait for its completion.

website/docs/r/online_archive.html.markdown
AgustinBettati: sorry for not stating beforehand, it is the same case for data sources so we keep docs aligned

mongodbatlas/resource_mongodbatlas_online_archive.go
lantoli: we haven't decided about where to do checks (client-side or server-side) but I tend to prefer to do changes in the server so logic is centralised.  

At the moment I don't see a big problem if an attribute is defined as optional in TF but at some moment server complains that it's required if not provided.

mongodbatlas/resource_mongodbatlas_online_archive_test.go
lantoli: i might be mistaken but this is what I concluded while working in search-index.
If the values are not computed there is not a real dependency, the values are known in advance as they won't change.

lantoli (approved):

website/docs/r/online_archive.html.markdown
marcosuma: sorry, it was my bad - thanks for catching that!

mongodbatlas/resource_mongodbatlas_online_archive.go
marcosuma: Agustin, I've expressed myself wrongly and mixed up two concepts. My sentence makes sense replacing `Optional` with `Required`, which is what you've said. Sorry about that.

In general, in fact, I think I support the idea that we should minimise validation in our provider because a) makes our code easier b) reduces code duplication c) keeps it consistent with API behaviour.

Said that, by having read this again I think we **need to fix this logic**.

marcosuma:
```suggestion
	if in.DataProcessRegion != nil && (in.DataProcessRegion.CloudProvider != nil || in.DataProcessRegion.Region != nil) {
```

@AgustinBettati @lantoli  wdyt about this change? I think it should be enough because if one of the two fields is nil we would pass it to the API as nil which shouldn't introduce any weird behaviour and additionally should return client errors from the API if needed. 

Open to other suggestions

mongodbatlas/resource_mongodbatlas_online_archive_test.go
marcosuma: Gonna resolve this because in any case this is not a new thing being introduced in this PR. If we realise it's like @lantoli is saying we can update it in all the places altogether.

mongodbatlas/resource_mongodbatlas_online_archive.go
AgustinBettati: Yes agree with your suggested change given that both fields are defined as optional.

AgustinBettati (approved):

PR #1640, opened 2023-11-20
fix!: handles paused clusters with errors when updating.

marcosuma:
## Description

Link to any related issue(s): https://jira.mongodb.org/browse/INTMDB-1307

## Type of change:

## Required Checklist:


marcosuma:
I tried to fix the same for `cluster` but APIs and logic inside are different than `advanced_cluster` - a fix would probably make it worse instead of better.

oarbusi (approved):
LGTM

mongodbatlas/resource_mongodbatlas_advanced_cluster_test.go
```diff
@@ -304,7 +309,7 @@ func TestAccClusterAdvancedCluster_PausedToUnpaused(t *testing.T) {
 		CheckDestroy:             testAccCheckMongoDBAtlasAdvancedClusterDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccMongoDBAtlasAdvancedClusterConfigSingleProviderPaused(orgID, projectName, rName, true),
+				Config: testAccMongoDBAtlasAdvancedClusterConfigSingleProviderPaused(orgID, projectName, rName, true, "M10"),
```
andreaangiolillo: Should the cluster tier be a constant reused by all the scenarios?

mongodbatlas/resource_mongodbatlas_advanced_cluster_test.go
```diff
@@ -327,6 +332,13 @@ func TestAccClusterAdvancedCluster_PausedToUnpaused(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "paused", "false"),
 				),
 			},
+			{
+				Config:      testAccMongoDBAtlasAdvancedClusterConfigSingleProviderPaused(orgID, projectName, rName, true, "M10"),
```
andreaangiolillo: Could you also add a scenario where we test that is not possible to update a pause cluster? I think that here we are testing that the cluster cannot be set to pause again after a short period of time

mongodbatlas/resource_mongodbatlas_advanced_cluster_test.go
```diff
@@ -327,6 +332,13 @@ func TestAccClusterAdvancedCluster_PausedToUnpaused(t *testing.T) {
...
+				ExpectError: regexp.MustCompile("CANNOT_PAUSE_RECENTLY_RESUMED_CLUSTER"),
+			},
+			{
+				Config: testAccMongoDBAtlasAdvancedClusterConfigSingleProviderPaused(orgID, projectName, rName, false, "M10"),
+			},
```
andreaangiolillo: Did we test the same behaviour with the cluster resource? they shared common functions so we should make sure that the behaviour there is correct

andreaangiolillo: > I tried to fix the same for cluster but APIs and logic inside are different than advanced_cluster - a fix would probably make it worse instead of better.

I just say this comment. Could you elaborate a little more? Maybe we can file a ticket for the Atlas Cluster team if this is a problem with the API

marcosuma: see my comment in the PR - the `cluster` has a different logic and has a different API behaviour, not planning to update that as I said it would make things likely worse because it would become a breaking change.

mongodbatlas/resource_mongodbatlas_advanced_cluster_test.go
marcosuma: 🆗

mongodbatlas/resource_mongodbatlas_advanced_cluster_test.go
marcosuma: I believe the test you are referring is just above this one. You can see I have an `ExpectError` for `CANNOT_UPDATE_PAUSED_CLUSTER`

andreaangiolillo: I haven't seen it. Thanks! resolving this comment

mongodbatlas/resource_mongodbatlas_advanced_cluster_test.go
andreaangiolillo: > not planning to update that as I said it would make things likely worse because it would become a breaking change.

Is this PR a breaking change? I see that the breaking change box is selected and you tried to add the breaking change label to the PR (I see how there is a problem with the automation, I will try to fix it).

What is the plan for the other cluster resource if we are not planning to update the logic? What are we going to do in case of a Help ticket about it? Does it make sense to update our cluster documentation to highlight this bug on the cluster resource? Happy to discuss this matter in another ticket.

marcosuma: > I just say this comment. Could you elaborate a little more? Maybe we can file a ticket for the Atlas Cluster team if this is a problem with the API

I don't think there is a problem with the APIs v1.0, I am just saying that right now there is no issue with `cluster`, but if we make the change the new behaviour would introduce an unnecessary breaking change for the `cluster` resource. 

Do we want that? I'd say no for one reason: I think customers are either using `cluster` or `advanced_cluster`, not both concurrently. In addition to that, once they'll have to migrate to `advanced_cluster` it'll be a breaking change for them no matter what.

I am open if you have a different opinion but only thing I can do is to create a separate Jira issue to align `cluster` to this - from a trial I did (that you can see in the commit history) it wasn't an immediate change hence it requires careful and tested changes and I don't want to slow this PR down (related to a HELP issue).

andreaangiolillo (approved):
LGTM, thanks for the clarification

lantoli (approved):
please fill the PR description

PR #1645, opened 2023-11-21
chore: Updates Atlas Go SDK

lantoli:
Automatic update for MongoDB Atlas Go Client SDK


andreaangiolillo (commented):
As discussed offline, let's see if we can use https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview  instead of the personal tokes so that the action won't fail the day that the token is deletes by the owner

lantoli:
please don't merge yet, I'm going to do some tests

PR #1643, opened 2023-11-20
PoC: extract project resource and data sources implementations into separate package

AgustinBettati:
## Description

**Summary**: This PoC shows the necessary changes need to enable grouping our resource and data source implementations into separate packages.

**Changes made**:
- Defining project resource and data source implementation in new `mongodbatlas/project` package.
- Use new `mongodbatlas/client` and  existing `mongodbatlas/framework` package for code that is required and defined in `mongodbatlas`. This code needs to be extract to a separate package to avoid `mongodbatlas/project` depending on `mongodbatlas` which results in cyclic import error.

**Benefits of having separate packages for each resource**:
- helps general organization, we avoid having all our implementation files in a single directory.
- gives us the flexibility of separating certain business logic into other files, with their associated unit testing files.

**Changes that were left out**:
- When trying to move the associated acceptance test and migration test files in the `mongodbatlas/project` package I did not find a way to break the cyclic import as tests rely on `testAccProviderV6Factories` which is tied to `mongodbatlas` package. As a follow up we can look more into other repos such as https://github.com/digitalocean/terraform-provider-digitalocean/pull/927 which include acceptance tests within each package in case we want to move in this direction.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [ ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals, I defined an isolated PR with a relevant title as it will be used in the auto-generated changelog.

PR #1653, opened 2023-11-22
test: including unit testing for search deployments state transition logic and model conversions

AgustinBettati:
## Description

Jira ticket: [INTMDB-1309](https://jira.mongodb.org/browse/INTMDB-1309)

Adds unit testing for search_deployment resource. As part of this the following tests were defined:

- test cases for logic that waits for deployment state to transition to valid state after create/update/delete operations.
- test cases for converting from terraform model to SDK model and vice versa.

## Type of change:

## Required Checklist:


mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -260,14 +277,15 @@ func splitSearchNodeImportID(id string) (projectID, clusterName string, err erro
 	return
 }
 
-func waitSearchNodeStateTransition(ctx context.Context, projectID, clusterName string, client *admin.APIClient, timeout time.Duration) (*admin.ApiSearchDeploymentResponse, error) {
+func waitSearchNodeStateTransition(
```
AgustinBettati: Having a separate file for this function, and its corresponding _unit.go file would be ideal here. Having more granular files will be viable if we separate implementation into packages (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1643)

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -304,7 +322,7 @@ func searchDeploymentRefreshFunc(ctx context.Context, projectID, clusterName str
 				return "", retrystrategy.RetryStrategyDeletedState, nil
 			}
 			if resp.StatusCode == 503 {
-				return "", retrystrategy.RetryStrategyPendingState, nil
+				return "", retrystrategy.RetryStrategyUpdatingState, nil
```
AgustinBettati: Detected a bug here thanks to test case. Not very relevant, simply a logic that continues the retry mechanism if a "Service Unavailable" is returned from API.

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -114,6 +115,22 @@ func (r *SearchDeploymentRS) Schema(ctx context.Context, req resource.SchemaRequ
 
 const defaultSearchNodeTimeout time.Duration = 3 * time.Hour
 
+func createUpdateTimeConfig(configuredTimeout time.Duration) retrystrategy.TimeConfig {
```
marcosuma: nit - just to align with the syntax below - said that, why did we use different config for the delete? why not just aligning it to all?
```suggestion
func createUpdateRetryTimeConfig(configuredTimeout time.Duration) retrystrategy.TimeConfig {
```

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
marcosuma: yes! I wonder if we can move forward with that change

mongodbatlas/fw_resource_mongodbatlas_search_deployment_state_transition_test.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+}
+
+type deleteStateTransitionTestCase struct {
+	name          string
+	mockResponses []SearchDeploymentResponse
+	expectedError bool
+}
+
+func TestTestSearchDeplomentStateTransitionForDelete(t *testing.T) {
```
andreaangiolillo: 👀 
```suggestion
func TestSearchDeploymentStateTransitionForDelete(t *testing.T) {
```

mongodbatlas/fw_resource_mongodbatlas_search_deployment_state_transition_test.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+type stateTransitionTestCase struct {
+	expectedResult *admin.ApiSearchDeploymentResponse
+	name           string
+	mockResponses  []SearchDeploymentResponse
+	expectedError  bool
+}
+
+func TestSearchDeplomentStateTransition(t *testing.T) {
```
andreaangiolillo: 👀 
```suggestion
func TestSearchDeploymentStateTransition(t *testing.T) {
```

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
marcosuma: ah, so nice to read this - music to our ears :)

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -366,3 +384,21 @@ func newTFSpecsModel(specs []admin.ApiSearchDeploymentSpec) []tfSearchNodeSpecMo
...
+	GetAtlasSearchDeployment(ctx context.Context, groupID, clusterName string) (*admin.ApiSearchDeploymentResponse, *http.Response, error)
+}
+
+type SearchDeploymentServiceFromClient struct {
+	client *admin.APIClient
+}
+
+func (a *SearchDeploymentServiceFromClient) GetAtlasSearchDeployment(ctx context.Context, groupID, clusterName string) (*admin.ApiSearchDeploymentResponse, *http.Response, error) {
+	return a.client.AtlasSearchApi.GetAtlasSearchDeployment(ctx, groupID, clusterName).Execute()
+}
```
marcosuma: if we move forward with (#1643) this can become a separate file.

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
```diff
@@ -280,22 +298,22 @@ func waitSearchNodeStateTransition(ctx context.Context, projectID, clusterName s
...
+		Timeout:    timeConfig.Timeout,
+		MinTimeout: timeConfig.MinTimeout,
+		Delay:      timeConfig.Delay,
 	}
 	_, err := stateConf.WaitForStateContext(ctx)
 	return err
 }
 
-func searchDeploymentRefreshFunc(ctx context.Context, projectID, clusterName string, client *admin.APIClient) retry.StateRefreshFunc {
+func searchDeploymentRefreshFunc(ctx context.Context, projectID, clusterName string, client SearchDeploymentService) retry.StateRefreshFunc {
```
andreaangiolillo: let's keep using reference here
```suggestion
func searchDeploymentRefreshFunc(ctx context.Context, projectID, clusterName string, client *SearchDeploymentService) retry.StateRefreshFunc {
```

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
andreaangiolillo: What was the strategy before? the name `RetryStrategyPendingState` is not super clear 🤔

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
andreaangiolillo: Was unit testing the only reason why you define the interface here?

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
```diff
@@ -0,0 +1,108 @@
...
+				Specs:       tfSpecsList(t, "S20_HIGHCPU_NVME", 2),
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resultModel, diags := newTFSearchDeployment(context.Background(), tc.clusterName, tc.SDKResp, nil)
+			if diags.HasError() {
+				t.Errorf("Unexpected errors found")
```
andreaangiolillo: should we print what was the error to help with the debug?

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
```diff
@@ -0,0 +1,108 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			apiReqResult := newSearchDeploymentReq(context.Background(), tc.tfModel)
+			if !reflect.DeepEqual(apiReqResult, tc.expectedSDKReq) {
+				t.Errorf("created sdk model did not match expected output")
+			}
+		})
+	}
+}
+
+func tfSpecsList(t *testing.T, instanceSize string, nodeCount int64) basetypes.ListValue {
```
marcosuma: nit: we don't have a team guideline hence so far it's up to engineer's preference. My suggestion is to avoid using separate method to "optimize" code in unit testing. Like Google does, we should maximize in-test context without deferring to separate methods.

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
```diff
@@ -0,0 +1,108 @@
...
+				ProjectID:   types.StringValue("222222222222222222222222"),
+				StateName:   types.StringValue("IDLE"),
+				Specs:       tfSpecsList(t, "S20_HIGHCPU_NVME", 2),
+			},
+			expectedSDKReq: admin.ApiSearchDeploymentRequest{
+				Specs: []admin.ApiSearchDeploymentSpec{
+					{
+						InstanceSize: "S20_HIGHCPU_NVME",
+						NodeCount:    2,
+					},
```
andreaangiolillo: Maybe you could define a list of const to reuse in your file

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
marcosuma: you don't have to follow this, I just wanted to bring this up.

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+type tfToSDKModelTestCase struct {
+	name           string
+	tfModel        *tfSearchDeploymentRSModel
+	expectedSDKReq admin.ApiSearchDeploymentRequest
+}
+
+func TestSearchDeploymentTFModelToSDK(t *testing.T) {
+	testCases := []tfToSDKModelTestCase{
```
marcosuma: worth it adding other test cases with not all fields?

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+type sdkToTFModelTestCase struct {
+	SDKResp         *admin.ApiSearchDeploymentResponse
+	expectedTFModel *tfSearchDeploymentRSModel
+	name            string
+	clusterName     string
+}
+
+func TestSearchDeploymentSDKToTFModel(t *testing.T) {
+	testCases := []sdkToTFModelTestCase{
```
marcosuma: same here... more test cases?

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
```diff
@@ -0,0 +1,108 @@
...
+func tfSpecsList(t *testing.T, instanceSize string, nodeCount int64) basetypes.ListValue {
+	tfSpecsList, diags := types.ListValueFrom(context.Background(), SpecObjectType, []tfSearchNodeSpecModel{
+		{
+			InstanceSize: types.StringValue(instanceSize),
+			NodeCount:    types.Int64Value(nodeCount),
+		},
+	})
+	if diags.HasError() {
+		t.Errorf("failed to create terraform spec lists model")
+	}
```
andreaangiolillo: 👀 same in adding more info on the reason why

mongodbatlas/fw_resource_mongodbatlas_search_deployment_state_transition_test.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+
+			if !reflect.DeepEqual(tc.expectedResult, resp) {
+				t.Errorf("Case %s: Response did not match expected output", tc.name)
+			}
+		})
+	}
+}
+
+type deleteStateTransitionTestCase struct {
```
andreaangiolillo: Can't you reuse `stateTransitionTestCase`? in other work, do you need two structs or only 1?

andreaangiolillo (commented):
Left some comments that are not blocking the PR

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
AgustinBettati: `RetryStrategyPendingState` is used for encryption at rest resource, but is not a valid state for the search deployment. The correct state for representing that a change is in progress is with `RetryStrategyUpdatingState`.

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
AgustinBettati: revised this and defined into a single function. The only difference is with the `MinTimeout` value, had defined a lower value for the delete operation as I could see during testing this operation was significantly faster. However I do imagine this depends a lot on the nature of each deployment.

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
AgustinBettati: I see your point. I extracted this method mainly for readability of the test case:
```
expectedTFModel: &tfSearchDeploymentRSModel{
  ID:          types.StringValue("111111111111111111111111"),
  ClusterName: types.StringValue("Cluster0"),
  ProjectID:   types.StringValue("222222222222222222222222"),
  StateName:   types.StringValue("IDLE"),
  Specs:       tfSpecsList(t, "S20_HIGHCPU_NVME", 2),
}
```
Unfortunately there is no way to define a terraform `types.List` inlined.

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
AgustinBettati: In the case of search deployment there are no optional parameters that can be provided. Everything is required, as well as spec lists which has to be of 1 element (validated by Atlas API).

mongodbatlas/fw_resource_mongodbatlas_search_deployment_state_transition_test.go
AgustinBettati: :) thanks

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
AgustinBettati: SearchDeploymentService is an interface type, which in turn is a pointer to struct that fulfills the interface (instantiated in `NewSearchDeploymentServiceFromClient`). More details of why pointer to interfaces is discouraged: https://stackoverflow.com/a/44372954

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
AgustinBettati: Yes. If we define a separate package structure can see this defined in a separate file to avoid noise in the main implementation file. Did you have any particular concern?

mongodbatlas/fw_resource_mongodbatlas_search_deployment_model_test.go
AgustinBettati: makes sense, adjusted

mongodbatlas/fw_resource_mongodbatlas_search_deployment_state_transition_test.go
AgustinBettati: Yep was able to reuse here. Simply do not use/define the field `expectedResult *admin.ApiSearchDeploymentResponse` for delete tests. Thanks!

mongodbatlas/fw_resource_mongodbatlas_search_deployment.go
andreaangiolillo: no, I was wondering if I was missing other reasons. I agree that a separate package structure would help here

marcosuma (approved):
LGTM - approve to unblock

lantoli (approved):
very nice!

PR #1655, opened 2023-11-22
feat!: Adds first and last name in backup_compliance_policy resource.

marcosuma:
## Description

Link to any related issue(s): https://jira.mongodb.org/browse/INTMDB-1072

## Type of change:

## Required Checklist:


mongodbatlas/resource_mongodbatlas_backup_compliance_policy.go
```diff
@@ -39,6 +39,16 @@ func resourceMongoDBAtlasBackupCompliancePolicy() *schema.Resource {
...
 			},
+			"authorized_user_first_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"authorized_user_last_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
andreaangiolillo: is the API returning these fields even when they are not set? if the answer is not, we can remove the compute here

mongodbatlas/resource_mongodbatlas_backup_compliance_policy.go
```diff
@@ -39,6 +39,16 @@ func resourceMongoDBAtlasBackupCompliancePolicy() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"authorized_user_first_name": {
```
andreaangiolillo: Should we update the documentation as well?

mongodbatlas/resource_mongodbatlas_backup_compliance_policy.go
```diff
@@ -39,6 +39,16 @@ func resourceMongoDBAtlasBackupCompliancePolicy() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"authorized_user_last_name": {
```
andreaangiolillo: Should we add these field to the data source as well?

mongodbatlas/resource_mongodbatlas_backup_compliance_policy.go
marcosuma: yes they are being returned as `unknown` hence that's why I used `Computed: true`.

andreaangiolillo: Thanks for the clarification here

mongodbatlas/resource_mongodbatlas_backup_compliance_policy.go
marcosuma: Good catch Andrea, thank you!

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/mwielbut/pointy v1.1.0
 	github.com/spf13/cast v1.5.1
 	github.com/zclconf/go-cty v1.14.1
-	go.mongodb.org/atlas v0.35.0
+	go.mongodb.org/atlas v0.36.0
```
lantoli, line 26: I normally prefer Atlas SDK updates in their own PR but it's not a big deal

lantoli (approved):

go.mod
marcosuma: ok thank you for the feedback!

marcosuma:
waiting confirmation from @Zuhairahmed to see if we know from which version we should make those fields mandatory.

andreaangiolillo (approved):
LGTM. Thanks for addressing all the comments

mongodbatlas/resource_backup_compliance_policy.go
```diff
@@ -534,3 +571,7 @@ func expandDemandBackupPolicyItem(d *schema.ResourceData) *matlas.PolicyItem {
 
 	return &onDemand
 }
+
+func printWarningMessageForFirstLastNameFields() {
+	log.Printf("[WARN] Note: Starting from v1.14.0 authorized_user_first_name and authorized_user_last_name will become mandatory. Please update your script accordingly.")
```
AgustinBettati: to make this more visible for the end user, have you considered adding a diagnostic with a warning severity?

AgustinBettati (commented):
small suggestion

Zuhairahmed:
> waiting confirmation from @Zuhairahmed to see if we know from which version we should make those fields mandatory.

suggest @marcosuma we make these fields mandatory as part of next v1.14.0 release. for users who are not ready to upgrade they can stay back on v1.13.1 until they are ready

mongodbatlas/resource_backup_compliance_policy.go
marcosuma: @AgustinBettati I considered it but it would require a bigger change because I would have to `append` the `warning` diag to all the `return` statements (unless I am missing something).

See this [extract example](https://gist.github.com/marcosuma/33c7201ae79e7ab77f6f34c90c6160a2) - do you think there's another way / it is worth it?

AgustinBettati: Yes that is the option I had in mind for adding warnings. It does have the benefit that users see the warning each time they run a terraform command similar to deprecation messages. If we go for defining attributes as required this will no longer be necessary.

marcosuma: Correct, it will no longer be necessary as per the conversation I had with Zuhair.

andreaangiolillo (approved):
LGTM

AgustinBettati (approved):
LGTM, would add breaking change label so we remember to mention this clearly in changelog/migration guide.

lantoli (approved):

PR #1664, opened 2023-11-23
fix: adjust root `project_id` attribute in `mongdbatlas_project_api_key` resource to optional

AgustinBettati:
## Description

Link to any related issue(s): [INTMDB-1296](https://jira.mongodb.org/browse/INTMDB-1296)

Making root `project_id` attribute in `mongdbatlas_project_api_key` optional enables users to unassign the root project from the created API Key without having to delete the resource all together. More details in the referenced ticket.

## Type of change:

## Required Checklist:


mongodbatlas/resource_mongodbatlas_project_api_key.go
```diff
@@ -269,28 +274,20 @@ func resourceMongoDBAtlasProjectAPIKeyDelete(ctx context.Context, d *schema.Reso
 		}
 	}
 
-	_, roleOk := d.GetOk("role_names")
-	if !roleOk {
```
AgustinBettati: legacy check, `role_names` no longer exists at root level.

website/docs/d/project_api_key.html.markdown
```diff
@@ -16,29 +16,32 @@ description: |-
 
 ## Example Usage
 
-### Using org_id attribute to query
```
AgustinBettati, line 19: no changes in data source, just doc fixes I found

website/docs/d/project_api_keys.html.markdown
```diff
@@ -17,29 +17,25 @@ description: |-
 ## Example Usage
 
 ```terraform
-resource "mongodbatlas_project_api_key" "test" {
```
AgustinBettati, line 20: no changes in data source, just doc fixes I found

examples/atlas-api-key/create-and-assign-pak-together/versions.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     mongodbatlas = {
       source  = "mongodb/mongodbatlas"
-      version = "~> 1.0"
+      version = "~> 1.13.2"
```
marcosuma, line 5: the downside of this is that when you merge this change this version won't be there. Are we ok with that? Another (longer) approach is that we leave the example as it is (by still setting the project_id) and once this is released to TF then we can update the example (or create a different one). LEt me know if I am missing something

mongodbatlas/resource_mongodbatlas_project_api_key.go
```diff
@@ -73,42 +73,41 @@ type APIProjectAssignmentKeyInput struct {
...
-				apiKey, resp, err = conn.ProjectAPIKeys.Create(ctx, projectID, createRequest)
-				if err != nil {
-					if resp != nil && resp.StatusCode == http.StatusNotFound {
-						d.SetId("")
-						return nil
-					}
-				}
+
+		// creates api key using project id of first defined project assignment
+		firstAssignment := projectAssignmentList[0]
```
marcosuma: should this new behaviour be documented? is it a breaking change? What happens with `project_id` when it is actually defined? from this code it seems like it won't be used aynmore

mongodbatlas/resource_mongodbatlas_project_api_key.go
```diff
@@ -25,7 +25,7 @@ func resourceMongoDBAtlasProjectAPIKey() *schema.Resource {
 		Schema: map[string]*schema.Schema{
```
lantoli: @AgustinBettati as there are significant changes in this file, I don't know if it's ok for you to wait until I merge the file structure PR.

mongodbatlas/resource_mongodbatlas_project_api_key.go
AgustinBettati: So this is more of an implementation detail: one project id is used for the creation of the api key, and then remaining projects assignments are assigned to the new api key. The previous implementation was essentially the same, only instead of taking the first element for the project assignment list it would grab the project which was defined in root project_id for the creation of the api key. From a users standpoint it is the same.

> What happens with project_id when it is actually defined? from this code it seems like it won't be used anymore

This is correct, it is no longer used. From docs changes and deprecation warning users will be aware this attribute is no longer being used (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1665/files#diff-31571e18c14ed5632e87c050342b195d9cbc5092b465b08abfb8b1dcf87d37bdL49)

mongodbatlas/resource_mongodbatlas_project_api_key.go
AgustinBettati: As long as this is merged before coming release no hurry with this PR.

examples/atlas-api-key/create-and-assign-pak-together/versions.tf
AgustinBettati: This is a good point which I had not contemplated. I do see this would apply to several scenarios where we modify examples, such as adding new attributes.
Users do have the possibility of seeing the examples section under a specific release tag, but understand this is probably not the regular flow.
One alternative I see to maintain examples compatible with latest release (in master branch) is to have new development in a separate branch (example `develop`), and merge into master during release process (this obviously has the downside of an additional step in our release process).

andreaangiolillo: Are you sure you need to update the version here? We have a [piece of logic](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/86df112a260ab00b9bbd188f1819ae9b23ad9eb7/.github/workflows/examples.yml#L31) in the action which overrides the version in the `version.tf` to be 99999 (which is the "current" built version)  so it should have your changes 


````
 name: go build
        run: go build -o terraform-plugin-dir/terraform-provider-mongodbatlas-dev_v99.99.99_x5 .
      - name: override plugin
        run: |
          # For newer versions
          mkdir -p ~/.terraform.d/plugins/registry.terraform.io/hashicorp/mongodbatlas-dev/99.99.99/"$(go env GOOS)"_"$(go env GOARCH)"/
          cp terraform-plugin-dir/terraform-provider-mongodbatlas-dev_v99.99.99_x5 ~/.terraform.d/plugins/registry.terraform.io/hashicorp/mongodbatlas-dev/99.99.99/"$(go env GOOS)"_"$(go env GOARCH)"/

````

Comment (AgustinBettati) [reply examples/atlas-api-key/create-and-assign-pak-together/versions.tf:5]: > Are you sure you need to update the version here?

There is no need from CI side, this change is more for users to know that this example (which does not define root project_id) can be used from 1.13.2 onwards. @marcosuma does raise a valid point that when this is merged, our master branch will contain examples which could cause confusion for users as it mentions a version that has not yet been released.

Comment (andreaangiolillo) [reply examples/atlas-api-key/create-and-assign-pak-together/versions.tf:5]: > master branch will contain examples which could cause confusion for users as it mentions a version that has not yet been released

it makes sense. We can create a ticket to add automation (action) to update the provider version for all the examples to a specified version before the release. In this case, you won't need to update the version here and the new version will be in master only between pre-release and final release. This will fix another issue where we have examples in our repo with older provider versions and we should make sure that our examples are relevant with the latest release.

it's worth noting that we haven't received any issues or help requests regarding examples in the repository using an unreleased version so far. (saying to help triage this piece of work if we agree on creating the ticket)

Comment (AgustinBettati) [reply examples/atlas-api-key/create-and-assign-pak-together/versions.tf:5]: > ticket to add automation (action) to update the provider version for all the examples

I believe the main issue is when the content of the example changes and is **only** compatible with a version that will be released in the future, causing potential confusion for users that see examples from master branch. I do agree I don't see it as priority issue, and cannot see a simple solution either.

Comment (marcosuma) [reply examples/atlas-api-key/create-and-assign-pak-together/versions.tf:5]: I don't have major concerns here. I prefer to unblock you and resolve this one if we haven't seen a similar issue before. Feel free to resolve if you all are ok.

Comment (marcosuma) [reply examples/atlas-api-key/create-and-assign-pak-together/versions.tf:5]: Also @andreaangiolillo's suggestion is ok to me as well

File: mongodbatlas/resource_project_api_key.go [hunk 7]
```diff
@@ -79,42 +79,41 @@ type APIProjectAssignmentKeyInput struct {
...
-				apiKey, resp, err = conn.ProjectAPIKeys.Create(ctx, projectID, createRequest)
-				if err != nil {
-					if resp != nil && resp.StatusCode == http.StatusNotFound {
-						d.SetId("")
-						return nil
-					}
-				}
+
+		// creates api key using project id of first defined project assignment
+		firstAssignment := projectAssignmentList[0]
```
lantoli, line 97: do we need to check len(projectAssignmentList) > 0?

lantoli (approved):

mongodbatlas/resource_project_api_key.go
AgustinBettati: The schema enforces that project_assignment must be defined with MinItem of 1 so there is no need really

AgustinBettati:
failing tests are related to "MAX_GROUPS_PER_ORG_EXCEEDED" issue in testing org, verified migration and acceptance tests are passing well.

PR #1665, opened 2023-11-23
deprecate: optional root `project_id` attribute in `mongdbatlas_project_api_key` is deprecated

AgustinBettati:
## Description

As a follow up to https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1664, this PR marks the adjusted attribute as deprecated. Created in separate PR to have explicit entry in change log.

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/resource_project_api_key.go
+++ b/mongodbatlas/resource_project_api_key.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/constant"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/config"
 	"go.mongodb.org/atlas-sdk/v20231115001/admin"
@@ -30,8 +31,9 @@ func ResourceProjectAPIKey() *schema.Resource {
 		},
 		Schema: map[string]*schema.Schema{
 			"project_id": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: fmt.Sprintf(constant.DeprecationParamByVersion, "1.16.0"),
 			},
 			"api_key_id": {
 				Type:     schema.TypeString,
--- a/website/docs/r/project_api_key.html.markdown
+++ b/website/docs/r/project_api_key.html.markdown
@@ -46,7 +46,7 @@ resource "mongodbatlas_project_api_key" "test" {
 ## Argument Reference
 
 * `description` - (Required) Description of this Project API key.
-* `project_id` - Unique 24-hexadecimal digit string that identifies your project.
+* `project_id` - Unique 24-hexadecimal digit string that identifies your project. **WARNING:** this parameter is deprecated as it no longer needs to be defined. It will be removed in version 1.16.0.
 
 ~> **NOTE:** Project created by API Keys must belong to an existing organization.
 


mongodbatlas/resource_mongodbatlas_project_api_key.go
```diff
@@ -24,8 +24,9 @@ func resourceMongoDBAtlasProjectAPIKey() *schema.Resource {
 		},
 		Schema: map[string]*schema.Schema{
 			"project_id": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: fmt.Sprintf(DeprecationByVersionMessageParameter, "1.16.0"),
```
AgustinBettati: @Zuhairahmed As this deprecation is not related to Atlas API or has any fixed date, similar to how we handled https://jira.mongodb.org/browse/INTMDB-1211 I set 3 major versions before removal. Let me know what you think.

andreaangiolillo: Do we have to update the data source as well?

website/docs/r/project_api_key.html.markdown
```diff
@@ -46,7 +46,7 @@ resource "mongodbatlas_project_api_key" "test" {
 ## Argument Reference
 
 * `description` - (Required) Description of this Project API key.
-* `project_id` - Unique 24-hexadecimal digit string that identifies your project.
+* `project_id` - Unique 24-hexadecimal digit string that identifies your project. **WARNING:** this parameter is deprecated as it no longer needs to be defined. It will be removed in version 1.16.0.
```
andreaangiolillo, line 49: Isn't this a breaking change? 🤔

mongodbatlas/resource_mongodbatlas_project_api_key.go
andreaangiolillo: We should update our acceptance tests to not use the deprecated field

andreaangiolillo: [if is not a breaking change] Can we add a migration test scenario to make sure that upgrading to the latest version with the deprecated field does not bring an error in the plan?

website/docs/r/project_api_key.html.markdown
andreaangiolillo: if the customers have to update their configuration to not use this field -> it is a breaking change

AgustinBettati: > if the customers have to update their configuration

Customers will not have to make any changes to there existing configuration with this change, the `project_id` is made optional and deprecated as it will be eventually deleted (this is when it will be a breaking change). For reference, HashiCorp versioning guidelines allow deprecating attributes in a minor version as existing functionality remains compatible: https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#example-minor-number-increments

mongodbatlas/resource_mongodbatlas_project_api_key.go
AgustinBettati: yes we can add this case

marcosuma: Can you create a jira issue for this and target 1.16.0? So we don't forget it

marcosuma (approved):

mongodbatlas/resource_mongodbatlas_project_api_key.go
AgustinBettati: > Do we have to update the data source as well?

Data source implementation remains the same. This is because the minimum attributes required to obtain information of the api key at the moment is using an api_key_id together with a project_id:
```
data "mongodbatlas_project_api_key" "test" {
  api_key_id = vars.api_key_id
  project_id = vars.project_id
}
```
(this is fetching the information of the api key, with all its project assignments)

> We should update our acceptance tests to not use the deprecated field

Yes can adjust so that eventually when field is removed less changes are required. Will however leave one test with old definition to ensure no breaking changes.

AgustinBettati: included in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1664

mongodbatlas/resource_project_api_key.go
```diff
@@ -30,8 +30,9 @@ func ResourceProjectAPIKey() *schema.Resource {
 		},
 		Schema: map[string]*schema.Schema{
 			"project_id": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Deprecated: fmt.Sprintf(DeprecationByVersionMessageParameter, "1.16.0"),
```
lantoli: take a look at the internal/common/constant package

AgustinBettati: thanks!

lantoli (approved):

AgustinBettati:
Verified failing test are not related (alert config and database user)

PR #1669, opened 2023-11-27
fix: Fixes project flaky tests

lantoli:
## Description

Jira ticket: [INTMDB-1341](https://jira.mongodb.org/browse/INTMDB-1341)

Fixes project flaky tests

## Type of change:

## Required Checklist:


.github/workflows/acceptance-tests.yml
```diff
@@ -335,7 +335,6 @@ jobs:
           MONGODB_ATLAS_ORG_ID: ${{ vars.MONGODB_ATLAS_ORG_ID_CLOUD_DEV }}
           MONGODB_ATLAS_BASE_URL: ${{ vars.MONGODB_ATLAS_BASE_URL }}
           MONGODB_ATLAS_PROJECT_OWNER_ID: ${{ vars.MONGODB_ATLAS_PROJECT_OWNER_ID }}
-          MONGODB_ATLAS_API_KEYS_IDS: ${{ vars.MONGODB_ATLAS_API_KEYS_IDS }}
```
lantoli, line 338: not used

internal/testutil/acc/pre_check.go
```diff
@@ -45,10 +46,26 @@ func PreCheckAtlasUsername(tb testing.TB) {
 	}
 }
 
-func PreCheckTeamsIds(tb testing.TB) {
-	if os.Getenv("MONGODB_ATLAS_TEAMS_IDS") == "" {
-		tb.Skip("`MONGODB_ATLAS_TEAMS_IDS` must be set for Projects acceptance testing")
+func PreCheckProjectTeamsIds(tb testing.TB, min int) {
```
lantoli: improve teamsIds checks, and fail tests if pre-check is not met

internal/service/project/data_source_projects.go
```diff
@@ -171,17 +171,14 @@ func (d *ProjectsDS) Read(ctx context.Context, req datasource.ReadRequest, resp
 }
 
 func populateProjectsDataSourceModel(ctx context.Context, conn *matlas.Client, connV2 *admin.APIClient, stateModel *tfProjectsDSModel, projectsRes *matlas.Projects) error {
```
lantoli, line 173: the fix is in this func

internal/service/project/data_source_project_test.go
```diff
@@ -15,26 +14,23 @@ import (
 )
 
 func TestAccProjectDSProject_byID(t *testing.T) {
-	projectName := acctest.RandomWithPrefix("test-acc")
-	orgID := os.Getenv("MONGODB_ATLAS_ORG_ID")
-	teamsIds := strings.Split(os.Getenv("MONGODB_ATLAS_TEAMS_IDS"), ",")
-	if len(teamsIds) < 2 {
-		t.Skip("`MONGODB_ATLAS_TEAMS_IDS` must have 2 team ids for this acceptance testing")
```
lantoli, line 22: improve pre-checks and fail if they're not met instead of just skipping the test

internal/service/project/data_source_projects.go
```diff
@@ -171,17 +171,14 @@ func (d *ProjectsDS) Read(ctx context.Context, req datasource.ReadRequest, resp
...
-	results := make([]*tfProjectDSModel, len(projectsRes.Results))
-
-	for i, project := range projectsRes.Results {
+	results := make([]*tfProjectDSModel, 0, len(projectsRes.Results))
+	for _, project := range projectsRes.Results {
 		atlasTeams, atlasLimits, atlasProjectSettings, err := getProjectPropsFromAPI(ctx, conn, connV2, project.ID)
-		if err != nil {
-			return fmt.Errorf("error while getting project properties for project %s: %v", project.ID, err.Error())
+		if err == nil { // if the project is still valid, e.g. could have just been deleted
```
marcosuma, line 177: should we fail if `err != nil` ?

internal/testutil/acc/pre_check.go
```diff
@@ -45,10 +46,26 @@ func PreCheckAtlasUsername(tb testing.TB) {
...
+		return
 	}
+	teamsIds := strings.Split(envVar, ",")
+	if count := len(teamsIds); count < min {
+		tb.Fatalf("`MONGODB_ATLAS_TEAMS_IDS` must have at least %d team ids for this acceptance testing, has %d", min, count)
+	}
+}
+
+func GetProjectTeamsIds(pos int) string {
+	envVar := os.Getenv("MONGODB_ATLAS_TEAMS_IDS")
```
marcosuma, line 62: nit: can you please pass the name of the environment variable as input of the function?

internal/testutil/acc/pre_check.go
marcosuma: nit: can you please pass the name of the environment variable as input of the function?

internal/service/project/data_source_project_test.go
```diff
@@ -15,26 +14,23 @@ import (
...
-	}
-
+	var (
+		projectName = acctest.RandomWithPrefix("test-acc")
+		orgID       = os.Getenv("MONGODB_ATLAS_ORG_ID")
+	)
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:                 func() { acc.PreCheckBasic(t); acc.PreCheckTeamsIds(t) },
+		PreCheck:                 func() { acc.PreCheckBasic(t); acc.PreCheckProjectTeamsIds(t, 2) },
```
marcosuma:
```suggestion
		PreCheck:                 func() { acc.PreCheckBasic(t); acc.PreCheckProjectTeamsIdsWithMinSize(t, 2) },
```

nit: for clarity I would suggest a more specific method name

internal/service/project/data_source_projects.go
lantoli: no, that was the problem before.  Imagine we get the list of projects for the org, but then some of them are deleted.  It's not really a failure, so we just ignore it if it's deleted after it was in the list.   This was the reason for project flaky test.

internal/testutil/acc/pre_check.go
marcosuma:
```suggestion
func PreCheckProjectTeamsIds(tb testing.TB, minProjectCount int) {
```

marcosuma (approved):
few minor comments, LGTM overall

internal/testutil/acc/pre_check.go
lantoli: I think it's better in this way as PreCheckProjectTeamsIds and GetProjectTeamsIds work together for teamsIds and they both use the same env.var

internal/testutil/acc/pre_check.go
lantoli: same, I think it's better to have the env.var in both methods as they're used only for teamsIds, these methods are not intended to be generic for other purporses.

internal/service/project/data_source_project_test.go
lantoli: I can do that but I normally like to say "with..." when there are multiple related methods, there is only one method in this case

internal/testutil/acc/pre_check.go
lantoli: I can change it, although I tried to follow "go idiomatic" in the sense the I think "min" in this context is clear, but I agree I can make it more specific here.

lantoli: renamed to minTeamsCount (although it looks a bit redudant to me as Teams is already in function name)

internal/service/project/data_source_project_test.go
lantoli: renamed to: func PreCheckProjectTeamsIdsWithMinCount(tb testing.TB, minTeamsCount int)

internal/service/project/data_source_projects.go
marcosuma: ok noted, but then I guess we're ok to ignore 5xx errors?
What I am trying to understand here is: if there is a 5xx error and we ignore that and we move on, will we get a different test behaviour that might hide some problem?

internal/testutil/acc/pre_check.go
marcosuma: > work together 

I think this is an assumption that might not be always true, unless you call the `precheck...` in the `get...` method right?

internal/service/project/data_source_projects.go
lantoli: I've got 500 sometimes when doing something in a project that is being closed, so we could get also a 500 and it doesn't mean a real error but that the project is closing.   So I think it's hard in this case to differentiate a "real" error from an error caused by the project being closed. Also chances are that if the project is in the list and it's not closed we'll get the details without any problem

internal/testutil/acc/pre_check.go
lantoli: if you don't call precheck you won't check that you have the min teams you need but it will still work (returns "" when you don't have that pos).  
but if you don't call the precheck method, i don't see how passing the env.var can help there.

(They have the same name part ProjectTeamsIds and are located together in the file to make it easier to see that they are intended to work together)

internal/service/project/data_source_projects.go
AgustinBettati: Understand this an edge case that is likely only appearing  in our acceptance tests execution: fetching and creating/deleting project in parallel. Curious if a potential fix to this problem could be defining the data source test to run sequentially (`resource.Test(` over `resource.ParallelTest(`) to reduce the chances of this scenario. (just to keep in mind for other flaky tests that we might have)

lantoli: that would probably fix the tests, also to use the concurrency facility in GH actions so project acc and migration tests don't run in parallel.  But at the end I thought this is a realistic use case that can happen sometimes and we don't want the DS to fail if that happens, that's why I focused on improving the prod. code.

PR #1673, opened 2023-11-28
chore: Increases project list limit for clean-up Github action

lantoli:
## Description

Jira ticket: [INTMDB-1398](https://jira.mongodb.org/browse/INTMDB-1398)

Increases project list limit for clean-up Github action

## Type of change:

## Required Checklist:


scripts/cleanup-test-env.sh
```diff
@@ -19,7 +19,7 @@ set -Eeou pipefail
 projectToSkip="${PROJECT_TO_NOT_DELETE:-NONE}"
 
 # Get all project Ids inside the organization
-projects=$(atlas project ls -o json)
+projects=$(atlas project ls --limit 500 -o json)
```
marcosuma: * Do we have a risk of increased latency and timeout from this? 
* What is the rationale in changing it? Is there any failure in our account right now?

lantoli: no latency problem, it's because we're getting errors:  "MAX_GROUPS_PER_ORG_EXCEEDED") Maximum number of groups per organization.
we want to try to delete all projects, not only the first 100 projects (default value for limit)

lantoli: I think the main problem is that we're not cleaning-up projects with network endpoints, i'm working on changing that

.github/workflows/acceptance-tests.yml
```diff
@@ -439,9 +439,9 @@ jobs:
           go-version-file: 'go.mod'
       - name: Acceptance Tests
         env:
-          MONGODB_ATLAS_PUBLIC_KEY: ${{ secrets.MONGODB_ATLAS_PUBLIC_KEY_CLOUD_DEV_NETWORK }}
```
AgustinBettati: was using the dev network org a mistake here?

scripts/cleanup-test-env.sh
AgustinBettati: regarding improving the cleanup process we do have INTMDB-1123 that contains more context and suggestions in case this is helpful

.github/workflows/acceptance-tests.yml
andreaangiolillo: we have two organizations in cloud-dev that we use for acceptance tests. The main reason was to decrease the changes of hitting the limit of projects that you can have in an org. In addition, the network org has a project with specific configurations needed for some of the test

scripts/cleanup-test-env.sh
lantoli: thx, as discussed offline I'll just do here the private endpoints deletion

.github/workflows/acceptance-tests.yml
lantoli: ok, now that we'll be deleting private network I'll undo this change

AgustinBettati (approved):
LGTM

scripts/cleanup-test-env.sh
marcosuma: yup, I remember we had INTMDB-1123 in our RFW and that's why I was asking since this PR wasn't attached to that. Thank you @AgustinBettati for the note and thanks @lantoli for leaving a comment in https://jira.mongodb.org/browse/INTMDB-1123

scripts/cleanup-test-env.sh
```diff
@@ -16,10 +16,33 @@
...
 
+if [ -z "${MONGODB_ATLAS_ORG_ID}" ]; then
+  echo "Error: MONGODB_ATLAS_ORG_ID environment variable must be set"
+  exit 1
+fi
+org_id="${MONGODB_ATLAS_ORG_ID}"
+
 # Get all project Ids inside the organization
-projects=$(atlas project ls -o json)
+projects=$(atlas project ls --limit 500 --orgId ${org_id}  -o json)
```
andreaangiolillo: nit: According to our documentation https://www.mongodb.com/docs/atlas/reference/atlas-limits/, we can only have max 250 projects per org

scripts/cleanup-test-env.sh
```diff
@@ -16,10 +16,33 @@
...
+    provider=$1
+    set +e
+    count=$(atlas privateEndpoints "${provider}" list --projectId "${clean_project_id}" -o=go-template="{{len .}}")
+    retVal=$?
+    if [ $retVal -ne 0 ]; then
+        count=0
+    fi
+    if [ "${count}" != "0" ]; then
+        echo "Project ${clean_project_id} contains ${provider} endpoints, will start deleting it now and will try to delete the project in the next execution"
+        id=$(atlas privateEndpoints "${provider}" list --projectId "${clean_project_id}" -o=go-template="{{(index . 0).Id}}")
```
andreaangiolillo, line 28: the list commands return the provider as well
```json
[
  {
    "cloudProvider": "AWS",
    "id": "64c90604ca62bc79f4554b5c",
    "regionName": "EU_WEST_1",
    "status": "AVAILABLE",
    "endpointServiceName": "com.amazonaws.vpce.eu-west-1.vpce-svc-032aa573f24af944c",
    "interfaceEndpoints": [
      "vpce-07d624b0203224579"
    ]
  }
]
```

instead of calling the function 3 times, could you just use the provider in the response as you are doing for the ID?

scripts/cleanup-test-env.sh
lantoli: i know but the limit in the commands is 500, so i assumed it didn't harm and it'll work if that limit is increased

scripts/cleanup-test-env.sh
lantoli: I'm not sure I can do that the provider is part of the command, the CLI cleanup also calls 3 times.

e.g.:  atlas privateEndpoints azure list --projectId my_project_id

I don't think you can ask for all endpoints in one call

PR #1691, opened 2023-12-01
feat: Runs Migration tests for different provider versions

lantoli:
## Description

Runs Migration tests for different provider versions.

Link to any related issue(s): [CLOUDP-215183](https://jira.mongodb.org/browse/CLOUDP-215183)

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/migration-tests.yml
+++ b/.github/workflows/migration-tests.yml
@@ -85,13 +85,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Get Last Release
-        id: get_last_release
-        run: |
-         LAST_RELEASE=$(curl -sSfL -X GET https://api.github.com/repos/mongodb/terraform-provider-mongodbatlas/releases/latest | jq -r '.tag_name | ltrimstr("v")')
-         echo "Last release: $LAST_RELEASE"
-         echo "MONGODB_ATLAS_LAST_VERSION=$LAST_RELEASE" >> $GITHUB_ENV
-        shell: bash
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
@@ -118,13 +111,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Get Last Release
-        id: get_last_release
-        run: |
-         LAST_RELEASE=$(curl -sSfL -X GET https://api.github.com/repos/mongodb/terraform-provider-mongodbatlas/releases/latest | jq -r '.tag_name | ltrimstr("v")')
-         echo "Last release: $LAST_RELEASE"
-         echo "MONGODB_ATLAS_LAST_VERSION=$LAST_RELEASE" >> $GITHUB_ENV
-        shell: bash
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-suite.yml
```diff
@@ -6,31 +6,31 @@ on:
     - cron: "0 2 * * *" # workflow runs every day at 02:00 AM
   
 jobs:
-  acc-tests:
+
+  mig-tests:
```
lantoli, line 10: run migration tests first as they take much less time

lantoli, line 10: short migration test name so it fits better in GH UI

.github/workflows/migration-tests.yml
```diff
@@ -57,7 +79,7 @@ jobs:
             - 'internal/service/advancedcluster/*.go'
 
   project: 
-    needs: [ change-detection ]
+    needs: [ change-detection, set-provider-version ]
```
maastha: Can we remove `Get Last Release` step in these jobs since we are now already doing that in `set-provider-version `?

.github/workflows/test-suite.yml
```diff
@@ -6,31 +6,31 @@ on:
...
     strategy:
       max-parallel: 1
       fail-fast: false
       matrix:
-        terraform_version: ['${{ vars.TF_VERSION_LATEST }}', '1.0.8']
-    name: acc-tests-${{ matrix.terraform_version }}
+        terraform_version: ['', '1.0.8'] # '' for latest version
+        provider_version: ['', '1.11.1'] # '' for latest version
```
maastha, line 16: is it possible to avoid hardcoding these? can we use variables?

.github/workflows/migration-tests.yml
lantoli: you're totally right, removed

.github/workflows/test-suite.yml
lantoli: I think this is what I'm doing, if pass empty then acc/mig decides the latest version in a centralised point

lantoli: or do you mean for the other versions?  i think it's ok to have them in the matrix, it's even clearer than having to go to see variable values

.github/workflows/test-suite.yml
```diff
@@ -6,31 +6,31 @@ on:
...
+    name: mig-tests-${{ matrix.terraform_version || 'latest' }}-${{ matrix.provider_version || 'latest' }}
```
marcosuma, line 17: nit: in this case when you see the name it won't be intuitive which version is referring to what

marcosuma (approved):
one minor comment, LGTM

.github/workflows/test-suite.yml
maastha: hmmm, my only concern is this is easy to miss few months from now where we easily miss out on which versions are defaulted to and we are already at 1.13.1 so would have been nice if this was captured somewhere, but I don't have a strong opinion on this.

maastha (approved):
minor comment, non-blocking, LGTM

.github/workflows/test-suite.yml
lantoli: you're right, you have to go into the jobs.  but I prefered this so the logic to get the latest version is centralized, and normally the latest version is not so important to know as the old ones.

PR #1693, opened 2023-12-01
define common pagination schema, and share schema definition for data source implementations

AgustinBettati:
## Description

This PR shows 2 improvements we can make to simplify the schema definition of plural data sources:

- Define a template for pagination used in plural data sources: This will ensure consistency in our plural data source implementations and remove boilerplate from schema definitions.
- Use schema definition of individual data source in plural data source: This will make sure that the singular and plural data sources will always maintain alignment.

As an end result the schema definition of the plural data source for stream_instance would be the following:
```
resp.Schema = tfschema.PaginatedDSSchema(
		map[string]schema.Attribute{
				Required: true,
			},
		},
		DSAttributes(false))
```

As a downside, the general readability of the schema definition is impacted.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/dsschema/pagination_schema.go
+++ b/internal/common/dsschema/pagination_schema.go
@@ -1,0 +1,34 @@
+package dsschema
+
+import (
+	"maps"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func PaginatedDSSchema(arguments, resultAttributes map[string]schema.Attribute) schema.Schema {
+	result := schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"page_num": schema.Int64Attribute{
+				Optional: true,
+			},
+			"items_per_page": schema.Int64Attribute{
+				Optional: true,
+			},
+			"total_count": schema.Int64Attribute{
+				Computed: true,
+			},
+			"results": schema.ListNestedAttribute{
+				Computed: true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: resultAttributes,
+				},
+			},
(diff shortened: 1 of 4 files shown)


internal/common/schema/pagination_schema_test.go
```diff
@@ -0,0 +1,63 @@
...
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"instance_name": schema.StringAttribute{
+				Computed: true,
+			},
+		})
+
+	if !reflect.DeepEqual(resultSchema, expectedSchema) {
+		t.Errorf("created schema does not matched expected")
```
lantoli: knitpit: it's normally a good idea to show expected and received values when error to help understand the problem although they might be a bit long here

lantoli (approved):
i love it, great job!

internal/common/schema/pagination_schema.go
```diff
@@ -0,0 +1,34 @@
+package schema
```
lantoli: recommend to change package name to avoid conflicts with TF packages, e.g. dsschema, schemads, or similar, so we don't need to use package aliases

marcosuma:
I don't disagree with the proposal but wanted to hear your thoughts: if our goal is to possibly full automate resource and DS creation, would this still be that necessary?

AgustinBettati:
> if our goal is to possibly full automate resource and DS creation, would this still be that necessary?

When schema generation automation comes into picture our schema definitions would only depend on the API spec + custom configuration, making this irrelevant. This change would be a small improvement over our existing way of defining plural data source schemas until we reach that state.

AgustinBettati:
closing as will include within https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1701

PR #1718, opened 2023-12-07
chore: Uses official CLI GH action in TF clean-up workflow

lantoli:
## Description

Uses official CLI GH action in TF clean-up workflow.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-215921

## Type of change:

## Required Checklist:


.github/workflows/cleanup-test-env.yml
```diff
@@ -8,46 +8,40 @@ jobs:
...
-        with:
-          public-key: ${{ secrets.MONGODB_ATLAS_PUBLIC_KEY_CLOUD_DEV }}
-          private-key: ${{ secrets.MONGODB_ATLAS_PRIVATE_KEY_CLOUD_DEV }}
-          org-id: ${{ vars.MONGODB_ATLAS_ORG_ID_CLOUD_DEV }}
       - name: Checkout
         uses: actions/checkout@v4
         with:
           sparse-checkout: |
             scripts
+      - uses: mongodb/atlas-github-action@main
```
andreaangiolillo:
```suggestion
      - uses: mongodb/atlas-github-action@v0.2.0
```

.github/workflows/cleanup-test-env.yml
```diff
@@ -8,46 +8,40 @@ jobs:
...
+      - shell: bash
+        run: atlas --version
```
andreaangiolillo: Why do you need this?

lantoli: i found it interesting in case something fails to compare with local atlas version

.github/workflows/cleanup-test-env.yml
lantoli: @andreaangiolillo it's the recommended way in the instructions: https://github.com/marketplace/actions/atlas-cli-github-action

also I suppose being our GH action, it's not so a big (security) problem to use main.

@matt-condon WDYT is better to use here? (if v0.2.0 , should we change example / doc?

.github/workflows/cleanup-test-env.yml
lantoli: but you're right we won't normally use it, removed

.github/workflows/cleanup-test-env.yml
andreaangiolillo: Using master is a bad practice as we may push a bug unintentionally. I will update the description to be at least @latest (still not a best practice but at least an official release)

andreaangiolillo: https://github.com/mongodb/atlas-github-action/pull/13

lantoli: @andreaangiolillo i see your point of master having a bug even if not published.  keeping v0.2.0 at the moment as latest doesn't work.

matt-condon: using `@v0` should also be an option if you always want the latest published version without breaking changes

Edit: for clarity, would need to publish a new version

andreaangiolillo (approved):
LGTM

PR #1724, opened 2023-12-07
fix: Fixes migration tests for backup_online_archive

lantoli:
## Description

Fixes migration tests for backup_online_archive.

`data_expiration_rule` was introduced in 1.12.2 so it can't be used before.

It also refactors mig (migration) helper methods. 

Link to any related issue(s):

## Type of change:

## Required Checklist:


internal/testutil/mig/provider.go
```diff
@@ -0,0 +1,61 @@
...
+			Source:            "mongodb/mongodbatlas",
+		},
+		"aws": {
+			VersionConstraint: awsVersion,
+			Source:            "hashicorp/aws",
+		},
+	}
+}
+
+func ValueIfVersionAtLeast[T any](tb testing.TB, val T, minVersion string) T {
```
AgustinBettati:
```suggestion
func ConfiguredProviderVersionAtLeast(tb testing.TB,  minVersion string) bool
```
nit: might find it easier to follow if this function returns a boolean, and then each function that generates a configuration uses that boolean to their preference (passing default value, or using the boolean itself to define a set of attributes in the config).

AgustinBettati (approved):
Like this approach 💯 , will definitely help for defining migrations tests for new attributes or breaking changes.

internal/testutil/mig/pre_check.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/acc"
+)
+
+func PreCheckBasic(tb testing.TB) {
+	checkLastVersion(tb)
+	acc.PreCheckBasic(tb)
+}
```
gssbzn: you should consider enabling `thelper` in the linter, this and all test helpers should be 
```suggestion
func PreCheckBasic(tb testing.TB) {
	tb.Helper()
	checkLastVersion(tb)
	acc.PreCheckBasic(tb)
}
```
It will report the failing test correctly instead of flagging this function

lantoli: ohhh. right, thanks!

lantoli: @gssbzn is there a problem if tb.Helper is called multiple times?     e.g. one helper method calls another, but both can be called from tests.

gssbzn: >b.Helper is called multiple times? 

No internally it flags to not consider the line but the caller the failure which will bubble up if a helper calls a helper

.golangci.yml
```diff
@@ -98,6 +98,13 @@ linters:
     - unconvert
     - unused
     - whitespace
+    - thelper
```
lantoli, line 101: add some linters as in CLI

internal/testutil/mig/provider.go
lantoli: makes sense, changed

internal/testutil/mig/pre_check.go
lantoli: thx, changed

PR #1733, opened 2023-12-11
fix: project delete function uses correct string representation of id

oarbusi:
## Description

Project delete was using wrong string representation of Id, with quotes (`"projectId"`) instead of plain string (`projectId`). 

Within our project delete function we have a logic that waits for all "dependants" to be deleted.
During the SDK migration there was a change with how to obtain the project id that is used to list the cluster of a project, which is now passing an incorrect value. [This API call](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/project/resource_project.go#L761) uses nonNullProjectID of type types.String, so nonNullProjectID.String() was responding with "6575af27f93c7a6a4b50b239" resulting in an error by the SDK: "groupId must have less than 24 elements”. Due to whats seems like incorrect logic, this error simply lead to retrying constantly until eventually the timeout is reached.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/project/resource_project.go
+++ b/internal/service/project/resource_project.go
@@ -729,7 +729,7 @@ func deleteProject(ctx context.Context, connV2 *admin.APIClient, projectID strin
 	stateConf := &retry.StateChangeConf{
 		Pending:    []string{projectDependentsStateDeleting, projectDependentsStateRetry},
 		Target:     []string{projectDependentsStateIdle},
-		Refresh:    resourceProjectDependentsDeletingRefreshFunc(ctx, projectID, connV2),
+		Refresh:    ResourceProjectDependentsDeletingRefreshFunc(ctx, projectID, ServiceFromClient(connV2)),
 		Timeout:    30 * time.Minute,
 		MinTimeout: 30 * time.Second,
 		Delay:      0,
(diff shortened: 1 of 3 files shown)


internal/service/project/resource_project.go
```diff
@@ -269,7 +269,7 @@ func (r *projectRS) Create(ctx context.Context, req resource.CreateRequest, resp
 
 		_, _, err := connV2.TeamsApi.AddAllTeamsToProject(ctx, project.GetId(), NewTeamRoleList(ctx, teams)).Execute()
 		if err != nil {
-			errd := deleteProject(ctx, r.Client.AtlasV2, project.Id)
+			errd := deleteProject(ctx, r.Client.AtlasV2, project.GetId())
```
lantoli, line 272: can we have a test that fails before this change and passes after?

oarbusi: we now have. The call we mock in the test is only mocked with correct projectId, so in case the string has extra quotes, tests fail. See image
Screenshot 2023-12-11 at 11 14 26

lantoli: great, thx!

lantoli (approved):

internal/service/project/resource_project.go
```diff
@@ -755,17 +755,16 @@ Else consider the aggregate dependents idle.
...
+func ResourceProjectDependentsDeletingRefreshFunc(ctx context.Context, projectID string, client GroupProjectService) retry.StateRefreshFunc {
 	return func() (any, string, error) {
-		nonNullProjectID := conversion.StringPtrNullIfEmpty(projectID)
-		clusters, _, err := connV2.ClustersApi.ListClusters(ctx, nonNullProjectID.String()).Execute()
+		clusters, _, listClustersErr := client.ListClusters(ctx, projectID)
 		dependents := AtlasProjectDependants{AdvancedClusters: clusters}
 
-		if _, ok := admin.AsError(err); ok {
-			return nil, "", err
+		if _, ok := admin.AsError(listClustersErr); ok {
```
AgustinBettati: (this would imply removing redudant if statement below)
```suggestion
		if listClustersErr != nil {
```
curious if we want to unify both error checks, don't see a reason why for ApiError we would return error, while for others SDK error types keep on retrying. Would simply return error for an encountered error.

AgustinBettati (approved):
LGTM, one change to consider

internal/service/project/resource_project.go
oarbusi: I didn't want to change this to avoid changing behaviour, but I think it makes sense to unify it.

PR #1736, opened 2023-12-11
feat: new `mongodbatlas_stream_connection` resource

AgustinBettati:
## Description

Link to any related issue(s): [CLOUDP-215322](https://jira.mongodb.org/browse/CLOUDP-215322)

Includes:

- Implementation of mongodbatlas_stream_connection resource
- Acceptance tests
- Unit tests for model conversions

Will followup with PR for documentation and examples to not make this PR bigger than it already is.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code
- [x] If changes include deprecations or removals, I defined an isolated PR with a relevant title as it will be used in the auto-generated changelog.

PR #1737, opened 2023-12-12
refactor: avoid depending on existing project for CI tests in search index and stream connection resources

AgustinBettati:
## Description

Related to https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1736, left in separate PR so this change is easier to follow.

This improves the `GetClusterInfo` helper function that returns cluster info for acceptance tests. Specifically this function no longer relies on having an existing project beforehand, so our CI environment does not need to have fixed projects created.

## Type of change:

## Required Checklist:


.github/workflows/acceptance-tests.yml
```diff
@@ -237,7 +237,6 @@ jobs:
           MONGODB_ATLAS_ORG_ID: ${{ vars.MONGODB_ATLAS_ORG_ID_CLOUD_DEV }}
           MONGODB_ATLAS_BASE_URL: ${{ vars.MONGODB_ATLAS_BASE_URL }}
           MONGODB_ATLAS_ENABLE_BETA: ${{ vars.MONGODB_ATLAS_ENABLE_BETA }}
-          MONGODB_ATLAS_PROJECT_ID: ${{ vars.MONGODB_ATLAS_ACCTEST_STREAMS_PROJECT_ID_CLOUD_DEV }}
```
AgustinBettati, line 240: both search_index and stream_connection will not depend on having a fixed project created in our dev environment.

marcosuma (approved):

internal/service/searchindex/resource_search_index_test.go
```diff
@@ -23,30 +23,30 @@ const (
 
 func TestAccSearchIndexRS_basic(t *testing.T) {
 	var (
-		projectID                                        = os.Getenv("MONGODB_ATLAS_PROJECT_ID")
-		clusterName, clusterNameStr, clusterTerraformStr = acc.GetClusterInfo(projectID)
-		indexName                                        = acctest.RandomWithPrefix("test-acc-index")
-		databaseName                                     = acctest.RandomWithPrefix("test-acc-db")
+		orgID                                                          = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectIDStr, clusterName, clusterNameStr, clusterTerraformStr = acc.GetClusterInfo(orgID)
```
lantoli: would it be useful to return also the projectID (like clusterName) to check it in test?

internal/service/searchindex/resource_search_index_test.go
```diff
@@ -23,30 +23,30 @@ const (
...
 		CheckDestroy:             acc.CheckDestroySearchIndex,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccSearchIndexConfigBasic(projectID, indexName, databaseName, clusterNameStr, clusterTerraformStr, false),
+				Config: testAccSearchIndexConfigBasic(projectIDStr, indexName, databaseName, clusterNameStr, clusterTerraformStr, false),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckSearchIndexExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "name", indexName),
-					resource.TestCheckResourceAttr(resourceName, "project_id", projectID),
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
```
lantoli, line 41: we could check projectID if returned by GetClusterInfo

lantoli (approved):

internal/testutil/acc/cluster.go
```diff
@@ -7,18 +7,29 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
 )
 
-func GetClusterInfo(projectID string) (clusterName, clusterNameStr, clusterTerraformStr string) {
```
lantoli, line 10: knit: don't know if it's better to create a clusterInfo struct to be returned as we're getting more returned values

internal/testutil/acc/cluster.go
```diff
@@ -7,18 +7,29 @@ import (
...
+// GetClusterInfo is used to obtain a project and cluster configuration resource.
+// If env variables `MONGODB_ATLAS_CLUSTER_NAME` and `MONGODB_ATLAS_PROJECT_ID` are defined, creation of resources is avoided (useful for local execution)
+func GetClusterInfo(orgID string) (projectIDStr, clusterName, clusterNameStr, clusterTerraformStr string) {
 	// Allows faster test execution in local, don't use in CI
```
lantoli: knitpit: maybe integrate/delete/modify this internal comment with the function comments

internal/testutil/acc/cluster.go
lantoli: also not in this PR but as it's getting bigger it may be nice to have some tests for it

internal/service/searchindex/resource_search_index_test.go
AgustinBettati: We cannot do that for the case of project id because it is a computed attribute (obtained once config is actually applied). It is similar to acceptance tests that create projects as part of the config and cannot check the actual value of the project id.

lantoli: you're right, and I guess project name is not so useful

internal/testutil/acc/cluster.go
AgustinBettati: agree, will define a struct

internal/testutil/acc/cluster.go
AgustinBettati: makes sense, will adjust

PR #1738, opened 2023-12-12
test: add unit test to database_user resource

oarbusi:
## Description

Add unit test to database_user resource

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-214963

## Type of change:

## Required Checklist:


internal/service/databaseuser/resource_database_user.go
```diff
@@ -43,7 +40,7 @@ func Resource() resource.Resource {
 	}
 }
 
-type tfDatabaseUserModel struct {
```
lantoli, line 46: I think we should use RequiresReplace also for Roles and Labels, for the same reason we did for Scopes, until the SDK is changed.

and ideally create a test that fails before this change and passes after

lantoli (approved):

internal/service/databaseuser/resource_database_user.go
oarbusi: I didn't include it to Roles and Labels because in the old sdk it was not like this it has omitempty. I understand it makes sense to do it, but I think that is better not to change the behaviour just now, at least until we have a more permanent solution for this cases

internal/service/databaseuser/model_database_user.go
```diff
@@ -0,0 +1,229 @@
...
+		AWSIAMType:       types.StringValue(dbUser.GetAwsIAMType()),
+		Roles:            NewTFRolesModel(dbUser.Roles),
+		Labels:           NewTFLabelsModel(dbUser.Labels),
+		Scopes:           NewTFScopesModel(dbUser.Scopes),
+	}
+
+	return databaseUserModel, nil
+}
+
+func NewTFDatabaseUsersMode(ctx context.Context, projectID string, dbUsers []admin.CloudDatabaseUser) (*TfDatabaseUsersDSModel, diag.Diagnostics) {
```
marcosuma: nit: 
```suggestion
func NewTFDatabaseUsersModel(ctx context.Context, projectID string, dbUsers []admin.CloudDatabaseUser) (*TfDatabaseUsersDSModel, diag.Diagnostics) {
```

internal/service/databaseuser/model_database_user_test.go
```diff
@@ -0,0 +1,347 @@
...
+			currentScopes:  []*databaseuser.TfScopeModel{},
+			expectedResult: []admin.UserScope{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			resultModel := databaseuser.NewMongoDBAtlasScopes(tc.currentScopes)
+
+			if !reflect.DeepEqual(resultModel, tc.expectedResult) {
```
marcosuma: @lantoli  didn't you suggest to use a different comparison way? Not sure if this might be related

marcosuma (approved):

internal/service/databaseuser/model_database_user.go
oarbusi: Fixed, thanks!

internal/service/databaseuser/model_database_user_test.go
oarbusi: @marcosuma you are right! Forgot to use the assert from testify library. Changed it.

PR #1756, opened 2023-12-14
test: unit test for project_ip_access_list resource

oarbusi:
## Description

unit test for project_ip_access_list resource

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-214982

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/projectipaccesslist/model_project_ip_access_list_test.go
+++ b/internal/service/projectipaccesslist/model_project_ip_access_list_test.go
@@ -1,0 +1,215 @@
+package projectipaccesslist_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/projectipaccesslist"
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/atlas-sdk/v20231115002/admin"
+)
+
+var (
+	AWSSecurityGroup = "AWSSecurityGroup"
+	CIDRBlock        = "CIDRBlock"
+	IPAddress        = "IPAddress"
+	Comment          = "Comment"
+	GroupID          = "GroupID"
+	IPAddressID      = conversion.EncodeStateID(map[string]string{
+		"entry":      IPAddress,
+		"project_id": GroupID,
+	})
+	CidrBlockID = conversion.EncodeStateID(map[string]string{
+		"entry":      CIDRBlock,
+		"project_id": GroupID,
+	})
+	AwsSecurityGroupID = conversion.EncodeStateID(map[string]string{
+		"entry":      AWSSecurityGroup,
+		"project_id": GroupID,
+	})
+)
+
+func TestNewMongoDBProjectIPAccessList(t *testing.T) {
+	testCases := []struct {
+		tfModel        *projectipaccesslist.TfProjectIPAccessListModel
+		expectedResult *[]admin.NetworkPermissionEntry
+		name           string
+	}{
+		{
+			name: "NewMongoDBProjectIPAccessList",
+			tfModel: &projectipaccesslist.TfProjectIPAccessListModel{
+				AWSSecurityGroup: types.StringValue(AWSSecurityGroup),
+				CIDRBlock:        types.StringValue(CIDRBlock),
+				IPAddress:        types.StringValue(IPAddress),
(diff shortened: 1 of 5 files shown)


internal/service/projectipaccesslist/resource_project_ip_access_list.go
```diff
@@ -200,50 +200,15 @@ func (r *projectIPAccessListRS) Create(ctx context.Context, req resource.CreateR
...
-		ProjectID:        types.StringValue(projectIPAccessList.GetGroupId()),
-		CIDRBlock:        types.StringValue(projectIPAccessList.GetCidrBlock()),
-		IPAddress:        types.StringValue(projectIPAccessList.GetIpAddress()),
-		AWSSecurityGroup: types.StringValue(projectIPAccessList.GetAwsSecurityGroup()),
-		Comment:          types.StringValue(projectIPAccessList.GetComment()),
-		Timeouts:         projectIPAccessListModel.Timeouts,
-	}
-}
-
-func newMongoDBProjectIPAccessList(projectIPAccessListModel *tfProjectIPAccessListModel) *[]admin.NetworkPermissionEntry {
```
lantoli, line 234: is it ok to assume projectIPAccessListModel is never nil?

internal/service/projectipaccesslist/model_project_ip_access_list_test.go
```diff
@@ -0,0 +1,215 @@
...
+		"project_id": GroupID,
+	})
+)
+
+func TestNewMongoDBProjectIPAccessList(t *testing.T) {
+	testCases := []struct {
+		tfModel        *projectipaccesslist.TfProjectIPAccessListModel
+		expectedResult *[]admin.NetworkPermissionEntry
+		name           string
+	}{
```
lantoli, line 40: i suppose it's ok to follow the same structure even if there is only one test

lantoli (approved):

internal/service/projectipaccesslist/resource_project_ip_access_list.go
oarbusi: yes, it will never be null. Also, it's attributes `AWSSecurityGroup`,  `CIDRBlock` and `IPAddress` are also checked

internal/service/projectipaccesslist/model_project_ip_access_list_test.go
oarbusi: Kept it like this for consistency even though it's not needed

AgustinBettati: think this is useful as it simplifies when a new test is added

internal/service/projectipaccesslist/model_project_ip_access_list_test.go
```diff
@@ -0,0 +1,215 @@
...
+			sdkModel: &admin.NetworkPermissionEntry{
+				IpAddress: &IPAddress,
+				Comment:   &Comment,
+				GroupId:   &GroupID,
+			},
+			expectedResult: &projectipaccesslist.TfProjectIPAccessListModel{
+				ID:               types.StringValue(IPAddressID),
+				ProjectID:        types.StringValue(GroupID),
+				AWSSecurityGroup: types.StringValue(""),
+				CIDRBlock:        types.StringValue(""),
```
AgustinBettati, line 90: Interesting that we are storing empty strings instead of defining terraform null value. Don't believe we need to take immediate action here, but nice to see how unit testing evidences this behaviour.

AgustinBettati (approved):
LGTM, small observation

internal/service/projectipaccesslist/model_project_ip_access_list_test.go
oarbusi: Yes its an interesting behaviour made explicit by the unit test. In this case, these attributes have `omitempty` (see [docs](https://pkg.go.dev/go.mongodb.org/atlas-sdk/v20231115002@v20231115002.1.0/admin#NetworkPermissionEntry)) so in this case when sent to the API they would be ignored (only one of `AWSSecurityGroup`, `CIDRBlock`, `IPAddress` can be set at the same time, as explained in the docs: `Your IP access list entry can be one **awsSecurityGroup**, one **cidrBlock**, or one **ipAddress**.` )

AgustinBettati: actually given that these fields are pointers, the `omitempty` will work by including the property in the json if the pointer is not `nil` (even if it is a pointer to ""). In this case we are not including in the request because in https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/projectipaccesslist/resource_project_ip_access_list.go#L237 we define the field as nil if state has an empty string.
TLDR: it is working well, but if we use proper terraform null values the state would be more "correct", and simplify handling these fields in model conversions.

PR #1761, opened 2023-12-15
chore: update the alert notifications regions for Datadog

jfmainville:
Since at least the version 2023-01-01 of the MongoDB Atlas REST API (https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/2023-01-01/#tag/Alert-Configurations/operation/createAlertConfiguration), there are more than 2 regions available with Datadog to send alerts.

Added the additional supported regions (US3 and US5) to the list for the `datadog_region` attribute.

### DIFF
--- a/internal/service/alertconfiguration/resource_alert_configuration.go
+++ b/internal/service/alertconfiguration/resource_alert_configuration.go
@@ -260,9 +260,6 @@ func (r *alertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						},
 						"datadog_region": schema.StringAttribute{
 							Optional: true,
-							Validators: []validator.String{
-								stringvalidator.OneOf("US", "EU", "US3", "US5"),
-							},
 						},
 						"delay_min": schema.Int64Attribute{
 							Optional: true,
--- a/internal/service/alertconfiguration/resource_alert_configuration.go
+++ b/internal/service/alertconfiguration/resource_alert_configuration.go
@@ -261,7 +261,7 @@ func (r *alertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						"datadog_region": schema.StringAttribute{
 							Optional: true,
 							Validators: []validator.String{
-								stringvalidator.OneOf("US", "EU"),
+								stringvalidator.OneOf("US", "EU", "US3", "US5"),
 							},
 						},
 						"delay_min": schema.Int64Attribute{
--- a/website/docs/r/alert_configuration.html.markdown
+++ b/website/docs/r/alert_configuration.html.markdown
@@ -195,7 +195,7 @@ List of notifications to send when an alert condition is detected.
 * `api_token` - Slack API token. Required for the SLACK notifications type. If the token later becomes invalid, Atlas sends an email to the project owner and eventually removes the token.
 * `channel_name` - Slack channel name. Required for the SLACK notifications type.
 * `datadog_api_key` - Datadog API Key. Found in the Datadog dashboard. Required for the DATADOG notifications type.
-* `datadog_region` - Region that indicates which API URL to use. Accepted regions are: `US`, `EU`. The default Datadog region is US.
+* `datadog_region` - Region that indicates which API URL to use. Accepted regions are: `US`, `EU`, `US3`, `US5`. The default Datadog region is US.
 * `delay_min` - Number of minutes to wait after an alert condition is detected before sending out the first notification.
 * `email_address` - Email address to which alert notifications are sent. Required for the EMAIL notifications type.
 * `email_enabled` - Flag indicating email notifications should be sent. This flag is only valid if `type_name` is set to `ORG`, `GROUP`, or `USER`.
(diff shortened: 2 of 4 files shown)


website/docs/r/alert_configuration.html.markdown
```diff
@@ -195,7 +195,7 @@ List of notifications to send when an alert condition is detected.
 * `api_token` - Slack API token. Required for the SLACK notifications type. If the token later becomes invalid, Atlas sends an email to the project owner and eventually removes the token.
 * `channel_name` - Slack channel name. Required for the SLACK notifications type.
 * `datadog_api_key` - Datadog API Key. Found in the Datadog dashboard. Required for the DATADOG notifications type.
-* `datadog_region` - Region that indicates which API URL to use. Accepted regions are: `US`, `EU`. The default Datadog region is US.
+* `datadog_region` - Region that indicates which API URL to use. Accepted regions are: `US`, `EU`, `US3`, `US5`. The default Datadog region is US.
```
marcosuma: @jfmainville thank you very much for your contribution! We also need to update the `d` documentation:

* website/docs/d/alert_configurations.html.markdown
* website/docs/d/alert_configuration.html.markdown

jfmainville: @marcosuma Tanks for the feedback! I added the missing references in the documentation for the `datadog_region` attribute

internal/service/alertconfiguration/resource_alert_configuration.go
```diff
@@ -261,7 +261,7 @@ func (r *alertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						"datadog_region": schema.StringAttribute{
 							Optional: true,
 							Validators: []validator.String{
-								stringvalidator.OneOf("US", "EU"),
+								stringvalidator.OneOf("US", "EU", "US3", "US5"),
```
AgustinBettati: @marcosuma thinking more long term, we could remove these client-side validations entirely if the API is returning proper error messages for incorrect values. (while also having the documentation reference the Atlas API docs to obtain supported values)

andreaangiolillo: I agree on removing the validation on the provider for fields that have multiple values (3 + and we know will have more values in the future). We did the same for the cluster region field

marcosuma: I am ok with that, @AgustinBettati @andreaangiolillo do we leave the documentation with the values though?

andreaangiolillo: I don't think we should. We would be forced to update it all the time that the endpoint add/remove a value. Instead, we should add a link to the endpoint documentation in the field description

AgustinBettati: Agree, this is an example https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/website/docs/r/search_deployment.html.markdown?plain=1#L44 but we have several others in our docs.

marcosuma: FYI @jfmainville - tl;dr: I think we'd prefer to remove this validation completely and update the documentation by referencing Datadog API supported values.

We can do two things:
* you make these changes
* we cancel this PR and we make the change on our side 

Either way should work, FYI @AgustinBettati @andreaangiolillo

jfmainville: @marcosuma I can make the changes, no issues, I'll remove the validation and add a link to the MongoDB Atlas API that shows the supported values for the Datadog regions (to have a reference). Thanks!

jfmainville: Also, to make sure that I comply with the appropriate structure for the documentation, is this line valid for the `datadog_region` attribute?

```
`datadog_region` - Region that indicates which API URL to use. See the `notifications` body parameter in the [MongoDB API Alert Configuration documentation](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Alert-Configurations/operation/createAlertConfiguration) for the details. The default Datadog region is US.`
```

marcosuma: Thank so much @jfmainville, we all really appreciate your commitment here, and sorry for late reply. I would say something like:

```
`datadog_region` - Region that indicates which API URL to use. See the `datadogRegion` field in the `notifications` request parameter of [MongoDB API Alert Configuration documentation](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Alert-Configurations/operation/createAlertConfiguration) for more details. The default Datadog region is US.
```

checking in with @andreaangiolillo @AgustinBettati so to make sure you don't have to do other changes later and we agree on it here.

AgustinBettati: SGTM, thank you @jfmainville for following up on this.

jfmainville: Awesome, thanks for the details! I updated the codebase accordingly, let me know if additional changes are required. Thanks!

andreaangiolillo (approved):
LGTM. 💯 Thanks for all the changes here

andreaangiolillo:
FYI: We plan to release the provider later today so this PR will be merged after the release.

jfmainville:
Great, thanks for the update!

marcosuma:
Thank you @jfmainville, great effort! We really appreciate your support and contribution here 🙏

marcosuma (approved):

PR #1769, opened 2023-12-18
chore: Runs tests for latest minor versions of each major TF version

lantoli:
## Description

Runs tests for latest minor versions of each major TF version.

- Allows to run manually Test Suite for a matrix of versions, by default it has the latest minor versions of each major TF version.
- Concurrency is limited to one Test Suite run at a time. You can always stop any run you don't need any more.
- However concurrency for Acceptance or Migration workflows (run manually or from a PR) is not limited.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-218618

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -4,10 +4,10 @@ on:
   workflow_dispatch:
     inputs:
       terraform_matrix:
-        description: 'Terraform matrix version (JSON array)'
+        description: 'Terraform version matrix (JSON array)'
         default: '["1.6.x", "1.5.x", "1.4.x", "1.3.x", "1.2.x"]'
       provider_matrix:
-        description: 'MongoDB Atlas Provider matrix version (JSON array)'
+        description: 'MongoDB Atlas Provider version matrix (JSON array)'
         default: '["", "1.12.3"]' # "" for latest version
 
   schedule:
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -7,7 +7,7 @@ on:
         description: 'Terraform version matrix (JSON array)'
         default: '["1.6.x", "1.5.x", "1.4.x", "1.3.x", "1.2.x"]'
       provider_matrix:
-        description: 'MongoDB Atlas Provider version matrix (JSON array)'
+        description: 'MongoDB Atlas Provider version matrix for running migration tests (JSON array)'
         default: '["", "1.12.3"]' # "" for latest version
 
   schedule:
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -12,30 +12,37 @@ on:
 
   schedule:
     - cron: "0 0 * * *" # workflow runs every day at midnight UTC
-  
+   
 concurrency:
   group: '${{ github.workflow }}'
   cancel-in-progress: false
- 
-env:
-    schedule_terraform_matrix: '["1.6.x", "1.0.8"]'
-    schedule_provider_matrix: '["", "1.12.3]' # "" for latest version
-  
+   
 jobs:
+  versions:
+    env:
+      schedule_terraform_matrix: '["1.6.x", "1.0.8"]'
+      schedule_provider_matrix: '["", "1.12.3]' # "" for latest version    
+    runs-on: ubuntu-latest
+    outputs:
+      terraform_matrix: ${{ inputs.terraform_matrix || vars.schedule_terraform_matrix }}
+      provider_matrix: ${{ inputs.provider_matrix || vars.schedule_provider_matrix }}
+    steps:
+      - if: false
+        run: echo jobs need steps
 
   clean-before:
     secrets: inherit
     uses: ./.github/workflows/cleanup-test-env.yml
 
   mig-tests:
-    needs: clean-before
+    needs: [clean-before, versions]
     if: ${{ !cancelled() }}
     strategy:
       max-parallel: 1
       fail-fast: false
       matrix:
-        terraform_version: ['', '1.0.8'] # '' for latest version
-        provider_version: ['', '1.11.1'] # '' for latest version
+        terraform_version: ${{ fromJSON(needs.versions.outputs.terraform_matrix) }}
+        provider_version: ${{ fromJSON(needs.versions.outputs.provider_matrix) }}
     name: mig-tests-${{ matrix.terraform_version || 'latest' }}-${{ matrix.provider_version || 'latest' }}
     secrets: inherit
     uses: ./.github/workflows/migration-tests.yml
@@ -44,13 +51,13 @@ jobs:
       provider_version: ${{ matrix.provider_version }}
 
   acc-tests:
-    needs: mig-tests
+    needs: [mig-tests, versions]
     if: ${{ !cancelled() }}
     strategy:
       max-parallel: 1
       fail-fast: false
       matrix:
-        terraform_version: ['', '1.0.8'] # '' for latest version
+        terraform_version: ${{ fromJSON(needs.versions.outputs.terraform_matrix) }}
     name: acc-tests-${{ matrix.terraform_version || 'latest' }}
     secrets: inherit
     uses: ./.github/workflows/acceptance-tests.yml
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -5,6 +5,10 @@ on:
   schedule:
     - cron: "0 0 * * *" # workflow runs every day at midnight UTC
   
+concurrency:
+  group: '${{ github.workflow }}'
+  cancel-in-progress: false
+    
 jobs:
 
   clean-before:
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -2,13 +2,25 @@ name: 'Test Suite'
 
 on:
   workflow_dispatch:
+    inputs:
+      terraform_matrix:
+        description: 'Terraform matrix version (JSON array)'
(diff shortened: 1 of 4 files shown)


.github/workflows/test-suite.yml
```diff
@@ -2,24 +2,47 @@ name: 'Test Suite'
 
 on:
   workflow_dispatch:
+    inputs:
+      terraform_matrix:
+        description: 'Terraform matrix version (JSON array)'
+        default: '["1.6.x", "1.5.x", "1.4.x", "1.3.x", "1.2.x"]'
```
lantoli, line 8: I checked that this notation works, e.g. 1.4.x will run version 1.4.7. 
TF version list is here: https://releases.hashicorp.com/terraform/

.tool-versions
```diff
@@ -1 +1,2 @@
 golang 1.21.3
+terraform 1.6.6
```
lantoli, line 2: opportunistic change not related to the main goal of the PR

.github/workflows/test-suite.yml
```diff
@@ -2,24 +2,47 @@ name: 'Test Suite'
...
+        description: 'MongoDB Atlas Provider version matrix (JSON array)'
+        default: '["", "1.12.3"]' # "" for latest version
+
   schedule:
     - cron: "0 0 * * *" # workflow runs every day at midnight UTC
-  
+   
+concurrency:
+  group: '${{ github.workflow }}'
+  cancel-in-progress: false
```
lantoli, line 18: only one Test Suite run at a time to avoid interference. new runs will wait until previous ones finish

.github/workflows/test-suite.yml
```diff
@@ -2,24 +2,47 @@ name: 'Test Suite'
...
+   
 jobs:
+  versions:
+    env:
+      schedule_terraform_matrix: '["1.6.x", "1.0.8"]'
+      schedule_provider_matrix: '["", "1.12.3]' # "" for latest version
```
lantoli, line 24: schedule triggers can't have input parameters so we define the values we want here

.github/workflows/test-suite.yml
```diff
@@ -2,24 +2,47 @@ name: 'Test Suite'
 
 on:
   workflow_dispatch:
+    inputs:
+      terraform_matrix:
+        description: 'Terraform version matrix (JSON array)'
+        default: '["1.6.x", "1.5.x", "1.4.x", "1.3.x", "1.2.x"]'
+      provider_matrix:
+        description: 'MongoDB Atlas Provider version matrix (JSON array)'
+        default: '["", "1.12.3"]' # "" for latest version
```
marcosuma, line 11: should we still have 1.12.3 hardcoded here? I think what we want ultimately is that our latest version is compatible... right @Zuhairahmed ?

.github/workflows/test-suite.yml
```diff
@@ -2,24 +2,47 @@ name: 'Test Suite'
...
+
   schedule:
     - cron: "0 0 * * *" # workflow runs every day at midnight UTC
```
marcosuma, line 14: can we do longer frequency? every month?

.github/workflows/test-suite.yml
lantoli: @marcosuma Zuhair told me to check our latest 2 minor patch versions, "" is the latest one (currently 1.13.1).
for example once we release 1.14.0,   we'll have to change this to:  "", "1.13.1", where "" will be 1.14.0

.github/workflows/test-suite.yml
lantoli: this frequency is for:
     - schedule_terraform_matrix: '["1.6.x", "1.0.8"]'
     - schedule_provider_matrix: '["", "1.12.3]' # "" for latest version  

the longer version test has to be run manually, for instance we can do it monthly or before a release or a major release.

.github/workflows/test-suite.yml
```diff
@@ -2,24 +2,47 @@ name: 'Test Suite'
 
 on:
   workflow_dispatch:
+    inputs:
+      terraform_matrix:
+        description: 'Terraform version matrix (JSON array)'
+        default: '["1.6.x", "1.5.x", "1.4.x", "1.3.x", "1.2.x"]'
+      provider_matrix:
+        description: 'MongoDB Atlas Provider version matrix (JSON array)'
```
AgustinBettati:
```suggestion
        description: 'MongoDB Atlas Provider version matrix for running migration tests (JSON array)'
```
WDYT about making it more explicit? Without too much context you can understand that the provider version used for running tests is defined here.

AgustinBettati (approved):
LGTM, one detail

.github/workflows/test-suite.yml
lantoli: sounds good, suggestion accepted

oarbusi (approved):

.github/workflows/test-suite.yml
marcosuma: I don't understand why 😅 and also this is something we'll have to manually update too often, every time we make a release (yet one more thing to do manually). Unless there is a strong reason (let's wait for Zuhair) that I am missing let's please remove it.

.github/workflows/test-suite.yml
marcosuma: If we look at [the requirement](https://docs.google.com/document/d/1Z5WnGGdDKpae9azaqJvWZgqbZuMouRHCly8xzcarKAs/edit?disco=AAABCnhRna4) we discussed yesterday, we want to test our latest version against the terraform versions that we declare we support. This was also stated in one of [the document](https://docs.google.com/document/d/1YRorIH4UdW0yIxoSRuoZg-DP0dAmBhF3RDOp5O5Sb4o/edit#heading=h.v0fmnq3qwxa) Zuhair has recently proposed.

So my ask is: can we please have a monthly-based testing that checks **our latest provider version** against `'["1.6.x", "1.5.x", "1.4.x", "1.3.x", "1.2.x"]'`. I'd prefer not to do this manually as we have already lots of things we do manually. 

Let me know if I am missing something

.github/workflows/test-suite.yml
Zuhairahmed: Feedback is to test last 2 Majors of Terraform Atlas Provider versions (BigMajor.Major.Minor format) so today the latest minor patch on each would be v1.13.1 and v1.12.3. This is to get ahead of any unintended breaking changes before our customers were to identify them so we can do this by ensuring the same tests pass last 2 Major versions. Would however explore ways to prevent hardcoding here if possible.

marcosuma: is it just me or I am really not seeing the value in this?

Once you've gone over 1.12.3, what is the point in keep testing the second last version? That will never change because we will never touch it.

To me it's more the extra manual effort we are introducing in having to edit the last version every time than the benefit of this

@lantoli @Zuhairahmed

Zuhairahmed: would consider the HashiCorp Terraform testing scenarios where: 
1) acceptance tests fail on Terraform Atlas Provider v1.13.1 and pass on v1.12.3   
2) acceptance tests fail on both Terraform Atlas Provider v1.13.1 and v1.12.3

if we only tested most recent version all we would know for certain is that tests have failed on v1.13.1 and we need to investigate why this occurred which can be from multiple sources. however by running testing on most 2 recent recent version of Terraform Atlas Provider we help accelerate triage of finding a bug by narrowing scope of issue. if scenario 1 one occurs, then we know that error is something that we introduced with most recent changes catching breaking changes ahead of customers. if it's scenario 2, then we know that issue most likely has to do with change made on HashiCorp Terraform side (for example they released bug) so we can then escalate to HashiCorp thereby saving time with bug resolution.  

Does this help? Agree again that we should be automating/avoiding hardcodes as much of this as possible.

.github/workflows/test-suite.yml
lantoli: we could but the problem I see is that somebody has to look at the results, and if it's done once per month I see it very easy to miss that execution result.  Also I find it interesting to run it at other occasions like before a release.

this is also happening with the daily Test Suite.  We run them automatically every night, but you have to go to the GH UI to see the results. 

If we run something monthly I think that execution will probably be missed. 
I don't see the benefit of running it monthly automatically.   I think it's better something like I did today: I run them manually and then I look at the results.

lantoli: this is also happening with the daily Test Suite.  We run them automatically every night, but you have to go the GH UI to see the results.

lantoli: that said, I you still think it's a good idea please confirm and I'll add the monthly schedule

.github/workflows/test-suite.yml
marcosuma: _removing previous comment_

Ok, I got some updates from Leo offline, so I think the main misunderstanding here is that these are migration tests, hence 1.12.3 is used to **first create the resource** and then "latest" is used to check if there are no plan changes.

Let us discuss this offline.

.github/workflows/test-suite.yml
marcosuma: On this, I don't have a strong opinion, I am ok to run it manually or before a release. @gssbzn  how do we do it for AtlasCLI?

gssbzn: > how do we do it for AtlasCLI?

We always run packaging on master (we call it the waterfall on evergreen), this means after every merge, and then all the OS compatibility depends on packaging succeeding. The compatibility matrix exploded on the CLI at some point and evergreen let's us generate test programmatically so we write the yaml via code [here](https://github.com/mongodb/mongodb-atlas-cli/blob/master/tools/genevergreen/main.go)

PR #1789, opened 2023-12-21
chore: Allows to to run acceptance tests for a published Atlas provider version

lantoli:
## Description

Allows to to run acceptance tests for a published Atlas provider version.

This allows to run acceptance test for already published MongoDB Atlas provider releases for a Terraform CLI version and check that it's compatible.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-219089

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -36,7 +36,6 @@ jobs:
 
   acc-tests-latest-published:
     needs: clean-before
-    if: ${{ github.event_name == 'schedule' && !cancelled() }}
     secrets: inherit
     uses: ./.github/workflows/acceptance-tests.yml
     with:
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -34,8 +34,16 @@ jobs:
     secrets: inherit
     uses: ./.github/workflows/cleanup-test-env.yml
 
+  acc-tests-latest-published:
+    needs: clean-before
+    if: ${{ github.event_name == 'schedule' && !cancelled() }}
+    secrets: inherit
+    uses: ./.github/workflows/acceptance-tests.yml
+    with:
+      ref: 'v1.14.0'
+
   mig-tests:
-    needs: [clean-before, versions]
+    needs: [acc-tests-latest-published, versions]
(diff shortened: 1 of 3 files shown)


.github/workflows/test-suite.yml
```diff
@@ -34,8 +34,16 @@ jobs:
     secrets: inherit
     uses: ./.github/workflows/cleanup-test-env.yml
 
+  acc-tests-latest-published:
+    needs: clean-before
+    if: ${{ github.event_name == 'schedule' && !cancelled() }}
+    secrets: inherit
+    uses: ./.github/workflows/acceptance-tests.yml
+    with:
+      ref: 'v1.14.0'
```
lantoli: pending to calculate instead of hardcoding

oarbusi, line 43: Will we need to change this value with every release?

.github/workflows/test-suite.yml
```diff
@@ -34,8 +34,16 @@ jobs:
     secrets: inherit
     uses: ./.github/workflows/cleanup-test-env.yml
 
+  acc-tests-latest-published:
```
AgustinBettati: I am struggling to see the value of adding this job to run daily, given that the results will be the same unless there is a new patch release from TF. 
If you can provide more context in the PR or reference where this requirement was discussed that would be great. I am thinking why we cannot rely on our master branch to see if the latest patch version of TF is breaking some functionality in our provider.

lantoli: the use case here is that Hashicorp releases a new version, e.g. 1.6.7 and breaks our latest published version, e.g. 1.14.0.    it's true that results won't change while Hashicorp doesn't publish a new version. But the alternative is that we have to be pending (or automate?) when a new version is published and then run on demand.

lantoli: @Zuhairahmed what do you think?
alternatively we can do that, don't run every night but on demand when we know that a new Hashicorp version came out.

lantoli: @AgustinBettati at the moment I'll leave only the option to run it manually

AgustinBettati (approved):
LGTM

PR #1792, opened 2023-12-27
test: Fixes tests after updating terraform-plugin-testing from 1.5.1 to 1.6.0

AgustinBettati:
## Description

Link to any related issue(s): [CLOUDP-219141](https://jira.mongodb.org/browse/CLOUDP-219141)

This PR uses the [dependabot PR](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1743) as the base branch which is updating `terraform-plugin-testing` from 1.5.1 to 1.6.0.
Due to some changes introduced in this new version some of our tests where impacted (seen in the following [CI execution](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/7274953242/job/19824564091) that was triggered manually):
- (caused by 1) TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_maintainsBackwardCompatibility
- (caused by 2) TestAccFederatedDatabaseInstance_basic
- (caused by 2) TestAccConfigRSCustomDBRoles_importBasic

The failing test are due to 2 causes.

### 1. Changes in the behaviour of `PlanOnly`

I raised an issue to HashiCorp with the details of the failing test and guidance was provided: https://github.com/hashicorp/terraform-plugin-testing/issues/256.
In summary, the failing case defines a migration test from an old provider version to the latest, and during these version one attribute is removed from the schema. After `terraform-plugin-testing` v1.6.0 when using `PlanOnly` the state is not automatically updated with the current provider schema before the destroy begins, leading to the error.
The suggested way moving forward is to avoid using `PlanOnly: true`, and instead use [plancheck.ExpectEmptyPlan()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/plancheck#ExpectEmptyPlan). This is only needed when the underlying provider schema changes between TestSteps.

### 2. Changes in the behaviour of `ImportState`

Examples of the failing test:
```
--- FAIL: TestAccConfigRSCustomDBRoles_importBasic (18.63s)
    /Users/agustin.bettati/mongodb/terraform-provider-mongodbatlas/internal/service/customdbrole/resource_custom_db_role_test.go:442: Step 2/3 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        + 	"actions.0.resources.0.cluster": "false",
          }
```

As seen in the message, after `terraform-plugin-testing` v1.6.0 the `ImportStateVerify: true,` config verifies if there are attributes that are present in the state, but not in the config. Before this version this check only verified missing attributes in the state. I verified that in both cases these additional attributes do not cause a non-empty plan (and also added in test cases), and add these additional attribute in the `ImportStateVerifyIgnore` array.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/cluster/resource_cluster_test.go
+++ b/internal/service/cluster/resource_cluster_test.go
@@ -1450,7 +1450,6 @@ func TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_main
 				Config:                   cfg,
 				ConfigPlanChecks: resource.ConfigPlanChecks{
 					PostApplyPreRefresh: []plancheck.PlanCheck{
-						acc.DebugPlan(),
 						plancheck.ExpectEmptyPlan(),
 					},
 				},
--- a/internal/service/cluster/resource_cluster_test.go
+++ b/internal/service/cluster/resource_cluster_test.go
@@ -1451,9 +1451,9 @@ func TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_main
 				ConfigPlanChecks: resource.ConfigPlanChecks{
 					PostApplyPreRefresh: []plancheck.PlanCheck{
 						acc.DebugPlan(),
+						plancheck.ExpectEmptyPlan(),
 					},
 				},
-				PlanOnly: true,
 			},
 		},
 	})
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/hashicorp/terraform-plugin-mux v0.13.0
 	github.com/hashicorp/terraform-plugin-sdk v1.17.2
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.31.0
-	github.com/hashicorp/terraform-plugin-testing v1.5.1
+	github.com/hashicorp/terraform-plugin-testing v1.6.0
 	github.com/mongodb-forks/digest v1.0.5
 	github.com/mwielbut/pointy v1.1.0
 	github.com/spf13/cast v1.6.0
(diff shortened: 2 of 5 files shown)


AgustinBettati:
failing test in cluster group (TestAccClusterRSCluster_tenant) is not related, will followup with a CLOUDP ticket for that issue as I am also seeing it in QA executions (flaky scenario). Also ran [all acceptance tests/migrations test](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/7337518724) to verify no other failing test was left out after the fix.

internal/service/cluster/resource_cluster_test.go
```diff
@@ -1451,9 +1451,9 @@ func TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_main
 				ConfigPlanChecks: resource.ConfigPlanChecks{
 					PostApplyPreRefresh: []plancheck.PlanCheck{
 						acc.DebugPlan(),
+						plancheck.ExpectEmptyPlan(),
 					},
 				},
-				PlanOnly: true,
```
andreaangiolillo, line 1456: Should we do a quick search/replace of `PlanOnly` in the repo? 👀

andreaangiolillo, line 1456: [not-blocking] Should we document these errors somewhere so that we do not forget them in the future? I am not sure about an ideal location 🤔  leave this here to think about it

andreaangiolillo (approved):
LGTM

internal/service/cluster/resource_cluster_test.go
```diff
@@ -1451,9 +1451,9 @@ func TestAccClusterRSCluster_withDefaultBiConnectorAndAdvancedConfiguration_main
 				ConfigPlanChecks: resource.ConfigPlanChecks{
 					PostApplyPreRefresh: []plancheck.PlanCheck{
 						acc.DebugPlan(),
```
maastha: do we still need this as well?

internal/service/cluster/resource_cluster_test.go
AgustinBettati: sorry I did not understand this point

AgustinBettati: Our terraform wiki can be good starting point, helpful for referencing in the future if it is encountered again.

internal/service/cluster/resource_cluster_test.go
AgustinBettati: thanks for noting, not necessary as ExpectEmptyPlan() gives details of the plan if it is non empty.

internal/service/cluster/resource_cluster_test.go
andreaangiolillo: just updated the message. I am wondering if we should search for `PlanOnly` in the repo and remove it

AgustinBettati: yeah, was waiting for confirmation [here](https://github.com/hashicorp/terraform-plugin-testing/issues/256#issuecomment-1870281163) as the migration test documentation still mentions `PlanOnly` usage. Can do this in a followup PR.

PR #1812, opened 2024-01-05
chore: Adjusts release process to run tests against QA before releasing

AgustinBettati:
## Description

Link to any related issue(s): [CLOUDP-220682](https://jira.mongodb.org/browse/CLOUDP-220682)

Used a [forked repository](https://github.com/AgustinBettati/terraform-provider-mongodbatlas/actions) to test logic of when acceptances test and release is run.

Scenarios that were tested:
- When providing an invalid input (1.0.0): execution of tests and release is skipped
- When doing a release (v1.0.0) or a pre release (v1.0.0-pre):
  - with skip value `true` -> tests are not run and release is made
  - with skip value empty -> tests are run
    - if test fails release job is not run
    - if test are successful release job runs

What has not been tested and will be tested on release?
- During testing in the forked repository execution of acceptance tests was replaced with a dummy job, so the call to the acceptance test workflow it not tested (was copied from working definition in [test-suite.yml#L63](https://github.com/AgustinBettati/terraform-provider-mongodbatlas/blob/testing-release-logic/.github/workflows/test-suite.yml#L63).

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,16 +1,39 @@
-name: release
+name: 'New Release'
+run-name: 'Release ${{ inputs.version_number }}'
+
 on:
-  push:
-    tags:
-      - 'v*'
-jobs: 
-  goreleaser:
+  workflow_dispatch:
+    inputs:
+      version_number:
+        description: 'Version number (e.g., v1.0.0, v1.0.0-pre, v1.0.0-pre1)'
+        required: true
+      skip_tests:
+        description: 'Define value to `true` to explicitly skip QA acceptance tests execution'
+
+jobs:
+    # QA acceptance tests are skipped for pre releases or when explicit input parameter is used
+  run-qa-acceptance-tests:
+    if: inputs.skip_tests != 'true' && !contains(inputs.version_number, 'pre')
+    secrets: inherit
+    uses: ./.github/workflows/acceptance-tests.yml
+    with:
+      atlas_cloud_env: "qa"
+
+  release:
     runs-on: ubuntu-latest
+    needs: [ run-qa-acceptance-tests ]
+    if: ${{ always() && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Unshallow
         run: git fetch --prune --unshallow
+      - name: Create release tag
+        uses: rickstaa/action-create-tag@v1.7.2 # will fail if existing tag is present
+        with:
+          tag: ${{ inputs.version_number }}
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg_passphrase: ${{ secrets.PASSPHRASE }}
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,18 @@ on:
         description: 'Skip QA acceptance tests, define value to `true` to explicitly skip'
 
 jobs:
+
+  validate-version-input:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validation of version format
+        run: |
+          echo "${{ inputs.version_number }}" | grep -P '^v\d+\.\d+\.\d+(-pre[A-Za-z0-9-]*)?$'  
+
     # QA acceptance tests are skipped for pre releases or when explicit input parameter is used
   run-qa-acceptance-tests:
-    if: inputs.skip_tests != 'true' && !contains(inputs.version_number, 'pre')
+    needs: [ validate-version-input ]
+    if: needs.validate-version-input.result == 'success' && inputs.skip_tests != 'true' && !contains(inputs.version_number, 'pre')
     secrets: inherit
     uses: ./.github/workflows/acceptance-tests.yml
     with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,16 +1,37 @@
...
+    if: ${{ always() && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Unshallow
         run: git fetch --prune --unshallow
+      - name: Create release tag
+        uses: rickstaa/action-create-tag@v1.7.2 # will fail if existing tag is present
+        with:
+          tag: ${{ inputs.version_number }}
```
AgustinBettati, line 43: as of now tag is not being signed, which is the behaviour we currently have when [creating tags for our releases](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/RELEASING.md#pre-release-the-provider). Will reach out to security team to verify if we can use the gpg signing feature of this action.

.github/workflows/release.yml
```diff
@@ -1,16 +1,37 @@
...
+  release:
     runs-on: ubuntu-latest
+    needs: [ run-qa-acceptance-tests ]
+    if: ${{ always() && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Unshallow
         run: git fetch --prune --unshallow
+      - name: Create release tag
```
andreaangiolillo, line 40: Do we want to add the logic to delete the tag in case of failure of a step after the tag creation?

.github/workflows/release.yml
```diff
@@ -1,16 +1,37 @@
...
 on:
-  push:
-    tags:
-      - 'v*'
-jobs: 
-  goreleaser:
+  workflow_dispatch:
+    inputs:
+      version_number:
+        description: 'Version number (e.g., v1.0.0, v1.0.0-pre, v1.0.0-pre1)'
```
andreaangiolillo, line 8: should we add a step to validate that the provided tag is correct? One common error is to forget the V

.github/workflows/release.yml
```diff
@@ -1,16 +1,37 @@
...
+        required: true
+      skip_tests:
+        description: 'Skip QA acceptance tests, define value to `true` to explicitly skip'
+
+jobs:
+    # QA acceptance tests are skipped for pre releases or when explicit input parameter is used
+  run-qa-acceptance-tests:
+    if: inputs.skip_tests != 'true' && !contains(inputs.version_number, 'pre')
```
andreaangiolillo: why do we always skip QA tests for pre-release? I see a scenario where we want to run the QA as part of the pre-release instead of the release. The pre-release will contain all the code changes except the changelog PR.

.github/workflows/release.yml
AgustinBettati: yes this should be simple to include and has value

.github/workflows/release.yml
AgustinBettati: did not want to slow down the process of generating the pre release, but I do see the value in detecting an issue in the prerelease vs seeing the issue in the final release. We can leave it to run by default, and have the option of skipping if you have already triggered manually for example.

AgustinBettati: adjusted workflow and documentation.

.github/workflows/release.yml
andreaangiolillo: Could you confirm that the action is creating an annotated tag?

.github/workflows/release.yml
```diff
@@ -1,16 +1,46 @@
...
+    runs-on: ubuntu-latest
+    needs: [ validate-version-input, run-qa-acceptance-tests ]
+    if: ${{ always() && needs.validate-version-input.result == 'success' && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Unshallow
         run: git fetch --prune --unshallow
+      - name: Create release tag
+        uses: rickstaa/action-create-tag@v1.7.2 # will fail if existing tag is present
```
andreaangiolillo, line 41: I've found this action https://github.com/mathieudutour/github-tag-action (520 starts) which seems to be used more than rickstaa/action-create-tag (60 starts). Maybe we could use that one (not a strong opinion)

.github/workflows/release.yml
AgustinBettati: yes verified that annotated tags were created in the forked repo used for testing, can also be seen in the code https://github.com/rickstaa/action-create-tag/blob/main/entrypoint.sh#L79

.github/workflows/release.yml
AgustinBettati: Yes that was an option I looked into, went for https://github.com/rickstaa/action-create-tag because it had the feature of signing tags (even though we are not using it yet) which this other action did not have.

andreaangiolillo: it makes sense. Thanks!

.github/workflows/release.yml
AgustinBettati: This can be an improvement but not sure if worth including in this PR. This case would only happen if `ghaction-import-gpg` or `goreleaser-action` action fails, which does not seem frequent from previous releases.

andreaangiolillo: I'd suggest documenting what we should do in case of a failure during the go releaser step in this PR: manually delete the tag and try again to run the action. Feel free to create a ticket to add this improvement if you don't want to do it as part of this PR.

andreaangiolillo (approved):
LGTM. Thanks for all the follow ups

.github/workflows/release.yml
AgustinBettati: leaving created jira ticket just for reference: https://jira.mongodb.org/browse/CLOUDP-221414

PR #1815, opened 2024-01-08
fix: Improves error message when improperly setting provider_region_name field

maastha:
## Description

This PR adds validation logic to Create & Update handlers of cluster resource to not set provider_region_name attribute for multi-region clusters.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-215125

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/cluster/resource_cluster.go
+++ b/internal/service/cluster/resource_cluster.go
@@ -18,12 +18,13 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/mwielbut/pointy"
+	"github.com/spf13/cast"
+
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/constant"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/config"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/advancedcluster"
-	"github.com/mwielbut/pointy"
-	"github.com/spf13/cast"
 )
 
 const (
@@ -463,10 +464,16 @@ func resourceMongoDBAtlasClusterCreate(ctx context.Context, d *schema.ResourceDa
 		return diag.FromErr(fmt.Errorf(errorClusterCreate, err))
 	}
 
+	clusterType := cast.ToString(d.Get("cluster_type"))
+	err = validateProviderRegionName(clusterType, providerSettings.RegionName, replicationSpecs)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf(errorClusterCreate, err))
+	}
+
 	clusterRequest := &matlas.Cluster{
 		Name:                     d.Get("name").(string),
 		EncryptionAtRestProvider: d.Get("encryption_at_rest_provider").(string),
-		ClusterType:              cast.ToString(d.Get("cluster_type")),
+		ClusterType:              clusterType,
 		BackupEnabled:            pointy.Bool(d.Get("backup_enabled").(bool)),
 		PitEnabled:               pointy.Bool(d.Get("pit_enabled").(bool)),
 		AutoScaling:              autoScaling,
@@ -803,13 +810,19 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
 		}
 	}
 
+	replicationSpecs, err := expandReplicationSpecs(d)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf(errorClusterUpdate, clusterName, err))
+	}
 	if d.HasChange("replication_specs") {
-		replicationSpecs, err := expandReplicationSpecs(d)
+		cluster.ReplicationSpecs = replicationSpecs
+	}
+
+	if cluster.ProviderSettings != nil || len(cluster.ReplicationSpecs) > 0 {
+		err := validateProviderRegionName(d.Get("cluster_type").(string), cluster.ProviderSettings.RegionName, replicationSpecs)
 		if err != nil {
 			return diag.FromErr(fmt.Errorf(errorClusterUpdate, clusterName, err))
 		}
-
-		cluster.ReplicationSpecs = replicationSpecs
 	}
 
 	cluster.AutoScaling = &matlas.AutoScaling{Compute: &matlas.Compute{}}
@@ -965,6 +978,29 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
 	}
 
 	return resourceMongoDBAtlasClusterRead(ctx, d, meta)
+}
+
+func isMultiRegionCluster(repSpecs []matlas.ReplicationSpec) bool {
+	if len(repSpecs) > 1 {
+		return true
+	}
+
+	for i := range repSpecs {
+		repSpec := repSpecs[i]
+
+		if len(repSpec.RegionsConfig) > 1 {
+			return true
+		}
+	}
+	return false
+}
+
+func validateProviderRegionName(clusterType, providerRegionName string, repSpecs []matlas.ReplicationSpec) error {
+	if conversion.IsStringPresent(&providerRegionName) && (clusterType == "GEOSHARDED" || isMultiRegionCluster(repSpecs)) {
+		return fmt.Errorf("provider_region_name attribute must be set ONLY for single-region clusters")
+	}
+
+	return nil
 }
 
 func didErrOnPausedCluster(err error) bool {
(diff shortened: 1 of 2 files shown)


internal/service/cluster/resource_cluster.go
```diff
@@ -803,15 +810,21 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
 		}
 	}
 
+	replicationSpecs, err := expandReplicationSpecs(d)
```
oarbusi, line 813: any specific reason for moving this check outside of the if `d.HasChange("replication_specs")` ?

internal/service/cluster/resource_cluster.go
```diff
@@ -803,15 +810,21 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
...
-			return diag.FromErr(fmt.Errorf(errorClusterUpdate, clusterName, err))
-		}
-
 		cluster.ReplicationSpecs = replicationSpecs
 	}
 
+	err = validateProviderRegionName(d.Get("cluster_type").(string), d.Get("provider_region_name").(string), replicationSpecs)
+	if err != nil {
+		if cluster.ProviderSettings != nil {
+			cluster.ProviderSettings.RegionName = ""
```
AgustinBettati: should we return an error as we are doing in the create operation?

oarbusi (approved):

internal/service/cluster/resource_cluster_test.go
```diff
@@ -608,6 +622,31 @@ func TestAccClusterRSCluster_MultiRegion(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet(resourceName, "replication_specs.0.regions_config.#"),
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "disk_size_gb", "100"),
+					resource.TestCheckResourceAttr(resourceName, "cluster_type", "REPLICASET"),
+					resource.TestCheckResourceAttr(resourceName, "replication_specs.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "replication_specs.0.regions_config.#", "2"),
+				),
+			},
+			{
+				Config: testAccMongoDBAtlasClusterConfigMultiRegion(orgID, projectName, name, "false", updatedRegionsConfig2),
```
AgustinBettati: should we add a test that expects the error?

internal/service/cluster/resource_cluster.go
marcosuma: I think it's done in order to be used for the `validateProviderRegionName` below

internal/service/cluster/resource_cluster.go
marcosuma: it feels like here we are swallowing an error and going ahead by resetting the field. If that's the right interpretation, should we just return an error?

maastha: We are swallowing the error for backwards compatibility here, so that if any existing users have the `provider_region_name` configured for a multi-region cluster (or updating a single region to multi-region) then they don't get any errors on running updates. But I guess this could lead to non-empty plans later if user has `provider_region_name` configured.

Let me cross check and will update this to return error.

internal/service/cluster/resource_cluster.go
maastha: right

internal/service/cluster/resource_cluster.go
maastha: Looked into this, we need to swallow the error during update. I have added a test where a single region cluster with this param specified is updated to a multi-region one, in such cases (or existing cases where users with multi-region clusters might have this param, if specified, in their state)

This is what happens in such cases:
1. a single-region cluster is created with `provider_region_name` set. User's state also has this param set.
2. user updates config to multi-region and adds more region-configs or replication_specs and removes `provider_region_name` from config and UPDATE handler is invoked.
3. Since `provider_region_name` is Optional & Computed, TF still reads the old value of this param from the state and user will get an error even though they have removed the param from their config.

To handle cases like above, we ignore `provider_region_name` param for multi-region clusters during update.

Please let me know if there's a better way to go about this.

internal/service/cluster/resource_cluster_test.go
maastha: added

internal/service/cluster/resource_cluster.go
AgustinBettati: Thanks for the details. Just to be sure, is `d.HasChange("provider_region_name")`/`d.GetChange("provider_region_name")` capable of detecting that the user has removed the attribute during the update operation?

maastha: No, verified that `d.HasChange("provider_region_name")/d.GetChange("provider_region_name")` don't detect that user has removed this attribute:

`old, new := d.GetChange("provider_region_name")`
log output:
```
2024/01/12 15:30:58 [INFO] old prn: US_EAST_1
2024/01/12 15:31:01 [INFO] new prn: US_EAST_1
```

AgustinBettati: I see, I believe we don't have another option. Just to be clear, currently a user is not able to transition from single region cluster with `provider_region_name` to a multi-region right? We would introduce potential inconsistency between users config and the state, but enable this transition.

If we go this way we can create a ticket to incorporate the error message in the update operation and link to the terraform framework migration epic. In that case we will be able to easily identify what is defined in the configuration vs the state when using the plugin framework.

maastha: >currently a user is not able to transition from single region cluster with provider_region_name to a multi-region right? 

right, they could potentially run into issues like https://jira.mongodb.org/browse/HELP-51429 

>If we go this way we can create a ticket to incorporate the error message in the update operation and link to the terraform framework migration epic.

Since this is cluster resource we don't have plans to migrate this to the Framework

internal/service/cluster/resource_cluster.go
```diff
@@ -803,13 +810,21 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
...
-		replicationSpecs, err := expandReplicationSpecs(d)
+		cluster.ReplicationSpecs = replicationSpecs
+	}
+
+	if v, ok := d.GetOk("provider_region_name"); ok {
+		err = validateProviderRegionName(d.Get("cluster_type").(string), v.(string), replicationSpecs)
 		if err != nil {
-			return diag.FromErr(fmt.Errorf(errorClusterUpdate, clusterName, err))
+			if cluster.ProviderSettings != nil {
+				cluster.ProviderSettings.RegionName = ""
```
marcosuma, line 827: I think this code will be hard to read in the future. Can we perhaps add an explanation of the why that we had in the comment?

internal/service/cluster/resource_cluster.go
```diff
@@ -967,6 +982,27 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
 	return resourceMongoDBAtlasClusterRead(ctx, d, meta)
 }
 
+func isMultiRegionCluster(repSpecs []matlas.ReplicationSpec) bool {
```
marcosuma: can we unit test these methods?

internal/service/cluster/resource_cluster.go
```diff
@@ -967,6 +982,27 @@ func resourceMongoDBAtlasClusterUpdate(ctx context.Context, d *schema.ResourceDa
...
+
+	for i := range repSpecs {
+		if len(repSpecs[i].RegionsConfig) > 1 {
+			return true
+		}
+	}
+	return false
+}
+
+func validateProviderRegionName(clusterType, providerRegionName string, repSpecs []matlas.ReplicationSpec) error {
```
marcosuma: this function seem to be targeting a specific validation for a specific case. Should we have a method name closer to what it does?

maastha: do you have any suggestions? I'd say it's ultimately validating "ProviderRegionName" for some scenarios, hence the current name

internal/service/cluster/resource_cluster.go
maastha: added

internal/service/cluster/resource_cluster.go
maastha: added

internal/service/cluster/resource_cluster_test.go
```diff
@@ -2412,3 +2559,130 @@ resource "mongodbatlas_cluster" "test" {
...
+func TestIsMultiRegionCluster(t *testing.T) {
+	tests := []struct {
+		name     string
+		repSpecs []matlas.ReplicationSpec
+		want     bool
+	}{
+		{
+			name:     "No ReplicationSpecs",
+			repSpecs: []matlas.ReplicationSpec{},
+			want:     false,
```
AgustinBettati, line 2572:
```suggestion
			expected:     false,
```
small nit

internal/service/cluster/resource_cluster_test.go
```diff
@@ -2412,3 +2559,130 @@ resource "mongodbatlas_cluster" "test" {
...
+	}
+}
+
+func TestValidateProviderRegionName(t *testing.T) {
+	tests := []struct {
+		name               string
+		clusterType        string
+		providerRegionName string
+		repSpecs           []matlas.ReplicationSpec
+		wantErr            bool
```
AgustinBettati, line 2630:
```suggestion
		expectedErr            bool
```

AgustinBettati (approved):
LGTM

PR #1831, opened 2024-01-12
chore: Uses new Atlas Go SDK version

lantoli:
## Description

Uses new Atlas Go SDK version.

The goal of this PR is to do as few changes as possible while keeping the same behavior as the current code.
Next PRs will improve update operation, createonly and readonly attributes of each TF resource using this SDK.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-222424

## Type of change:

## Required Checklist:


internal/common/conversion/type_conversion.go
```diff
@@ -56,3 +56,17 @@ func IsStringPresent(strPtr *string) bool {
...
 }
+
+func SlicePtrToSlice[T any](v *[]T) []T {
+	if v == nil {
+		return nil
+	}
+	return *v
+}
+
+func NonEmptySliceToPtrSlice[T any](v []T) *[]T {
```
lantoli: @AgustinBettati idea to keep same current behavior

internal/service/onlinearchive/resource_online_archive.go
```diff
@@ -466,15 +466,6 @@ func resourceMongoDBAtlasOnlineArchiveUpdate(ctx context.Context, d *schema.Reso
 		}
 	}
 
-	if dataProcessRegionHasChange {
```
lantoli, line 469: TEMPORARY deletion to make it compile and be able to run tests

internal/service/onlinearchive/resource_online_archive.go
```diff
@@ -559,22 +550,12 @@ func fromOnlineArchiveToMap(in *admin.BackupOnlineArchive) map[string]any {
 		schemaVals["data_expiration_rule"] = []any{dataExpirationRule}
 	}
 
-	var dataProcessRegion map[string]any
```
lantoli, line 562: TEMPORARY deletion to make it compile and be able to run tests

internal/common/conversion/type_conversion.go
marcosuma: curious: why are you specifying `NonEmpty...` in the method name? the name makes me thinking you can't call the method for an empty slice, which in reality you can

lantoli: it's to highlight that it will return nil for an empty slice instead of a pointer to an empty slice.  This is to keep current behaviour.    (normally you would expect it to return the pointer to the slice even if empty -but not nil-)

we want to send nil pointer for empty slices so it's not send in JSON to the API as current behavior. 

in next PRs we'll improve the update operation so we can update the slice including emptying it (which is not possible in current code).

the expected implementation (without nonempty...) would be:

if v ==  nil {
		return nil
	}
return &v

marcosuma:
```suggestion
func NonEmptySliceToSlicePtr[T any](v []T) *[]T {
```

nit: can we keep consistency of how to use `Ptr`? either prefix or suffix (see previous method).

internal/common/conversion/type_conversion.go
```diff
@@ -56,3 +56,17 @@ func IsStringPresent(strPtr *string) bool {
 func MongoDBRegionToAWSRegion(region string) string {
 	return strings.ReplaceAll(strings.ToLower(region), "_", "-")
 }
+
+func SlicePtrToSlice[T any](v *[]T) []T {
```
marcosuma: can we please unit test these two methods?

internal/common/conversion/type_conversion.go
marcosuma: ok I get it.

lantoli: oops, thanks, changing to NonEmptySliceToSlicePtr

internal/common/conversion/type_conversion.go
lantoli: added

internal/common/conversion/type_conversion_test.go
```diff
@@ -67,3 +68,31 @@ func TestMongoDBRegionToAWSRegion(t *testing.T) {
...
+		{nil, nil},
+		{&[]string{"hello", "there"}, []string{"hello", "there"}},
+	}
+	for _, test := range tests {
+		assert.Equal(t, test.slice, conversion.SlicePtrToSlice(test.ptr))
+		assert.Equal(t, test.ptr, conversion.NonEmptySliceToSlicePtr(test.slice))
+	}
+}
+
+func TestSlice_DifferentBehavior(t *testing.T) {
```
lantoli: in slice tests I'm focusing in the different behavior (because the "nonempty" part) instead of testing each method separately, so it's more clear the different behavior, testing each method in its test method it's more challenging to see the differences

internal/common/conversion/type_conversion_test.go
```diff
@@ -67,3 +68,31 @@ func TestMongoDBRegionToAWSRegion(t *testing.T) {
...
+		ptr   *[]string
+		slice []string
+	}{
+		{nil, nil},
+		{&[]string{"hello", "there"}, []string{"hello", "there"}},
+	}
+	for _, test := range tests {
+		assert.Equal(t, test.slice, conversion.SlicePtrToSlice(test.ptr))
+		assert.Equal(t, test.ptr, conversion.NonEmptySliceToSlicePtr(test.slice))
+	}
```
gssbzn: when doing test matrix please use `t.Run` for the test case

internal/common/conversion/type_conversion_test.go
```diff
@@ -67,3 +68,31 @@ func TestMongoDBRegionToAWSRegion(t *testing.T) {
...
 }
+
+func TestSlice_SameBehavior(t *testing.T) {
+	tests := []struct {
+		ptr   *[]string
+		slice []string
+	}{
+		{nil, nil},
+		{&[]string{"hello", "there"}, []string{"hello", "there"}},
+	}
```
gssbzn: when using test matrix please consider naming your args with names like `given|want`, `in|out` or something more clear on expected input/output

internal/common/conversion/type_conversion_test.go
```diff
@@ -67,3 +68,31 @@ func TestMongoDBRegionToAWSRegion(t *testing.T) {
...
+	for _, test := range tests {
+		assert.Equal(t, test.slice, conversion.SlicePtrToSlice(test.ptr))
+		assert.Equal(t, test.ptr, conversion.NonEmptySliceToSlicePtr(test.slice))
```
gssbzn: you are testing two unrelated functions in the same test, please use `Test<FuncName>` as best practice, ie `TestSlicePtrToSlice` and `TestNonEmptySliceToSlicePtr` as separate tests

internal/service/atlasuser/data_source_atlas_users_test.go
```diff
@@ -223,8 +223,8 @@ func dataSourceChecksForUsers(dataSourceName, orgID string, users *admin.Paginat
 		resource.TestCheckResourceAttr(dataSourceName, "org_id", orgID),
 		resource.TestCheckResourceAttr(dataSourceName, "total_count", fmt.Sprintf("%d", totalCountValue)),
 	}
-	for i := range users.Results {
-		checks = append(checks, dataSourceChecksForUser(dataSourceName, fmt.Sprintf("results.%d.", i), &users.Results[i])...)
+	for i := range *users.Results {
+		checks = append(checks, dataSourceChecksForUser(dataSourceName, fmt.Sprintf("results.%d.", i), &(*users.Results)[i])...)
```
gssbzn: wouldn't this be the same as
```suggestion
		checks = append(checks, dataSourceChecksForUser(dataSourceName, fmt.Sprintf("results.%d.", i), users.Results)[i])...)
```

internal/service/alertconfiguration/data_source_alert_configuration.go
```diff
@@ -319,7 +319,7 @@ func outputAlertConfigurationResourceHcl(label string, alert *admin.GroupAlertsC
 		resource.SetAttributeValue("enabled", cty.BoolVal(*alert.Enabled))
 	}
 
-	for _, matcher := range alert.Matchers {
+	for _, matcher := range conversion.SlicePtrToSlice(alert.Matchers) {
```
gssbzn: why this function over
```suggestion
	for _, matcher := range *alert.Matchers {
```

internal/service/databaseuser/model_database_user.go
```diff
@@ -40,24 +40,24 @@ func NewMongoDBDatabaseUser(ctx context.Context, dbUserModel *TfDatabaseUserMode
 		OidcAuthType: dbUserModel.OIDCAuthType.ValueStringPointer(),
 		LdapAuthType: dbUserModel.LDAPAuthType.ValueStringPointer(),
 		DatabaseName: dbUserModel.AuthDatabaseName.ValueString(),
-		Roles:        NewMongoDBAtlasRoles(rolesModel),
-		Labels:       NewMongoDBAtlasLabels(labelsModel),
-		Scopes:       NewMongoDBAtlasScopes(scopesModel),
+		Roles:        conversion.NonEmptySliceToSlicePtr(NewMongoDBAtlasRoles(rolesModel)),
+		Labels:       conversion.NonEmptySliceToSlicePtr(NewMongoDBAtlasLabels(labelsModel)),
+		Scopes:       conversion.NonEmptySliceToSlicePtr(NewMongoDBAtlasScopes(scopesModel)),
```
gssbzn: why not have the `New..` function return the pointer?

internal/common/conversion/type_conversion_test.go
lantoli: I normally do but in this case I'm calling the two methods and the args are expected/output for one call but the opposite for the other call.  They're like symmetric methods and in this test I test that they behave the same for those use cases.

internal/service/databaseuser/model_database_user.go
lantoli: I'm trying to minimize the changes here, then there will a PR per resource to fix the update bugs, and check the readonly / createonly attributes, and improve the clarity as in your suggestion.   Those changes will be more located (per resource)

internal/common/conversion/type_conversion_test.go
lantoli: @gssbzn  this is what I was trying here instead of testing each method in its own test function. I did that at first but then I thought this would be better.

they're related methods that behave similarly except when dealing with empty non-nil slices.

internal/service/alertconfiguration/data_source_alert_configuration.go
AgustinBettati:
```suggestion
	for _, matcher := range alert.GetMatchers() {
```
This will avoid a potential nil pointer dereference if `matchers` property is not defined. Potentially all usages of `SlicePtrToSlice` could be replaced with the respective .Get methods provided by the SDK right?

internal/common/conversion/type_conversion_test.go
lantoli: thx, done

internal/common/conversion/type_conversion_test.go
lantoli: mentioning you in the comment with the explanation

internal/service/atlasuser/data_source_atlas_users_test.go
lantoli: doing users.Results[i] gives this error:  "invalid operation: cannot index users.Results (variable of type *[]"go.mongodb.org/atlas-sdk/v20231115003/admin".CloudAppUser)"

internal/service/alertconfiguration/data_source_alert_configuration.go
lantoli: correct @AgustinBettati , that will be in the next PRs specific to each resource. 

@gssbzn it's what Agustin is saying, alert.Matchers could be nil so * panics.

with this PR we're trying to keep the same behavior as before the PR

internal/service/databaseuser/model_database_user.go
lantoli: as Agusting said before we'll probably get rid of all NonEmptySliceToSlicePtr in the next PRs

internal/service/onlinearchive/data_source_online_archive.go
```diff
@@ -216,12 +216,12 @@ func schemaOnlineArchive() map[string]*schema.Schema {
 }
 
 func dataSourceMongoDBAtlasOnlineArchiveRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	connV2 := meta.(*config.MongoDBClient).AtlasV2
+	connOldV2 := meta.(*config.MongoDBClient).OldAtlasV2
```
AgustinBettati: Just curious if we have a ticket to followup on this so that the resource uses the latest API version.

lantoli: not yet apart from the specific online_archive SDK update ticket . I'm waiting to collect more info (i asked backup team), will do after I have more info

internal/service/alertconfiguration/data_source_alert_configuration.go
gssbzn: I'm with @AgustinBettati here that use the Getters, they are nil safe

lantoli: I'll use them in the next PR

gssbzn: Why not avoid introducing `NonEmptySliceToSlicePtr`  in the first place? I'm still nitpicking how the test is written and a lot of those conversations disappear by not introducing the helper?

internal/common/conversion/type_conversion_test.go
lantoli: at then end I'm only having one method, changed to follow your suggestion

internal/common/conversion/type_conversion_test.go
lantoli: at then end I'm only having one method, changed to follow your suggestion

internal/service/alertconfiguration/data_source_alert_configuration.go
lantoli: at the end I'm made the changes in this PR to use Get functions so no need to have SlicePtrToSlice method.  
thank you both!

internal/service/onlinearchive/data_source_online_archive.go
andreaangiolillo: Could we use the name `v20231001002AtlasSDK` instead of `OldAtlasV2`? It is not easy to understand what we consider the old atlas v2. In addition, the new name will allow us to support multiple older versions if needed

lantoli: I'm following the convention @oarbusi did in his previous PR, e.g. it's being used in federated_settings_identity_provider service

andreaangiolillo: I missed that PR 😭  What is our opinion on the name used for the Atlas SDK?

lantoli: maybe it's not the best name but I couldn't find a better name to suggest, the idea is that it's an SDK a little older than the current one used in most places.   
also the idea is that this old SDK is only used until the corresponding service team fixes their API so use of this old sdk should be short duration

internal/common/conversion/type_conversion.go
```diff
@@ -56,3 +56,10 @@ func IsStringPresent(strPtr *string) bool {
 func MongoDBRegionToAWSRegion(region string) string {
 	return strings.ReplaceAll(strings.ToLower(region), "_", "-")
 }
+
+func NonEmptySliceToSlicePtr[T any](v []T) *[]T {
```
gssbzn: if slices only let's be more correct about it 
```suggestion
func NonEmptySliceToSlicePtr[S ~[]E, E any]](v S) *S {
```

internal/service/alertconfiguration/data_source_alert_configuration.go
```diff
@@ -438,16 +439,14 @@ func convertNotificationToCtyValues(notification *admin.AlertsNotificationRootFo
 		values["sms_enabled"] = cty.BoolVal(*notification.SmsEnabled)
 	}
 
-	if len(notification.Roles) > 0 {
-		roles := make([]cty.Value, 0)
-
-		for _, r := range notification.Roles {
+	if roles := notification.GetRoles(); len(roles) > 0 {
+		roleList := make([]cty.Value, 0)
```
gssbzn: can this be pre-allocated?
```suggestion
		roleList := make([]cty.Value, 0, len(roles))
```

internal/service/atlasuser/data_source_atlas_users_test.go
gssbzn: what if you use the getter?

internal/service/onlinearchive/data_source_online_archive.go
gssbzn: I think the correct naming would include the dates to avoid confusion if more parallel sdks are mantained

internal/common/conversion/type_conversion.go
AgustinBettati: we will likely have a similar approach in CFN and CLI for this version upgrade, could be worth incorporating within SDK as a helper function and mention in upgrade guide.

AgustinBettati (commented):
LGTM besides pending comments

internal/common/conversion/type_conversion_test.go
lantoli: at the end there is only one method, there is a function test for it as you say

internal/service/atlasuser/data_source_atlas_users_test.go
lantoli: this works: &users.GetResults()[i]
changed

internal/service/onlinearchive/data_source_online_archive.go
lantoli: right now I think it's a premature optimization as we only have 2 sdks, y we intend to have the old one a short period of time.  if more than 2 would be needed at some moment then I agree.

internal/service/alertconfiguration/data_source_alert_configuration.go
lantoli: in this PR i'm trying to change as few things as possible, you're right there are many loops than can be improved, but I prefer to do that in next PRs.

edit: ok, changed as it doesn't affect code logic.

internal/common/conversion/type_conversion.go
lantoli: @gssbzn what do you think?

on the other hand I think I'll be able to delete most of the uses of this function in next PRs

internal/service/alertconfiguration/data_source_alert_configuration.go
lantoli: @gssbzn SlicePtrToSlice is already deleted, do you mean that or NonEmptySliceToSlicePtr?

gssbzn: all good, my comment was before the deletion

internal/service/alertconfiguration/data_source_alert_configuration.go
gssbzn: SGTM

internal/common/conversion/type_conversion.go
lantoli: name changed to NonEmptyToPtr as it might called for slices or also arrays.

internal/config/client.go
```diff
@@ -31,7 +31,7 @@ var (
 type MongoDBClient struct {
 	Atlas      *matlasClient.Client
 	AtlasV2    *atlasSDK.APIClient
-	OldAtlasV2 *oldAtlasSDK.APIClient // Needed to avoid sudden breaking changes in federated_settings_identity_provider resource. Will be removed in terraform-provider-1.16.0
+	OldAtlasV2 *oldAtlasSDK.APIClient // Needed to avoid sudden breaking changes in federated_settings_identity_provider and online_archive resources. Will be removed in terraform-provider-1.16.0
```
lantoli: @andreaangiolillo @AgustinBettati @gssbzn I've added online_archive to the old sdk comment.   but I would prefer to keep this PR changes as small as possible, I think old sdk renaming belongs to a different PR, it would affect client file and federated files apart from online_archive.

marcosuma:
LGTM on my side, will wait once the comments are addressed.

AgustinBettati (approved):

internal/config/client.go
lantoli: @andreaangiolillo @AgustinBettati @gssbzn hey, at the end I've added the date to old sdk in this PR in case you want to take a look. thanks

internal/service/onlinearchive/data_source_online_archive.go
lantoli: at the end I've added the date to the old sdk as per your recommendation

PR #1833, opened 2024-01-12
doc: Adjust documentation of search_deployment resource and data source

andreaangiolillo:
## Description

Ticket: https://jira.mongodb.org/browse/CLOUDP-221359

This PR updates the `search_deployment` doc via `make generate-doc`

## Type of change:

## Required Checklist:


internal/service/searchdeployment/data_source_search_deployment_schema.go
```diff
@@ -9,6 +9,7 @@ import (
 
 func DataSourceSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
+		MarkdownDescription: "Provides a Search Deployment data source.",
```
lantoli: I thought at the end we would generate both MarkdowDescription and Description cc @AgustinBettati

website/docs/d/search_deployment.html.markdown
```diff
@@ -1,37 +1,77 @@
 ---
 layout: "mongodbatlas"
-page_title: "MongoDB Atlas: search deployment"
-sidebar_current: "docs-mongodbatlas-datasource-search-deployment"
+page_title: "MongoDB Atlas: mongodbatlas_search_deployment"
```
lantoli, line 3: were these deleted/added headers wrong? should we change them in all resources / ds?

lantoli (approved):

website/docs/d/search_deployment.html.markdown
andreaangiolillo: I double-checked this with Zuhair and he confirmed that this metadata is no longer used by Hashicorp. Hashicorp used this metadata for screen readers which does not seem to be the case anymore.

internal/service/searchdeployment/data_source_search_deployment_schema.go
andreaangiolillo: Actually, I added this manually, we could also move this to the template instead. Wondering if there is a way to get the resource description from the open api spec instead. CC @AgustinBettati in case you know how to do this

andreaangiolillo: looking at the Hashicorp doc https://developer.hashicorp.com/terraform/plugin/code-generation/openapi-generator#generator-config it does not seem possible. I will open an issue on their repo. In the meantime I will remove this field from the schema and move the description to the template

andreaangiolillo: Created this issue https://github.com/hashicorp/terraform-plugin-codegen-openapi/issues/123

website/docs/r/search_deployment.html.markdown
```diff
@@ -1,57 +1,99 @@
...
+- `state_name` (String) Human-readable label that indicates the current operating condition of this search deployment.
 
-### Specs
+<a id="nestedatt--specs"></a>
+### Nested Schema for `specs`
 
-Specs list is defined as a ["list nested attribute"](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/list-nested) containing a single element.
+Required:
 
-* `instance_size` - (Required) Hardware specification for the search node instance sizes. The [MongoDB Atlas API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Atlas-Search/operation/createAtlasSearchDeployment) describes the valid values. More details can also be found in the [Search Node Documentation](https://www.mongodb.com/docs/atlas/cluster-config/multi-cloud-distribution/#search-tier).
```
marcosuma, line 44: Interesting: are we losing some valuable information by removing this?

website/docs/r/search_deployment.html.markdown
```diff
@@ -1,57 +1,99 @@
...
+
+### Required
+
+- `cluster_name` (String) Label that identifies the cluster to return the search nodes for.
+- `project_id` (String) Unique 24-hexadecimal character string that identifies the project.
+- `specs` (Attributes List) List of settings that configure the search nodes for your cluster. (see [below for nested schema](#nestedatt--specs))
+
+### Optional
+
+- `timeouts` (Attributes) (see [below for nested schema](#nestedatt--timeouts))
```
marcosuma, line 71: nit: can you just check that we are not losing some relevant information here with the change of timeouts field compared to how it was before?

internal/service/searchdeployment/data_source_search_deployment_schema.go
AgustinBettati: Yes I agree moving to template will be the best option for now

website/docs/r/search_deployment.html.markdown
andreaangiolillo: IMO we are getting more information for `timeouts`
Screenshot 2024-01-15 at 09 27 53. 

I see that this does not apply to `instance_size`. If we think that the description is not good enough, we should update the open API spec of the endpoint

website/docs/d/search_deployment.html.markdown
```diff
@@ -1,37 +1,81 @@
...
+- `specs` (Attributes List) List of settings that configure the search nodes for your cluster. (see [below for nested schema](#nestedatt--specs))
+- `state_name` (String) Human-readable label that indicates the current operating condition of this search deployment.
 
-* `specs` - List of settings that configure the search nodes for your cluster. See [specs](#specs).
-* `state_name` - Human-readable label that indicates the current operating condition of this search node deployment.
+<a id="nestedatt--specs"></a>
+### Nested Schema for `specs`
 
-### Specs
-* `instance_size` - (Required) Hardware specification for the search node instance sizes. The [MongoDB Atlas API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Atlas-Search/operation/createAtlasSearchDeployment) describes the valid values. More details can also be found in the [Search Node Documentation](https://www.mongodb.com/docs/atlas/cluster-config/multi-cloud-distribution/#search-tier).
```
AgustinBettati, line 33: I would preserve this more detailed description of the instance size attribute as it also references valid values, we can adjust the schema description fields and also include a [description override config](https://developer.hashicorp.com/terraform/plugin/code-generation/openapi-generator#overriding-attribute-descriptions) in the generator_config.yml file.

website/docs/r/search_deployment.html.markdown
andreaangiolillo: If the description is not good enough, we should update the open API spec of the endpoint. In this case, I will open a follow-up PR to add "More details can also be found in the [Search Node Documentation](https://www.mongodb.com/docs/atlas/cluster-config/multi-cloud-distribution/#search-tier)" to the endpoint. Thanks!

andreaangiolillo: FYI created this ticket https://jira.mongodb.org/browse/CLOUDP-222701 to update the open api spec

website/docs/r/search_deployment.html.markdown
```diff
@@ -1,57 +1,99 @@
...
+    "Provides a Search Deployment resource."
 ---
 
 # Resource: mongodbatlas_search_deployment
 
-`mongodbatlas_search_deployment` provides a Search Deployment resource. The resource lets you create, edit and delete dedicated search nodes in a cluster.
 
--> **NOTE:** For details on supported cloud providers and existing limitations you can visit the [Search Node Documentation](https://www.mongodb.com/docs/atlas/cluster-config/multi-cloud-distribution/#search-nodes-for-workload-isolation).
 
--> **NOTE:** Only a single search deployment resource can be defined for each cluster.
```
AgustinBettati, line 15: Would prefer to not remove this content as it is valuable for the docs. Could we add this in the template, or as part of the description field of the schema?

website/docs/r/search_deployment.html.markdown
```diff
@@ -1,57 +1,99 @@
...
+- `specs` (Attributes List) List of settings that configure the search nodes for your cluster. (see [below for nested schema](#nestedatt--specs))
+
+### Optional
+
+- `timeouts` (Attributes) (see [below for nested schema](#nestedatt--timeouts))
+
+### Read-Only
 
-* `project_id` - (Required) Unique 24-hexadecimal digit string that identifies your project.
-* `cluster_name` - (Required) Label that identifies the cluster to create search nodes for.
```
AgustinBettati, line 36: I believe we should preserve this description as it makes more sense within the context of a resource.

AgustinBettati (commented):
Some comments after comparing generated documentation using https://registry.terraform.io/tools/doc-preview

andreaangiolillo:
@AgustinBettati  I will open a follow-up PR to address these comments.

website/docs/d/search_deployment.html.markdown
andreaangiolillo: I was planning to modify this in the open api spec but this may be better. I will also update our doc to make sure this is documented

website/docs/r/search_deployment.html.markdown
andreaangiolillo: Could you clarify the content that you want to include in the template? is the NOTE?

AgustinBettati: Yes the 2 `NOTE:` provide valuable info/clarification.

andreaangiolillo: yes, we should add them in the template in this case

PR #1834, opened 2024-01-12
refactor: Migrates mongodbatlas_organization resource & data sources to new Atlas SDK

maastha:
## Description

This PR migrates mongodbatlas_organization resource & data sources to new Atlas SDK and makes minor improvements to tests.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-219771

## Type of change:

## Required Checklist:


internal/service/organization/data_source_organizations.go
```diff
@@ -22,8 +24,9 @@ func PluralDataSource() *schema.Resource {
 				Optional: true,
 			},
 			"include_deleted_orgs": {
-				Type:     schema.TypeBool,
-				Optional: true,
+				Type:       schema.TypeBool,
+				Optional:   true,
+				Deprecated: fmt.Sprintf(constant.DeprecationParamByDate, "January 2025"),
```
maastha, line 29: @Zuhairahmed could you please confirm the removal date for this as per comment here https://jira.mongodb.org/browse/CLOUDP-219771?focusedCommentId=6007024&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6007024

AgustinBettati: q: if the attribute is no longer usable, would it be better to remove directly? I see the value in deprecating if behaviour is maintained but I believe this is not the case.

AgustinBettati (commented):
LGTM, just one question regarding `include_deleted_orgs` attribute

internal/service/organization/resource_organization_test.go
```diff
@@ -56,39 +63,71 @@ func testAccCheckMongoDBAtlasOrganizationExists(resourceName string) resource.Te
...
+
 		ids := conversion.DecodeStateID(rs.Primary.ID)
 
-		organizationOptions := &matlas.OrganizationsListOptions{}
-		orgs, _, err := conn.Organizations.List(context.Background(), organizationOptions)
+		cfg := config.Config{
+			PublicKey:  rs.Primary.Attributes["public_key"],
+			PrivateKey: rs.Primary.Attributes["private_key"],
+			BaseURL:    acc.TestAccProviderSdkV2.Meta().(*config.MongoDBClient).Config.BaseURL,
+		}
```
marcosuma: side question: how was it working before with previous SDK without using public and private key?

internal/service/organization/resource_organization_test.go
```diff
@@ -56,39 +63,71 @@ func testAccCheckMongoDBAtlasOrganizationExists(resourceName string) resource.Te
...
-			for _, val := range orgs.Results {
-				if val.ID == ids["org_id"] {
-					return fmt.Errorf("Organization (%s) still exists", ids["role_name"])
+			for _, val := range *orgs.Results {
+				if *val.Id == ids["org_id"] {
+					return nil
 				}
 			}
-			return nil
+			return fmt.Errorf("Organization (%s) doesn't exist", ids["org_id"])
```
marcosuma: can you clarify the change in the test context here? are we changing the nature of the error from "an org still exist" to "an org doesn't exist"? And why so?

internal/service/organization/resource_organization_test.go
marcosuma: oh maybe it's because you need to change creds here? if so maybe worth adding a comment

internal/service/organization/resource_organization_test.go
maastha: this method is supposed to check if the organization exists (return error otherwise), instead it was returning an error if the organization existed. 
Since the API keys being used to call the ListOrganizations() was not associated with the newly created org, it would not be returned, hence we now use the API keys of the new org - https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/2023-11-15/#tag/Organizations/operation/listOrganizations `To use this resource, the requesting API Key must have the Organization Member role`
Will add a comment

internal/service/organization/resource_organization_test.go
maastha: Since the API keys being used to call the ListOrganizations() was not associated with the newly created org, it would not be returned, hence we now use the API keys of the new org - https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/2023-11-15/#tag/Organizations/operation/listOrganizations To use this resource, the requesting API Key must have the Organization Member role
Will add a comment

internal/service/organization/data_source_organizations.go
maastha: removing directly would be a breaking change if users are using this in their configs

internal/service/organization/data_source_organization.go
```diff
@@ -63,11 +64,11 @@ func dataSourceMongoDBAtlasOrganizationRead(ctx context.Context, d *schema.Resou
 		return diag.FromErr(fmt.Errorf("error setting `is_deleted`: %s", err))
 	}
 
-	if err := d.Set("links", flattenOrganizationLinks(organization.Links)); err != nil {
+	if err := d.Set("links", flattenOrganizationLinks(organization.GetLinks())); err != nil {
 		return diag.FromErr(fmt.Errorf("error setting `is_deleted`: %s", err))
 	}
 
-	d.SetId(organization.ID)
+	d.SetId(*organization.Id)
```
lantoli: does it also work with organization.GetId() ?

internal/service/organization/resource_organization_test.go
```diff
@@ -56,39 +63,71 @@ func testAccCheckMongoDBAtlasOrganizationExists(resourceName string) resource.Te
...
+		ctx := context.Background()
+		clients, _ := cfg.NewClient(ctx)
+		conn := clients.(*config.MongoDBClient).AtlasV2
+
+		orgs, _, err := conn.OrganizationsApi.ListOrganizations(ctx).Execute()
 		if err == nil {
-			for _, val := range orgs.Results {
-				if val.ID == ids["org_id"] {
-					return fmt.Errorf("Organization (%s) still exists", ids["role_name"])
+			for _, val := range *orgs.Results {
```
lantoli: would it work with orgs.GetResults() ?

internal/service/organization/resource_organization_test.go
```diff
@@ -56,39 +63,71 @@ func testAccCheckMongoDBAtlasOrganizationExists(resourceName string) resource.Te
...
+				if *val.Id == ids["org_id"] {
```
lantoli: val.GetId() ?

internal/service/organization/resource_organization_test.go
```diff
@@ -56,39 +63,71 @@ func testAccCheckMongoDBAtlasOrganizationExists(resourceName string) resource.Te
...
 
 func testAccCheckMongoDBAtlasOrganizationDestroy(s *terraform.State) error {
-	conn := acc.TestAccProviderSdkV2.Meta().(*config.MongoDBClient).Atlas
-
 	for _, rs := range s.RootModule().Resources {
 		if rs.Type != "mongodbatlas_organization" {
 			continue
 		}
 
+		if rs.Primary.Attributes["public_key"] == "" {
```
lantoli: as these 2 checks are done in a couple of places, don't know if it makes sense to extract them together

lantoli (approved):

lantoli:
PR description and ticket link is missing

internal/service/organization/resource_organization_test.go
marcosuma: sounds good, feel free to resolve after you add the comment.

internal/service/organization/data_source_organizations.go
AgustinBettati: correct, but if they are using `include_deleted_orgs` this functionality would not work anymore because the attribute is not being used when calling the API. Both options would be a breaking change in some form unless we preserve using the old sdk for a deprecation period (but I understand this would not allow to support new attributes in the plural data source).

maastha: >Both options would be a breaking change 

so as per [this comment here](https://jira.mongodb.org/browse/CLOUDP-219771?focusedCommentId=6006899&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6006899) this param has always been a no-op, so by not removing this right away and not using it anymore, we aren't introducing any breaking/behavior changes

AgustinBettati: Sorry had missed the part that it was a no-op for atlas, sound goods thanks for clarifying

internal/service/organization/resource_organization_test.go
maastha: moved to a new method and added comment, resolving

internal/service/organization/data_source_organization.go
maastha: yes, updated

internal/service/organization/resource_organization_test.go
maastha: updated

internal/service/organization/resource_organization_test.go
maastha: updated

internal/service/organization/resource_organization_test.go
maastha: updated

internal/service/organization/data_source_organizations.go
```diff
@@ -110,7 +109,7 @@ func dataSourceMongoDBAtlasOrganizationsRead(ctx context.Context, d *schema.Reso
 	return nil
 }
 
-func flattenOrganizationLinks(links []*matlas.Link) []map[string]any {
+func flattenOrganizationLinks(links []admin.Link) []map[string]any {
```
lantoli, line 112: not related to this PR, I think we're not exposing Links in most of the resources, don't know if it's useful to expose it in TF

internal/service/organization/resource_organization_test.go
```diff
@@ -108,3 +121,26 @@ func testAccMongoDBAtlasOrganizationConfigBasic(orgOwnerID, name, description, r
 	  }
 	`, orgOwnerID, name, description, roleNames)
 }
+
+// getTestClientWithNewOrgCreds This method creates a new Atlas client with credentials for the newly created organization.
```
lantoli:
```suggestion
// getTestClientWithNewOrgCreds creates a new Atlas client with credentials for the newly created organization.
```

internal/service/organization/resource_organization_test.go
```diff
@@ -108,3 +121,26 @@ func testAccMongoDBAtlasOrganizationConfigBasic(orgOwnerID, name, description, r
...
+// This is required to call relevant API methods for the new organization, for example ListOrganizations requires that the requesting API
```
lantoli:
```suggestion
// getTestClientWithNewOrgCreds is required to call relevant API methods for the new organization, for example ListOrganizations requires that the requesting API
```

lantoli (approved):

PR #1394, opened 2023-08-16
feat: POC on how to autogenerate the documentation using tfplugindocs

andreaangiolillo:
## Description

This PR shows how we can leverage [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) to autogenerate the terraform documentation.

### How it works

1) Add `MarkdownDescription` to each element in the resource/datasource scema
**NOTE**: Add `Description` if the resource/datasourfce uses SDKv2

2) Add the schema description in `mongodbatlas/description/resource_name.go`
**NOTE: In the future we will reuse the field description in the Open API spec.**

3) Define the template to use for the resource/datasource in `templates/`
**NOTE** we can define a general template to use for all the resources (or datasource) by adding `templates/resources.md.tmpl`.

4) Add the resource/datasource examples in `examples` dir
**NOTE** we reuse the example in `examples` to populate the example section of the doc page

5) run `tfplugindocs-arm generate`
**NOTE:** `tfplugindocs-arm` will generate all the resource/datasource in the `docs` folder if `--website-source-dir` is not specified, make sure to create the `docs` directory

6) Copy the resource/datasource doc into the `websites/` folder
**NOTE:** Once all the resources/datasource have `MarkdownDescription` in their schema we won't need to run step 6 since `tfplugindocs-arm generate --website-source-dir websites/` will generate the documentation for all the resources/datasources

**NOTE** **`tfplugindocs-arm` can be use with both SDKv2 and the new framework**

### Pros
1. This will be a step closer to the complete autogeneration of our documentation. The next step will be to automatically add the description in `MarkdownDescription` from the open API spec
2. We will reuse our example in `examples/` which will make consistent the example in our documentation with the ones that we provide in the repo. 
3. Force us to improve our examples in the repo to make sure to cover all the scenarios
4. Less code to write when adding/updating resources/datasource 
5. The autogeneration will force format consistency within our docs, and will likely be very similar to other providers using the same tool.

### Cons

1.  [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) does not support creating doc for only specific resources/datasources but it will generate the doc for all the resources/datasources in the provider: This forces us to have the step `n6` until all the resources/datasources use `MarkdownDescription`
2.  [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) has several open issues in the repo. We do not know as of today how much investment Hashicorp will put in the product. **This can be easily discovered by contacting Hashicorp via our Slack channel.**
[...]

PR #1847, opened 2024-01-17
chore: Define default resource and data source documentation template

andreaangiolillo:
## Description

Ticket: https://jira.mongodb.org/browse/CLOUDP-222897

**Context:**
 - Currently all generated documentation requires a new template to be defined.

**Proposed Solution:**
-  Define a generic template for resources as well as single/plural data sources.

For documentation that requires no specific changes the default template can be used and no new template needs to be created. In case specific adjustments are needed, the default template will serve as a starting point to define the new template.

## Type of change:

## Required Checklist:

## Testing
I ran `export resource_name=cluster && make generate-doc` and chcke that the default template was used.


templates/data-source.html.md.tmpl
```diff
@@ -0,0 +1,75 @@
...
+    {{ else if eq .Name "mongodbatlas_encryption_at_rest" }}
+    {{ else if eq .Name "mongodbatlas_ldap_configuration" }}
+    {{ else if eq .Name "mongodbatlas_teams" }}
+    {{ else if eq .Name "mongodbatlas_serverless_instance" }}
+    {{ else if eq .Name "mongodbatlas_maintenance_window" }}
+    {{ else if eq .Name "mongodbatlas_ldap_verify" }}
+    {{ else if eq .Name "mongodbatlas_third_party_integration" }}
+    {{ else if eq .Name "mongodbatlas_x509_authentication_database_user" }}
+    {{ else if eq .Name "mongodbatlas_privatelink_endpoint_service_data_federation_online_archive" }}
+    {{ else }}
```
marcosuma, line 68: could you explain why we need all these `else if` and not simply one `else` like the one at the bottom?

andreaangiolillo: sure, these resources do not have an example in the `examples` folder and the gen doc tool fails if the example is not there. The last else instead is accessing the example in the "default" location (`examples/[resource_name]/main.tf`)

marcosuma: ahhh gotcha ok, I thought all those else if would end up in the last statemnet. ok for me

marcosuma (approved):

templates/resources.md.tmpl
```diff
@@ -0,0 +1,77 @@
...
+        {{ tffile (printf "examples/%s/main.tf" .Name )}}
+    {{ end }}
+{{ end }}
+
+{{ if .SchemaMarkdown }}{{ .SchemaMarkdown | trimspace }}{{ end }}
+
+# Import 
+## Add the import
+
+For more information see: [MongoDB Atlas API - Search Node](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Atlas-Search/operation/createAtlasSearchDeployment) Documentation.
```
AgustinBettati, line 77: I believe we can remove this from the default template right? Not sure if we are capable of leaving this generic. Applies to the data source template as well.

AgustinBettati (commented):
A small detail

templates/resources.md.tmpl
andreaangiolillo: > I believe we can remove this from the default template right? Not sure if we are capable of leaving this generic.

we can have the import autogenerated if we define an `import.txt` in the example folder. similar to https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1394/files#diff-ab6cba42e3e4a9f3642faec48e42d41f668902c8ffa3041a36bbc56e01c59913. What do you think? 

> applies to the data source template as well.

data sources do not have the import

PR #1848, opened 2024-01-17
chore: Updates search_index resource with new SDK

lantoli:
## Description

Updates search_index resource with new SDK.

- Status added to documentation
- Testing improved
-  Synonyms, Analyzers, Mapping_Fields can be updated to empty
- Checked read-only fields: IndexD, Status 

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-222718

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/searchindex/resource_search_index.go
+++ b/internal/service/searchindex/resource_search_index.go
@@ -55,7 +55,6 @@ func returnSearchIndexSchema() map[string]*schema.Schema {
 		"index_id": {
 			Type:     schema.TypeString,
 			Computed: true,
-			Required: false,
 		},
 		"analyzer": {
 			Type:     schema.TypeString,
@@ -113,7 +112,6 @@ func returnSearchIndexSchema() map[string]*schema.Schema {
 		},
 		"status": {
 			Type:     schema.TypeString,
-			Optional: true,
 			Computed: true,
 		},
 		"wait_for_index_build_completion": {
@@ -412,7 +410,6 @@ func resourceMongoDBAtlasSearchIndexCreate(ctx context.Context, d *schema.Resour
 		Database:       d.Get("database").(string),
 		Name:           d.Get("name").(string),
 		SearchAnalyzer: conversion.StringPtr(d.Get("search_analyzer").(string)),
-		Status:         conversion.StringPtr(d.Get("status").(string)),
 	}
 
 	if indexType == vectorSearch {
--- a/internal/service/searchindex/resource_search_index.go
+++ b/internal/service/searchindex/resource_search_index.go
@@ -319,14 +319,38 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 		return diag.Errorf("error setting `analyzer` for search index (%s): %s", d.Id(), err)
 	}
 
-	if analyzers := searchIndex.GetAnalyzers(); len(analyzers) > 0 {
+	if searchIndex.Type != nil && *searchIndex.Type == vectorSearch {
+		fields, err := marshalSearchIndex(searchIndex.GetFields())
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("fields", fields); err != nil {
+			return diag.Errorf("error setting `fields` for for search index (%s): %s", d.Id(), err)
+		}
+	} else {
+		analyzers := searchIndex.GetAnalyzers()
 		searchIndexMappingFields, err := marshalSearchIndex(analyzers)
 		if err != nil {
 			return diag.FromErr(err)
 		}
-
 		if err := d.Set("analyzers", searchIndexMappingFields); err != nil {
 			return diag.Errorf("error setting `analyzers` for search index (%s): %s", d.Id(), err)
+		}
+
+		if searchIndex.Mappings != nil {
+			if err := d.Set("mappings_dynamic", searchIndex.Mappings.Dynamic); err != nil {
+				return diag.Errorf("error setting `mappings_dynamic` for search index (%s): %s", d.Id(), err)
+			}
+
+			if len(searchIndex.Mappings.Fields) > 0 {
+				searchIndexMappingFields, err := marshalSearchIndex(searchIndex.Mappings.Fields)
+				if err != nil {
+					return diag.FromErr(err)
+				}
+				if err := d.Set("mappings_fields", searchIndexMappingFields); err != nil {
+					return diag.Errorf("error setting `mappings_fields` for for search index (%s): %s", d.Id(), err)
+				}
+			}
 		}
 	}
 
@@ -350,34 +374,6 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 		return diag.Errorf("error setting `synonyms` for search index (%s): %s", d.Id(), err)
 	}
 
-	if searchIndex.Mappings != nil {
-		if err := d.Set("mappings_dynamic", searchIndex.Mappings.Dynamic); err != nil {
-			return diag.Errorf("error setting `mappings_dynamic` for search index (%s): %s", d.Id(), err)
-		}
-
-		if len(searchIndex.Mappings.Fields) > 0 {
-			searchIndexMappingFields, err := marshalSearchIndex(searchIndex.Mappings.Fields)
-			if err != nil {
-				return diag.FromErr(err)
-			}
-
-			if err := d.Set("mappings_fields", searchIndexMappingFields); err != nil {
-				return diag.Errorf("error setting `mappings_fields` for for search index (%s): %s", d.Id(), err)
-			}
-		}
-	}
-
-	if fields := searchIndex.GetFields(); len(fields) > 0 {
-		fields, err := marshalSearchIndex(fields)
-		if err != nil {
-			return diag.FromErr(err)
-		}
-
-		if err := d.Set("fields", fields); err != nil {
-			return diag.Errorf("error setting `fields` for for search index (%s): %s", d.Id(), err)
-		}
-	}
-
 	return nil
 }
 
@@ -395,7 +391,11 @@ func flattenSearchIndexSynonyms(synonyms []admin.SearchSynonymMappingDefinition)
 
 func marshalSearchIndex(fields any) (string, error) {
 	bytes, err := json.Marshal(fields)
-	return string(bytes), err
+	str := string(bytes)
+	if str == "[]" || str == "{}" { // empty JSON array or object
+		str = ""
+	}
+	return str, err
 }
 
 func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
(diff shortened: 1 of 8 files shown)


internal/service/searchindex/resource_search_index.go
```diff
@@ -404,30 +403,27 @@ func resourceMongoDBAtlasSearchIndexCreate(ctx context.Context, d *schema.Resour
...
 	indexType := d.Get("type").(string)
-
 	searchIndexRequest := &admin.ClusterSearchIndex{
 		Type:           conversion.StringPtr(indexType),
 		Analyzer:       conversion.StringPtr(d.Get("analyzer").(string)),
 		CollectionName: d.Get("collection_name").(string),
 		Database:       d.Get("database").(string),
 		Name:           d.Get("name").(string),
 		SearchAnalyzer: conversion.StringPtr(d.Get("search_analyzer").(string)),
-		Status:         conversion.StringPtr(d.Get("status").(string)),
```
lantoli, line 415: Status is read-only, don't pass in Create

internal/testutil/acc/attribute_checks.go
```diff
@@ -36,17 +36,21 @@ func IntGreatThan(value int) resource.CheckResourceAttrWithFunc {
 	}
 }
 
-func JSONEquals[T any](value T) resource.CheckResourceAttrWithFunc {
+func JSONEquals(expected string) resource.CheckResourceAttrWithFunc {
```
lantoli, line 39: changed to make it easier to use

internal/service/searchindex/resource_search_index.go
```diff
@@ -320,15 +319,39 @@ func resourceMongoDBAtlasSearchIndexRead(ctx context.Context, d *schema.Resource
 		return diag.Errorf("error setting `analyzer` for search index (%s): %s", d.Id(), err)
 	}
 
-	if analyzers := searchIndex.GetAnalyzers(); len(analyzers) > 0 {
+	if searchIndex.Type != nil && *searchIndex.Type == vectorSearch {
```
AgustinBettati: q: if these values are coming from the API, is there a need to add this conditional depending on search index type?

internal/service/searchindex/resource_search_index.go
```diff
@@ -396,38 +391,39 @@ func flattenSearchIndexSynonyms(synonyms []admin.SearchSynonymMappingDefinition)
...
 	indexType := d.Get("type").(string)
-
 	searchIndexRequest := &admin.ClusterSearchIndex{
 		Type:           conversion.StringPtr(indexType),
 		Analyzer:       conversion.StringPtr(d.Get("analyzer").(string)),
 		CollectionName: d.Get("collection_name").(string),
 		Database:       d.Get("database").(string),
 		Name:           d.Get("name").(string),
 		SearchAnalyzer: conversion.StringPtr(d.Get("search_analyzer").(string)),
-		Status:         conversion.StringPtr(d.Get("status").(string)),
```
AgustinBettati, line 415: If status is read only we can remove the `Optional` property in the schema and leave it as `Computed`. Also saw it is currently not defined in the docs.

internal/service/searchindex/resource_search_index.go
```diff
@@ -396,38 +391,39 @@ func flattenSearchIndexSynonyms(synonyms []admin.SearchSynonymMappingDefinition)
 
 func marshalSearchIndex(fields any) (string, error) {
 	bytes, err := json.Marshal(fields)
-	return string(bytes), err
+	str := string(bytes)
+	if str == "[]" || str == "{}" { // empty JSON array or object
```
AgustinBettati: q: why was this check added in these changes?

internal/service/searchindex/resource_search_index.go
marcosuma: why are we changing the order and logic of the if statements here?

internal/service/searchindex/resource_search_index.go
```diff
@@ -545,35 +543,35 @@ func validateSearchAnalyzersDiff(k, old, newStr string, d *schema.ResourceData)
 	return true
 }
 
-func unmarshalSearchIndexMappingFields(mappingString string) (map[string]any, diag.Diagnostics) {
-	if mappingString == "" {
-		return nil, nil
+func unmarshalSearchIndexMappingFields(str string) (map[string]any, diag.Diagnostics) {
```
marcosuma, line 543: maybe an opportunity to unit test all these changed methods?

internal/service/searchindex/resource_search_index.go
```diff
@@ -396,38 +391,39 @@ func flattenSearchIndexSynonyms(synonyms []admin.SearchSynonymMappingDefinition)
 
 func marshalSearchIndex(fields any) (string, error) {
```
marcosuma: unit test?

internal/service/searchindex/resource_search_index_test.go
```diff
@@ -332,76 +411,30 @@ func testAccSearchIndexConfigMapping(projectIDStr, indexName, databaseName, clus
...
 			cluster_name     = %[1]s
 			project_id       = %[2]s
 			index_id 				 = mongodbatlas_search_index.test.index_id
 		}
-	`, clusterNameStr, projectIDStr, indexName, databaseName, collectionName, searchAnalyzer)
+	`, clusterNameStr, projectIDStr, indexName, databaseName, collectionName, searchAnalyzer, analyzersTF, mappingsFieldsTF)
 }
 
-func testAccSearchIndexConfigSynonyms(projectIDStr, indexName, databaseName, clusterNameStr, clusterTerraformStr string) string {
+func configSynonyms(projectIDStr, indexName, databaseName, clusterNameStr, clusterTerraformStr string, has bool) string {
```
marcosuma: nit: maybe a better name for `has`? Unless I am missing something obvious here

internal/service/searchindex/resource_search_index_test.go
```diff
@@ -159,7 +152,7 @@ func TestAccSearchIndexRS_withSynonyms(t *testing.T) {
 		CheckDestroy:             acc.CheckDestroySearchIndex,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccSearchIndexConfigSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr),
+				Config: configSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, true),
```
marcosuma: very nit: but for a good reading style we (not just you) should avoid passing bool values directtly to methods and instead use bool vars that has meaningful names. 
```suggestion
				Config: configSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, hasOrIsSomething),
```

internal/service/searchindex/resource_search_index.go
lantoli: yes, because depending on the type only some fields are allowed to be filled.  before empty slices would be removed from call so it was nice, now that are slice pointers we can send an empty slice

internal/service/searchindex/resource_search_index.go
lantoli: I think the intention was not to expose it in the TF schema as not in the doc, is it ok if i remove it from the schema as it's not documented?

internal/service/searchindex/resource_search_index.go
lantoli: because now some fields can be empty arrays or objects instead of nil pointer

internal/service/searchindex/resource_search_index.go
lantoli: now we want to have empty arrays (instead of nil pointers) so for instance they're updated correctly.  but depending on the vector type some arrays can't not be sent even empty.    basically here we're doing the same logic as we do in Create.

the order is not important, at the end we need to set all fields in any order, the important part is to do different sets depending on the search type

internal/service/searchindex/resource_search_index.go
lantoli: not a strong opinion but don't know if they add big value here as in other cases because they're local functions that are simple wrappers to json functionality

internal/service/searchindex/resource_search_index_test.go
lantoli: if it's fine I'll rename to configWithSynonyms to be consistent with other functions

internal/service/searchindex/resource_search_index_test.go
lantoli: i hear you.  here being configWithSynonyms I thought it could be understood.  WDYT about defining constansts:
withSynonyms = true
withoutSynonyms = true

and call like: 

configWithSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, withSynonyms),

configWithSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, withoutSynonyms),


or being he function already with "synoyms" on it, what about constant to call:


configWithSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, with),

configWithSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, without),



i see it a bit redundant but maybe it's clearer

internal/service/searchindex/resource_search_index.go
lantoli: answered below

internal/service/searchindex/resource_search_index_test.go
```diff
@@ -195,6 +188,99 @@ func TestAccSearchIndexRS_withSynonyms(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckSearchIndexExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "synonyms.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "synonyms.0.analyzer", "lucene.simple"),
+					resource.TestCheckResourceAttr(resourceName, "synonyms.0.name", "synonym_test"),
+					resource.TestCheckResourceAttr(resourceName, "synonyms.0.source_collection", collectionName),
+				),
+			},
+			{
+				Config: configWithSynonyms(clusterInfo.ProjectIDStr, indexName, databaseName, clusterInfo.ClusterNameStr, clusterInfo.ClusterTerraformStr, without),
```
lantoli, line 214: @marcosuma using with / without instead of true / false

internal/service/searchindex/resource_search_index.go
lantoli: it's strange that Optional is only in DS not in RS.  removed Optional.
Added to documentation, thanks for noticing it!

internal/service/searchindex/resource_search_index.go
AgustinBettati: But for this case that it is processing the response from the api into the terraform state it would not be necessary right?

website/docs/r/search_index.html.markdown
```diff
@@ -197,6 +197,7 @@ EOF
 * `synonyms` - Synonyms mapping definition to use in this index.
 
 * `fields` - Array of [Fields](https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/#std-label-fts-data-types-knn-vector) to configure this `vectorSearch` index. It is mandatory for vector searches and it must contain at least one `vector` type field. This field needs to be a JSON string in order to be decoded correctly.
+* `status` - (Computed) Current status of the index.
```
AgustinBettati: for consistency with our docs, should we define an Arguments section and place the computed attribute there? example: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/project#attributes-reference

internal/service/searchindex/resource_search_index.go
AgustinBettati: I see, seems a little hacky to include in this function. Maybe we can check when calling this function that the input parameter is non empty? That way it is clear when the terraform attribute will be left empty.

internal/service/searchindex/resource_search_index.go
lantoli: it is because the marshals now return empty slices (that are not nil pointers) so if we don't do this there will be an error because an unexpected change plan

internal/service/searchindex/resource_search_index.go
lantoli: it's really the same idea as we're already doing in the unmarshal functions like:
	if str == "" {
		return fields, nil
	}

not sure about checking this in each caller as we really want always this functionality, i.e. empty arrays to be serialized to empty strings

website/docs/r/search_index.html.markdown
lantoli: thx, done

internal/service/searchindex/resource_search_index.go
lantoli: I've changed it to only check for empty arrays that it's the real need:

if str == "[]" { // empty JSON array serialized to empty string

there is no need to check for empty objects here as it's a local function, in a generic utiliy function it could make sense

AgustinBettati: Looking into the usages I struggle to understand why this is now needed vs the previous implementation, taking into account that the input of this function are using the .Get of the SDK to pass a slice (not pointer). Just want to make sure we have no change in behaviour.

lantoli: in the prev. implementation empty arrays were nil but now they are len 0 arrays so they serialize to "[]" when we want to serialize to "".

if we don't do that we get this error:
```
 Step 1/1 error:  After applying this test step, the plan was not empty.
...
# mongodbatlas_search_index.test will be updated in-place
          ~ resource "mongodbatlas_search_index" "test" {
              - analyzers        = jsonencode([])
...
 resource_search_index_test.go:231: Step 2/2 error: Check failed: Check 2/2 error: mongodbatlas_search_index.test: Attribute 'analyzers' expected "", got "[]"
FAIL

```

internal/service/searchindex/resource_search_index.go
```diff
@@ -320,15 +319,39 @@ func resourceMongoDBAtlasSearchIndexRead(ctx context.Context, d *schema.Resource
 		return diag.Errorf("error setting `analyzer` for search index (%s): %s", d.Id(), err)
 	}
 
-	if analyzers := searchIndex.GetAnalyzers(); len(analyzers) > 0 {
```
AgustinBettati: As a follow up to our conversation, as seen here before calling `marshalSearchIndex` we checked that analyzers array was non empty. This non empty check has been removed which is forcing to handle the empty array value in the `marshal`. I guess the question would be if the condition was removed intentionally, and from there we can define if `marshalSearchIndex` effectively need to handle this case

lantoli: in Read it doesn't really matter but the logic len(array) > 0 in Update was wrong as it caused the array not to be updated if it's emptied.   
I'm trying to simplify the logic and havae a 0-len array as a normal array citizen with not additional logic needed for it, we shouldn't need to do specific logic for 0-len arrays (except serialize to an empty string).

lantoli: at the end I've followed your recommendation and don't make this extra case for marshalSearchIndex, thx!

internal/service/searchindex/resource_search_index.go
lantoli: at the end I've returned to the original order and logic following @AgustinBettati advice and using len > 0 again :-)

AgustinBettati (approved):
LGTM

marcosuma (approved):

PR #1859, opened 2024-01-19
chore: Increase parallelism in Test Suite

lantoli:
## Description

Increases parallelism in Test Suite to make test execution faster.

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -39,7 +39,7 @@ jobs:
     needs: [clean-before, versions]
     if: ${{ !cancelled() }}
     strategy:
-      max-parallel: 2
+      max-parallel: 1
       fail-fast: false
       matrix:
         terraform_version: ${{ fromJSON(needs.versions.outputs.terraform_matrix) }}
@@ -55,7 +55,7 @@ jobs:
     needs: [clean-before, versions]
     if: ${{ !cancelled() }}
     strategy:
-      max-parallel: 2
+      max-parallel: 1
       fail-fast: false
       matrix:
         terraform_version: ${{ fromJSON(needs.versions.outputs.terraform_matrix) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-suite.yml
```diff
@@ -52,10 +52,10 @@ jobs:
       provider_version: ${{ matrix.provider_version }}
 
   acc-tests:
-    needs: [mig-tests, versions]
+    needs: [clean-before, versions]
```
oarbusi, line 55: why the change from `mig-tests` to `clean-before`? I see that you added `mig-tests` in `clean-after` but not in the `clean-before`

lantoli: so mig and acc run in parallel instead of being a chain of mig then acc

maastha:
[q] just for my knowledge, how do we test this?

lantoli:
@maastha I ran manually the Test Suite GH action in this branch.   for this change, i didn't expect until the end, i saw that both tests were running in parallel and seemed right and canceled.   for more complex changes i wait until the GH action in this branch finishes completely

AgustinBettati (approved):
LGTM, only risk I see is that organization reaches limit of projects but we can always revert and look into other options in that case.

lantoli:
@AgustinBettati i don't think we reach org limit but could happen, however i think it will probably fail in other places.  anyway i'll wait until Mon so it doesn't interfere with TF 1.7 in case something fails

.github/workflows/test-suite.yml
lantoli: mig_test already had:

 mig-tests:
    needs: [clean-before, versions]

lantoli:
@AgustinBettati it could happen but i guess it'll fail because other things.  so to go a bit more smoother I'm going to have mig and acc tests in paralel, but each on in serial and see what happens. if it works we can increase parallelism inside acc and mig tests

PR #1874, opened 2024-01-25
feat: Adds support to new Federated Auth parameters for OIDC 

oarbusi:
## Description

Adds support for new federated auth attributes to support OIDC IdP.

- Resource `federated_settings_identity_provider` now supports both types of IdP: SAML and OIDC (workforce only)
- Data source `federated_settings_identity_provider` now supports both types of IdP: SAML and OIDC (workforce only)
- Data source `federated_settings_identity_providers` now returns both types of IdP: SAML and OIDC (workforce only)

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-220798

## Type of change:

## Required Checklist:


internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
```diff
@@ -360,95 +462,12 @@ func resourceMongoDBAtlasFederatedSettingsIdentityProviderImportState(ctx contex
 		return oldSDKImport(federationSettingsID, idpID, d, meta)
 	}
 
-	federatedSettingsIdentityProvider, _, err := connV2.FederatedAuthenticationApi.GetIdentityProvider(context.Background(), *federationSettingsID, *idpID).Execute()
```
oarbusi, line 363: For consistency with other resources, I am removing the non necessary code in import. Only necessary attribute to set in import is the Id

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
```diff
@@ -36,6 +36,35 @@ func TestAccFedDSFederatedSettingsIdentityProvider_basic(t *testing.T) {
 	})
 }
 
+func TestAccFedDSFederatedSettingsIdentityProvider_oidcBasic(t *testing.T) {
+	acc.SkipTestExtCred(t)
```
lantoli, line 44: would it be possible easy to have a migration test even with acc.SkipTestExtCred(t) or it's hard to do?

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
```diff
@@ -36,6 +36,35 @@ func TestAccFedDSFederatedSettingsIdentityProvider_basic(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckFederatedSettings(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccMongoDBAtlasDataSourceFederatedSettingsIdentityProviderConfig(federatedSettingsID, idpID),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckMongoDBAtlasFederatedSettingsIdentityProvidersExists(resourceName),
+
+					resource.TestCheckResourceAttrSet(resourceName, "federation_settings_id"),
```
lantoli, line 59: can we also check for the expected value or you think it's not so important?

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_providers.go
```diff
@@ -227,15 +255,27 @@ func dataSourceMongoDBAtlasFederatedSettingsIdentityProvidersRead(ctx context.Co
...
-	if err != nil {
-		return diag.Errorf("error getting federatedSettings IdentityProviders assigned (%s): %s", federationSettingsID, err)
+	oidcParams := &admin.ListIdentityProvidersApiParams{
+		FederationSettingsId: federationSettingsID.(string),
+		Protocol:             conversion.StringPtr("OIDC"),
+	}
+	samlParams := &admin.ListIdentityProvidersApiParams{
+		FederationSettingsId: federationSettingsID.(string),
+		Protocol:             conversion.StringPtr("SAML"),
+	}
```
lantoli, line 266: can both protocols be enabled at the same time?

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_providers_test.go
```diff
@@ -30,7 +30,11 @@ func TestAccFedDSFederatedSettingsIdentityProviders_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(resourceName, "federation_settings_id"),
 					resource.TestCheckResourceAttrSet(resourceName, "results.#"),
 					resource.TestCheckResourceAttrSet(resourceName, "results.0.acs_url"),
-					resource.TestCheckResourceAttrSet(resourceName, "results.0.display_name"),
+					resource.TestCheckResourceAttrSet(resourceName, "results.1.audience_claim.0"),
```
lantoli, line 33: same as above, is it interesting check for the expected values?

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
```diff
@@ -114,42 +145,76 @@ func resourceMongoDBAtlasFederatedSettingsIdentityProviderRead(ctx context.Conte
...
+			return diag.FromErr(fmt.Errorf("error setting response signature algorithm (%s): %s", d.Id(), err))
+		}
 
-	if err := d.Set("okta_idp_id", federatedSettingsIdentityProvider.OktaIdpId); err != nil {
-		return diag.FromErr(fmt.Errorf("error setting OktaIdpID (%s): %s", d.Id(), err))
+		if err := d.Set("sso_debug_enabled", federatedSettingsIdentityProvider.SsoDebugEnabled); err != nil {
+			return diag.FromErr(fmt.Errorf("error setting sso debug enabled (%s): %s", d.Id(), err))
+		}
+
+		if err := d.Set("sso_url", federatedSettingsIdentityProvider.SsoUrl); err != nil {
```
lantoli, line 164: not specific to this PR, there are many error checks like this that seem a bit verbose and we might be able to extract to some helper functions

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
oarbusi: I think that testing that the attribute itself is set it's enough. In the unit tests in model_federated_settings_identity_provider_test.go we are already testing the values.

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
```diff
@@ -230,6 +295,11 @@ func resourceMongoDBAtlasFederatedSettingsIdentityProviderUpdate(ctx context.Con
 		return diag.FromErr(fmt.Errorf("error retreiving federation settings identity provider (%s): %s", federationSettingsID, err))
 	}
 
+	if d.HasChange("protocol") {
```
lantoli, line 301: related to a question above, can be only 1 or 2 protocols at the same time?

lantoli: if only one, have you tried to change the protocol and see how it's handled the attributes that can only be in one type of protocol?   e.g. if just protocol changes, it would probably fail because you would need to delete the fields only used in the old protocol and add the ones used in the new one

lantoli (approved):

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
oarbusi: I would say that it's not straight forward even with acc.SkipTestExtCred(t) but I am open to discussing it. Maybe as part of enabling acceptance tests in the CI we could create migration tests?

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_providers.go
oarbusi: One federation can have multiple IdPs, so you can have one SAML IdP and one OIDC IdP in the same federation. But one IdP can be either SAML or OIDC, not both

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
oarbusi: Yes, if you only change the protocol it fails

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_providers_test.go
oarbusi: I think that testing that the attribute itself is set it's enough. In the unit tests in model_federated_settings_identity_provider_test.go we are already testing the values.

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
AgustinBettati: This could be good test to add to ensure no breaking changes are encountered. It can import a resource with previous version and then update to the latest and expect an empty plan.

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
```diff
@@ -41,11 +41,11 @@ func Resource() *schema.Resource {
 			},
```
AgustinBettati, line 44: For resource testing we can consider adding test that imports a config with SAML (using new id format) and then adjust attributes to update to OIDC.

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
```diff
@@ -360,95 +462,12 @@ func resourceMongoDBAtlasFederatedSettingsIdentityProviderImportState(ctx contex
 		return oldSDKImport(federationSettingsID, idpID, d, meta)
```
AgustinBettati: Is there a need to have this if? I believe both cases are handled the same way now

AgustinBettati (commented):
LGTM, small comment in import operation

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
marcosuma: how does it fail? is it a readable error?

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
marcosuma: +1 with the test. Perhaps no need to block the release (and this PR) but it's a good one to have, we should actually aim at always have a test like this.

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider.go
```diff
@@ -233,8 +263,56 @@ func dataSourceMongoDBAtlasFederatedSettingsIdentityProviderRead(ctx context.Con
 		return diag.Errorf("error getting federatedSettings IdentityProviders assigned (%s): %s", federationSettingsID, err)
 	}
 
-	if err := d.Set("acs_url", federatedSettingsIdentityProvider.AcsUrl); err != nil {
-		return diag.FromErr(fmt.Errorf("error setting `acs_url` for federatedSettings IdentityProviders: %s", err))
+	if federatedSettingsIdentityProvider.GetProtocol() == "SAML" {
```
marcosuma: can we use const for SAML and OIDC?

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_providers.go
```diff
@@ -227,15 +255,27 @@ func dataSourceMongoDBAtlasFederatedSettingsIdentityProvidersRead(ctx context.Co
...
 	}
-	// There is no available paginated method
-	// If no protocol is specified, only SAML identity providers will be returned. Use ListIdentityProvidersWithParams to specify protocol
-	federatedSettingsIdentityProviders, _, err := connV2.FederatedAuthenticationApi.ListIdentityProviders(ctx, federationSettingsID.(string)).Execute()
 
-	if err != nil {
-		return diag.Errorf("error getting federatedSettings IdentityProviders assigned (%s): %s", federationSettingsID, err)
+	oidcParams := &admin.ListIdentityProvidersApiParams{
+		FederationSettingsId: federationSettingsID.(string),
+		Protocol:             conversion.StringPtr("OIDC"),
```
marcosuma: we can reuse those const here

marcosuma (approved):

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
marcosuma: I agree with you we're using unit tests and we shouldn't really "test the other components", but think about an API change that changes a computed value (like it's been happening recently): how would we detect it if not through these tests?

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
oarbusi: I checked the [backend code](https://github.com/10gen/mms/blob/6850f30e48c88c6bd5d596ec386276c06e79705c/server/src/main/com/xgen/cloud/federation/_public/svc/IdentityProviderSvcProxy.java#L81) and seems like this operation is not allowed. If existing IdP is OIDC the update must be OIDC. So this case can't happen. 
Thanks you for raising this!

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
oarbusi: You are right! Will remove it

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider.go
oarbusi: You are right! Done

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_providers.go
oarbusi: Done!

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
oarbusi: Unexpected error from the server. From looking at the [code](https://github.com/10gen/mms/blob/6850f30e48c88c6bd5d596ec386276c06e79705c/server/src/main/com/xgen/cloud/federation/_public/svc/IdentityProviderSvcProxy.java#L81), updating protocol or trying to update from SAML to OIDC or the other way around is not possible

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
oarbusi: Good point! I included some value checks in both OIDC and SAML acceptance test to check most important attributes values

internal/service/federatedsettingsidentityprovider/resource_federated_settings_identity_provider.go
AgustinBettati: Thanks for checking this, as a small detail we might want to mention this in the docs in the protocol attribute (the fact that it cannot be modified from one to another).

internal/service/federatedsettingsidentityprovider/data_source_federated_settings_identity_provider_test.go
oarbusi: I tried to add the migration test with the help of @lantoli and it seems that there is an issue when running a migration test using `ExternalProvider` and `ImportState` in the acceptance test step at the same time. We will look again after release.

PR #1876, opened 2024-01-26
fix: Sets replication_specs IDs when updating them.

marcosuma:
## Description

Passes the replicationSpec item id to the PATCH request during updates. This fixes the error:

```
400 (request "INVALID_CLUSTER_CONFIGURATION") The specified cluster configuration is not valid: Updating id of existing replication spec.
```

Link to any related issue(s): [CLOUDP-227483](https://jira.mongodb.org/browse/CLOUDP-227483)

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedcluster/resource_advanced_cluster.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster.go
@@ -850,6 +850,10 @@ func expandAdvancedReplicationSpec(tfMap map[string]any) *matlas.AdvancedReplica
 		RegionConfigs: expandRegionConfigs(tfMap["region_configs"].([]any)),
 	}
 
+	if tfMap["id"].(string) != "" {
+		apiObject.ID = tfMap["id"].(string)
+	}
+
 	return apiObject
 }
 
--- a/internal/service/advancedcluster/resource_advanced_cluster_test.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster_test.go
@@ -606,8 +606,8 @@ func TestAccClusterAdvancedClusterConfig_ReplicationSpecsAndShardUpdating(t *tes
 		orgID            = os.Getenv("MONGODB_ATLAS_ORG_ID")
 		projectName      = acctest.RandomWithPrefix("test-acc")
 		rName            = acctest.RandomWithPrefix("test-acc")
-		numShards        = 1
-		numShardsUpdated = 1
+		numShards        = "1"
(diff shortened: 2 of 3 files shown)


lantoli (approved):

AgustinBettati (approved):
LGTM

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -850,6 +850,10 @@ func expandAdvancedReplicationSpec(tfMap map[string]any) *matlas.AdvancedReplica
 		RegionConfigs: expandRegionConfigs(tfMap["region_configs"].([]any)),
 	}
 
+	if tfMap["id"].(string) != "" {
```
AgustinBettati, line 853: Just to confirm, we are setting the id in update request of `mongodbatlas_cluster` resource right?

marcosuma: @AgustinBettati this method is called in both update and create. The reason why it should only run on the update is because of the if itself. I am assuming that at the creation phase this field is empty. Any concerns?

andreaangiolillo: The function is called `expandAdvancedReplicationSpec` but this PR is also updating the cluster test 🤔 . Is `expandAdvancedReplicationSpec` called also in the cluster resource? if not, is the cluster resource already sending the `id` during the update?

andreaangiolillo (approved):

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: Yes that statement seems correct to me. My doubt was more related to how we handle this id in `mongodbatlas_cluster`, not a blocker just something to keep in mind.

marcosuma: > The function is called expandAdvancedReplicationSpec but this PR is also updating the cluster test
@andreaangiolillo @AgustinBettati 

Adding tests to `cluster` was more like to increase test coverage. Sorry if it created confusion. confirming that `expandAdvancedReplicationSpec` is only affecting `advanced_cluster`.

regarding cluster: that is a completely different story: while in here we use list, in that one we are representing replication_specs as a set and we're also using v1.0 - I added the test on purpose: to see if it was able to handle replicationspecs updates and indeed it does without having to change the logic.

PR #1885, opened 2024-01-30
chore: New released Atlas Go SDK can't be used until several hours later

gssbzn:
## Description

CLOUDP-221977

using latest is known to cause cache issues with go.dev, using directly the tag should help reduce those problems


wtrocki (approved):
LGTM

lantoli (approved):
thanks!

PR #1947, opened 2024-02-14
chore: Upgrades `advanced_cluster` resource to auto-generated SDK

lantoli:
## Description

Upgrades `advanced_cluster` resource to auto-generated SDK

Link to any related issue(s): CLOUDP-229554

## Type of change:

## Required Checklist:


internal/service/cluster/model_cluster.go
```diff
@@ -0,0 +1,528 @@
...
+	providerSettings := &matlas.ProviderSettings{
+		InstanceSizeName: cast.ToString(d.Get("provider_instance_size_name")),
+		ProviderName:     providerName,
+		RegionName:       region,
+		VolumeType:       cast.ToString(d.Get("provider_volume_type")),
+	}
+
+	if d.HasChange("provider_disk_type_name") {
+		_, newdiskTypeName := d.GetChange("provider_disk_type_name")
+		diskTypeName := cast.ToString(newdiskTypeName)
```
oarbusi, line 453: more of a general comment: I see cast.ToString/Bool/etc being used a lot and also see .(bool)/.(string)/etc being used. I think the latter is the more idiomatic way of casting in Go and in my opinion more clear. WDYT?  No need to change in this PR but I think we should probably only use one of the ways moving forward

oarbusi (approved):
LGTM overall

internal/service/cluster/model_cluster.go
lantoli: you're totally right, it's something i also discussed with @gssbzn that we're using both styles all across the code and we might want to unify

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -689,19 +671,23 @@ func resourceMongoDBAtlasAdvancedClusterUpdate(ctx context.Context, d *schema.Re
...
+		cluster.VersionReleaseSystem = conversion.StringPtr(d.Get("version_release_system").(string))
 	}
 
 	if d.HasChange("accept_data_risks_and_force_replica_set_reconfig") {
-		cluster.AcceptDataRisksAndForceReplicaSetReconfig = d.Get("accept_data_risks_and_force_replica_set_reconfig").(string)
+		t, ok := conversion.StringToTime(d.Get("accept_data_risks_and_force_replica_set_reconfig").(string))
+		if !ok {
+			return diag.FromErr(fmt.Errorf(errorUpdate, clusterName, "accept_data_risks_and_force_replica_set_reconfig time format is incorrect"))
+		}
+		cluster.AcceptDataRisksAndForceReplicaSetReconfig = &t
```
AgustinBettati: Do we have to additionally contemplate the possibility the user wants to remove this attribute? As of now I believe that would result in an error because empty string will fail to parse as a Time struct.

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -229,78 +230,76 @@ func DataSource() *schema.Resource {
...
 		return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "disk_size_gb", clusterName, err))
 	}
 
-	if err := d.Set("encryption_at_rest_provider", cluster.EncryptionAtRestProvider); err != nil {
+	if err := d.Set("encryption_at_rest_provider", cluster.GetEncryptionAtRestProvider()); err != nil {
 		return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "encryption_at_rest_provider", clusterName, err))
 	}
 
-	if err := d.Set("labels", FlattenLabels(RemoveLabel(cluster.Labels, DefaultLabel))); err != nil {
+	if err := d.Set("labels", flattenLabels(cluster.GetLabels())); err != nil {
```
AgustinBettati, line 274: When removing the logic of handling the default label, how does this impact existing terraform state/clusters? I see the risk that now that the read does not filter out the default label, users could encounter a non empty plan.

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -0,0 +1,955 @@
...
+	}
+	// v == MinInt
+	return 0
+}
+
+func IsSharedTier(instanceSize string) bool {
+	return instanceSize == "M0" || instanceSize == "M2" || instanceSize == "M5"
+}
+
+func UpgradeCluster(ctx context.Context, connV2 *admin.APIClient, request *admin.LegacyAtlasTenantClusterUpgradeRequest, projectID, name string, timeout time.Duration) (*admin.LegacyAtlasCluster, *http.Response, error) {
```
AgustinBettati: curious why this was moved over to the model_* file, as we usually keep this for model conversion logic.

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -0,0 +1,955 @@
...
+		if containers[i].GetProviderName() == cluster.GetProviderName() &&
+			containers[i].GetRegion() == cluster.GetRegionName() || // For Azure
+			containers[i].GetRegionName() == cluster.GetRegionName() { // For AWS
+			return containers[i].GetId()
+		}
+	}
+	return ""
+}
+
+func expandProcessArgs(d *schema.ResourceData, p map[string]any) *admin.ClusterDescriptionProcessArgs {
```
AgustinBettati: nit: any reason why we return a pointer here? Asking because in the usages the value is being accessed directly.

internal/service/advancedcluster/resource_advanced_cluster_migration_test.go
```diff
@@ -4,90 +4,65 @@ import (
...
-				ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
-				Config:                   testAccMongoDBAtlasAdvancedClusterConfigSingleProvider(orgID, projectName, rName),
-				ConfigPlanChecks: resource.ConfigPlanChecks{
-					PreApply: []plancheck.PlanCheck{
-						acc.DebugPlan(),
-						plancheck.ExpectEmptyPlan(),
-					},
-				},
-			},
+			mig.TestStep(config),
```
AgustinBettati: 💯, I see this has already been introduced in previous PR but renaming to something more explicity like `TestEmptyPlanStep` would be nice imo.

AgustinBettati (changes requested):
main comment to double check is the one related to `accept_data_risks_and_force_replica_set_reconfig`

internal/service/advancedcluster/data_source_advanced_cluster.go
AgustinBettati: I see that the expander and flattener have the logic of removing/validating this default label, so the only change is that now in the create we do not add this default label. Would this be correct?

internal/service/advancedcluster/resource_advanced_cluster_migration_test.go
lantoli: that's a good idea, I'll change for the tests in test groups changing in this PR

lantoli: i'll use mig.TestStepCheckEmptyPlan if that's fine

internal/service/advancedcluster/resource_advanced_cluster.go
lantoli: you're right, code changed to keep behavior,  if it's removed it won't be sent now

internal/service/advancedcluster/data_source_advanced_cluster.go
lantoli: correct, the logic has been simplified but we're keeping the logic to filter out the special label so no unexpected plans will happen

internal/service/advancedcluster/model_advanced_cluster.go
lantoli: you're right, moved

internal/service/advancedcluster/model_advanced_cluster.go
lantoli: thx, changed to make it more explicit that a value is always returned

internal/service/advancedcluster/resource_advanced_cluster_migration_test.go
lantoli: changed for adv_cluster

AgustinBettati (approved):
LGTM, thanks for the follow ups

internal/service/advancedcluster/data_source_advanced_cluster.go
maastha: @lantoli Wanted to call out that we currently mention in the documentation for cluster and advanced cluster that users should not set `Infrastructure Tool` label, should we update this in documentation as well?
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/d7b15fc556943b25eeeee2c08e894c15780ddae2/website/docs/d/advanced_cluster.html.markdown?plain=1#L91 
Or can users still not set it based on the logic?

lantoli: i've kept that behavior so the doc is correct, user will get an error if trying to set that label

PR #2010, opened 2024-03-12
chore: Creates project for test execution

lantoli:
## Description

Creates project for test execution. As an example it's used for `privatelink_endpoint_service_data_federation_online_archive` resource tests.

Link to any related issue(s): CLOUDP-237263

This creates a project for each resource test execution (if needed), example log:

```
Creating execution project: test-acc-tf-p-430893557017739723, resource: privatelinkendpointservicedatafederationonlinearchive, global project (not found): test-acc-tf-p-keep-privatelinkendpointservicedatafederationonlin
--- PASS: TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basic (11.40s)
=== RUN   TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basicWithRegionDnsName
--- PASS: TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basicWithRegionDnsName (8.42s)
PASS
Deleting execution project: test-acc-tf-p-430893557017739723, resource: privatelinkendpointservicedatafederationonlinearchive
```

Optionally it can use an existing project for that resource if it exists, example log:

```
Reusing global project: test-acc-tf-p-keep-privatelinkendpointservicedatafederationonlin, resource: privatelinkendpointservicedatafederationonlinearchive
```

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/privatelinkendpointservicedatafederationonlinearchive/data_source_privatelink_endpoint_service_data_federation_online_archive_test.go
+++ b/internal/service/privatelinkendpointservicedatafederationonlinearchive/data_source_privatelink_endpoint_service_data_federation_online_archive_test.go
@@ -15,7 +15,7 @@ var (
 
 func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchiveDS_basic(t *testing.T) {
 	var (
-		projectID               = acc.ProjectIDGlobal(t)
+		projectID               = acc.ProjectIDExecution(t)
 		endpointID              = os.Getenv("MONGODB_ATLAS_PRIVATE_ENDPOINT_ID")
 		customerEndpointDNSName = os.Getenv("MONGODB_ATLAS_PRIVATE_ENDPOINT_DNS_NAME")
 	)
--- a/internal/service/privatelinkendpointservicedatafederationonlinearchive/resource_privatelink_endpoint_service_data_federation_online_archive_test.go
+++ b/internal/service/privatelinkendpointservicedatafederationonlinearchive/resource_privatelink_endpoint_service_data_federation_online_archive_test.go
@@ -18,9 +18,13 @@ var (
 	atlasRegion  = "US_EAST_1"
 )
 
+func TestMain(m *testing.M) {
+	acc.TestMainExecution(m)
+}
+
 func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basic(t *testing.T) {
 	var (
-		projectID  = acc.ProjectIDGlobal(t)
+		projectID  = acc.ProjectIDExecution(t)
 		endpointID = os.Getenv("MONGODB_ATLAS_PRIVATE_ENDPOINT_ID")
 	)
 
@@ -51,7 +55,7 @@ func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basic(t
 }
 func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_updateComment(t *testing.T) {
 	var (
-		projectID      = acc.ProjectIDGlobal(t)
+		projectID      = acc.ProjectIDExecution(t)
 		endpointID     = os.Getenv("MONGODB_ATLAS_PRIVATE_ENDPOINT_ID")
 		commentUpdated = "Terraform Acceptance Test Updated"
 	)
@@ -98,7 +102,7 @@ func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_updateC
 
 func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basicWithRegionDnsName(t *testing.T) {
 	var (
-		projectID               = acc.ProjectIDGlobal(t)
+		projectID               = acc.ProjectIDExecution(t)
 		endpointID              = os.Getenv("MONGODB_ATLAS_PRIVATE_ENDPOINT_ID")
 		customerEndpointDNSName = os.Getenv("MONGODB_ATLAS_PRIVATE_ENDPOINT_DNS_NAME")
 	)
--- a/internal/service/privatelinkendpointservicedatafederationonlinearchive/resource_privatelink_endpoint_service_data_federation_online_archive_test.go
+++ b/internal/service/privatelinkendpointservicedatafederationonlinearchive/resource_privatelink_endpoint_service_data_federation_online_archive_test.go
@@ -19,7 +19,10 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	acc.TestMainExecution(m)
+	acc.SetupSharedResources()
+	exitCode := m.Run()
+	acc.CleanupSharedResources()
+	os.Exit(exitCode)
 }
 
 func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basic(t *testing.T) {
--- a/internal/testutil/acc/atlas.go
+++ b/internal/testutil/acc/atlas.go
@@ -4,88 +4,11 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"regexp"
-	"runtime"
-	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/atlas-sdk/v20231115007/admin"
 )
-
-// TestMainExecution must be called from TestMain in the test package if ProjectIDExecution is going to be used.
-func TestMainExecution(m *testing.M) {
-	atlasInfo.init = true
-	atlasInfo.resourceName = resourceName()
-
-	exitCode := m.Run()
-
-	if atlasInfo.needsDeletion {
-		fmt.Printf("Deleting execution project: %s, resource: %s\n", atlasInfo.projectName, atlasInfo.resourceName)
-		deleteProject(atlasInfo.projectID)
-	}
-
-	os.Exit(exitCode)
-}
-
-// ProjectIDExecution returns a project id created for the execution of the resource tests.
-func ProjectIDExecution(tb testing.TB) string {
-	tb.Helper()
-	SkipInUnitTest(tb)
-	require.True(tb, atlasInfo.init, "TestMainExecution must called to be able to use ProjectIDExecution")
-
-	atlasInfo.mu.Lock()
-	defer atlasInfo.mu.Unlock()
-
-	// lazy creation so it's only done if really needed
-	if atlasInfo.projectName == "" {
-		var globalName, globalID string
-		if atlasInfo.resourceName != "" {
-			globalName = (prefixProjectKeep + "-" + atlasInfo.resourceName)[:projectNameMaxLen]
-			globalID = projectID(globalName)
-		}
-
-		if globalID == "" {
-			atlasInfo.projectName = RandomProjectName()
-			tb.Logf("Creating execution project: %s, resource: %s, global project (not found): %s\n", atlasInfo.projectName, atlasInfo.resourceName, globalName)
-			atlasInfo.projectID = createProject(tb, atlasInfo.projectName)
-			atlasInfo.needsDeletion = true
-		} else {
-			atlasInfo.projectName = globalName
-			tb.Logf("Reusing global project: %s, resource: %s\n", atlasInfo.projectName, atlasInfo.resourceName)
-			atlasInfo.projectID = globalID
-		}
-	}
-
-	return atlasInfo.projectID
-}
-
-var atlasInfo = struct {
-	projectID     string
-	projectName   string
-	resourceName  string
-	mu            sync.Mutex
-	init          bool
-	needsDeletion bool
-}{}
-
-const (
-	projectNameMaxLen = 64
-)
-
-func resourceName() string {
-	pc, _, _, ok := runtime.Caller(2)
-	if !ok {
-		return ""
-	}
-	pattern := `([^/]+)_test\.TestMain$`
-	re := regexp.MustCompile(pattern)
-	matches := re.FindStringSubmatch(runtime.FuncForPC(pc).Name())
-	if len(matches) <= 1 {
-		return ""
-	}
-	return matches[1]
-}
 
 func createProject(tb testing.TB, name string) string {
 	tb.Helper()
@@ -106,7 +29,11 @@ func deleteProject(id string) {
 	}
 }
 
-func projectID(name string) string {
+func projectID(tb testing.TB, name string) string {
+	tb.Helper()
+	SkipInUnitTest(tb)
 	resp, _, _ := ConnV2().ProjectsApi.GetProjectByName(context.Background(), name).Execute()
-	return resp.GetId()
+	id := resp.GetId()
+	require.NotEmpty(tb, id, "Project name not found: %s", name)
+	return id
 }
--- a/internal/testutil/acc/atlas.go
+++ b/internal/testutil/acc/atlas.go
@@ -1,0 +1,116 @@
+package acc
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"runtime"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/atlas-sdk/v20231115007/admin"
+)
+
+// TestMainExecution must be called from TestMain in the test package if ProjectIDExecution is going to be used.
+func TestMainExecution(m *testing.M) {
+	if !InUnitTest() {
+		atlasInfo.init = true
+		atlasInfo.resourceName = resourceName()
+	}
+
+	exitCode := m.Run()
+
+	if !InUnitTest() && atlasInfo.needsDeletion {
+		fmt.Printf("Deleting execution project: %s, resource: %s\n", atlasInfo.projectName, atlasInfo.resourceName)
+		deleteProject(atlasInfo.projectID)
+		atlasInfo.projectID = ""
+		atlasInfo.projectName = ""
+	}
+
+	os.Exit(exitCode)
+}
+
+// ProjectIDExecution returns a project id created for the execution of the resource tests.
+func ProjectIDExecution(tb testing.TB) string {
+	tb.Helper()
+	SkipInUnitTest(tb)
+	require.True(tb, atlasInfo.init, "TestMainExecution must called to be able to use ProjectIDExecution")
+
+	atlasInfo.mu.Lock()
+	defer atlasInfo.mu.Unlock()
+
+	// lazy creation so it's only done if really needed
+	if atlasInfo.projectName == "" {
+		var globalName, globalID string
+		if atlasInfo.resourceName != "" {
+			globalName = (prefixProjectKeep + "-" + atlasInfo.resourceName)[:projectNameMaxLen]
+			globalID = projectID(globalName)
+		}
+
+		if globalID == "" {
+			atlasInfo.projectName = RandomProjectName()
+			tb.Logf("Creating execution project: %s, resource: %s, global project (not found): %s\n", atlasInfo.projectName, atlasInfo.resourceName, globalName)
+			atlasInfo.projectID = createProject(tb, atlasInfo.projectName)
+			atlasInfo.needsDeletion = true
+		} else {
+			atlasInfo.projectName = globalName
+			tb.Logf("Reusing global project: %s, resource: %s\n", atlasInfo.projectName, atlasInfo.resourceName)
+			atlasInfo.projectID = globalID
+		}
+	}
+
+	return atlasInfo.projectID
+}
+
+var atlasInfo = struct {
+	projectID     string
+	projectName   string
+	resourceName  string
+	mu            sync.Mutex
+	init          bool
+	needsDeletion bool
+}{}
+
+const (
+	projectNameMaxLen = 64
+)
+
+func resourceName() string {
+	pc, _, _, ok := runtime.Caller(2)
+	if !ok {
+		return ""
+	}
+	pattern := `([^/]+)_test\.TestMain$`
+	re := regexp.MustCompile(pattern)
+	matches := re.FindStringSubmatch(runtime.FuncForPC(pc).Name())
+	if len(matches) <= 1 {
+		return ""
+	}
+	return matches[1]
+}
+
+func createProject(tb testing.TB, name string) string {
+	tb.Helper()
+	orgID := os.Getenv("MONGODB_ATLAS_ORG_ID")
+	require.NotNil(tb, "Project creation failed: %s, org not set", name)
+	params := &admin.Group{Name: name, OrgId: orgID}
+	resp, _, err := ConnV2().ProjectsApi.CreateProject(context.Background(), params).Execute()
+	require.NoError(tb, err, "Project creation failed: %s, err: %s", name, err)
+	id := resp.GetId()
+	require.NotEmpty(tb, id, "Project creation failed: %s", name)
+	return id
+}
+
+func deleteProject(id string) {
+	_, _, err := ConnV2().ProjectsApi.DeleteProject(context.Background(), id).Execute()
+	if err != nil {
+		fmt.Printf("Project deletion failed: %s, error: %s", id, err)
+	}
+}
+
+func projectID(name string) string {
+	resp, _, _ := ConnV2().ProjectsApi.GetProjectByName(context.Background(), name).Execute()
+	return resp.GetId()
+}
--- a/internal/testutil/acc/atlas.go
+++ b/internal/testutil/acc/atlas.go
@@ -15,14 +15,12 @@ import (
 
 // TestMainExecution must be called from TestMain in the test package if ProjectIDExecution is going to be used.
 func TestMainExecution(m *testing.M) {
-	if !InUnitTest() {
-		atlasInfo.init = true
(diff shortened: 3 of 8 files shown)


internal/service/privatelinkendpointservicedatafederationonlinearchive/resource_privatelink_endpoint_service_data_federation_online_archive_test.go
```diff
@@ -18,9 +18,13 @@ var (
 	atlasRegion  = "US_EAST_1"
 )
 
+func TestMain(m *testing.M) {
```
lantoli, line 21: Go runs this function before any test in the package

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,116 @@
...
+	os.Exit(exitCode)
+}
+
+// ProjectIDExecution returns a project id created for the execution of the resource tests.
+func ProjectIDExecution(tb testing.TB) string {
+	tb.Helper()
+	SkipInUnitTest(tb)
+	require.True(tb, atlasInfo.init, "TestMainExecution must called to be able to use ProjectIDExecution")
+
+	atlasInfo.mu.Lock()
```
lantoli: we serialize access to avoid problems while creating the project

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,116 @@
...
+	init          bool
+	needsDeletion bool
+}{}
+
+const (
+	projectNameMaxLen = 64
+)
+
+func resourceName() string {
+	pc, _, _, ok := runtime.Caller(2)
```
lantoli: this is to get the full function name of the test and extract the resource name

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,116 @@
...
+	"runtime"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/atlas-sdk/v20231115007/admin"
+)
+
+// TestMainExecution must be called from TestMain in the test package if ProjectIDExecution is going to be used.
+func TestMainExecution(m *testing.M) {
```
marcosuma: suggest improving both method names. This one sounds more like an init() and the other one I don't have a suggestion because I am not really sure I understand it. It looks like it's a `getProjectID()` kind of method

marcosuma: can't this class also be called by multiple places in parallel?

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,116 @@
...
+	if !InUnitTest() {
+		atlasInfo.init = true
+		atlasInfo.resourceName = resourceName()
+	}
+
+	exitCode := m.Run()
+
+	if !InUnitTest() && atlasInfo.needsDeletion {
```
marcosuma: reading this `if` makes me thinking that we're designing a non-trivial set of rules for which resources should be re-used or deleted. I see a potential risk here that tests could start _randomly_ failing due to concurrency issues and figuring what happened is hard.

Let me start first with:
1. Why this `if`? What is the rationale?
2. Could we make things simpler? e.g. re-use resource for every test file and make it just a serial process?
3. Is there an alternative way to write this?

internal/testutil/acc/atlas.go
lantoli: no, Go test calls TestMain and we call this function from there, it's like the main function but when running tests

lantoli: any suggestion?  This method is the implementation of TestMain in the tests, it basically do some setup, execute the tests and teardown. 

I called the methods TestMainExecution and ProjectIDExecution with the intention that Execution means during the test execution of the package.     So for example ProjectIDExecution provides the id of a project that is created before the tests run and deleted after.

internal/testutil/acc/atlas.go
lantoli: by default a project is created when you call ProjectIDExecution and it's deleted after all tests run. This `if` deletes the project created for this test execution.

i made a further optimization to allow reuse of a global project for that resource if defined so we avoid creating the project during the test run. 

i'm not sure the code can be simplified to fulfil those requirements, but we'll probably learn more as we use this in other resources and might be able to simplify

lantoli: i think in next PRs it'll be more clear how this is intended to be used, the thing with this PR is that there is more implementation details than real use of the feature.

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,116 @@
...
+		fmt.Printf("Deleting execution project: %s, resource: %s\n", atlasInfo.projectName, atlasInfo.resourceName)
+		deleteProject(atlasInfo.projectID)
+		atlasInfo.projectID = ""
```
lantoli, line 28: it's not really needed to empty these string as we're about to exit the tests, but it's to make it more clear that those values are invalid from that moment

internal/testutil/acc/atlas.go
lantoli: i've simplified the `if` condition a little bit, let me know if it helps

1 - the if is to delete the project if it was created before running the tests
2 - we're reusing the resource (project) for all tests in the resource but we allow the tests to run in parallel as now so performance is not affected
3 - surely it can be written in other ways, but I think this approach is ok at the moment, we'll probably learn more soon and will be able to improve it. any suggestions?

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,112 @@
...
+		if atlasInfo.resourceName != "" {
+			globalName = (prefixProjectKeep + "-" + atlasInfo.resourceName)[:projectNameMaxLen]
+			globalID = projectID(globalName)
+		}
+
+		if globalID == "" {
+			atlasInfo.projectName = RandomProjectName()
+			tb.Logf("Creating execution project: %s, resource: %s, global project (not found): %s\n", atlasInfo.projectName, atlasInfo.resourceName, globalName)
+			atlasInfo.projectID = createProject(tb, atlasInfo.projectName)
+			atlasInfo.needsDeletion = true
```
lantoli: this flags that we created a project for the test execution and we'll need to delete it after all tests finish

internal/testutil/acc/atlas.go
marcosuma: ok noted. I am curious to know more: `TestMain` is called once per test package. So if in a subsequent PR you also add `func TestMain(m *testing.M) {...}` to another package, will this still work? 
From how I am getting this (also based on your comment `Go runs this function before any test in the package`), it looks like there can be multiple TestMain calls from all the different test packages (which I think we have more than one)

marcosuma: looking at what this file is managing, I think this should be about project resource initialisation.

`TestMainExecution` --> `TestResourceInit`
`ProjectIDExecution` --> `GetOrCreateProjectID`

I would avoid having this generic `TestMain` logic when in reality it is solely focusing on project resource re-usage (reminds me of [YAGNI](https://martinfowler.com/bliki/Yagni.html)). 

I still have not digested this file in full (as I mentioned somewhere, it looks too complicated), looping in the rest of the team @maastha @AgustinBettati @EspenAlbert @oarbusi

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,112 @@
...
+	mu            sync.Mutex
+	init          bool
+	needsDeletion bool
+}{}
+
+const (
+	projectNameMaxLen = 64
+)
+
+func resourceName() string {
```
marcosuma: I am probably on the side of making things as simple as possible. Can't we just pass the name of the resource when things are initialised?

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,112 @@
...
+	pattern := `([^/]+)_test\.TestMain$`
+	re := regexp.MustCompile(pattern)
+	matches := re.FindStringSubmatch(runtime.FuncForPC(pc).Name())
+	if len(matches) <= 1 {
+		return ""
+	}
+	return matches[1]
+}
+
+func createProject(tb testing.TB, name string) string {
```
marcosuma, line 13: [nit] this looks like a more generic util method to be put outside this file.

internal/testutil/acc/atlas.go
lantoli: yeah, the idea is to add TestMain to all packages.
when you run tests you choose the package, and as TestMain is a public function (uppercase) there can be only one TestMain for that package, and that's the function to be called when running the test for that package.

So it can't happen that two TestMain are run in a test execution,   the execution is only for one package, and that can have at most one TestMain

internal/testutil/acc/atlas.go
lantoli: i think this is less error prone than passing the resource name as a string in each call. i think it's better if it can be got automatically.

(something similar happens in the GH Actions where we have some errors from time to time because copying/pasting incorrectly to pass redundant info)

internal/testutil/acc/atlas.go
lantoli: TestMainExecution is really for initialising and destroying resources, and calling test runs in the middle.

It's our custom implementation for TestMain adding resource lifecycle management

lantoli: some info here about TestMain: https://medium.com/goingogo/why-use-testmain-for-testing-in-go-dafb52b406bc

internal/testutil/acc/name.go
```diff
@@ -40,17 +37,3 @@ func RandomIP(a, b, c byte) string {
...
-}
-
-func projectID(tb testing.TB, name string) string {
-	tb.Helper()
-	SkipInUnitTest(tb)
-	resp, _, _ := ConnV2().ProjectsApi.GetProjectByName(context.Background(), name).Execute()
-	id := resp.GetId()
-	require.NotEmpty(tb, id, "Project name not found: %s", name)
-	return id
-}
```
AgustinBettati: [q] Why are we removing the logic of having a Global project? I would imagine we would have a mix of both have a global project, and also projects that are created per resource execution

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,112 @@
...
+	atlasInfo.mu.Lock()
+	defer atlasInfo.mu.Unlock()
+
+	// lazy creation so it's only done if really needed
+	if atlasInfo.projectName == "" {
+		var globalName, globalID string
+		if atlasInfo.resourceName != "" {
+			globalName = (prefixProjectKeep + "-" + atlasInfo.resourceName)[:projectNameMaxLen]
+			globalID = projectID(globalName)
+		}
```
AgustinBettati: [q] I am struggling to understand in which case resourceName would be empty. If it is empty it is likely an indicator that TestMainExecution was not called, and in this case I would opt for stopping the execution all together.

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if globalID == "" {
+			atlasInfo.projectName = RandomProjectName()
+			tb.Logf("Creating execution project: %s, resource: %s, global project (not found): %s\n", atlasInfo.projectName, atlasInfo.resourceName, globalName)
+			atlasInfo.projectID = createProject(tb, atlasInfo.projectName)
+			atlasInfo.needsDeletion = true
+		} else {
+			atlasInfo.projectName = globalName
+			tb.Logf("Reusing global project: %s, resource: %s\n", atlasInfo.projectName, atlasInfo.resourceName)
+			atlasInfo.projectID = globalID
```
AgustinBettati: Is there a possibility of having a separate function for providing the global project? I believe this would help simplify this function and make its usage more clear. In that case I understand that `needsDeletion` property could be removed as well.

internal/testutil/acc/atlas.go
AgustinBettati: I do agree that we should focus on more concrete naming so that it is more clear what is being done. I was more on the side of a name such as `SetupSharedProject` over `TestMainExecution`.

> TestMainExecution is really for initialising and destroying resources

Regarding this point I was slightly confused because this function does not contain all the initialising logic, for example the creation of the project is handled in `ProjectIDExecution`. Was this for a specific reason?

AgustinBettati: I do agree that we should focus on more concrete naming so that it is more clear what is being done. I was more on the side of a name such as SetupSharedProject over TestMainExecution.

> TestMainExecution is really for initialising and destroying resources

Regarding this point I was slightly confused because this function does not contain all the initialising logic, for example the creation of the project is handled in ProjectIDExecution. Was this for a specific reason?

internal/testutil/acc/atlas.go
marcosuma: ok noted, this method makes me thinking that even if this is code written for testing infra I think it should be unit tested. If we don't test it, I suspect it will become hard to touch for other engineers.

internal/testutil/acc/atlas.go
lantoli: yes, because i realized that for instance right now we always run all packages even if one test won't be run.   so I lazy initialize the project creation in case test is being executed but doesn't need to get the execution project. in that case we don't need to create the project

internal/testutil/acc/atlas.go
lantoli: also this method is not really very important, it could fail and return "" and everything would work (except trying to reuse the global project for that resource).

but you have a point, let me see if it can be tested, it's a bit tricky because it depends on the resource test structure.

internal/testutil/acc/name.go
lantoli: because I realized that projects can't have more than 25 clusters so I don't think that's an option any more, at least for generic projects that don't need any special configuration.
for projects than need some setup (like triggers) we'll probably have a global project for triggers.

but in the resource in this PR it doesn't need any special config so i think it's better not to use a global project for it (it can use a global project for that resource, but not shared by all/many resources)

internal/testutil/acc/atlas.go
lantoli: it's just a precaution, imagine a resource doesn't follow the expected structure.  but i agree it shouldn't happen but in some edge cases

internal/testutil/acc/atlas.go
lantoli: i had in mind that projects just ask for a project and a new one or a resource-global one is given but they don't need  to know.  and there is a nice fallback mechanism so you can create global-resource clusters for the most demanding resources and maybe for resources that just create a couple of projects you don't want to create a global-resource one as the benefit is not so big and you need that project always created.
 
do you think it's better to remove this abstraction and projects decide if they want an execution project or a global-project?

lantoli: in any way there will be a way to ask for specific projects when they need some extra config like triggers, but i don't think it's so useful to allow this for regular projects

internal/service/privatelinkendpointservicedatafederationonlinearchive/resource_privatelink_endpoint_service_data_federation_online_archive_test.go
```diff
@@ -18,9 +18,13 @@ var (
...
+	acc.TestMainExecution(m)
```
gssbzn: don't handle the exist status internally let `TestMain` handle that which is supposed to do

lantoli: it's a bit more verbose but I like the idea and it's more explicit, will change to something like:
```
func TestMain(m *testing.M) {
	acc.SetupSharedResources()
	exitCode := m.Run()
	acc.CleanupSharedResources()
	os.Exit(exitCode)
}
```

thx!

internal/service/privatelinkendpointservicedatafederationonlinearchive/data_source_privatelink_endpoint_service_data_federation_online_archive_test.go
```diff
@@ -15,7 +15,7 @@ var (
 
 func TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchiveDS_basic(t *testing.T) {
 	var (
-		projectID               = acc.ProjectIDGlobal(t)
+		projectID               = acc.ProjectIDExecution(t)
```
lantoli, line 18: all functions like this will start with ProjectID,    we have ProjectIDExecution, ProjectIDGlobal and might have ProjectIDTrigger, etc.

internal/testutil/acc/atlas.go
lantoli: moved

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/atlas-sdk/v20231115007/admin"
+)
+
+func createProject(tb testing.TB, name string) string {
+	tb.Helper()
+	orgID := os.Getenv("MONGODB_ATLAS_ORG_ID")
+	require.NotNil(tb, "Project creation failed: %s, org not set", name)
```
gssbzn, line 16:
```suggestion
	require.NotNilf(tb, "Project creation failed: %s, org not set", name)
```

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,39 @@
...
+	params := &admin.Group{Name: name, OrgId: orgID}
+	resp, _, err := ConnV2().ProjectsApi.CreateProject(context.Background(), params).Execute()
+	require.NoError(tb, err, "Project creation failed: %s, err: %s", name, err)
```
gssbzn, line 19:
```suggestion
	require.NoErrorf(tb, err, "Project creation failed: %s, err: %s", name, err)
```

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,39 @@
...
+	id := resp.GetId()
+	require.NotEmpty(tb, id, "Project creation failed: %s", name)
```
gssbzn, line 21:
```suggestion
	require.NotEmptyf(tb, id, "Project creation failed: %s", name)
```

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,39 @@
...
+		fmt.Printf("Project deletion failed: %s, error: %s", id, err)
+	}
+}
+
+func projectID(tb testing.TB, name string) string {
+	tb.Helper()
+	SkipInUnitTest(tb)
+	resp, _, _ := ConnV2().ProjectsApi.GetProjectByName(context.Background(), name).Execute()
+	id := resp.GetId()
+	require.NotEmpty(tb, id, "Project name not found: %s", name)
```
gssbzn, line 37: can you make sure all assertions where you do string formatting end with `f` for format, other wise is a concatenation

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,39 @@
...
+	require.NotNil(tb, "Project creation failed: %s, org not set", name)
+	params := &admin.Group{Name: name, OrgId: orgID}
+	resp, _, err := ConnV2().ProjectsApi.CreateProject(context.Background(), params).Execute()
+	require.NoError(tb, err, "Project creation failed: %s, err: %s", name, err)
+	id := resp.GetId()
+	require.NotEmpty(tb, id, "Project creation failed: %s", name)
+	return id
+}
+
+func deleteProject(id string) {
```
gssbzn, line 25: out of curiosity why create is a helper but delete is not?

internal/testutil/acc/atlas.go
```diff
@@ -0,0 +1,39 @@
...
+	_, _, err := ConnV2().ProjectsApi.DeleteProject(context.Background(), id).Execute()
+	if err != nil {
+		fmt.Printf("Project deletion failed: %s, error: %s", id, err)
```
gssbzn, line 28: which is delete is a helper you can then do 
```suggestion
		tb.Logf("Project deletion failed: %s, error: %s", id, err)
```

internal/testutil/acc/atlas.go
maastha: How about `ProjectIDForExecution` or `GetProjectIDForCurrentExecution` or something?

internal/testutil/acc/shared_resource.go
```diff
@@ -0,0 +1,48 @@
...
+
+// SetupSharedResources must be called from TestMain test package in order to use ProjectIDExecution.
+func SetupSharedResources() {
+	sharedInfo.init = true
+}
+
+// CleanupSharedResources must be called from TestMain test package in order to use ProjectIDExecution.
+func CleanupSharedResources() {
+	if sharedInfo.projectID != "" {
+		fmt.Printf("Deleting execution project: %s, id: %s\n", sharedInfo.projectName, sharedInfo.projectID)
```
gssbzn, line 19: how are we running tests? normally you want logs when running in verbose mode `-v`

internal/testutil/acc/atlas.go
maastha: Additionally, I think some methods/files could use more comments as to the "why/how", for example shared_resource.go (or the TestMain() but not sure if we need to do for every resource),  as the mechanism could be hard for someone new to understand or use or extend completely or in case we run into any issues later.

internal/testutil/acc/atlas.go
lantoli: I think they do the same, e.g.:

```
a := "hello"
	require.Empty(t, a, "a is not empty: %s, watchout", a)

Error:      	Should be empty, but was hello
        	Test:       	TestAccNetworkPrivatelinkEndpointServiceDataFederationOnlineArchive_basic
        	Messages:   	a is not empty: hello, watchout
```

the signature is: func Empty(t TestingT, object interface{}, **msgAndArgs** ...interface{}) {

example of implementation:
```
//	assert.NotNilf(t, err, "error message %s", "formatted")
func NotNilf(t TestingT, object interface{}, msg string, args ...interface{}) bool {
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	return NotNil(t, object, append([]interface{}{msg}, args...)...)
}
```

internal/testutil/acc/atlas.go
lantoli: the creation is done lazily inside a test so we have the T, but deletion is done in TestMain when all test have finished and we don't have test context anymore

internal/testutil/acc/atlas.go
lantoli: can't be because it's called from TestMain not from a test

internal/testutil/acc/atlas.go
lantoli: i think it's just to make sure that the first param is a string (with the msg)

gssbzn: yeah I see internally does `fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)` I assumed this was closer to `t.Error` vs `t.Errorf` 🤷

internal/testutil/acc/atlas.go
gssbzn: I see, this may make this approach a bit noisy on the logs since you usually only want logs when running on verbose mode (-v) and right now with printf is harder to control

gssbzn (approved):

internal/testutil/acc/shared_resource.go
```diff
@@ -0,0 +1,48 @@
...
+func ProjectIDExecution(tb testing.TB) string {
+	tb.Helper()
+	SkipInUnitTest(tb)
+	require.True(tb, sharedInfo.init, "SetupSharedResources must called from TestMain test package")
+
+	sharedInfo.mu.Lock()
+	defer sharedInfo.mu.Unlock()
+
+	// lazy creation so it's only done if really needed
+	if sharedInfo.projectID == "" {
```
EspenAlbert, line 35: Maybe support `os.Getenv("MONGODB_ATLAS_PROJECT_ID")` similar to `acc.GetClusterInfo`?

lantoli: I think that is a good idea, but i'm going to do in a next PR so this doesn't get more complicated, thx!

internal/testutil/acc/atlas.go
lantoli: we're running with -v in GH Actions, but you have a point, will consider in next PRs if delete it, thanks!

internal/testutil/acc/atlas.go
lantoli: I'm not sure about document in our code how Go features like TestMain work, but I'm fine to document more our code. I'm going to merge but please feel free to suggest doc, and i'll think in more docs in the following PRs, thx!

internal/testutil/acc/shared_resource.go
lantoli: we use verbose flag:

TF_ACC=1 go test $(ACCTEST_PACKAGES) -run '$(ACCTEST_REGEX_RUN)' -skip '$(ACCTEST_REGEX_SKIP)' -v -parallel $(PARALLEL_GO_TEST) $(TESTARGS) -timeout $(ACCTEST_TIMEOUT) -ldflags="$(LINKER_FLAGS)"

AgustinBettati (approved):
LGTM, thanks for followups simplifying this initial approach

PR #2054, opened 2024-03-20
fix: Uses `overwriteBackupPolicies` in `mongodbatlas_backup_compliance_policy` to avoid overwriting non complying backup policies in updates

maastha:
## Description

This PR adds support for overwriteBackupPolicies in `mongodbatlas_backup_compliance_policy` which is set to `false` to avoid creation/update of conflicting policies.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-229775

## Type of change:

## Required Checklist:

## Further comments
User receives error as below when trying to create/update a conflicting policy:
```
╷
│ Error: error updating a Backup Compliance Policy: .....: https://cloud-dev.mongodb.com/api/atlas/v2/groups/....../backupCompliancePolicy PUT: HTTP 400 Bad Request (Error code: "BACKUP_POLICIES_NOT_MEETING_BACKUP_COMPLIANCE_POLICY_REQUIREMENTS") Detail: The following backup policies do not comply with the specified backup compliance policy: [Differences for Cluster: test1, [The backup policy does not meet the Backup Compliance Policy requirements for DAILY policy item]]. Reason: Bad Request. Params: [[Differences for Cluster: test1, [The backup policy does not meet the Backup Compliance Policy requirements for DAILY policy item]]]
│ 
│   with mongodbatlas_backup_compliance_policy.backup_compliance_policy,
│   on cluster_backup_compliance_conflict.tf line 6, in resource "mongodbatlas_backup_compliance_policy" "backup_compliance_policy":
│    6: resource "mongodbatlas_backup_compliance_policy" "backup_compliance_policy" {
```


internal/service/backupcompliancepolicy/resource_backup_compliance_policy_test.go
```diff
@@ -41,7 +42,7 @@ func TestAccGenericBackupRSBackupCompliancePolicy_basic(t *testing.T) {
 	})
 }
 
-func TestAccGenericBackupRSBackupCompliancePolicy_withFirstLastName(t *testing.T) {
+func TestAccGenericBackupRSBackupCompliancePolicy_update(t *testing.T) {
```
maastha, line 46: First and Last name are already tested in `TestAccGenericBackupRSBackupCompliancePolicy_basic ` so updating this test to instead test update functionality

marcosuma:
@maastha WDYT about tagging this as breaking change? we should probably bump this to 1.16.0 and explain in the migration guide the change in behaviour. 
On the other side: this can be seen as a bug fixing because current behaviour can lead to inconsistent state

internal/service/backupcompliancepolicy/resource_backup_compliance_policy_test.go
marcosuma, line 46: would it be possible to have a test that captures the behaviour:
1. have a TF with consistent backup and backup policy data
2. try to change backup policy
3. catch the validation error

internal/service/backupcompliancepolicy/resource_backup_compliance_policy.go
```diff
@@ -8,13 +8,15 @@ import (
 	"net/http"
```
AgustinBettati, line 8: Regarding the possibility of exposing `overwrite_backup_policies` attribute we can wait for confirmation from the API team but it would seem reasonable that only the UPDATE request receives this parameter.
With that said we can expose the attribute with the default value of false, and this will ensure that if a customer was relying on the existing behaviour they can quickly adjust to that behaviour with the explicit `overwrite_backup_policies = true`.

maastha: Added [comment to ticket](https://jira.mongodb.org/browse/CLOUDP-229775?focusedCommentId=6198851&focusedId=6198851&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6198851) about Backup team's response, this attribute is not persisted and hence cannot be returned in API response.
Hence the idea is to not expose this to user to be able to update

Not sure if this addresses your concern?

AgustinBettati: My question is more related to why we would not add this attribute, because from technical stand point it does seem possible. We can define it as an optional attribute with a default value, include the value for patch requests and in the case of a read simply preserve the value defined in the tf config.
Personally I see value in exposing the attribute in case customers have preference for the existing behaviour of having their backup definitions automatically change to the new policy.

maastha:
>WDYT about tagging this as breaking change? we should probably bump this to 1.16.0 and explain in the migration guide the change in behaviour.
On the other side: this can be seen as a bug fixing because current behaviour can lead to inconsistent state

@marcosuma This seems more like a bug to me (currently targeted for 1.16.0) that is now being fixed from the backend for the users, essentially if existing users make any "conflicting" updates to their policy (which is avoiding an incorrect behavior) they will get an error from the API. 
I'll update the PR title but lmk

internal/service/backupcompliancepolicy/resource_backup_compliance_policy.go
maastha: @AgustinBettati  To make sure I understand correctly, you mean even though this attribute isn't returned from the API, we should persist in the TF state and re-use that every time for PATCH, is that correct? 
If yes, personally unless there's a strong reason, I think we should avoid having such attributes as these can be problematic during imports (for example if user does a `terraform state rm` and then wants to re-import for some reason or to get out of an inconsistent state)

>Personally I see value in exposing the attribute in case customers have preference for the existing behaviour of having their backup definitions automatically change to the new policy.

I understand this but I'm also thinking that:
1. this can potentially leave them in an inconsistent state (even if they intended the behavior)
2. and in addition of the inconsistent state, they now also have an additional attribute to deal with in case they want to re-import the resource for any reason.

I think we should only add this as a new attribute to be used by customers if we get a customer request for such a use-case. WDYT?

Zuhairahmed:
> > WDYT about tagging this as breaking change? we should probably bump this to 1.16.0 and explain in the migration guide the change in behaviour.
> > On the other side: this can be seen as a bug fixing because current behaviour can lead to inconsistent state
> 
> @marcosuma This seems more like a bug to me (currently targeted for 1.16.0) that is now being fixed from the backend for the users, essentially if existing users make any "conflicting" updates to their policy (which is avoiding an incorrect behavior) they will get an error from the API. I'll update the PR title but lmk

Agree with @maastha, this feels more like a bug fix to users back to correct / intended behavior

internal/service/backupcompliancepolicy/resource_backup_compliance_policy.go
Zuhairahmed: `overwriteBackupPolicies` parameter was purely created to help downstream tools like AKO and TF to behave as intended. upstream Atlas API team did not intend for API users to manipulate directly as is evidenced that it is not return back in GET/UPDATE responses. in this case suggest we keep design simple and do not allow users to directly modify parameter. 

if this feedback comes in from a customer we can always re-evaluate.

marcosuma: Looping in @colm-quinn  for awareness since he was interested in this

internal/service/backupcompliancepolicy/resource_backup_compliance_policy_test.go
maastha: added test to catch error

andreaangiolillo (approved):

PR #2031, opened 2024-03-14
fix: Ensure required field RegionName is set when updating network conatiner

BrettFolkins:
## Description

The `regionName` field is required when updating a network container: https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Network-Peering/operation/updatePeeringContainer

However the current code only sets it if there is a change to the region name. This has the effect of making any change that only affects the cidr fail:
```
2024-03-13T21:07:31.906Z [INFO]  plugin.terraform-provider-mongodbatlas_v1.11.1: 2024/03/13 21:07:31 [DEBUG] MongoDB Atlas API Request Details:
---[ REQUEST ]---------------------------------------
PATCH /api/atlas/v1.0/groups/5e84b906246b5a0e3926c1f4/containers/65b8318cda738f151bd9d8b1 HTTP/1.1Host: cloud.mongodb.comUser-Agent: terraform-provider-mongodbatlas/1.11.1 go-mongodbatlas/0.33.0 (linux;amd64)Content-Length: 57Accept: application/jsonContent-Type: application/jsonAccept-Encoding: gzip{
 "atlasCidrBlock": "10.248.86.0/23",
 "providerName": "AWS"
}-----------------------------------------------------: timestamp=2024-03-13T21:07:31.906Z
---[ RESPONSE ]--------------------------------------
HTTP/2.0 500 Internal Server ErrorContent-Length: 122Content-Type: application/jsonDate: Wed, 13 Mar 2024 21:07:31 GMTReferrer-Policy: strict-origin-when-cross-originServer: mdbwsStrict-Transport-Security: max-age=31536000; includeSubdomains;X-Content-Type-Options: nosniffX-Envoy-Upstream-Service-Time: 46X-Frame-Options: DENYX-Mongodb-Service-Version: gitHash=2a321c7a1e0c84ef86357fa877294b946492ec03; versionString=v20240306X-Permitted-Cross-Domain-Policies: none{
 "detail": "Unexpected error.",
 "error": 500,
 "errorCode": "UNEXPECTED_ERROR",
 "parameters": [],
 "reason": "Internal Server Error"
}
-----------------------------------------------------: timestamp=2024-03-13T21:07:31.983Z 
```

This PR adds a test case that attempts to change the cidr without changing the region, and updates the code to always set the `regionName` field.

Link to any related issue(s): https://jira.mongodb.org/browse/HELP-56981

## Type of change:

[...]

PR #2055, opened 2024-03-21
fix: Fixes `network_container` resource update

lantoli:
## Description

Fixes `network_container` resource update when only one field changes.

Based on this PR:  https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2031

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/networkcontainer/resource_network_container.go
+++ b/internal/service/networkcontainer/resource_network_container.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"log"
 	"net/http"
-	"reflect"
 	"strings"
 	"time"
 
@@ -212,40 +211,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 }
 
 func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if !d.HasChange("atlas_cidr_block") && !d.HasChange("region_name") && !d.HasChange("region") && !d.HasChange("regions") {
+		return resourceRead(ctx, d, meta)
+	}
+
 	connV2 := meta.(*config.MongoDBClient).AtlasV2
 	ids := conversion.DecodeStateID(d.Id())
 	projectID := ids["project_id"]
 	containerID := ids["container_id"]
+	regionName, _ := conversion.ValRegion(d.Get("region_name"))
+	region, _ := conversion.ValRegion(d.Get("region"))
 
-	container := new(admin.CloudProviderContainer)
-
-	if d.HasChange("atlas_cidr_block") {
-		atlasCidrBlock := d.Get("atlas_cidr_block").(string)
-		providerName := d.Get("provider_name").(string)
-		container.AtlasCidrBlock = &atlasCidrBlock
-		container.ProviderName = &providerName
+	params := &admin.CloudProviderContainer{
+		ProviderName:   conversion.StringPtr(d.Get("provider_name").(string)),
+		AtlasCidrBlock: conversion.StringPtr(d.Get("atlas_cidr_block").(string)),
+		RegionName:     conversion.StringPtr(regionName),
+		Region:         conversion.StringPtr(region),
 	}
 
-	if d.HasChange("provider_name") {
-		providerName := d.Get("provider_name").(string)
-		container.ProviderName = &providerName
-	}
-
-	if d.HasChange("region_name") {
-		regionName, _ := conversion.ValRegion(d.Get("region_name"))
-		container.RegionName = &regionName
-	}
-
-	if d.HasChange("region") {
-		region, _ := conversion.ValRegion(d.Get("region"))
-		container.Region = &region
-	}
-
-	if !reflect.DeepEqual(container, admin.CloudProviderContainer{}) {
-		_, _, err := connV2.NetworkPeeringApi.UpdatePeeringContainer(ctx, projectID, containerID, container).Execute()
-		if err != nil {
-			return diag.FromErr(fmt.Errorf(errorContainerUpdate, containerID, err))
+	if regionList, ok := d.GetOk("regions"); ok {
+		if regions := cast.ToStringSlice(regionList); regions != nil {
+			params.SetRegions(regions)
 		}
+	}
+
+	_, _, err := connV2.NetworkPeeringApi.UpdatePeeringContainer(ctx, projectID, containerID, params).Execute()
+	if err != nil {
+		return diag.FromErr(fmt.Errorf(errorContainerUpdate, containerID, err))
 	}
 
 	return resourceRead(ctx, d, meta)
--- a/internal/service/networkcontainer/resource_network_container.go
+++ b/internal/service/networkcontainer/resource_network_container.go
@@ -210,7 +210,7 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 }
 
 func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	if !d.HasChange("region_name") && !d.HasChange("atlas_cidr_block") && !d.HasChange("region_name") && !d.HasChange("region") && !d.HasChange("regions") {
+	if !d.HasChange("provider_name") && !d.HasChange("atlas_cidr_block") && !d.HasChange("region_name") && !d.HasChange("region") && !d.HasChange("regions") {
 		return resourceRead(ctx, d, meta)
 	}
 
--- a/internal/service/networkcontainer/resource_network_container.go
+++ b/internal/service/networkcontainer/resource_network_container.go
@@ -51,6 +51,7 @@ func Resource() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      constant.AWS,
+				ForceNew:     true,
 				ValidateFunc: validation.StringInSlice([]string{constant.AWS, constant.GCP, constant.AZURE}, false),
 			},
 			"region_name": {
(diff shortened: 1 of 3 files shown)


lantoli:
thanks @BrettFolkins for your contribution, this PR is based on yours

internal/service/networkcontainer/resource_network_container.go
```diff
@@ -51,6 +50,7 @@ func Resource() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      constant.AWS,
+				ForceNew:     true,
```
lantoli: provider change forces a replace

internal/service/networkcontainer/resource_network_container.go
```diff
@@ -211,40 +211,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
 
-	if d.HasChange("region_name") {
-		regionName, _ := conversion.ValRegion(d.Get("region_name"))
-		container.RegionName = &regionName
-	}
-
-	if d.HasChange("region") {
-		region, _ := conversion.ValRegion(d.Get("region"))
-		container.Region = &region
+	if regionList, ok := d.GetOk("regions"); ok {
```
lantoli: regions was not updated before

internal/service/networkcontainer/resource_network_container_test.go
```diff
@@ -24,7 +24,7 @@ func TestAccNetworkContainer_basicAWS(t *testing.T) {
 		projectID        = acc.ProjectIDExecution(t)
 		randInt          = acctest.RandIntRange(0, 255)
 		cidrBlock        = fmt.Sprintf("10.8.%d.0/24", randInt)
-		randIntUpdated   = acctest.RandIntRange(0, 255)
+		randIntUpdated   = (randInt + 1) % 256
```
lantoli, line 27: to make sure cidr and updated cidr are always different

internal/service/networkcontainer/resource_network_container.go
```diff
@@ -211,40 +211,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 }
 
 func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if !d.HasChange("atlas_cidr_block") && !d.HasChange("region_name") && !d.HasChange("region") && !d.HasChange("regions") {
```
marcosuma: missing `provider_name`?

internal/service/networkcontainer/resource_network_container.go
```diff
@@ -211,40 +211,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
-		container.ProviderName = &providerName
-	}
-
-	if d.HasChange("provider_name") {
-		providerName := d.Get("provider_name").(string)
-		container.ProviderName = &providerName
+	regionName, _ := conversion.ValRegion(d.Get("region_name"))
+	region, _ := conversion.ValRegion(d.Get("region"))
+
+	params := &admin.CloudProviderContainer{
```
marcosuma, line 225: One behaviour change compared to before is that we're going to pass all the values no matter if they have changed or not (As long as at least one has changed) - why are we changing this behaviour?

internal/service/networkcontainer/resource_network_container.go
lantoli: no because I've set it to ForceNew, so if it changes it'll delete and create instead of update

internal/service/networkcontainer/resource_network_container.go
lantoli: it's same behavior, we're not really passing all values, only the ones that are set, it will pass nil if the string is empty so it won't go as now (because StringPtr).
and for the values that are set, the change is precisely to pass all of them because the API will fail if we don't

internal/service/networkcontainer/resource_network_container.go
AgustinBettati: Just to understand the why behind this change, the API does not support updating the providerName?

internal/service/networkcontainer/resource_network_container.go
AgustinBettati: From https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2031 it would seem that regionName is the only property that must be included in the update. For the rest we can include them only if they have mainly for consistency with how we handle other attributes in this same function.

internal/service/networkcontainer/resource_network_container.go
lantoli: i was under that impression but i checked again and it looks like it's allowed (although it's a bit tricky because they need different attributes).
updated

internal/service/networkcontainer/resource_network_container.go
lantoli: if you change the provider name you might want to change other attributes as different providers require different attributes.  i've change it to send the require argument depending on the provider.  It's the same idea as we're doing in Create

internal/service/networkcontainer/resource_network_container.go
```diff
@@ -211,42 +210,49 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
-		_, _, err := connV2.NetworkPeeringApi.UpdatePeeringContainer(ctx, projectID, containerID, container).Execute()
+	switch providerName {
+	case constant.AWS:
+		regionName, err := conversion.ValRegion(d.Get("region_name"))
 		if err != nil {
 			return diag.FromErr(fmt.Errorf(errorContainerUpdate, containerID, err))
 		}
+		params.SetRegionName(regionName)
+	case constant.AZURE:
+		region, err := conversion.ValRegion(d.Get("region"))
```
lantoli, line 238: checking errors, it was not done before

internal/service/networkcontainer/resource_network_container_test.go
```diff
@@ -218,6 +218,53 @@ func TestAccNetworkContainer_importBasic(t *testing.T) {
...
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: configBasic(projectID, cidrBlock, constant.AWS, region),
+				Check: resource.ComposeTestCheckFunc(
+					checkExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
+					resource.TestCheckResourceAttr(resourceName, "provider_name", constant.AWS),
+					resource.TestCheckResourceAttrSet(resourceName, "provisioned"),
```
EspenAlbert: These lines seems to be repeated many times. Maybe refactor into a helper?

EspenAlbert (approved):

internal/service/networkcontainer/resource_network_container_test.go
lantoli: i think there was a similar comment in another PR, I don't disagree but let's talk to use the same approach in all resources

internal/service/networkcontainer/resource_network_container.go
```diff
@@ -211,42 +210,49 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 }
 
 func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if !d.HasChange("region_name") && !d.HasChange("atlas_cidr_block") && !d.HasChange("region_name") && !d.HasChange("region") && !d.HasChange("regions") {
```
marcosuma: I think `region_name` is repeated twice? and missing `provider_name`? Sorry if I am missing some further context.

internal/service/networkcontainer/resource_network_container.go
marcosuma: looping in @BrettFolkins for this just for awareness

marcosuma:
Let's make sure @BrettFolkins has the opportunity to review this PR (already sent a Slack DM)

internal/service/networkcontainer/resource_network_container.go
BrettFolkins: I'm not familiar enough with best practices in terraform providers to have an opinion about sending the request with only the necessary changes or the full target state, but this version lgtm. Thanks!

BrettFolkins (approved):
LGTM, thanks for picking up where I left off and getting a full solution to this ready so quickly!

internal/service/networkcontainer/resource_network_container_test.go
```diff
@@ -35,31 +35,11 @@ func TestAccNetworkContainer_basicAWS(t *testing.T) {
...
-					resource.TestCheckResourceAttr(dataSourceName, "provider_name", constant.AWS),
-					resource.TestCheckResourceAttrSet(dataSourceName, "provisioned"),
-
-					resource.TestCheckResourceAttrWith(dataSourcePluralName, "results.#", acc.IntGreatThan(0)),
-					resource.TestCheckResourceAttrSet(dataSourcePluralName, "results.0.id"),
-					resource.TestCheckResourceAttrSet(dataSourcePluralName, "results.0.atlas_cidr_block"),
-					resource.TestCheckResourceAttrSet(dataSourcePluralName, "results.0.provider_name"),
-					resource.TestCheckResourceAttrSet(dataSourcePluralName, "results.0.provisioned"),
-				),
+				Check:  resource.ComposeTestCheckFunc(commonChecks(constant.AWS)...),
```
lantoli, line 38: @EspenAlbert at then end i did the refactor you recommended following @AgustinBettati example in search_deployment.   a lot of duplication gone and clearer to read.  (still this resource might have more specific attribute checks, not only that they exist).    cc @marcosuma

lantoli:
@EspenAlbert +91 -111 ;-)

internal/service/networkcontainer/resource_network_container.go
lantoli: added after reverting force new in region_name

lantoli: good catch, thanks! i was doing multiple tests and at the end i didn't revert this, updated

PR #2085, opened 2024-03-26
chore: Improves user agent in the provider to include Terraform CLI version

maastha:
## Description

Improves user agent in the provider to include Terraform CLI version

Link to any related issue(s): [CLOUDP-237005](https://jira.mongodb.org/browse/CLOUDP-237005)

## Type of change:

## Required Checklist:

## Further comments

TF Plugin Framework based resource API call logs:
```
2024/04/05 15:17:50 
GET /api/atlas/v2/groups/6602bedbc32461337d92e2a5/settings HTTP/1.1
Host: cloud-dev.mongodb.com
User-Agent: terraform-provider-mongodbatlas/devel Terraform/1.7.0-rc1
Accept: application/vnd.atlas.2023-01-01+json
Accept-Encoding: gzip
```

TF Plugin SDK based resource API call logs:
```
2024/04/05 15:17:52 
POST /api/atlas/v2/groups/6602d0208210bb7b10f5f8a8/clusters HTTP/1.1
Host: cloud-dev.mongodb.com
User-Agent: terraform-provider-mongodbatlas/devel Terraform/1.7.0-rc1
Content-Length: 302
Accept: application/vnd.atlas.2023-02-01+json
```


internal/config/client.go
```diff
@@ -175,3 +183,18 @@ func (c *MongoDBClient) GetRealmClient(ctx context.Context) (*realm.Client, erro
...
+		if currentUA != "" {
+			currentUA += " "
+		}
+		currentUA += additionalInfo
+	}
+	c.UserAgent = currentUA
+}
+
+func TerraformVersionUserAgentInfo(tfVersion string) string {
+	return fmt.Sprintf("Terraform/%s", tfVersion)
```
oarbusi: small nit: create and use const like `toolName`
```
return fmt.Sprintf("%s/%s", terraformName, tfVersion)
```

oarbusi (approved):
Great work! LGTM

internal/provider/provider_sdk2.go
```diff
@@ -133,11 +133,15 @@ func NewSdkV2Provider(proxyPort *int) *schema.Provider {
...
-		ResourcesMap:         getResourcesMap(),
-		ConfigureContextFunc: providerConfigure(proxyPort),
+		DataSourcesMap: getDataSourcesMap(),
+		ResourcesMap:   getResourcesMap(),
 	}
 	addPreviewFeatures(provider)
+
+	provider.ConfigureContextFunc = func(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
+		return providerConfigure(provider, proxyPort)(ctx, d)
+	}
```
AgustinBettati:
```suggestion
	provider.ConfigureContextFunc = providerConfigure(provider, proxyPort)
```

internal/provider/provider_sdk2.go
```diff
@@ -296,6 +300,7 @@ func providerConfigure(proxyPort *int) func(ctx context.Context, d *schema.Resou
 			BaseURL:      d.Get("base_url").(string),
 			RealmBaseURL: d.Get("realm_base_url").(string),
 			ProxyPort:    proxyPort,
+			UserAgent:    config.TerraformVersionUserAgentInfo(provider.TerraformVersion),
```
AgustinBettati: a little hard to follow, but I assume that when ConfigureContextFunc is called it mutates the `TerraformVersion` property of the schema.Provider?

internal/provider/provider.go
```diff
@@ -231,6 +231,7 @@ func (p *MongodbtlasProvider) Configure(ctx context.Context, req provider.Config
 		PrivateKey:   data.PrivateKey.ValueString(),
 		BaseURL:      data.BaseURL.ValueString(),
 		RealmBaseURL: data.RealmBaseURL.ValueString(),
+		UserAgent:    config.TerraformVersionUserAgentInfo(req.TerraformVersion),
```
AgustinBettati: [q] Could both the provider version and the terraform version be defined in a single place so it is easy to see the complete user agent format?

internal/config/client.go
```diff
@@ -175,3 +183,18 @@ func (c *MongoDBClient) GetRealmClient(ctx context.Context) (*realm.Client, erro
...
+	if !strings.Contains(currentUA, additionalInfo) {
+		if currentUA != "" {
+			currentUA += " "
+		}
+		currentUA += additionalInfo
+	}
+	c.UserAgent = currentUA
+}
+
+func TerraformVersionUserAgentInfo(tfVersion string) string {
```
AgustinBettati: Curious if from the analytics team there is any preference in how this terraform information is added so we don't impact any existing queries that make use of the user agent. Maybe worth sending a message to Victor to verify.

AgustinBettati (approved):
LGTM, just some follow up doubts

internal/provider/provider.go
maastha: it can be but I have kept it separate. Ideally user agent should be taken care of when creating new client (like we already do today) but in this case I had to set the new Terraform version user agent in the Provider because that's only where it's available.
Defining both in single place would mean we take provider version out of the client creation which I don't think we should do

internal/provider/provider_sdk2.go
maastha: right

internal/config/client.go
maastha: Looks like Victor is out until after next week, I think let's merge this for now and in the meantime I'll send a ping to him anyway so we can make any updates later if required

internal/provider/provider.go
AgustinBettati: I see, one option could be to define a property TerraformVersion in the Config struct (instead of UserAgent), and then when creating a new client define the user agent with both the terraform and provider versions available. Let me know if you feel this makes it clearer or not.

internal/config/client.go
maastha: Reached out to Victor/natalya over Slack, included you both in the conversation so he can respond whenever he is back

internal/config/client.go
```diff
@@ -6,25 +6,28 @@ import (
...
+	"github.com/mongodb/terraform-provider-mongodbatlas/version"
 )
 
 const (
 	toolName = "terraform-provider-mongodbatlas"
 )
 
 var (
-	userAgent = fmt.Sprintf("%s/%s", toolName, version.ProviderVersion)
+	userAgentProviderVersion = fmt.Sprintf("%s/%s", toolName, version.ProviderVersion)
```
lantoli: would it be useful to add some tests in the PR?

EspenAlbert (approved):

internal/config/client.go
```diff
@@ -175,3 +183,18 @@ func (c *MongoDBClient) GetRealmClient(ctx context.Context) (*realm.Client, erro
 
 	return realmClient, nil
 }
+
+func (c *Config) appendToUserAgent(additionalInfo string) {
+	currentUA := c.UserAgent
+	if !strings.Contains(currentUA, additionalInfo) {
```
marcosuma: why do we need to check if it's already there? is the code being called more than once?

internal/provider/provider.go
marcosuma: I was thinking the same thing after spending some time reviewing this. Having `TerraformVersion` would help a lot in terms of readability. In fact during client creation you can "compose" the user agent in-place without even having to have a special append function IMO. 

Also this solution would help to handle logs / problems: e.g. if `TerraformVersion` is empty during the client creation phase then we should handle it with a warning probably since it's something that needs attention

internal/config/client.go
maastha: >is the code being called more than once?

no, but was trying to keep the method generic

internal/provider/provider.go
maastha: I have added TerraformVersion in the Config struct, let me know what you think

AgustinBettati (approved):
LGTM, just to confirm the conclusion with analytics team was to have the provider version first and then the terraform version?

maastha:
>LGTM, just to confirm the conclusion with analytics team was to have the provider version first and then the terraform version?

This was Gus's suggestion, yes. I have updated the testing example in the PR accordingly.

marcosuma (approved):
Awesome job!

PR #2094, opened 2024-03-27
test: Adds migration test to networkpeering for AWS

EspenAlbert:
## Description

Refactors names & checks, reuses ProjectID, adds a migration test, and enables `TestAccNetworkRSNetworkPeering_AWSDifferentRegionName` in the CI

Link to any related issue(s): CLOUDP-237027

## Type of change:

## Required Checklist:


internal/service/networkpeering/resource_network_peering_test.go
```diff
@@ -148,12 +104,11 @@ func TestAccNetworkRSNetworkPeering_basicGCP(t *testing.T) {
 }
 
 func TestAccNetworkRSNetworkPeering_AWSDifferentRegionName(t *testing.T) {
-	acc.SkipTestForCI(t)
```
EspenAlbert, line 151: Enabling this test since we can use `ProjectIDExecution`

internal/service/networkpeering/resource_network_peering_test.go
```diff
@@ -12,60 +12,16 @@ import (
...
-			{
-				ResourceName:            resourceName,
-				ImportStateIdFunc:       testAccCheckMongoDBAtlasNetworkPeeringImportStateIDFunc(resourceName),
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"accepter_region_name", "container_id"},
-			},
-		},
-	})
+func TestMigNetworkNetworkPeering_basicAWS(t *testing.T) {
```
EspenAlbert: Choosing to put the migration test next to the acceptance test as I think the new `mig.CreateAndRunTest` makes it overkill to have a separate file.
But could go back to convention of a separate file?

internal/service/networkpeering/resource_network_peering_test.go
```diff
@@ -244,41 +246,41 @@ func testAccCheckMongoDBAtlasNetworkPeeringExists(resourceName string, peer *mat
...
-			provider_name		  = "%[3]s"
-			region_name			  = "%[7]s"
-		}
+	resource "mongodbatlas_project" "my_project" {
+		name   = %[2]q
+		org_id = %[1]q
+	}
+	resource "mongodbatlas_network_container" "test" {
+		project_id   		  = mongodbatlas_project.my_project.id
+		atlas_cidr_block  = "192.168.208.0/21"
```
EspenAlbert: This value is hardcoded and doesn't match the value I find in the UI by going to:
Project -> NetworkAccess -> Peering -> AddPeeringConnection

image

There I find:
dev: 192.168.240.0/21
prod: 192.168.248.0/21

Any idea why?

internal/testutil/mig/test_case.go
```diff
@@ -4,16 +4,19 @@ import (
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/acc"
 	"github.com/stretchr/testify/require"
 )
 
 func CreateAndRunTest(t *testing.T, test *resource.TestCase) {
 	t.Helper()
+	acc.SkipInUnitTest(t)
```
oarbusi: What's the reason for this? Do we want this for all mig tests?
From the method definition I see:
`SkipInUnitTest allows skipping a test entirely when TF_ACC=1 is not defined. This can be useful for acceptance tests that define logic prior to resource.Test/resource.ParallelTest functions as this code would always be run.`

internal/service/networkpeering/resource_network_peering_test.go
oarbusi: No strong opinion here but probably we should think about it because we recently decided to put resource and data source tests in the same test file. Having also migration tests in the same file could be inconvenient. It could be a good topic to bring to the technical discussion of tomorrow.

internal/service/networkpeering/resource_network_peering_test.go
oarbusi: I think in the UI it shows a default/available CIDR block, but you can use a custom one if it adheres to the constraints.

internal/service/networkpeering/resource_network_peering_test.go
AgustinBettati: I would favor sticking to existing convention for now for consistency. Feel free to bring up the topic with pros/cons and we can take decision from there.

internal/testutil/mig/test_case.go
EspenAlbert: I think we want to skip all mig tests when running unit tests since they will create resources in the cloud. 
All migration tests need the `MONGODB_ATLAS_LAST_VERSION` (see [Here for a failed unit test](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/8450359863/job/23146358963)).
Therefore, I think this is in line with our current practices since this env-var is NOT set in our unit test pipeline.

internal/service/networkpeering/resource_network_peering_test.go
AgustinBettati: one question, would `ProjectIDExecution` be a safe usage here? If a new tests uses `ProjectIDExecution` they will share the same project, and I am not sure how this resource would behave in that case. I am seeing that for other tests in this resource we create a unique project for each one. (@lantoli might have additional inputs as he was looking into reusing projects in all our resources)

oarbusi (approved):
LGTM overall! Feel free to ignore but IMO the PR title seems like a full description. Could we maybe rephrase so it's shorter and explain the exact details in the description?

internal/service/networkpeering/resource_network_peering_test.go
EspenAlbert: True, I should have warned about that.
Basically, this means that if another test in this package uses the `ProjectIDExecution`, one of them will fail.
This is because the tests creates a `mongodbatlas_network_container` which can only be created **once** per project

AgustinBettati: In this case I would favor the pattern used in other tests of defining the project in the terraform config. This way we ensure there is no risk of new test breaking existing ones. Usage of `ProjectIDExecution` will be limited to resources where a project can be safely shared for multiple tests executions. I can follow up with a PR to clarify this in our contributing guide.

EspenAlbert: Updated in [bb0f628](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2094/commits/bb0f628f1a6d69b8b9e52b46eb234b4304edf68e) ✅

internal/service/networkpeering/resource_network_peering_test.go
EspenAlbert: Moved in [1015e62](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2094/commits/1015e6206d7addfb2561ed7e4c9bcf6b14fddfb4)

internal/service/networkpeering/resource_network_peering_test.go
```diff
@@ -176,42 +137,65 @@ func TestAccNetworkRSNetworkPeering_AWSDifferentRegionName(t *testing.T) {
...
+func commonChecksAWS(vpcID, providerName, awsAccountID, vpcCIDRBlock, regionPeer string) []resource.TestCheckFunc {
+	attributes := map[string]string{
+		"vpc_id":                 vpcID,
+		"provider_name":          providerName,
+		"aws_account_id":         awsAccountID,
+		"route_table_cidr_block": vpcCIDRBlock,
+	}
+	checks := []resource.TestCheckFunc{checkExists(resourceName, &matlas.Peer{})}
+	checks = acc.AddAttrChecks(resourceName, checks, attributes)
+	checks = append(checks, resource.TestCheckResourceAttr(resourceName, "accepter_region_name", regionPeer))
```
EspenAlbert, line 190: Only checking this on the resource now.
By testing this on the datasource I get:

- when passing: `us-east-1`
    - `data.mongodbatlas_network_peering.test: Attribute 'accepter_region_name' not found`
- when passing: `US_EAST_1`
  - `Check failed: Check 2/20 error: mongodbatlas_network_peering.test: Attribute 'accepter_region_name' expected "US_EAST_1", got "us-east-1"`
  
  I can see a comment `//Workaround until fix` in https://github.com/mongodb/terraform-provider-mongodbatlas/blob/1015e6206d7addfb2561ed7e4c9bcf6b14fddfb4/internal/service/networkpeering/data_source_network_peering.go#L127
  
  @AgustinBettati what do you think? Is this a known issue? Can you reproduce it?

internal/service/networkpeering/resource_network_peering_test.go
```diff
@@ -148,23 +104,28 @@ func TestAccNetworkRSNetworkPeering_basicGCP(t *testing.T) {
...
+	checks = acc.AddAttrSetChecks(resourceNameContainer, checks, "project_id", "atlas_cidr_block", "provider_name", "region_name")
+	mapChecksContainer := map[string]string{
+		"atlas_cidr_block": "192.168.208.0/21",
+		"provider_name":    providerName,
+		"region_name":      containerRegion,
+	}
+	checks = acc.AddAttrChecks(resourceNameContainer, checks, mapChecksContainer)
 
-	resource.Test(t, resource.TestCase{
+	resource.ParallelTest(t, resource.TestCase{
```
marcosuma, line 126: why are we moving this back to Parallel? I lost track with what should be Test and what should be ParallelTest

internal/testutil/mig/test_case.go
marcosuma: is it worth it adding a comment in the file itself? it might be confusing and lead the developer going through files to understand why

internal/service/networkpeering/resource_network_peering_test.go
EspenAlbert: `Test`: Can only run one test at a time
`ParallelTest`: Can run multiple tests in parallel

This test is refactored to create and use its own project instead of reading a project from `MONGODB_ATLAS_PROJECT_ID`, making it safe to run it in parallel.

internal/testutil/mig/test_case.go
EspenAlbert: Added comment in [8d5f7d0](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2094/commits/8d5f7d0db20e9ff2f2824df020691417a5f351e5)

lantoli: we want to use SkipInUnitTest when it's not detected automatically for acc or mig tests.
For example resource.Test/ParallelTest handle that, but it's needed before calling them, e.g. when declaring vars in the test, or here in this mig function that is called before  resource.Test/ParallelTest .

it's also done for instance in ProjectIDExecution or ClusterNameExecution as they're used in var declaration.  

this is normally needed in "infrastructure code" that it's used though multiple tests, it's not normally needed to be called inside a test itself

lantoli (approved):

internal/service/networkpeering/resource_network_peering_test.go
EspenAlbert: This has now been fixed in [ca77513](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2094/commits/ca7751301b880f14e8cc23c1bb421f03851b06a4)
Thank you @lantoli for the help 👏

internal/service/networkpeering/resource_network_peering_test.go
marcosuma: ok thanks, my concern is more around having a precise guideline of what can run in parallel and what not - right now it seems it's on a test-level basis 

cc @lantoli

marcosuma (approved):

PR #2101, opened 2024-04-02
chore: Reuses project in tests - ldap

lantoli:
## Description

Reuses project in tests - ldap

Link to any related issue(s): CLOUDP-240659

## Type of change:

## Required Checklist:


internal/service/ldapverify/resource_ldap_verify_test.go
```diff
@@ -14,48 +14,22 @@ import (
...
-					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
-					resource.TestCheckResourceAttrSet(resourceName, "request_id"),
-					resource.TestCheckResourceAttr(resourceName, "hostname", hostname),
-					resource.TestCheckResourceAttr(resourceName, "bind_username", username),
-					resource.TestCheckResourceAttr(resourceName, "port", port),
-				),
-			},
-		},
-	})
+	resource.ParallelTest(t, *basicTestCase(t))
```
AgustinBettati:
```suggestion
	resource.Test(t, *basicTestCase(t)) // creating multiple ldap_verify resources for the same project at the same time leads to 500 errors.
```
defining basic tests (in ldap_verify and ldap_configuration) to run in parallel was not intentional right? In addition for the case of the migration tests I believe it is also running in parallel (not a significant issue currently as there is only 1 migration test).
As a good practice we can leave a small comment clarifying why we do not run in parallel, for the case of ldap_verify I was able to confirm that when running in parallel some return 500 error code during create.

lantoli: you're right, it's the only parallel test here so it's not a different in this case but it's better to be explicit, thanks

AgustinBettati (approved):
LGTM

PR #2128, opened 2024-04-09
chore: Adds workflow that generates changelog after PR is merged

oarbusi:
## Description

Adds a workflow to automatically update the unreleased section of the changelog after a PR with a new changelog entry is merged or there is a new entry in the changelog in the repository and the workflow is triggered manually.
Link to any related issue(s): CLOUDP-241727

## Type of change:

## Required Checklist:

## Further comments
Tested the workflow in a [fork](https://github.com/oarbusi/terraform-provider-mongodbatlas)
- Manual trigger of workflow: https://github.com/oarbusi/terraform-provider-mongodbatlas/commit/f717afa11bc7505aafb3fdf07e050bc97fb4e0e0
- Automatic trigger of workflow after PR is merged: https://github.com/oarbusi/terraform-provider-mongodbatlas/commit/bc819700f7be137ccf2aa756442c6cf7c6c0a59d
- Full commit history: https://github.com/oarbusi/terraform-provider-mongodbatlas/commits/master/


contributing/changelog-process.md
```diff
@@ -151,3 +151,8 @@ A PR check is included to validate the changelog entry file.
 If a PR doesn't need a changelog entry its check can be skipped:
 - Adding the label `skip-changelog-check` to the PR.
 - Check in PRs with title `chore`, `test`, `doc` or `ci` is automatically skipped. However a changelog can still be added if needed.
+
+## Unreleased section of CHANGELOG.md automatic update
+
+After a PR is merged to with a new entry in `.changelog` directory, [Generate CHANGELOG workflow](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/workflows/generate-changelog.yml) will be triggered and it will update the `CHANGELOG.md` file with the new entry.
```
oarbusi: link will not work until this PR is merged

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,30 @@
+name: Generate CHANGELOG
+on:
+  pull_request:
+    types: [closed]
+  workflow_dispatch:
+jobs:
+  GenerateChangelog:
+    if: github.event.pull_request.merged || github.event_name == 'workflow_dispatch'
```
lantoli, line 11: knitpit: this is strictly correct and efficient because it's not run if a PR is closed but not merged.

alternatively if we prefer it to be simpler we can remove that if condition.  It'll run sometimes when it's not necessary (when a pr is closed without merge) but there is not problem as this action is idempotent.

lantoli, line 11: knitpit: this is strictly correct and efficient because it's not run if a PR is closed but not merged.

alternatively if we prefer it to be simpler we can remove that if condition.  It'll run sometimes when it's not necessary (when a pr is closed without merge) but there is not problem as this action is idempotent.

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,30 @@
+name: Generate CHANGELOG
+on:
```
lantoli, line 2: @oarbusi cc @AgustinBettati as this is creating a commit in master there might be a race condition, e.g. if 2 PRs are merged almost at the same time.  I know this is highly unlikely but also it's very easy to avoid.
e.g.: 
- PR 1 is merged
- job 1 starts and get current status with only entry file 1
- PR 2 is merged
- job 2 starts, get 1 and 2, and it's faster than job 1 and commits
- job 1 commits, info about PR 2 is lost

We might want to use concurrency limit for all GH actions writing to master: https://docs.github.com/en/actions/using-jobs/using-concurrency

e.g.:

concurrency:
  group: master-write
  cancel-in-progress: false

lantoli: (for example release process might also use this concurrency group)

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,30 @@
...
+          go-version-file: 'go.mod'
+      - run: make tools update-changelog-unreleased-section
+      - run: |
+          if [[ $(git status --porcelain) ]]; then
+            if ${{github.event_name == 'workflow_dispatch'}}; then
+              MSG="Update CHANGELOG.md (Manual Trigger)"
+            else
+              MSG="Update CHANGELOG.md for #${{ github.event.pull_request.number }}"
+            fi
+            git config --local user.email changelogbot@hashicorp.com
```
lantoli: we might want to use a MongoDB email address

.github/workflows/generate-changelog.yml
lantoli: also similarly to example.yaml we can use something like:

  pull_request:
 ...
    paths:
      - .changelog/**

so it will run only if the PR has an associated changelo entry file

but as my previous comment, I don't think it's very important to optimize run of this GH action

oarbusi: I would avoid removing the if just in case a PR is closed with a new file in .changelog directory which could cause issues in the changelog

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,30 @@
...
+            git config --local user.name changelogbot
+            git add CHANGELOG.md
+            git commit -m "$MSG"
+            git push
```
lantoli, line 32: what happens if there are not entries yet? (e.g. in a new release and the PR is a chore with no entry file)

is an empty commit pushed?

lantoli (approved):
LGTM, some minor comments

.github/workflows/generate-changelog.yml
lantoli: doesn't it run with latest master content?

.github/workflows/generate-changelog.yml
oarbusi: I might be wrong but in the case you explained:

> - job 2 starts, get 1 and 2, and it's faster than job 1 and commits

here changes from PR 1 and 2 are commited (and pushed)

> - job 1 commits, info about PR 2 is lost

job 1 would fail to push because tip of current branch would be behind its remote counterpart

so I think the issue would not happen in this specific case.
If the order was reversed (job 1 commits and pushes before job 2) the push from job 2 would fail and PR2 info would be "lost". This would be easily fixable by manually triggering the workflow.
All in all I think it a very unlikely edge case but we can use concurrency as you suggested since it has an easy fix

.github/workflows/generate-changelog.yml
oarbusi: I used the [same](https://github.com/hashicorp/terraform-provider-aws/blob/60cf7fd65c391de1e8fc200207131eb0f35fe028/.github/workflows/generate_changelog.yml#L26) as AWS uses in their provider. Happy to change it. 
Would `changelogbot@mongodb.com` be better?

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,30 @@
+name: Generate CHANGELOG
+on:
+  pull_request:
+    types: [closed]
+  workflow_dispatch:
+jobs:
+  GenerateChangelog:
```
AgustinBettati:
```suggestion
  generate-changelog:
```

.github/workflows/generate-changelog.yml
oarbusi: Nothing would happen. 
`update-changelog-unreleased-section.sh` only updates changelog if there are new entries. If the script does not do any changes to changelog the script would not enter inside the outermost if where the commit and push are performed`if [[ $(git status --porcelain) ]]; then`

contributing/changelog-process.md
AgustinBettati:
```suggestion
After a PR is merged to master with a new entry in `.changelog` directory, [Generate CHANGELOG workflow](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/workflows/generate-changelog.yml) will be triggered and it will update the `CHANGELOG.md` file with the new entry.
```

contributing/changelog-process.md
```diff
@@ -151,3 +151,8 @@ A PR check is included to validate the changelog entry file.
 If a PR doesn't need a changelog entry its check can be skipped:
 - Adding the label `skip-changelog-check` to the PR.
 - Check in PRs with title `chore`, `test`, `doc` or `ci` is automatically skipped. However a changelog can still be added if needed.
+
+## Unreleased section of CHANGELOG.md automatic update
```
AgustinBettati, line 156: We can add this title in the index at the top of the file so it has visibility.

.github/workflows/generate-changelog.yml
oarbusi: yes you are correct, my bad. Even if it's run in latest master branch, I think it's better to keep the if. It will avoid  unnecessary jobs being triggered and help to keep the action history cleaner

contributing/changelog-process.md
oarbusi: This made me realise that with current implementation, the action would be triggered if a PR is merged into any branch (not only master). Thanks! [Fixed here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2128/commits/25e59b9483d71887ca1fcde827f4828671d10af0)

.github/workflows/generate-changelog.yml
lantoli: ok, then you might also want to add "paths" so it's only run when there is a changelog file as most of PRs won't have a changelog file

contributing/changelog-process.md
oarbusi: Done!

.github/workflows/generate-changelog.yml
lantoli: you're right push would fail.

you brought an interesting point, how can we know if this GH action fails?  if it runs and fail i don't think we'll be watching (manually) if they're working

.github/workflows/generate-changelog.yml
lantoli: not sure what email it's best to use, maybe @gssbzn can help.
as an example I found in CLI:
maintainer: MongoDB Inc. <support@mongodb.com>

contributing/changelog-process.md
lantoli: what about doing something similar to examples.yaml and run on push instead of pull_request:

push:
    branches:
      - master

.github/workflows/generate-changelog.yml
```diff
@@ -0,0 +1,31 @@
+name: Generate CHANGELOG
```
lantoli:
```suggestion
name: Generate Changelog
```

contributing/changelog-process.md
```diff
@@ -151,3 +152,8 @@ A PR check is included to validate the changelog entry file.
...
+
+After a PR is merged to master with a new entry in `.changelog` directory, [Generate CHANGELOG workflow](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/workflows/generate-changelog.yml) will be triggered and it will update the `CHANGELOG.md` file with the new entry.
+The workflow can be triggered manually and it will update the `CHANGELOG.md` file with all entries in `.changelog` directory that are not present in the `CHANGELOG.md` file.
```
marcosuma:
```suggestion
This workflow can also be triggered manually and it will update the `CHANGELOG.md` file with all entries in `.changelog` directory that are not present in the `CHANGELOG.md` file.
```

.github/workflows/generate-changelog.yml
oarbusi: Done!

EspenAlbert (approved):

.github/workflows/generate-changelog.yml
AgustinBettati: To make sure I understand, the concurrency group configuration would prevent this potential push failure as one wound run after the other right?
It is true we will likely not keep an eye on the execution of this workflow, would a slack message to our notifications channel be sufficient? Not suggesting to do as part of this PR of course.

oarbusi: Good point, we won't be actively looking at this GH action failing.
I see two ways of avoiding forgetting about this GH action failing
- Add a step to notify team/oncall person when it fails
- Trigger this workflow as part of the release process/workflow. 

I feel the second option is better because we avoid disrupting team/oncall with a notification where the fix is trivial(retrigger workflow) and we automate the fix. If in the release process the step fails we would notice it because when releasing whoever is working on it will notice and will fix it.
WDYT?
cc. @AgustinBettati

AgustinBettati: I like the idea of keeping the changelog up-to-date after each PR as it will update the `Unreleased` section of changelog (helpful for us and external users).
I would not anticipate this workflow having frequent failures, meaning it would no be disruptive for on call. Seeing [cloudflares history](https://github.com/cloudflare/terraform-provider-cloudflare/actions/workflows/generate-changelog.yml) for this action the only failures I see is because of `Updates were rejected because the remote contains work that you do not have locally` (exact issue we where discussing here).

.github/workflows/generate-changelog.yml
gssbzn: We have cloud bot (https://github.com/mms-build-account) but you have to consider in the near future all shipped product needs it's commit signed according to SSDLC cc @marcosuma

.github/workflows/generate-changelog.yml
lantoli: @AgustinBettati yes, concurrency is like a mutex so only one job is executed and the others wait (similar to the mutex in ProjectIDExecution).   

i don't dislike the idea of notifying in Slack when this fails as i don't anticipate many errors either, and it can help us to understand better the problems.

and it might be useful for other GH actions that are also unattended

lantoli: about "Trigger this workflow as part of the release process/workflow" we could do that in addition to do it in each PR in case the latest ones failed, but it's probably not worth it and it's better just to get notified if it fails

.github/workflows/generate-changelog.yml
lantoli: (wow, i'm surprised AWS uses that address)

.github/workflows/generate-changelog.yml
oarbusi: Thanks for the input! Seems like we reached a decision: Notifying through slack when the job fails. I can do it in a follow up PR. Thanks again

.github/workflows/generate-changelog.yml
oarbusi: @gssbzn what does _near future_ mean in this context? so we have an idea if we have to address this now or a bit later.
 @AgustinBettati for context

gssbzn: June next year all ship products need by be compliant with SSDLC

.github/workflows/generate-changelog.yml
lantoli: ok not to implement concurrency at the moment and see if we get errors

.github/workflows/generate-changelog.yml
AgustinBettati: We can make use of ghaction-import-gpg action to sign commits https://github.com/actions/runner/issues/667#issuecomment-775277257. We currently use this action for signing our release. @oarbusi We can include in a followup PR but seems reasonable.

Regarding
> Would changelogbot@mongodb.com be better?

I agree.

PR #2133, opened 2024-04-10
fix: Removes default comment when creating resource `mongodbatlas_privatelink_endpoint_serverless` and adds update support to `mongodbatlas_privatelink_endpoint_service_serverless`

### LINKED ISSUE #2011: [Bug]: CDKTF is trying to recreate PrivatelinkEndpointServiceServerless when it's already created

- [X] I have searched the existing issues

### Provider Version

v1.15.0

### Terraform Version

v1.7.0

### Terraform Edition

Terraform Cloud

### Current Behavior

At first, I deployed PrivatelinkEndpoint and PrivatelinkEndpointServiceServerless via CDKTF. Then, I apply some changes in other resources other than PrivatelinkEndpointServiceServerless. When I ran CDKTF deploy, it yields the following error:

` Error: error adding MongoDB Serverless PrivateLink Endpoint Connection(65e78dc712946a729ac815a9): PATCH https://cloud.mongodb.com/api/atlas/v1.0/groups/<GROUP_ID>/privateEndpoint/serverless/instance/<COMPANY>/endpoint/65e78dc712946a729ac815a9: 400 (request "TENANT_PRIVATE_ENDPOINT_PROVIDER_UPDATE_INVALID") The private endpoint can not be updated in its current state AVAILABLE.
                         │ 
                         │   with mongodbatlas_privatelink_endpoint_service_serverless.serverless  (serverless),
                         │   on cdk.tf.json line 341, in resource.mongodbatlas_privatelink_endpoint_service_serverless.serverless (serverless):
                         │  341:       }
                         │ `

It seems like CDKTF does not catch that the PrivatelinkEndpointSeviceServerless is already created. 

```shell
# Setup VPC PrivateLink
pl_endpoint_serverless = PrivatelinkEndpointServerless(
    self,
    "ple-serverless",
    project_id=Token().as_string(project.id),
    provider_name="AWS",
    instance_name=Token().as_string(db_cluster.name),
)

# Security Group for Private Endpoint
security_group = SecurityGroup(
    self,
    "sg",
    name=get_service_id(cluster_id=cluster_id, service="sg"),
    description="Security Group for Private Endpont",
    vpc_id=Token().as_string(vpc.vpc_id_output),
    ingress_with_source_security_group_id=[
        {
            "rule": "all-tcp",
            "source_security_group_id": Token().as_string(
                eks.node_security_group_id_output
            ),
        }
    ],
)

vpc_endpoint = VpcEndpoint(
    self,
    "endpoint",
    service_name=Token().as_string(
        pl_endpoint_serverless.endpoint_service_name
    ),
    vpc_id=Token().as_string(vpc.vpc_id_output),
    vpc_endpoint_type="Interface",
    subnet_ids=Token().as_list(vpc.private_subnets_output)
    if private_subnet_enabled
    else Token().as_list(vpc.public_subnets_output),
    security_group_ids=[
        Token().as_string(
            security_group.security_group_id_output
        )
    ],
)

pl_endpoint_service_serverless = (
    PrivatelinkEndpointServiceServerless(
        self,
        "serverless",
        provider_name="AWS",
        project_id=Token().as_string(project.id),
        endpoint_id=Token().as_string(
            pl_endpoint_serverless.endpoint_id
        ),
        cloud_provider_endpoint_id=Token().as_string(
            vpc_endpoint.id
        ),
        instance_name=Token().as_string(db_cluster.name),
    )
)
```

### Steps To Reproduce

1. Run TF apply
2. Update SG
3. Run TF apply again

### Logs

```hcl
` Error: error adding MongoDB Serverless PrivateLink Endpoint Connection(65e78dc712946a729ac815a9): PATCH https://cloud.mongodb.com/api/atlas/v1.0/groups/<GROUP_ID>/privateEndpoint/serverless/instance/<COMPANY>/endpoint/65e78dc712946a729ac815a9: 400 (request "TENANT_PRIVATE_ENDPOINT_PROVIDER_UPDATE_INVALID") The private endpoint can not be updated in its current state AVAILABLE.
                         │ 
                         │   with mongodbatlas_privatelink_endpoint_service_serverless.serverless  (serverless),
                         │   on cdk.tf.json line 341, in resource.mongodbatlas_privatelink_endpoint_service_serverless.serverless (serverless):
                         │  341:       }
                         │ `
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

maastha:
## Description

Based on investigations [here](https://jira.mongodb.org/browse/CLOUDP-237302?focusedCommentId=6242448&focusedId=6242448&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6242448) a "create" comment was being added when a `mongodbatlas_privatelink_endpoint_serverless` resource is created. This comment created a drift when `mongodbatlas_privatelink_endpoint_service_serverless` resource was retrieved (in read) as it refers the same "comment" attribute of the private endpoint.

This PR:
- Removes the logic to add that default "create" comment when creating mongodbatlas_privatelink_endpoint_serverless resource.
- Adds support for Update operation ONLY for "comment" attribute in `mongodbatlas_privatelink_endpoint_service_serverless` resource. 
           - - This will help unblock any existing users who run into the issue in https://github.com/mongodb/terraform-provider-mongodbatlas/issues/2011 as they can now perform an in-place update for the comment instead of having to replace the resource due to `ForceNew` that was previously configured for the `comment` attribute. Note that users will not be able to set the comment to null/"" as API returns a 500. We are already working with the respective API team to improve error messaging for such cases ([CLOUDP-242264](https://jira.mongodb.org/browse/CLOUDP-242264)).
- Adds acceptance and migration tests to create Serverless instance private endpoints with AWS VPC configured.

Link to any related issue(s): https://jira.mongodb.org/browse/CLOUDP-237302
Fixes https://github.com/mongodb/terraform-provider-mongodbatlas/issues/2011 

## Type of change:

## Required Checklist:


.github/workflows/acceptance-tests-runner.yml
```diff
@@ -419,6 +419,8 @@ jobs:
           terraform_wrapper: false    
       - name: Acceptance Tests
         env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.aws_access_key_id }}
```
lantoli, line 422: can we also add this to mig test config so they're synchronized even if no mig tests need it now?

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
```diff
@@ -79,11 +80,7 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	projectID := d.Get("project_id").(string)
 	instanceName := d.Get("instance_name").(string)
 
-	serverlessTenantCreateRequest := &admin.ServerlessTenantCreateRequest{
-		Comment: conversion.StringPtr("create"),
-	}
-
-	endPoint, _, err := connV2.ServerlessPrivateEndpointsApi.CreateServerlessPrivateEndpoint(ctx, projectID, instanceName, serverlessTenantCreateRequest).Execute()
```
lantoli, line 86: not to do in this PR but I wonder why comment field is not exposed in TF resource schema

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
```diff
@@ -79,11 +80,7 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+	endPoint, _, err := connV2.ServerlessPrivateEndpointsApi.CreateServerlessPrivateEndpoint(ctx, projectID, instanceName, &admin.ServerlessTenantCreateRequest{}).Execute()
```
lantoli, line 83: just curious, can you try if it works passing nil instead of  &admin.ServerlessTenantCreateRequest{} ?

.changelog/2133.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Removes default comment when creating resource mongodbatlas_privatelink_endpoint_serverless
```
lantoli:
```suggestion
resource/mongodbatlas_privatelink_endpoint_serverless: Removes default comment when creating 
```

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
```diff
@@ -79,11 +80,7 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	projectID := d.Get("project_id").(string)
 	instanceName := d.Get("instance_name").(string)
 
-	serverlessTenantCreateRequest := &admin.ServerlessTenantCreateRequest{
-		Comment: conversion.StringPtr("create"),
```
lantoli, line 83: there is not comment attribute in resource schema that can be passed in create?

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
```diff
@@ -146,6 +148,51 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	return resourceRead(ctx, d, meta)
 }
 
+func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	connV2 := meta.(*config.MongoDBClient).AtlasV2
+	projectID := d.Get("project_id").(string)
+	instanceName := d.Get("instance_name").(string)
+	endpointID := d.Get("endpoint_id").(string)
+
+	_, _, err := connV2.ServerlessPrivateEndpointsApi.GetServerlessPrivateEndpoint(ctx, projectID, instanceName, endpointID).Execute()
```
lantoli: is this call needed? we don't use any data from the response

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
```diff
@@ -146,6 +148,51 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+	if err != nil {
+		return diag.Errorf("error getting Serverless PrivateLink Endpoint Information: %s", err)
+	}
+
+	// only "comment" attribute update is supported, updating other attributes forces replacement of this resource
```
lantoli, line 161: we might want to do in the beginning of update function something like:

if !d.HasChange("comment") {
	return resourceRead(ctx, d, meta)	
	}

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
```diff
@@ -146,6 +148,51 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+		Comment:      conversion.StringPtr(d.Get("comment").(string)),
+		ProviderName: d.Get("provider_name").(string),
+	}
+
+	endPoint, _, err := connV2.ServerlessPrivateEndpointsApi.UpdateServerlessPrivateEndpoint(ctx, projectID, instanceName, endpointID, &updateRequest).Execute()
+	if err != nil {
+		return diag.Errorf(ErrorServerlessServiceEndpointAdd, endpointID, err)
+	}
+
+	stateConf := &retry.StateChangeConf{
```
lantoli: is this needed when only comment is changed or the comment change in the service is synchronous and there is no need to poll the state?

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless_test.go
```diff
@@ -51,6 +65,56 @@ func TestAccServerlessPrivateLinkEndpointService_basic(t *testing.T) {
 	})
 }
 
+func TestAccServerlessPrivateLinkEndpointService_AWSEndpointCommentUpdate(t *testing.T) {
+	var (
+		resourceName            = "mongodbatlas_privatelink_endpoint_service_serverless.test"
+		datasourceEndpointsName = "data.mongodbatlas_privatelink_endpoints_service_serverless.test"
+		orgID                   = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName             = acc.RandomProjectName()
```
lantoli: i guess we can't use ProjectIDExecution because it affects the project tests will affect each other?

lantoli (approved):

.changelog/2133.txt
AgustinBettati: should we add an additional entry for mongodbatlas_privatelink_endpoint_service_serverless mentioning it now supports updating `comment` in place?

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
AgustinBettati: I believe there is not. I am curious if we have context on why this hardcoded value is defined.

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless_migration_test.go
```diff
@@ -38,3 +39,32 @@ func TestMigServerlessPrivateLinkEndpointService_basic(t *testing.T) {
 		},
 	})
 }
+
+func TestMigServerlessPrivateLinkEndpointService_noComment(t *testing.T) {
```
AgustinBettati: [q] not sure from the title and steps of this migration test what scenario it is testing. I see it defines a "test comment" on both steps (no update), so I would assume since no update is made this was also working before?

.github/workflows/acceptance-tests-runner.yml
maastha: good catch, will add

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
maastha: yeah I tried, that didn't work

.changelog/2133.txt
maastha: yes I will update

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
maastha: >I believe there is not. I am curious if we have context on why this hardcoded value is defined.

Yeah I couldn't figure it out, I didn't get any such details from the API team as well on this

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
maastha: yes, it is needed because this resource is supposed to be used in conjunction with another resource `mongodbatlas_privatelink_endpoint_serverless` which creates the resource that is further updated by this service resource.
This actually should be more like a module but we have two resources instead.

cc @Zuhairahmed this is a good candidate for a module

marcosuma:
> We are already working with the respective API team to improve error messaging for such cases.

@maastha can you share in the description the CLOUDP?

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
marcosuma: > This comment created a drift
Aastha if we decided to add this comment field in the schema, would the problem of drift still persist I guess? Can you create a separate CLOUDP as a follow-up on this?

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
marcosuma: Great explanation Aastha - do we have this documented for customers? or is it intuitive when they are using it?

marcosuma: so I am guessing you are using this get just to see if it exists? as Leo said, if the response is not used it's weird we are calling the get.

At minimum we should explain it in the code

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
```diff
@@ -28,6 +30,7 @@ func Resource() *schema.Resource {
 		CreateWithoutTimeout: resourceCreate,
 		ReadWithoutTimeout:   resourceRead,
 		DeleteWithoutTimeout: resourceDelete,
+		UpdateWithoutTimeout: resourceUpdate,
```
marcosuma, line 33: very weird there wasn't an update here

maastha: yes, possibly because apart from the comment attribute no other values can be updated

lantoli: i think it is because all attributes had ForceNew

internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go
maastha: >Aastha if we decided to add this comment field in the schema, would the problem of drift still persist I guess?

Do you mean add it to `mongodbatlas_privatelink_endpoint_serverless` schema? The drift detection currently happens in the other resource which is `mongodbatlas_privatelink_endpoint_service_serverless` so the issue will persist.

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
maastha: I'll add a comment in the code. Essentially, the response isn't used but this is more for error handling I'd say.

As for users, in the documentation this is how it is called out that these resources are to be used together:

>This is the first of two resources required to configure PrivateLink for Serverless, the second is [mongodbatlas_privatelink_endpoint_service_serverless](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/privatelink_endpoint_service_serverless).

Ref: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/privatelink_endpoint_serverless

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
maastha: yeah so because there was no Update operation configured, all the attributes had to be set with ForceNew otherwise TF throws schema validation error

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
marcosuma: > this is more for error handling I'd say.

mmm what type of error handling? the only error you can get here is either the call fails for networking/atlas reasons (which means we are just making it worse) or because it doesn't exist..

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless_test.go
```diff
@@ -65,16 +129,17 @@ func checkDestroy(state *terraform.State) error {
...
 
-func configBasic(orgID, projectName, instanceName, comment string) string {
+func configBasic(orgID, projectID, instanceName, comment string) string {
 	return fmt.Sprintf(`
 
-	resource "mongodbatlas_project" "test" {
-		name   = %[2]q
-		org_id = %[1]q
-	}
+	#resource "mongodbatlas_project" "test" {
```
marcosuma: is this intentional?

maastha: yess I was trying out if ProjectExecution causes any issues, cleaning up the PR, will update when RFAL

.github/workflows/acceptance-tests-runner.yml
maastha: added

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
maastha: added

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless_test.go
maastha: updated to use ProjectIDExecution, I don't think it should be a problem

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
maastha: removed

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless_migration_test.go
maastha: right, updated the name of the test to better reflect the scenario, lmk in case of concerns

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless.go
maastha: My bad, I was thinking about the create operation, for Update you are both right, the GET call should not be required.
Removed.

internal/service/privatelinkendpointserviceserverless/resource_privatelink_endpoint_service_serverless_test.go
maastha: cleaned up!

marcosuma (approved):

marcosuma (approved):

corryroot (approved):
LGTM

PR #2134, opened 2024-04-10
chore: Adds new step in release process for updating header of changelog

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-241731

Defines a new job that automatically adjusts the CHANGELOG.md header section during a new release.

This PR contemplates the possibility of doing pre-releases (by skipping this job), this will not be needed for our regular release process but could be useful in the future if we want to share a preview to customers (pointed out by Melissa in the scope).

## Testing

- Triggered a release process in a forked repository: [Generated commit which is included in release](https://github.com/AgustinBettati/terraform-provider-mongodbatlas/commit/8b1ecfc6cedf83e1d79a3862e997ae7154467583), [triggered release action](https://github.com/AgustinBettati/terraform-provider-mongodbatlas/actions/runs/8643807283).
- Triggered a pre-release making sure it will not modify the changelog: https://github.com/AgustinBettati/terraform-provider-mongodbatlas/actions/runs/8643839887

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,9 +32,27 @@ jobs:
       atlas_cloud_env: "qa"
       ref: ${{ inputs.use_existing_tag == 'true' && inputs.version_number  || github.ref }}
 
+  update-changelog-header:
+    needs: [ run-qa-acceptance-tests ]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633
+        with:
+          fetch-depth: 0
+      - run: ./scripts/update-changelog-header-for-release.sh ${{inputs.version_number}}
+      - run: |
+          if [[ $(git status --porcelain) ]]; then
+            MSG="Update CHANGELOG.md header for ${{inputs.version_number}} release"
+            git config --local user.email changelogbot@mongodb.com
+            git config --local user.name changelogbot
+            git add CHANGELOG.md
+            git commit -m "$MSG"
+            git push
+          fi
+    
   release:
     runs-on: ubuntu-latest
-    needs: [ validate-version-input, run-qa-acceptance-tests ]
+    needs: [ validate-version-input, run-qa-acceptance-tests, update-changelog-header ]
     if: ${{ always() && needs.validate-version-input.result == 'success' && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
     steps:
       - name: Checkout
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,7 +66,7 @@ jobs:
         run: git fetch --prune --unshallow
       - name: Get the latest commit SHA
         id: get-sha
-        run: echo "sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
       - name: Create release tag
         uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,8 @@ jobs:
 
   run-qa-acceptance-tests:
     needs: [ validate-version-input ]
-    # Skipped when explicit input parameter is used
+    # QA acceptance tests are skipped when explicit input parameter is used
+    # As this job may be skipped following jobs require using 'always()' to make sure they are still run
     if: needs.validate-version-input.result == 'success' && inputs.skip_tests != 'true'
     secrets: inherit
     uses: ./.github/workflows/acceptance-tests.yml
@@ -35,8 +36,13 @@ jobs:
   update-changelog-header:
     runs-on: ubuntu-latest
     needs: [ validate-version-input, run-qa-acceptance-tests ]
-    # Skipped if use_existing_tag is defined to true or previous jobs failed.
-    if: ${{ always() && inputs.use_existing_tag != 'true' && needs.validate-version-input.result == 'success' && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
+    # Skipped if use_existing_tag is defined, is a pre-release, or previous jobs failed.
+    if: >-
+      always() 
+      && inputs.use_existing_tag != 'true' 
+      && !contains(inputs.version_number, 'pre') 
+      && needs.validate-version-input.result == 'success' 
+      && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success')
     steps:
       - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
...
+      always() 
+      && needs.validate-version-input.result == 'success' 
+      && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') 
+      && (needs.update-changelog-header.result == 'skipped' || needs.update-changelog-header.result == 'success')
     steps:
       - name: Checkout
         uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633
         with:
-          ref: ${{ inputs.use_existing_tag == 'true' && inputs.version_number || github.ref }}
+          ref: ${{ inputs.use_existing_tag == 'true' && inputs.version_number || 'master' }}
```
AgustinBettati, line 74: need to define master to make sure latest commit is used, more details here https://github.com/actions/checkout/issues/439#issuecomment-830862188

.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
...
       - name: Unshallow
         run: git fetch --prune --unshallow
+      - name: Get the latest commit SHA
+        id: get-sha
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
       - name: Create release tag
         uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72
         with:
           tag: ${{ inputs.version_number }}
+          commit_sha: ${{ steps.get-sha.outputs.sha }}
```
AgustinBettati, line 84: Needed so tag is added into latest commit, if not the sha used for triggering the workflow is used: https://github.com/actions/checkout/issues/439#issuecomment-830862188

scripts/update-changelog-header-for-release.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
lantoli, line 1: what happens if the Unreleased section was already modified before, e.g. this already ran and we're running release workflow again?

will it fail or will do nothing silently?

.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
         run: |
           echo "${{ inputs.version_number }}" | grep -P '^v\d+\.\d+\.\d+(-pre[A-Za-z0-9-]*)?$'
```
lantoli, line 23: once the release is finished successfully, do we need to put the Unrelease header again in the file?

.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
...
       atlas_cloud_env: "qa"
       ref: ${{ inputs.use_existing_tag == 'true' && inputs.version_number  || github.ref }}
 
-  release:
+  update-changelog-header:
     runs-on: ubuntu-latest
     needs: [ validate-version-input, run-qa-acceptance-tests ]
-    if: ${{ always() && needs.validate-version-input.result == 'success' && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
+    # Skipped if use_existing_tag is defined, is a pre-release, or previous jobs failed.
+    if: >-
```
lantoli, line 40: knitpit: this condition is a bit complex

.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
...
+          fetch-depth: 0
+      - run: ./scripts/update-changelog-header-for-release.sh ${{inputs.version_number}}
+      - run: |
+          if [[ $(git status --porcelain) ]]; then
+            MSG="Update CHANGELOG.md header for ${{inputs.version_number}} release"
+            git config --local user.email changelogbot@mongodb.com
+            git config --local user.name changelogbot
+            git add CHANGELOG.md
+            git commit -m "$MSG"
+            git push
```
lantoli, line 58: related to previous comment, if there is no change to the file because was already updated, is an empty commit pushed or no action is done?

.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
...
+          fi
+    
+  release:
+    runs-on: ubuntu-latest
+    needs: [ validate-version-input, run-qa-acceptance-tests, update-changelog-header ]
+    # Release is skipped if there are failures in previous steps
+    if: >- 
+      always() 
+      && needs.validate-version-input.result == 'success' 
+      && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success')
```
lantoli, line 68: which are the possible result values?

lantoli (approved):

.github/workflows/release.yml
oarbusi: if no change to the file is done `$(git status --porcelain)` will be false and no action is done

.github/workflows/release.yml
```diff
@@ -22,31 +22,66 @@ jobs:
...
     with:
       atlas_cloud_env: "qa"
       ref: ${{ inputs.use_existing_tag == 'true' && inputs.version_number  || github.ref }}
 
-  release:
+  update-changelog-header:
     runs-on: ubuntu-latest
     needs: [ validate-version-input, run-qa-acceptance-tests ]
-    if: ${{ always() && needs.validate-version-input.result == 'success' && (needs.run-qa-acceptance-tests.result == 'skipped' || needs.run-qa-acceptance-tests.result == 'success') }}
+    # Skipped if use_existing_tag is defined, is a pre-release, or previous jobs failed.
```
oarbusi, line 39: here you are using `or` but if condition is all `and`. Which one is correct?

.github/workflows/release.yml
oarbusi: from what I see [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2134/files#diff-08c41df88f0b622e21d5036ae226546549ef4bcf07b674a920a76d97c3c9a9d1R26), Unreleased header is never removed and the new versions are always put after the unreleased header

.github/workflows/release.yml
lantoli: i see. a small issue that could happen is that we start release one afternoon, there are some problems, we finish the release next day, but the changelog release section date is yesterday

.github/workflows/release.yml
oarbusi: success, failure, cancelled, or skipped from [docs](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context)

oarbusi (approved):

EspenAlbert (approved):

.github/workflows/release.yml
lantoli: it was in case condition could be simplied, but it can't if there is more than one. thx

scripts/update-changelog-header-for-release.sh
AgustinBettati: The script will output a message like `CHANGELOG already has a header defined for 1.16.0, no changes made to changelog.` (example: https://github.com/AgustinBettati/terraform-provider-mongodbatlas/actions/runs/8638524825/job/23683156506).

.github/workflows/release.yml
AgustinBettati: Correct the Unreleased header is preserved at the top of the changelog, [example commit](https://github.com/AgustinBettati/terraform-provider-mongodbatlas/commit/8b1ecfc6cedf83e1d79a3862e997ae7154467583).

.github/workflows/release.yml
AgustinBettati: true, not sure if the concern is worth making the solution more complex (check if header of current release is present and updating the date). If it ever becomes a significant gap a commit can adjust the date before or after the release.

.github/workflows/release.yml
AgustinBettati: This is because the comment mentions when the job is skipped, while the if statement defines when it is run.

.github/workflows/release.yml
lantoli: agree not to address it now

PR #2135, opened 2024-04-10
feat: Adds project tags

EspenAlbert:
## Description

Adds project tags to `resource_project`, `data_source_project`, and `data_source_projects`
Updates the project example and documentation for resource and data sources.

See more internal changes below

Link to any related issue(s): CLOUDP-242261

## Type of change:

## Required Checklist:

## Further comments

- Adds a new acceptance test `TestAccProject_withTags` in `resource_project_test.go`
- Adds a `testutil/acc/hcl.go`.
  - I use this to simplify the acceptance test for tags. See `TestAccProject_withTags` and `hcl_test.go` for usage.
- Adds a new `acc.AddNoAttrSetChecks`. Used when checking for tag deletion
  - Also unit tested in: `TestAddNoAttrSetChecks`


internal/service/project/data_source_project.go
```diff
@@ -166,6 +167,11 @@ func (d *projectDS) Schema(ctx context.Context, req datasource.SchemaRequest, re
 					},
 				},
 			},
+			"tags": schema.MapAttribute{
+				ElementType: types.StringType,
+				Optional:    true,
```
maastha: In data source fields should be only Computed

internal/service/project/model_project.go
```diff
@@ -202,3 +206,32 @@ func UpdateProjectBool(plan, state types.Bool, setting **bool) bool {
...
+	}
+	return types.MapValueMust(types.StringType, typesTags)
+}
+
+func NewAdminTags(ctx context.Context, tags types.Map) *[]admin.ResourceTag {
+	if tags.IsNull() || len(tags.Elements()) == 0 {
+		return &[]admin.ResourceTag{}
+	}
+	elements := make(map[string]types.String, len(tags.Elements()))
+	// todo: want to improve this.
```
EspenAlbert: I have an idea of improving this. But I think it belongs in a separate PR

internal/service/project/data_source_projects.go
```diff
@@ -163,6 +163,11 @@ func (d *ProjectsDS) Schema(ctx context.Context, req datasource.SchemaRequest, r
 								},
 							},
 						},
+						"tags": schema.MapAttribute{
+							ElementType: types.StringType,
+							Optional:    true,
```
maastha:
```suggestion
```

internal/service/project/model_project.go
```diff
@@ -202,3 +206,32 @@ func UpdateProjectBool(plan, state types.Bool, setting **bool) bool {
...
+	}
+
+	typesTags := make(map[string]attr.Value, len(*tags))
+	for _, tag := range *tags {
+		typesTags[tag.Key] = types.StringValue(tag.Value)
+	}
+	return types.MapValueMust(types.StringType, typesTags)
+}
+
+func NewAdminTags(ctx context.Context, tags types.Map) *[]admin.ResourceTag {
```
maastha:
```suggestion
func NewResourceTags(ctx context.Context, tags types.Map) *[]admin.ResourceTag {
```

internal/service/project/model_project.go
maastha: pending?

internal/service/project/model_project_test.go
```diff
@@ -234,6 +235,7 @@ func TestProjectDataSourceSDKToDataSourceTFModel(t *testing.T) {
 				Limits:                                      limitsTF,
 				IPAddresses:                                 ipAddressesTF,
 				Created:                                     types.StringValue("0001-01-01T00:00:00Z"),
+				Tags:                                        types.MapNull(types.StringType),
```
maastha: should we have non-null case here for Tags? I see all are MapNull

.changelog/2135.txt
```diff
@@ -0,0 +1,8 @@
+```release-note:enhancement
+resource/mongodbatlas_project: Adds `tags` attribute
+```
+
+```release-note:enhancement
+data-source/mongodbatlas_project: Adds `tags` attribute
+```
+
```
lantoli, line 8:
```suggestion
```release-note:enhancement
resource/mongodbatlas_project: Adds `tags` attribute
```

```release-note:enhancement
data-source/mongodbatlas_project: Adds `tags` attribute
```

```release-note:enhancement
data-source/mongodbatlas_projects: Adds `tags` attribute
```

```

Comment (lantoli) [inline .changelog/2135.txt:8]:
```suggestion
```release-note:enhancement
resource/mongodbatlas_project: Adds `tags` attribute
```

```release-note:enhancement
data-source/mongodbatlas_project: Adds `tags` attribute
```

```release-note:enhancement
data-source/mongodbatlas_projects: Adds `tags` attribute
```

```

File: .changelog/2135.txt [hunk 7]
```diff
@@ -0,0 +1,8 @@
+```release-note:enhancement
+resource/mongodbatlas_project: Adds `tags` attribute
+```
+
+```release-note:enhancement
+data-source/mongodbatlas_project: Adds `tags` attribute
```
lantoli, line 6: plural ds is missing in changelog

internal/service/project/model_project.go
```diff
@@ -202,3 +206,32 @@ func UpdateProjectBool(plan, state types.Bool, setting **bool) bool {
 	}
 	return false
 }
+
+func NewTFTags(tags *[]admin.ResourceTag) types.Map {
```
lantoli: i guess conversion FlattenTags & ExpandTagsFromSetSchema can't be used because different types.  
Don't know if we should extract this function or better wait until it's needed in another resource

examples/mongodbatlas_project/main.tf
```diff
@@ -11,4 +11,15 @@ resource "mongodbatlas_project" "test" {
     name  = "atlas.project.deployment.nodesPerPrivateLinkRegion"
     value = 3
   }
+  tags = {
+    Owner       = "Terraform"
+    Environment = "Example"
+    Team        = "tf-experts"
```
lantoli, line 17: 😄

examples/mongodbatlas_project/main.tf
```diff
@@ -11,4 +11,15 @@ resource "mongodbatlas_project" "test" {
...
+  }
+
+  lifecycle {
+    ignore_changes = [
+      tags["CostCenter"] # useful if `CostCenter` is managed outside terraform
+    ]
+  }
 }
```
lantoli: new line at the end

internal/service/project/resource_project.go
```diff
@@ -245,6 +254,16 @@ func (r *projectRS) Schema(ctx context.Context, req resource.SchemaRequest, resp
 					},
 				},
 			},
+			"tags": schema.MapAttribute{
+				ElementType: types.StringType,
+				Optional:    true,
+				Computed:    true,
```
maastha: As discussed offline, not sure if this should be Computed too, we can discuss more in the meeting if needed

internal/service/project/model_project.go
```diff
@@ -34,6 +35,7 @@ func NewTFProjectDataSourceModel(ctx context.Context, project *admin.Group, proj
 		Teams:                                       NewTFTeamsDataSourceModel(ctx, projectProps.Teams),
 		Limits:                                      NewTFLimitsDataSourceModel(ctx, projectProps.Limits),
 		IPAddresses:                                 ipAddressesModel,
+		Tags:                                        NewTFTags(project.Tags),
```
lantoli: in general it's better to user getters in sdk, e.g.: project.GetTags()

internal/service/project/model_project_test.go
```diff
@@ -540,3 +545,43 @@ func TestUpdateProjectBool(t *testing.T) {
 		})
 	}
 }
+
+func TestNewAdminTags(t *testing.T) {
```
lantoli: this is correct and what we're doing. However I'm seeing a slightly different approach that I like more with name being the key for a map, e.g.:

	testCases := map[string]struct {
		plan     types.Map
		expected []admin.ResourceTag
	}{
		"tags null": {types.MapNull(types.StringType), []admin.ResourceTag{}},
...

internal/service/project/resource_project.go
```diff
@@ -301,12 +320,13 @@ func (r *projectRS) Create(ctx context.Context, req resource.CreateRequest, resp
 	if resp.Diagnostics.HasError() {
 		return
 	}
-
+	tags := *NewAdminTags(ctx, projectPlan.Tags)
```
lantoli: it's normally risky to take the content without checking nil reference, i guess in this case we know NewAdminTags always returns a non-nil pointer, but if somebody changes NewAdminTags later that could not be true anymore

internal/service/project/resource_project.go
```diff
@@ -301,12 +320,13 @@ func (r *projectRS) Create(ctx context.Context, req resource.CreateRequest, resp
...
 	projectGroup := &admin.Group{
 		OrgId:                     projectPlan.OrgID.ValueString(),
 		Name:                      projectPlan.Name.ValueString(),
 		WithDefaultAlertsSettings: projectPlan.WithDefaultAlertsSettings.ValueBoolPointer(),
 		RegionUsageRestrictions:   projectPlan.RegionUsageRestrictions.ValueStringPointer(),
+		Tags:                      &tags,
```
lantoli, line 315: related to above comment, could we just have:  tags := NewAdminTags(ctx, projectPlan.Tags) if we later take the address? and here just Tags: tags

internal/service/project/resource_project.go
```diff
@@ -245,6 +254,16 @@ func (r *projectRS) Schema(ctx context.Context, req resource.SchemaRequest, resp
 					},
 				},
 			},
+			"tags": schema.MapAttribute{
+				ElementType: types.StringType,
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.Map{
+					mapvalidator.KeysAre(stringvalidator.LengthBetween(1, 255)),
```
lantoli: this is good because we send better error messages to the client.

on the flip side as we're duplicating the validation logic with server, it could get out of sync if it changes in the server at some moment

marcosuma:
food for thoughts (and for the next time): could have this PR been split in multiple smaller PRs?

internal/service/project/resource_project.go
AgustinBettati: Agree that `Computed` does not seem accurate here. I would assume users will handle all configuration through terraform, and if there are exceptions then `ignore_changes` lifecycle can be used to ignore the `tags` attribute or specific tag keys. This forces the config to be explicit of what values are managed externally from terraform.
For reference [aws tags schema](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/tags/tags.go#L11), as well as our existing [tags schema](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/5a7d85cbe9695abc89392a1c7d72d76b0b272d47/internal/service/advancedcluster/model_advanced_cluster.go#L38).

internal/service/project/model_project.go
AgustinBettati: Good point, could also apply for the schema definition of tags.

website/docs/d/project.html.markdown
```diff
@@ -86,6 +86,7 @@ In addition to all arguments above, the following attributes are exported:
 * `org_id` - The ID of the organization you want to create the project within.
 * `cluster_count` - The number of Atlas clusters deployed in the project.
 * `created` - The ISO-8601-formatted timestamp of when Atlas created the project.
+* `tags` - Map that contains key-value pairs between 1 to 255 characters in length for tagging and categorizing the cluster. To learn more, see [Resource Tags](https://dochub.mongodb.org/core/add-cluster-tag-atlas)
```
AgustinBettati: all new doc changes are making references to categorizing a cluster, would make sure to mention project instead.

internal/service/project/model_project.go
EspenAlbert: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2135#discussion_r1560638070

internal/service/project/resource_project.go
EspenAlbert: Hmm. Better to drop them maybe?

website/docs/d/project.html.markdown
EspenAlbert: Great catch 🔍

internal/service/project/resource_project.go
EspenAlbert: Great point and references.
I'll do a little investigation 🔍

internal/service/project/resource_project.go
lantoli: in this case i'd probably drop it because i don't think this error will happen very often, so good to go simple here

internal/service/project/model_project_test.go
EspenAlbert: See `TestNewResourceTags` and `TestNewTFTags`

examples/mongodbatlas_project/main.tf
```diff
@@ -11,4 +11,15 @@ resource "mongodbatlas_project" "test" {
...
   }
+  tags = {
+    Owner       = "Terraform"
+    Environment = "Example"
+    Team        = "tf-experts"
+  }
+
+  lifecycle {
+    ignore_changes = [
+      tags["CostCenter"] # useful if `CostCenter` is managed outside terraform
```
marcosuma: CostCenter is not in the tags above, is it?

internal/service/project/data_source_projects.go
```diff
@@ -163,6 +163,11 @@ func (d *ProjectsDS) Schema(ctx context.Context, req datasource.SchemaRequest, r
...
+							Computed:    true,
```
marcosuma, line 168: I guess we are removing this after yesterday's conversation? @maastha @EspenAlbert

internal/service/project/model_project.go
marcosuma: I would suggest to wait until we have a need (YAGNI) - unless we do have a need now :D

internal/service/project/model_project.go
marcosuma: avoid this generic comment in the code. Try to be more specific and preferably link to a CLOUDP jira issue

marcosuma: also I am curious of what you'd like to improve and why we can't just do it now :)

internal/service/project/model_project_test.go
marcosuma: also this name should be updated: TestNewResourceTags @maastha @EspenAlbert

internal/service/project/resource_project.go
marcosuma: yes, no validation unless strictly necessary - let's make sure though we have a step in an existing test that while trying to create a > 255 it fails with a meaningful error (it's our responsibility to make sure the error is readable)

internal/service/project/resource_project_test.go
```diff
@@ -980,12 +980,87 @@ func TestAccProject_withInvalidLimitNameOnUpdate(t *testing.T) {
...
+		envUnchanged = "unchanged"
+		tags1        = map[string]string{
+			"Name":        "my-tag-name",
+			"Environment": envUnchanged,
+			"Deleted":     "short-lived",
+		}
+		tagsOneUpdatedOneDeleted = map[string]string{
+			"Name":        nameUpdated,
+			"Environment": envUnchanged,
+			"NewKey":      "new-value",
```
marcosuma, line 999: referring of what I was saying before, one last step could be to check if a >255char long tag fails

website/docs/d/projects.html.markdown
```diff
@@ -61,6 +61,7 @@ data "mongodbatlas_projects" "test" {
 * `org_id` - The ID of the organization you want to create the project within.
 * `cluster_count` - The number of Atlas clusters deployed in the project.
 * `created` - The ISO-8601-formatted timestamp of when Atlas created the project.
+* `tags` - Map that contains key-value pairs between 1 to 255 characters in length for tagging and categorizing the cluster. To learn more, see [Resource Tags](https://dochub.mongodb.org/core/add-cluster-tag-atlas)
```
marcosuma:
```suggestion
* `tags` - Map that contains key-value pairs between 1 to 255 characters in length for tagging and categorizing the project. To learn more, see [Resource Tags](https://dochub.mongodb.org/core/add-cluster-tag-atlas)
```

marcosuma: link seems wrong - please make sure you check carefully after copy/pasting

website/docs/r/project.html.markdown
```diff
@@ -66,6 +67,25 @@ resource "mongodbatlas_project" "test" {
...
+  }
+  
+  lifecycle {
+    ignore_changes = [
+      tags["CostCenter"] # useful if `CostCenter` is managed outside terraform
+    ]
+  }
+```
+
+Key-value pairs between 1 to 255 characters in length for tagging and categorizing the cluster.
```
marcosuma: project

website/docs/r/project.html.markdown
```diff
@@ -57,6 +57,7 @@ resource "mongodbatlas_project" "test" {
 * `name` - (Required) The name of the project you want to create.
 * `org_id` - (Required) The ID of the organization you want to create the project within.
 * `project_owner_id` - (Optional) Unique 24-hexadecimal digit string that identifies the Atlas user account to be granted the [Project Owner](https://docs.atlas.mongodb.com/reference/user-roles/#mongodb-authrole-Project-Owner) role on the specified project. If you set this parameter, it overrides the default value of the oldest [Organization Owner](https://docs.atlas.mongodb.com/reference/user-roles/#mongodb-authrole-Organization-Owner).
+* `tags` - (Optional) Map that contains key-value pairs between 1 to 255 characters in length for tagging and categorizing the cluster. See [below](#tags).
```
marcosuma: here

website/docs/r/project.html.markdown
```diff
@@ -66,6 +67,25 @@ resource "mongodbatlas_project" "test" {
...
+  lifecycle {
+    ignore_changes = [
+      tags["CostCenter"] # useful if `CostCenter` is managed outside terraform
+    ]
+  }
+```
+
+Key-value pairs between 1 to 255 characters in length for tagging and categorizing the cluster.
+
+To learn more, see [Resource Tags](https://dochub.mongodb.org/core/add-cluster-tag-atlas).
```
marcosuma: here

internal/testutil/acc/hcl.go
```diff
@@ -0,0 +1,52 @@
+package acc
```
marcosuma: what's this hcl? sorry if I am missing something

internal/service/project/data_source_project.go
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/data_source_projects.go
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/model_project.go
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/model_project_test.go
EspenAlbert: Also, added null/empty test 👍 in 35d729e4e186219cab55b274cb1af334b82e1645

.changelog/2135.txt
EspenAlbert: Great catch. added

examples/mongodbatlas_project/main.tf
EspenAlbert: 😅

examples/mongodbatlas_project/main.tf
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/model_project.go
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/model_project_test.go
EspenAlbert: I like your suggestion 😄 fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/resource_project.go
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

internal/service/project/resource_project.go
EspenAlbert: fixed in 35d729e4e186219cab55b274cb1af334b82e1645

website/docs/d/projects.html.markdown
EspenAlbert: Thank you. Good catch. 
Fixed in 51a0496a14a971c51daebda9bef68575e288dead
Also, added a step to my personal PR review checklist ✅

website/docs/r/project.html.markdown
EspenAlbert: Fixed in https://github.com/mongodb/terraform-provider-mongodbatlas/commit/51a0496a14a971c51daebda9bef68575e288dead

website/docs/r/project.html.markdown
EspenAlbert: Fixed in https://github.com/mongodb/terraform-provider-mongodbatlas/commit/51a0496a14a971c51daebda9bef68575e288dead

website/docs/r/project.html.markdown
EspenAlbert: Fixed in 679dee5c16fe14b3e76fce885ef359b46a6ba173

internal/testutil/acc/hcl.go
EspenAlbert: [HCL is the HashiCorp configuration language.](https://github.com/hashicorp/hcl)

examples/mongodbatlas_project/main.tf
maastha: nice:)

examples/mongodbatlas_project/main.tf
EspenAlbert: Tried to clarify in 625707e97aac902e62fbc98c34f2bb15c3fb6957 and in b1fef55eeef1b560dce2999a167d056c8190723d
WDYT?

internal/service/project/data_source_projects.go
maastha: This is data source so this would be Computed-only

examples/mongodbatlas_project/README.md
```diff
@@ -1,7 +1,56 @@
...
-This example creates a Project and defines custom values for certain limits.
+This example creates a Project with tags and defines custom values for certain limits.
+It also shows how you can ignore a specific tag key.
 
 Variables Required to be set:
 - `public_key`: Atlas public key
 - `private_key`: Atlas  private key
 - `org_id`: Organization ID where project will be created
+
+## How to ignore tags managed outside of Terraform
```
maastha:
```suggestion
```

maastha: As per discussion yesterday, I don't think we should encourage users to manage infra outside of their configs. Looking at ignore_changes [documentation](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes), it's not intended so that resources can be managed outside of TF configs, it is intended for other scenarios:

>The ignore_changes feature is intended to be used when a resource is created with references to data that may change in the future, but should not affect said resource after its creation. In some rare cases, settings of a remote object are modified by processes outside of Terraform, which Terraform would then attempt to "fix" on the next run. In order to make Terraform share management responsibilities of a single object with a separate process, the ignore_changes meta-argument specifies resource attributes that Terraform should ignore when planning updates to the associated remote object.


It's really great that we are supporting and being vary of useful TF constructs like `ignore_changes` but at the same time it's important to not miscommunicate the intention here.

Would remove this mention from the example as well.

Let me know WDYT? Happy to discuss more on this

internal/service/project/model_project.go
EspenAlbert: I'll present it in the next tech meeting (`API-> TF` & `TF -> API` model conversions). 
For now,  I removed the comment.

internal/service/project/model_project.go
EspenAlbert: BTW: There is a difference in behavior. GetTags() will always return a list. That simplifies the code 😁 (`project.Tags` can be nil)

internal/service/project/resource_project.go
EspenAlbert: ~~Yes, no point in using `Computed`.~~
But I had to add an extra flag to have a difference between MapNull and MapEmpty since the `project.GetTags()` always returns an empty list
`func NewTFTags(tags []admin.ResourceTag, useNullForEmpty bool) types.Map {`
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-242261_project_tags/internal/service/project/model_project.go#L210

UPDATE: Discussion continued in [comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2135#discussion_r1562698554)

internal/service/project/model_project.go
```diff
@@ -202,3 +206,38 @@ func UpdateProjectBool(plan, state types.Bool, setting **bool) bool {
 	}
 	return false
 }
+
+func NewTFTags(tags []admin.ResourceTag, useNullForEmpty bool) types.Map {
+	if tags == nil {
+		if useNullForEmpty {
```
maastha: [q] Why are we using `useNullForEmpty`?

examples/mongodbatlas_project/README.md
```diff
@@ -1,7 +1,56 @@
...
+}
+```
+
+- Notice how `CostCenter` does **not** exist in the tags defined above.
+  - If someone adds the tag outside Terraform and we run `terraform apply`, the output will be `No Changes`
+  - However, if `tag["CostCenter]` is excluded from the `ignore_changes` section, `terraform apply` would detect a plan drift and remove the `CostCenter` tag
+- Notice how `CurrentDRI` exists in the tags defined above
+  - This can be useful when you want an initial empty value, but someone will update the value outside of Terraform
+  - Terraform will only respect the `tags["CurrentDRI"] = "unset"` during resource creation. If you add this after the resource exists, the `tags["CurrentDRI"]` will **not** be set
+
```
maastha: not sure if we want to explain how ignore_changes works, I think just adding an example and link to TF documentation ignore_changes should suffice 
Either ways would have this reviewed by docs team

examples/mongodbatlas_project/README.md
maastha: (sorry about the noise on this) okay so looking at other provider's documentation regarding ignoring tags, I think the intention is conveyed much better [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#ignoring-changes-to-specific-tags) maybe we can include something along those lines as well

internal/service/project/model_project.go
EspenAlbert: It took me a while to wrap my head around this 😅

We need the flag `useNullForEmpty` to separate two cases:

1. `tags` is not specified:
```hcl
resource "mongodbatlas_project" "tags_example" {
  name   = "tags-readme-example1"
  org_id = var.org_id
}
```

2. `tags` is specified but empty

```hcl
resource "mongodbatlas_project" "tags_example" {
  name   = "tags-readme-example1"
  org_id = var.org_id
  tags = {}
}
```

When calling `project.GetTags()` in the SDK, we **always** get an empty `[]admin.ResourceTag{}.`
To avoid provider bugs such as:

image

Does that answer your question?

examples/mongodbatlas_project/README.md
EspenAlbert: Yea. I'm not sure about this either 🤔 

- Pro: It is helpful to know about this method, but it is not intuitive until you see an example.
- Con: It is a bad practice.

I can see two options:
1. Remove the section
2. Add a disclaimer saying this is not recommended and a link to that reference or rewrite it.

I'm curious to hear what other people think, too.

maastha: I'm not saying that we should say that it isn't recommended but maybe just add more about what scenarios this could be helpful for instead of just "manage it outside of Terraform"

internal/service/project/resource_project_test.go
EspenAlbert: Added this and got a 500 :S
https://jira.mongodb.org/browse/CLOUDP-242986

examples/mongodbatlas_project/README.md
AgustinBettati: I would agree this can be a smaller mention, simply having the usage of tags in the examples/mongodbatlas_project/main.tf file and include a ignore lifecycle block with a comment mentioning that it is possible to ignore specific tag keys.

internal/service/project/model_project.go
EspenAlbert: This is related to the removal of `computed`, see [discussion above](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2135#discussion_r1560681693)

After a discussion with @lantoli and @maastha offline, I have the following finding:

> Using `Optional` and `Computed` simplifies the code and the user experience:

1. User experience: Cases 1 & 2 above lead to the same Terraform state. By introducing `tags = {}` there will be no state changes.
2. Code simplified: We don't have to differentiate between Null & EmptyList/EmptyMap. We can then use `ProjectApi.GetTags()`, which always returns an empty list and never nil.

WDYT? @AgustinBettati @maastha @marcosuma 
Can I go back to using both `Optional` and `Computed`?

internal/testutil/acc/hcl.go
marcosuma: yes, but why we have a file named like that? even the method names themselves are not quite clear on what they are doing - suggest renaming both method names and the file name

internal/service/project/model_project.go
maastha: So if we only use `project.Tags` instead of `project.GetTags()` and have Tags attribute as optional-only, can the issue with empty `tags = {}` be handled better? would we still need `useNullForEmpty`?

examples/mongodbatlas_project/README.md
marcosuma: How can we unblock this? should we go with a small mention mostly calling the official documentation and saying that it is possible (but not encouraged)?

internal/service/project/model_project.go
maastha: One scenario where using Optional & computed might be problematic is if user wants to delete all created tags, then if they remove the tags attribute completely from their config to do this, TF might not detect a plan diff as it could interpret the previous value in the state as a "computed" value, @EspenAlbert can you confirm if this happens?

examples/mongodbatlas_project/README.md
marcosuma: Yes, I think this puts too much emphasis on ignore_changes, let's remove all this explanation

examples/mongodbatlas_project/main.tf
```diff
@@ -11,4 +11,4 @@ resource "mongodbatlas_project" "test" {
     name  = "atlas.project.deployment.nodesPerPrivateLinkRegion"
     value = 3
   }
-}
```
marcosuma, line 14: what changed here?

internal/service/project/model_project.go
marcosuma: So are we suggesting to use `Computed` because the API response always return an empty tag list?
If so, how would having a default value (empty) work in this case? would that solve the non-empty plan?

examples/mongodbatlas_project/README.md
maastha: I think we should remove this entire section and simply update the `main.tf` in example to include below in the project resource:

```
  tags = {
    Owner       = "Terraform"
    Environment = "Example"
    Team        = "tf-experts"
    CurrentDRI  = "unset"
  }
  lifecycle {
    ignore_changes = [
      tags["CostCenter"]
    ]
  }
```

internal/service/project/model_project.go
maastha: my understanding is the API response doesn't  return an empty tag list, but the GetTags() method does

EspenAlbert: #### Understanding what is the api response

1. What happens when we create a project with no tags
2. What happens when we create a project with empty tags

- In both cases, when reading the project (either by name or id), tags are always set to the empty list (`"tags": []` in the JSON response)

### Understanding the impact of `computed`

If changes happen outside terraform, will terraform plan detect the change?

#### Case 1: Create a project, update a new tag from the UI. Result: Changes will be detected in the terraform plan and reverted after terraform apply

```hcl
resource "mongodbatlas_project" "tags_test2" {
  name   = "espen-tags-test"
  org_id = var.org_id

  tags = {
    Owner       = "Terraform"
    Environment = "Example"
  }
}
```
image-1

- Updating the value of `Environment=UI` and adding a `UI=yes` tag

image-2

- re-running `terraform apply`

image-3

- UI after `yes`
  
image-4


#### Case 2: Import a project and run without the `tags` attribute. Result: Changes will be detected and terraform will remove the tags

- Existing project with tags

image-4

- terraform configuration

```hcl
resource "mongodbatlas_project" "tags_test2" {
  name   = "espen-tags-test"
  org_id = var.org_id
}
```

- Import the project

image-6

- running `terraform apply`

image-5

- UI after `yes`

image-7

EspenAlbert: Next, I will research how the aws-provider handles the tags.
@maastha @marcosuma does this ☝️ address the concerns about using `Computed`?

examples/mongodbatlas_project/main.tf
EspenAlbert: Extra line break I think. https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2135#discussion_r1560680191

internal/service/project/model_project.go
AgustinBettati: I am struggling to understand why in `case 2` a plan diff is shown when the attribute is marked as computed and not defined in the config.
For reference was able to find in [google best practices](https://github.com/hashicorp/terraform-provider-google/wiki/Developer-Best-Practices#optionalcomputed-lists-of-objects) a mention related to this risk (`removing an attribute completely from the config and having no plan diff`) when having a list of objects that are optional + computed. The guide mentions this usage for cases when there are default values coming from the API, but I do not believe this would be the case.

maastha: >I am struggling to understand why in case 2 a plan diff is shown when the attribute is marked as computed

@AgustinBettati I think the plan diff is shown because for Computed value it is going to show` (known after apply)`, notice that same happens for ip_addresses. The actual diff seems to be triggered by `with_default_alert_settings` which has a default value i think.


Also, based on the cases that we are looking at here, it seems we see adding `Computed` because we want to understand if it can potentially handle changes outside of Terraform (is that correct @EspenAlbert ?) If yes, then we don't need to that because that would imply every attribute in every resource should always be marked as Computed.

@EspenAlbert  I think we can close on this over a call with the team WDYT?

EspenAlbert: @AgustinBettati 
Tried these two cases here:

### Case 3: Create a project without tags, add a tag in the UI, result: `no changes`

However, if we change the project name, then the tags will be updated to empty

### Case 4: Create project without tags or `tags={}`, add a tag in the UI, update/keep tags to `tags = {}`, result: Changes seen and removed

examples/mongodbatlas_project/README.md
EspenAlbert: Removed

examples/mongodbatlas_project/README.md
EspenAlbert: done in 30aa29163faa90e3fe2c93d467669dd8205f7ec4

internal/testutil/acc/hcl.go
EspenAlbert: maybe `terraform_config.go` is better?
And renaming:
`acc.HclMap` -> `acc.TfConfigVariableMap`
`acc.HclLifecycleIgnore` - >`acc.TfConfigLifecycleIgnore`
WDYT?

internal/service/project/model_project.go
EspenAlbert: @maastha no, I was trying to address the `cons` of using `Computed` with the cases above. 
Right now, I cannot see the `cons` of using `Computed`.

The "pros" for adding `Computed` are (also mentioned [in the comment above](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2135#discussion_r1562698554)):

1. Improve user experience: Avoid plan change when `tags` is not defined  -> and `tags = {}`
2. Improve the code readability. Can always represent tags as an empty list/map in the state. (Simplifies code since the API always returns `tags=[]` when tags are not defined or empty)

Looking forward to the call 😁

internal/service/project/model_project_test.go
```diff
@@ -266,6 +268,7 @@ func TestProjectDataSourceSDKToDataSourceTFModel(t *testing.T) {
 				Limits:                                      limitsTF,
 				IPAddresses:                                 ipAddressesTF,
 				Created:                                     types.StringValue("0001-01-01T00:00:00Z"),
+				Tags:                                        emptyTfTags(),
```
marcosuma, line 271: these model tests should not have changed IMO - unless customers are adding tests, there shouldn't be a need to specify empty tags. If the API is returning an empty set even when not required we should ignore that

website/docs/r/project.html.markdown
```diff
@@ -66,6 +67,25 @@ resource "mongodbatlas_project" "test" {
...
+ ```terraform
+ tags = {
+    Owner       = "Terraform"
+    Environment = "Example"
+    Team        = "tf-experts"
+  }
+  
+  lifecycle {
+    ignore_changes = [
+      tags["CostCenter"] # useful if `CostCenter` is managed outside terraform
```
marcosuma:
```suggestion
      tags["CostCenter"] # useful if `CostCenter` key is managed outside terraform
```

internal/testutil/acc/hcl.go
marcosuma: it looks like `config_formatter.go`
Better to use `formatToHCLMap`

internal/service/project/model_project.go
EspenAlbert: Discussed offline and resolved ✅

examples/mongodbatlas_project/README.md
EspenAlbert: Removed and moved example to `main.tf`

internal/testutil/acc/hcl.go
EspenAlbert: Renamed in 17981e8ed017543d8c287cc1bc2bfbc73404b9ed

internal/service/project/model_project.go
```diff
@@ -202,3 +206,27 @@ func UpdateProjectBool(plan, state types.Bool, setting **bool) bool {
...
+	}
+	return types.MapValueMust(types.StringType, typesTags)
+}
+
+func NewResourceTags(ctx context.Context, tags types.Map) []admin.ResourceTag {
+	if tags.IsNull() || len(tags.Elements()) == 0 {
+		return []admin.ResourceTag{}
+	}
+	elements := make(map[string]types.String, len(tags.Elements()))
+	_ = tags.ElementsAs(ctx, &elements, false)
```
oarbusi, line 223: couldn't this throw an error? should we check it?

EspenAlbert: This is done 8 times in the resource_project.go and it is never checked 😅
E.g., https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-242261_project_tags/internal/service/project/resource_project.go#L353
Not sure if we should add the check here. WDYT?

oarbusi (approved):
LGTM

internal/service/project/model_project.go
oarbusi: If it fails and we ignore the error, in this case `elements` will not be populated and tagsAdmin will remain empty. Not sure about the other 8 occurances. I think it would be better to do something like [this](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/78c807dbef5db7750ff384d36ee84cfb002dbe4c/internal/service/streamconnection/model_stream_connection.go#L46) if possible.
Can be addressed in another PR since this is already a big one.

internal/service/project/model_project_test.go
```diff
@@ -540,3 +545,48 @@ func TestUpdateProjectBool(t *testing.T) {
...
+}
+
+func TestNewTFTags(t *testing.T) {
+	var (
+		tfMapEmpty     = emptyTfTags()
+		apiListEmpty   = []admin.ResourceTag{}
+		apiSingleTag   = []admin.ResourceTag{*admin.NewResourceTag("key1", "value1")}
+		tfMapSingleTag = types.MapValueMust(types.StringType, map[string]attr.Value{"key1": types.StringValue("value1")})
+	)
+	testCases := map[string]struct {
```
lantoli, line 576: nice the string map 👏

internal/service/project/resource_project_test.go
```diff
@@ -980,12 +981,106 @@ func TestAccProject_withInvalidLimitNameOnUpdate(t *testing.T) {
...
+			"Name": nameUpdated,
+		}
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyProject,
+		Steps: []resource.TestStep{
+			// no tags
```
lantoli: knit: have comments at the right of Config lines (if needed)

internal/testutil/acc/attribute_checks_test.go
```diff
@@ -56,3 +57,14 @@ func TestJSONEquals(t *testing.T) {
 		})
 	}
 }
+
+func TestAddNoAttrSetChecks(t *testing.T) {
+	asserter := assert.New(t)
```
lantoli, line 62: knit: i prefer assert := ..., one the few exceptions I think it's ok to shadow, then assert.Empty, etc.

internal/testutil/acc/config_formatter.go
```diff
@@ -0,0 +1,52 @@
...
+		strings.Join(ignoreParts, "\n"),
+		"\t\t]",
+		"\t}",
+	}
+	return strings.Join(lines, "\n")
+}
+
+// make test deterministic
+func sortStringMapKeys(m map[string]string) []string {
+	keys := make([]string, 0)
```
lantoli, line 46: knit: as you know the len it's a bit more efficient:

keys := make([]string, len(m))
for k := range m {
		keys[m] = k
}

lantoli: knit: as you know the len it's a bit more efficient to define capacity: 
keys := make([]string, 0, len(m))

lantoli (approved):

internal/testutil/acc/attribute_checks_test.go
EspenAlbert: Hmm. I couldn't find this practice in other places in the code base. (searching for `assert := assert.New(t)` )
And the `golangci-lint run` will complain about it:

```
internal/testutil/acc/attribute_checks_test.go:62:2: importShadow: shadow of imported from 'github.com/stretchr/testify/assert' package 'assert' (gocritic)
	assert := assert.New(t)
	^
```
Therefore, I prefer keeping it as is

internal/testutil/acc/config_formatter.go
EspenAlbert: Fixed in c8208d9

internal/service/project/resource_project_test.go
EspenAlbert: Agree. They don't add much value. Removed in c8208d9

internal/service/project/model_project_test.go
EspenAlbert: Hmm. I believe they have to change.
Since we have added `Tags` to the structs and we do a full comparison in the assertion, I don't see how we can avoid that.
Or maybe you meant something else? That it should be a `types.MapNull()` ?

internal/service/project/resource_project.go
```diff
@@ -79,6 +80,11 @@ type TFTeamModel struct {
 	RoleNames types.Set    `tfsdk:"role_names"`
 }
 
+type TFTagModel struct {
+	Key   types.String `tfsdk:"key"`
+	Value types.String `tfsdk:"value"`
+}
```
AgustinBettati: This can be removed right?

AgustinBettati (approved):
LGTM

internal/service/project/resource_project.go
EspenAlbert: Yes, good spot!
Removed in 5b37865

internal/service/project/model_project_test.go
marcosuma: ok got it, I see it now

examples/mongodbatlas_project/README.md
```diff
@@ -1,5 +1,6 @@
 # MongoDB Atlas Provider -- Atlas Project with custom limits
-This example creates a Project and defines custom values for certain limits.
+This example creates a Project with tags and defines custom values for certain limits.
```
JuliaMongo: Project can be lower case

JuliaMongo (approved):
LGTM, also will ask Corry or Zach to take a look if needed.

maastha (approved):
LGTM! Great job!

zach-carr (approved):
LGTM

examples/mongodbatlas_project/README.md
EspenAlbert: Thank you. Fixed in 78f2475

PR #2266, opened 2024-05-10
chore: Fail fast when migration test is detecting TF_CLI_CONFIG_FILE

EspenAlbert:
## Description

Using TF_CLI_CONFIG_FILE might cause your local "latest" build version of TF provider to be used instead of latest released version (e.g., 1.16.0)

Link to any related issue(s): CLOUDP-248697

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/testutil/mig/provider.go
+++ b/internal/testutil/mig/provider.go
@@ -1,6 +1,7 @@
 package mig
 
 import (
+	"log"
 	"os"
 	"testing"
 
@@ -32,11 +33,19 @@ func ExternalProvidersWithAWS() map[string]resource.ExternalProvider {
 
 func checkLastVersion(tb testing.TB) {
 	tb.Helper()
+	validateConflictingEnvVars()
 	if os.Getenv("MONGODB_ATLAS_LAST_VERSION") == "" {
 		tb.Fatal("`MONGODB_ATLAS_LAST_VERSION` must be set for migration acceptance testing")
 	}
 }
 
 func versionConstraint() string {
+	validateConflictingEnvVars()
 	return os.Getenv("MONGODB_ATLAS_LAST_VERSION")
+}
+
+func validateConflictingEnvVars() {
+	if os.Getenv("TF_CLI_CONFIG_FILE") != "" {
+		log.Fatal("found `TF_CLI_CONFIG_FILE` in env-var when running migration tests, this might override the terraform provider for MONGODB_ATLAS_LAST_VERSION and instead use local latest build!")
+	}
 }
--- a/internal/testutil/mig/provider.go
+++ b/internal/testutil/mig/provider.go
@@ -1,7 +1,6 @@
 package mig
 
 import (
-	"log"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


internal/testutil/mig/provider.go
```diff
@@ -32,11 +33,19 @@ func ExternalProvidersWithAWS() map[string]resource.ExternalProvider {
...
 }
 
 func versionConstraint() string {
+	validateConflictingEnvVars()
 	return os.Getenv("MONGODB_ATLAS_LAST_VERSION")
 }
+
+func validateConflictingEnvVars() {
+	if os.Getenv("TF_CLI_CONFIG_FILE") != "" {
+		log.Fatal("found `TF_CLI_CONFIG_FILE` in env-var when running migration tests, this might override the terraform provider for MONGODB_ATLAS_LAST_VERSION and instead use local latest build!")
```
EspenAlbert: Would have liked to use `tb.Fatal` instead of `log.Fatal` but I think there is a possibility of `versionConstraint` being called without `checkLastVersion`, therefore, I had to use `log.Fatal`

lantoli: sorry, not sure I understand this, can you elaborate or explain in a different way?  We shouldn't use log.Fatal in tests but tb.Fatal

EspenAlbert: I was afraid of a case where some migration tests would call `ExternalProvidersWithAWS` and `ExternalProviders` which call `versionConstraint` and the migration test would not be skipped.
However, looking at the code, everywhere those are used we expect to see `mig.PreCheck` or `mig.SkipIfVersionBelow` used which would check the TF_CLI_CONFIG_FILE env-var and skip the test

Changed in [11ecbc6](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2266/commits/11ecbc61b46a42bfe5c1a33acb3adad63fc45d02)

marcosuma (approved):

internal/testutil/mig/provider.go
```diff
@@ -11,6 +11,7 @@ import (
 
 func SkipIfVersionBelow(tb testing.TB, minVersion string) {
 	tb.Helper()
+	validateConflictingEnvVars(tb)
```
lantoli, line 14: do we need to call it here?   I assume that checkLastVersion will be called at some point in a mig test

lantoli (approved):

internal/testutil/mig/provider.go
EspenAlbert: I found some examples where only `acc` pre checks are used, e.g., [here.](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/streaminstance/resource_stream_instance_migration_test.go#L19)

lantoli: oh i see, so we have some inconsistencies as we should call checkLastVersion before calling versionConstraint, but it's not happening in the example you're mentioning.

mig.ExternalProviders calls versionConstraint but checkLastVersion is not called because the test is missing a mig precheck

lantoli: all mig tests should call a mig precheck but not sure if it's easy to enforce

EspenAlbert: Yes, it would be nice if we could enforce it.
For now, I pefer keeping the `validateConflictingEnvVars` in the `SkipIfVersionBelow` 👍

PR #2282, opened 2024-05-15
fix: Removes update logic for the global_cluster_config resource.

marcosuma:
## Description

More info on https://jira.mongodb.org/browse/HELP-59447 

tl;dr: updating the global_cluster_config resource leaves the collection and index in an inconsistent state because they are on the data plane side and they would remain unchanged. 
That's also why the API does not have any PATCH endpoint.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/globalclusterconfig/resource_global_cluster_config_test.go
+++ b/internal/service/globalclusterconfig/resource_global_cluster_config_test.go
@@ -103,7 +103,7 @@ func TestAccClusterRSGlobalCluster_database(t *testing.T) {
 				),
 			},
 			{
-				Config:      configBasic(&clusterInfo, true, false),
+				Config:      configWithDBConfig(&clusterInfo, customZoneUpdated),
 				ExpectError: regexp.MustCompile("Updating a global cluster configuration resource is not allowed."),
 			},
 			{
--- a/internal/service/globalclusterconfig/resource_global_cluster_config_test.go
+++ b/internal/service/globalclusterconfig/resource_global_cluster_config_test.go
@@ -39,7 +39,7 @@ func TestAccClusterRSGlobalCluster_basic(t *testing.T) {
 			},
 			{
 				Config:      configBasic(&clusterInfo, true, false),
-				ExpectError: regexp.MustCompile("Updating a global cluster configuration resource is not allowed."),
+				ExpectError: regexp.MustCompile("Updating a global cluster configuration resource is not allowed"),
 			},
 		},
 	})
@@ -104,7 +104,7 @@ func TestAccClusterRSGlobalCluster_database(t *testing.T) {
 			},
 			{
 				Config:      configWithDBConfig(&clusterInfo, customZoneUpdated),
-				ExpectError: regexp.MustCompile("Updating a global cluster configuration resource is not allowed."),
+				ExpectError: regexp.MustCompile("Updating a global cluster configuration resource is not allowed"),
 			},
 			{
 				ResourceName:            resourceName,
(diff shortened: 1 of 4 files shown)


internal/service/globalclusterconfig/resource_global_cluster_config_test.go
```diff
@@ -116,19 +103,8 @@ func TestAccClusterRSGlobalCluster_database(t *testing.T) {
...
-					resource.TestCheckResourceAttrSet(resourceName, "custom_zone_mapping.%"),
-					resource.TestCheckResourceAttrSet(resourceName, "custom_zone_mapping.US"),
-					resource.TestCheckResourceAttrSet(resourceName, "custom_zone_mapping.IE"),
-					resource.TestCheckResourceAttrSet(resourceName, "custom_zone_mapping.DE"),
-					resource.TestCheckResourceAttrSet(resourceName, "custom_zone_mapping.JP"),
-					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
-					resource.TestCheckResourceAttr(resourceName, "cluster_name", clusterInfo.ClusterName),
-				),
+				Config:      configWithDBConfig(&clusterInfo, customZoneUpdated),
+				ExpectError: regexp.MustCompile("Updating a global cluster configuration resource is not allowed"),
```
lantoli, line 107: knit: testing the update error in one test should be enough, no need to test it twice

website/docs/r/global_cluster_config.html.markdown
```diff
@@ -13,6 +13,7 @@ description: |-
 
 -> **NOTE:** Groups and projects are synonymous terms. You may find group_id in the official documentation.
 
+~> **IMPORTANT:** A Global Cluster Configuration, once created, can only be deleted and never re-created again with the same data if not without the help of the Atlas UI. This is because the configuration and its related collection with shard key and indexes are managed separately and they would end up in an inconsistent state. [Read more about Global Cluster Configuration](https://www.mongodb.com/docs/atlas/global-clusters/)
```
lantoli: i guess we can't prevent the resource for being created again once deleted?  it'd be nice if we could detect that conditions and block the create (the second time)

lantoli (approved):

oarbusi (approved):

website/docs/r/global_cluster_config.html.markdown
maastha: Should we include about this in the release guide too? 
Not sure if this should be considered a breaking change, I understand we are removing an incorrect behavior (just thinking out loud)

marcosuma: yes, but I can't think of a way on our side. It should be something the Atlas itself need to implement a block at API level.

marcosuma: It should not be considered a breaking change because it's a bug. I wasn't planning to put it in the release guide, but let me know otherwise

sarahsimpers: Suggestion: the wording is confusing here, suggest we clarify it:
```suggestion
~> **IMPORTANT:** A Global Cluster Configuration, once created, can only be deleted. You can recreate the Global Cluster with the same data only in the Atlas UI. This is because the configuration and its related collection with shard key and indexes are managed separately and they would end up in an inconsistent state. [Read more about Global Cluster Configuration](https://www.mongodb.com/docs/atlas/global-clusters/)
```

sarahsimpers (commented):
LGTM with minor suggestion

sarahsimpers (approved):

PR #2291, opened 2024-05-21
fix: Marks `project.region_usage_restrictions` as OptionalComputed to avoid provider upgrade errors

EspenAlbert:
## Description

- Marks `project.region_usage_restrictions` as OptionalComputed to avoid provider upgrade errors
- Enables `TestAccGovProject_withProjectOwner` in CI
- Adds `TestMigGovProject_regionUsageRestrictionsDefault` (also enabled in the CI, error reproduction)

Link to any related issue(s): #2263 and CLOUDP-248517

Marking the attribute as Computed as earlier versions of the resource will have an empty state for this attribute.
Since the state is empty and the default server value depends on cloud-gov or not:

- If cloud-gov, default value is: `COMMERCIAL_FEDRAMP_REGIONS_ONLY`
- else, default value is `null`

Example of error and solution: 
```
mongodbatlas_project.project: Modifying... [id=664c6820c4391e7c9f58377c]
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to mongodbatlas_project.project, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new value: .region_usage_restrictions: was null, but now cty.StringVal("COMMERCIAL_FEDRAMP_REGIONS_ONLY").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

After adding `Computed: true`

```
mongodbatlas_project.project: Refreshing state... [id=664c6820c4391e7c9f58377c]

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
```

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
[...]

PR #2302, opened 2024-05-24
chore: Supports running cloud gov tests in CI

EspenAlbert:
## Description

Supports running cloud gov tests in CI.

Follow-up PR with a gov migration and acceptance test: #2291 

## Solution discussion
Uses a fully configured provider when using cloud-gov, why?

1. Enables running gov tests in the same process as cloud-dev/cloud-qa tests
2. Keeps the CI simple with a single acceptance-tests job and we can run the tests often

Alternatives considered:
1. Modify env-vars for each cloud-gov test run (similar to [`t.Setenv("MONGODB_ATLAS_LAST_VERSION", "1.2.3")`](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/testutil/mig/test_case_test.go#L14) but with MONGODB_ATLAS_PRIVATE/PUBLIC_KEY/BASE_URL++
  - Does not allow running tests in parallel since env-vars would impact the "normal" acceptance tests

2. Run a separate process/workflow
  - Would require different CI workflows, but config would be a bit shorter as the provider would read env-vars directly

Link to any related issue(s): CLOUDP-250271

## Type of change:

## Required Checklist:


internal/testutil/acc/provider.go
```diff
@@ -38,3 +41,23 @@ func providerAWS() *resource.ExternalProvider {
 		Source:            "hashicorp/aws",
 	}
 }
+
+// ConfigProvider creates a new provider with credentials explicit in config.
+//
+// This can be used when you want credentials different from the default env-vars.
+func ConfigProvider(publicKey, privateKey, baseURL string) string {
```
lantoli: is this func only used by ConfigGovProvider? if so, does it need to be public?

lantoli (approved):

internal/testutil/acc/factory.go
```diff
@@ -51,6 +51,17 @@ func ConnV2UsingProxy(proxyPort *int) *admin.APIClient {
 	return client.(*config.MongoDBClient).AtlasV2
 }
 
+func ConnV2UsingGov() *admin.APIClient {
+	cfg := config.Config{
+		PublicKey:    os.Getenv("MONGODB_ATLAS_GOV_PUBLIC_KEY"),
+		PrivateKey:   os.Getenv("MONGODB_ATLAS_GOV_PRIVATE_KEY"),
+		BaseURL:      os.Getenv("MONGODB_ATLAS_GOV_BASE_URL"),
+		RealmBaseURL: os.Getenv("MONGODB_ATLAS_GOV_BASE_URL"),
```
AgustinBettati: I believe realm uses a different URL, meaning `BaseURL` != `RealmBaseURL`. Would simply leave the realm url undefined and we can configure this if we have the need in the future.

contributing/testing-best-practices.md
```diff
@@ -32,6 +32,15 @@
...
 - Helper functions such as `resource.TestCheckTypeSetElemNestedAttrs` or `resource.TestCheckTypeSetElemAttr` can be used to check resource and data source attributes more easily, e.g. [resource_serverless_instance_test.go](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/66c44e62c9afe04ffe8be0dbccaec682bab830e6/internal/service/serverlessinstance/resource_serverless_instance_test.go#L61).
 
+### Cloud Gov tests
+
+1. Use [`PreCheck: PreCheckGovBasic`](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-250271_cloud_gov/internal/testutil/acc/pre_check.go#L98)
+2. Use the [`acc.ConfigGovProvider`](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-250271_cloud_gov/internal/testutil/acc/provider.go#L61) together with your normal terraform config
+3. Modify the `checkExist` and `CheckDestroy` to use `acc.ConnV2UsingGov`
+4. Follow naming convention:
+   1. `TestAccGovProject_withProjectOwner`, note prefix: `TestAccGov`
```
AgustinBettati, line 41:
```suggestion
   1. `TestAccProject_govWithProjectOwner`, note prefix: `TestAccGov`
```
not a strong opinion but like the idea of leaving the first part before the underscore only with the resource, and second part giving detail of the test case (as shown in [tf docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/testcase#creating-acceptance-tests-functions))

internal/testutil/acc/provider.go
AgustinBettati: nice 👍

AgustinBettati (approved):

internal/testutil/acc/factory.go
marcosuma: I believe Agustin is right, the URL is not the same. Also agree on the action

marcosuma (approved):

contributing/testing-best-practices.md
EspenAlbert: I think using it in the first part makes `regex` selection simpler. 
Although with this approach it might not be necessary 🤔 WDYT?

internal/testutil/acc/provider.go
EspenAlbert: A good point, changed to private 👍

internal/testutil/acc/factory.go
EspenAlbert: Changed ✅

contributing/testing-best-practices.md
AgustinBettati: > I think using it in the first part makes regex selection simpler

Not sure if I understand this point given we wont need to identify gov tests from others

EspenAlbert: Good discussion. Added a bit more context in c888aff

maastha: I think we should be executing tests by resource and not by environment (gov vs. non-gov). For example, if we change something in the project resource, we would want to run tests in gov too to ensure nothing broke.

Would `Test(Acc|Mig)Gov` do that with the current implementation?

PR #2324, opened 2024-06-04
fix: Adds create error handling to push-based log export resource

maastha:
## Description

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/pushbasedlogexport/resource.go
+++ b/internal/service/pushbasedlogexport/resource.go
@@ -60,6 +60,9 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
 			return
 		}
 
+		logConfig, _, _ := connV2.PushBasedLogExportApi.GetPushBasedLogConfiguration(ctx, projectID).Execute()
+		log.Printf("new state.. %s", *logConfig.State)
+
 		resp.State.RemoveResource(ctx)
 		return
 	}
--- a/internal/service/pushbasedlogexport/resource.go
+++ b/internal/service/pushbasedlogexport/resource.go
@@ -60,9 +60,6 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
 			return
 		}
 
-		logConfig, _, _ := connV2.PushBasedLogExportApi.GetPushBasedLogConfiguration(ctx, projectID).Execute()
-		log.Printf("new state.. %s", *logConfig.State)
-
 		resp.State.RemoveResource(ctx)
 		return
 	}
@@ -78,6 +75,13 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
 	if err != nil {
 		resp.Diagnostics.AddError("Error when creating push-based log export configuration", err.Error())
 
+		log.Printf("[INFO] Unconfiguring push-based log export for project due to create failure: %s", projectID)
+		if _, err := connV2.PushBasedLogExportApi.DeletePushBasedLogConfiguration(ctx, projectID).Execute(); err != nil {
+			resp.Diagnostics.AddError("Error when unconfiguring push-based log export configuration", err.Error())
+			return
+		}
+
+		resp.State.RemoveResource(ctx)
 		return
 	}
 
--- a/internal/service/pushbasedlogexport/resource.go
+++ b/internal/service/pushbasedlogexport/resource.go
@@ -51,7 +51,15 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
 	connV2 := r.Client.AtlasV2
 	projectID := tfPlan.ProjectID.ValueString()
 	if _, err := connV2.PushBasedLogExportApi.CreatePushBasedLogConfiguration(ctx, projectID, logExportConfigReq).Execute(); err != nil {
-		resp.Diagnostics.AddError("eError when creating push-based log export configuration", err.Error())
+		resp.Diagnostics.AddError("Error when creating push-based log export configuration", err.Error())
+
+		// log.Printf("[INFO] Unconfiguring push-based log export for project due to create failure: %s", projectID)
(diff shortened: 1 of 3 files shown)


EspenAlbert (approved):
Do we have a test covering this situation?

internal/service/pushbasedlogexport/resource.go
```diff
@@ -51,7 +52,15 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
 	connV2 := r.Client.AtlasV2
 	projectID := tfPlan.ProjectID.ValueString()
 	if _, err := connV2.PushBasedLogExportApi.CreatePushBasedLogConfiguration(ctx, projectID, logExportConfigReq).Execute(); err != nil {
-		resp.Diagnostics.AddError("eError when creating push-based log export configuration", err.Error())
+		resp.Diagnostics.AddError("Error when creating push-based log export configuration", err.Error())
+
+		log.Printf("[INFO] Unconfiguring push-based log export for project due to create failure: %s", projectID)
+		if _, err := connV2.PushBasedLogExportApi.DeletePushBasedLogConfiguration(ctx, projectID).Execute(); err != nil {
```
AgustinBettati: just curious, in what scenario does the create fail but we still have to follow up with a delete of the resource?

maastha: This is an example of the [failure](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/9334168181/job/25692105230). 
The fix is basically to address this error: `Push-based log export
        configuration is already configured (current state =
        BUCKET_VERIFICATION_FAILED) for group 665bbaee73190531728ff741. Push-based
        log export configuration can only be created from the "UNCONFIGURED" state.`

To elaborate, in case of failure the status of log export can be set to BUCKET_VERIFICATION_FAILED or ASSUME_ROLE_FAILED or something else based on the kind of error, for such scenarios we need to reset the state to UNCONFIGURED

oarbusi (approved):

internal/service/pushbasedlogexport/resource.go
marcosuma: @maastha I am not 100% clear after reading your answer. If this create failed because "is already configured", wouldn't the delete affect some other configuration that was perhaps created elsewhere?

internal/service/pushbasedlogexport/resource.go
```diff
@@ -51,7 +52,15 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
 	connV2 := r.Client.AtlasV2
 	projectID := tfPlan.ProjectID.ValueString()
 	if _, err := connV2.PushBasedLogExportApi.CreatePushBasedLogConfiguration(ctx, projectID, logExportConfigReq).Execute(); err != nil {
-		resp.Diagnostics.AddError("eError when creating push-based log export configuration", err.Error())
+		resp.Diagnostics.AddError("Error when creating push-based log export configuration", err.Error())
+
+		log.Printf("[INFO] Unconfiguring push-based log export for project due to create failure: %s", projectID)
```
marcosuma: I think this should be part of the `AddError` message (eventually)

internal/service/pushbasedlogexport/resource.go
```diff
@@ -2,6 +2,7 @@ package pushbasedlogexport
 
 import (
 	"context"
+	"log"
```
marcosuma, line 5: are these changes coming from the recent episode of the test failure?

internal/service/pushbasedlogexport/resource.go
AgustinBettati: > wouldn't the delete affect some other configuration that was perhaps created elsewhere?

We could limit this follow-up delete operation to only be called if the create operation resulted in `BUCKET_VERIFICATION_FAILED` or `ASSUME_ROLE_FAILED`, providing this information to the user. This way if creation fails due to these cases, the create operation will bring the configuration back to `UNCONFIGURED` (currently we fail but the configuration is also left in a failed state).

internal/service/pushbasedlogexport/resource.go
maastha: yes

internal/service/pushbasedlogexport/resource.go
maastha: Could be, but I chose to log this separately so it's understood why we need to make this API call

internal/service/pushbasedlogexport/resource.go
maastha: Agree with @AgustinBettati that we can limit the delete to be called only if status is BUCKET_VERIFICATION_FAILED or ASSUME_ROLE_FAILED.

maastha: added

internal/service/pushbasedlogexport/resource.go
```diff
@@ -51,7 +53,18 @@ func (r *pushBasedLogExportRS) Create(ctx context.Context, req resource.CreateRe
...
+		resp.Diagnostics.AddError("Error when creating push-based log export configuration", err.Error())
+
+		logConfig, _, _ := connV2.PushBasedLogExportApi.GetPushBasedLogConfiguration(ctx, projectID).Execute()
+		if logConfig != nil && slices.Contains(failureStates, *logConfig.State) {
+			log.Printf("[INFO] Unconfiguring push-based log export for project due to create failure: %s", projectID)
+			if _, err := connV2.PushBasedLogExportApi.DeletePushBasedLogConfiguration(ctx, projectID).Execute(); err != nil {
+				resp.Diagnostics.AddError("Error when unconfiguring push-based log export configuration", err.Error())
+				return
+			}
+		}
```
AgustinBettati: could we extract to a function so it does not diverge with identical logic below?

AgustinBettati (approved):
LGTM

internal/service/pushbasedlogexport/resource.go
maastha: done

PR #2333, opened 2024-06-06
chore: Sends only one notification when Terraform Compatibility Matrix fails

oarbusi:
## Description

Sends only one notification when Terraform Compatibility Matrix fails
- Add boolean input `send_notification` in Test suite workflow, defaults to true
- Test suite sends notification only if `send_notification == true`
- terraform compatibility matrix sets the value of `send_notification` to `false` when calling test suite

Link to any related issue(s): CLOUDP-248667

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -13,7 +13,7 @@ on:
       send_notification:
         description: 'Send the Slack notification if any of the tests fail.'
         type: boolean
-        default: true
+        default: false
   workflow_call:
     inputs:
       terraform_matrix:
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -81,7 +81,7 @@ jobs:
   
   slack-notification:
     needs: [tests, clean-after]
-    if: ${{ !cancelled() && github.event_name == 'schedule' && needs.tests.result == 'failure' && inputs.send_notification == true}}
+    if: ${{ !cancelled() && github.event_name == 'schedule' && needs.tests.result == 'failure' && inputs.send_notification == 'true' }}
     runs-on: ubuntu-latest
     permissions: {}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-suite.yml
```diff
@@ -10,6 +10,10 @@ on:
       provider_matrix:
         description: 'Previous MongoDB Atlas Provider version matrix for migration tests (JSON array)'
         default: '[""]' # "" for latest version
+      send_notification:
+        description: 'Send the Slack notification if any of the tests fail. Can be either `true` or `false`.'
+        type: boolean
+        default: true
```
AgustinBettati: In case we trigger a test suite execution manually from GH UI, even if we leave this in true is wont run because notifications are sent when `github.event_name == 'schedule'`. Can this parameter be removed for the case of dispatch to avoid confusion?

marcosuma (approved):

.github/workflows/test-suite.yml
oarbusi: Makes sense! Removed

AgustinBettati: I believe the parameter is still in dispatch, but maybe I am missing the reasoning behind the latest change.

oarbusi: I might have misunderstood. I removed `github.event_name == 'schedule'` from the condition to send the notification, but added the input `send_notification` in both workflow_call and workflow_displatch (defaulting to true in both) so that we can choose if we want to send it when running it manually. 
You meant to remove the input completely from `workflow_dispatch`?

AgustinBettati: I see, yes I was thinking that for the case of dispatching an execution manually there would be no need of notifying in slack. Maybe we can just adjust the default to `false` for this case?

oarbusi: Okay, I changed the default to false. Thanks!

AgustinBettati (approved):
LGTM

PR #2381, opened 2024-07-01
feat: Support new API version for read operation in advanced cluster resource

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-258264

This PR modified the read operation of advanced cluster to start using the new API for all cases in which terraform schema is compatible with the new API (num_shards = 1). This enables defining acceptance tests for symmetric and asymmetric sharded clusters (without updates).

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedcluster/data_source_advanced_cluster.go
+++ b/internal/service/advancedcluster/data_source_advanced_cluster.go
@@ -318,7 +318,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
 	}
 
-	// TODO: update to use connLatest to call below API
+	// TODO: CLOUDP-258711 update to use connLatest to call below API
 	processArgs, _, err := connV220231115.ClustersApi.GetClusterAdvancedConfiguration(ctx, projectID, clusterName).Execute()
 	if err != nil {
 		return diag.FromErr(fmt.Errorf(ErrorAdvancedConfRead, clusterName, err))
--- a/internal/service/advancedcluster/data_source_advanced_cluster.go
+++ b/internal/service/advancedcluster/data_source_advanced_cluster.go
@@ -288,7 +288,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 			return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
 		}
 
-		diags := setCommonSchemaFields(d, convertClusterDescToLatestExcludeRepSpecs(clusterDescOld))
+		diags := setRootFields(d, convertClusterDescToLatestExcludeRepSpecs(clusterDescOld))
 		if diags.HasError() {
 			return diags
 		}
@@ -308,7 +308,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 			return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
 		}
 
-		diags := setCommonSchemaFields(d, clusterDescLatest)
+		diags := setRootFields(d, clusterDescLatest)
 		if diags.HasError() {
 			return diags
 		}
@@ -332,7 +332,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	return nil
 }
 
-func setCommonSchemaFields(d *schema.ResourceData, cluster *admin.ClusterDescription20240710) diag.Diagnostics {
+func setRootFields(d *schema.ResourceData, cluster *admin.ClusterDescription20240710) diag.Diagnostics {
 	clusterName := *cluster.Name
 
 	if err := d.Set("backup_enabled", cluster.GetBackupEnabled()); err != nil {
--- a/internal/service/advancedcluster/data_source_advanced_cluster.go
+++ b/internal/service/advancedcluster/data_source_advanced_cluster.go
@@ -250,8 +250,8 @@ func DataSource() *schema.Resource {
 }
 
 func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	connV2 := meta.(*config.MongoDBClient).AtlasV2
-	connLatest := meta.(*config.MongoDBClient).AtlasV2Preview
+	connV220231115 := meta.(*config.MongoDBClient).AtlasV2
+	connV2 := meta.(*config.MongoDBClient).AtlasV2Preview
 
 	projectID := d.Get("project_id").(string)
 	clusterName := d.Get("name").(string)
@@ -264,7 +264,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	if !useReplicationSpecPerShard {
-		clusterDescOld, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
+		clusterDescOld, resp, err := connV220231115.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
 		if err != nil {
 			if resp != nil {
 				if resp.StatusCode == http.StatusNotFound {
@@ -283,7 +283,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 			return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "disk_size_gb", clusterName, err))
 		}
 
-		replicationSpecs, err = FlattenAdvancedReplicationSpecsOldSDK(ctx, clusterDescOld.GetReplicationSpecs(), clusterDescOld.GetDiskSizeGB(), d.Get("replication_specs").([]any), d, connLatest)
+		replicationSpecs, err = FlattenAdvancedReplicationSpecsOldSDK(ctx, clusterDescOld.GetReplicationSpecs(), clusterDescOld.GetDiskSizeGB(), d.Get("replication_specs").([]any), d, connV2)
 		if err != nil {
 			return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
 		}
@@ -293,7 +293,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 			return diags
 		}
 	} else {
-		clusterDescLatest, resp, err := connLatest.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
+		clusterDescLatest, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
 		if err != nil {
 			if resp != nil && resp.StatusCode == http.StatusNotFound {
 				return nil
@@ -303,7 +303,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 		clusterID = clusterDescLatest.GetId()
 
-		replicationSpecs, err = flattenAdvancedReplicationSpecsDS(ctx, clusterDescLatest.GetReplicationSpecs(), d, connLatest)
+		replicationSpecs, err = flattenAdvancedReplicationSpecsDS(ctx, clusterDescLatest.GetReplicationSpecs(), d, connV2)
 		if err != nil {
 			return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
 		}
@@ -319,7 +319,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	// TODO: update to use connLatest to call below API
-	processArgs, _, err := connV2.ClustersApi.GetClusterAdvancedConfiguration(ctx, projectID, clusterName).Execute()
+	processArgs, _, err := connV220231115.ClustersApi.GetClusterAdvancedConfiguration(ctx, projectID, clusterName).Execute()
 	if err != nil {
 		return diag.FromErr(fmt.Errorf(ErrorAdvancedConfRead, clusterName, err))
 	}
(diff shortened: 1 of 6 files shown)


internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -507,10 +529,14 @@ func FlattenAdvancedReplicationSpecsOldSDK(ctx context.Context, apiObjects []adm
...
 }
 
-func doesAdvancedReplicationSpecMatchAPI(tfObject map[string]any, apiObject *admin20231115.ReplicationSpec) bool {
+func doesAdvancedReplicationSpecMatchAPIOldSDK(tfObject map[string]any, apiObject *admin20231115.ReplicationSpec) bool {
 	return tfObject["id"] == apiObject.GetId() || (tfObject["id"] == nil && tfObject["zone_name"] == apiObject.GetZoneName())
 }
 
+func doesAdvancedReplicationSpecMatchAPI(tfObject map[string]any, apiObject *admin.ReplicationSpec20240710) bool {
+	return tfObject["external_id"] == apiObject.GetId()
+}
```
AgustinBettati, line 538: clear separation of how we match replication spec state with API object depending on API which is called.

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -977,8 +1004,11 @@ func flattenAdvancedReplicationSpec(ctx context.Context, apiObject *admin.Replic
 	}
 
 	tfMap := map[string]any{}
-	tfMap["id"] = apiObject.GetId()
```
AgustinBettati, line 980: when using new api we explicitly don't set `id` attribute so it is clear we are relying on `external_id`.

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -407,10 +407,7 @@ func TestAccClusterAdvancedClusterConfig_replicationSpecsAndShardUpdating(t *tes
 				Config: configMultiZoneWithShards(orgID, projectName, clusterName, 1, 1, false),
 				Check:  checkMultiZoneWithShards(clusterName, 1, 1),
 			},
-			{
-				Config: configMultiZoneWithShards(orgID, projectName, clusterName, 2, 1, false),
-				Check:  checkMultiZoneWithShards(clusterName, 2, 1),
-			},
+			// TODO: updating from single sharded to using old schema should throw an error here
```
AgustinBettati: this second update is no longer working because the first config is compatible with the new schema (`num_shards` = 1) and is populating `external_id` (and not `id`). This is a case where we will throw an error if user tries to update going into the old schema structure (`num_shards` > 1)

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -339,6 +347,7 @@ func schemaSpecs() *schema.Schema {
 		Type:     schema.TypeList,
 		MaxItems: 1,
 		Optional: true,
+		Computed: true,
```
AgustinBettati, line 350: **TLDR**: defining the attribute as optional/computed aligns with how the API behaves (returns value even if not defined), and avoids generating an non empty plan after creation. 

Why is this change needed now? In our existing implementation we were able to match replication specs with api object using `zoneName` ([matcher](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/ac3f1fde3a80a42855766139221cb6e7c2f38cfb/internal/service/advancedcluster/model_advanced_cluster.go#L495)), so even during initial creation if `id` was not present `zoneName` was used to match each replication spec. This matching is being used to only define the value in state if defined in the config ([reference](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/ac3f1fde3a80a42855766139221cb6e7c2f38cfb/internal/service/advancedcluster/model_advanced_cluster.go#L571)). 
Now with the new API, matching is only done with id because zoneName is not unique for identifying each replication spec. During creation no existing state is present, so no matching is done and default is to populate all values in the state. If left only as optional the state in the initial read is populated with all info from API, and if some of the attributes (`analytics_specs`, `electable_specs`, `read_only_specs`) are not defined in the config a non-empty plan is returned.

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -354,6 +363,7 @@ func schemaSpecs() *schema.Schema {
 				"ebs_volume_type": {
 					Type:     schema.TypeString,
 					Optional: true,
+					Computed: true,
```
AgustinBettati, line 366: same case as https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2381/files#r1662130558

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -977,8 +1003,11 @@ func flattenAdvancedReplicationSpec(ctx context.Context, apiObject *admin.Replic
...
 	tfMap["external_id"] = apiObject.GetId()
+
+	// define num_shards for backwards compatibility as this attribute has default value of 1.
+	tfMap["num_shards"] = 1
```
maastha, line 1009: now that we are deciding to add deprecation message, do we still want to do this? or should we wait?

AgustinBettati: Independent if we deprecate or not, I believe we still have to define this value to avoid non-empty plans: the schema defines a default value of 1, meaning existing state has a value of 1 and if value is not set in the read a diff is detected in the optional attribute.

internal/service/advancedcluster/model_advanced_cluster.go
EspenAlbert: Nice. Like how much simpler the newer API is 👍

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -490,21 +500,77 @@ func CreateStateChangeConfig(ctx context.Context, connV2 *admin20231115.APIClien
...
+
+		replicationSpecs, err = FlattenAdvancedReplicationSpecsOldSDK(ctx, clusterOldSDK.GetReplicationSpecs(), clusterOldSDK.GetDiskSizeGB(), d.Get("replication_specs").([]any), d, connV2)
+		if err != nil {
+			return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
+		}
+
+		diags := setResourceRootFields(d, convertClusterDescToLatestExcludeRepSpecs(clusterOldSDK))
+		if diags.HasError() {
+			return diags
+		}
```
EspenAlbert: [nit] Move outside of if condition by using a shared `clusterDesc`?
*update possible also the error check above too

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -513,35 +579,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
 	}
 
 	if err := d.Set("create_date", conversion.TimePtrToStringPtr(cluster.CreateDate)); err != nil {
 		return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "create_date", clusterName, err))
 	}
 
-	if err := d.Set("disk_size_gb", cluster.GetDiskSizeGB()); err != nil {
-		return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "disk_size_gb", clusterName, err))
-	}
+	// TODO: set disk size gb at root level
```
EspenAlbert: remove comment, done at the root level

internal/service/advancedcluster/resource_advanced_cluster_test.go
EspenAlbert: Can this be captured in the test?

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -490,6 +487,46 @@ func TestAccClusterAdvancedClusterConfig_selfManagedShardingIncorrectType(t *tes
...
+		clusterName = acc.RandomClusterName()
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyCluster,
+		Steps: []resource.TestStep{
+			{
+				Config: configShardedNewSchema(orgID, projectName, clusterName, "M30", "M40", 3000, 3000), // TODO: disk iops is failing if value is different
```
EspenAlbert, line 523: Is the comment still relevant?

EspenAlbert (approved):
Great work! Love the explanation comments too.

internal/service/advancedcluster/resource_advanced_cluster_test.go
AgustinBettati: yes, reaching out to AD team here, as per my understanding disk IOPS should also be able to be modified "independently" from other replication spec.

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: Actually this comment is relevant because when calling new API it is not populated. Moved the comment so it is clear for what path this logic is missing.

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: makes sense, adjusted!

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -264,7 +264,7 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	if !useReplicationSpecPerShard {
-		clusterDescOld, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
+		clusterDescOld, resp, err := connV220231115.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
```
marcosuma, line 267: nit for future refactoring: maybe we can use some [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern) here? it could make the code cleaner and less relying on "ifs" 

if you agree, always good to track with TODO and CLOUDP

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -451,6 +451,28 @@ func flattenProcessArgs(p *admin20231115.ClusterDescriptionProcessArgs) []map[st
...
 func FlattenAdvancedReplicationSpecsOldSDK(ctx context.Context, apiObjects []admin20231115.ReplicationSpec, rootDiskSizeGB float64, tfMapObjects []any,
 	d *schema.ResourceData, connV2 *admin.APIClient) ([]map[string]any, error) {
+	replicationSpecFlattener := func(ctx context.Context, sdkModel *admin20231115.ReplicationSpec, tfModel map[string]any, resourceData *schema.ResourceData, client *admin.APIClient) (map[string]any, error) {
+		return flattenAdvancedReplicationSpecOldSDK(ctx, sdkModel, rootDiskSizeGB, tfModel, resourceData, connV2)
+	}
+	return flattenAdvancedReplicationSpecsLogic[admin20231115.ReplicationSpec](ctx, apiObjects, tfMapObjects, d,
+		doesAdvancedReplicationSpecMatchAPIOldSDK, replicationSpecFlattener, connV2)
+}
+
+func flattenAdvancedReplicationSpecs(ctx context.Context, apiObjects []admin.ReplicationSpec20240710, tfMapObjects []any,
```
marcosuma, line 461: same here: `flattenAdvancedReplicationSpecs` and `flattenAdvancedReplicationSpecsOldSDK` suite very well for some sort of design pattern and I believe Strategy is a good one but there might be others

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -451,6 +451,28 @@ func flattenProcessArgs(p *admin20231115.ClusterDescriptionProcessArgs) []map[st
 
 func FlattenAdvancedReplicationSpecsOldSDK(ctx context.Context, apiObjects []admin20231115.ReplicationSpec, rootDiskSizeGB float64, tfMapObjects []any,
```
marcosuma, line 452: do you think we can intensify unit tests around all this logic?

internal/service/advancedcluster/resource_advanced_cluster.go
marcosuma: will this be a breaking change?

marcosuma: in any case it's good to take a note and add some e2e [migration] test around this

internal/service/advancedcluster/data_source_advanced_cluster.go
AgustinBettati: Yes this could be an option for removing nested logic, applies for the read operation of the resource as well. On the other hand I do see that the logic within the if/else is quite simple so could be over complicating the solution. Will create a ticket so we can track and revisit this.

internal/service/advancedcluster/model_advanced_cluster.go
AgustinBettati: In the case of `flattenAdvancedReplicationSpecs` and `FlattenAdvancedReplicationSpecsOldSDK` both share the same logic, just different model and function for matching tf model with API model. This is why I went with the approach of a common function with generics so we don't need to duplicate any of that logic.

internal/service/advancedcluster/model_advanced_cluster.go
AgustinBettati: FlattenAdvancedReplicationSpecsOldSDK does have unit testing, will create ticket so we include for FlattenAdvancedReplicationSpecs as well (prefer to handle in separate PR).

internal/service/advancedcluster/resource_advanced_cluster_test.go
AgustinBettati: Yes, as part of https://jira.mongodb.org/browse/CLOUDP-259828 we will add the validation and assert the error in this test

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: This is not a breaking change: The attributes are now more permissive as state can be populated if not defined in the config, does not impact exiting config/states.

marcosuma (approved):

PR #2466, opened 2024-07-30
feat: Support update transitioning from old to new sharding schema

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-260664

Supports transition of going from usage of numShards > 1 to having split replication specs. 
The main complexity here is that configs with numShards > 1 do not have `replication_specs.*.external_id` defined, so once a users transitions to the new schema we invoke the new API for the PATCH request but need to obtain the correct ids to send in the request.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedcluster/resource_advanced_cluster_test.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster_test.go
@@ -591,6 +591,54 @@ func TestAccClusterAdvancedClusterConfig_asymmetricShardedNewSchema(t *testing.T
 	})
 }
 
+func TestAccClusterAdvancedClusterConfig_shardedTransitionFromOldToNewSchema(t *testing.T) {
+	var (
+		orgID       = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName = acc.RandomProjectName()
+		clusterName = acc.RandomClusterName()
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyCluster,
+		Steps: []resource.TestStep{
+			{
+				Config: configShardedTransitionOldToNewSchema(orgID, projectName, clusterName, false),
+				Check:  checkShardedTransitionOldToNewSchema(false),
+			},
+			{
+				Config: configShardedTransitionOldToNewSchema(orgID, projectName, clusterName, true),
+				Check:  checkShardedTransitionOldToNewSchema(true),
+			},
+		},
+	})
+}
+
+func TestAccClusterAdvancedClusterConfig_geoShardedTransitionFromOldToNewSchema(t *testing.T) {
+	var (
+		orgID       = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName = acc.RandomProjectName()
+		clusterName = acc.RandomClusterName()
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyCluster,
+		Steps: []resource.TestStep{
+			{
+				Config: configGeoShardedTransitionOldToNewSchema(orgID, projectName, clusterName, false),
+				Check:  checkGeoShardedTransitionOldToNewSchema(false),
+			},
+			{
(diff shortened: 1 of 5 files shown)


internal/service/advancedcluster/resource_update_logic.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+	return true
+}
+
+func populateIDValuesUsingNewAPI(ctx context.Context, projectID, clusterName string, connV2CLusterAPI admin.ClustersApi, replicationSpecs *[]admin.ReplicationSpec20250101) (*[]admin.ReplicationSpec20250101, diag.Diagnostics) {
+	if replicationSpecs == nil || len(*replicationSpecs) == 0 {
+		return replicationSpecs, nil
+	}
+	cluster, _, err := connV2CLusterAPI.GetCluster(ctx, projectID, clusterName).Execute()
```
maastha:
```suggestion
	cluster, _, err := connV2ClusterAPI.GetCluster(ctx, projectID, clusterName).Execute()
```

internal/service/advancedcluster/resource_update_logic.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func AddIDsToReplicationSpecs(replicationSpecs []admin.ReplicationSpec20250101, zoneToReplicationSpecsIDs map[string][]string) []admin.ReplicationSpec20250101 {
+	for zoneName, availableIDs := range zoneToReplicationSpecsIDs {
+		var indexOfIDToUse = 0
+		for i := range replicationSpecs {
+			if indexOfIDToUse >= len(availableIDs) {
+				break // all available ids for this zone have been used
+			}
+			if replicationSpecs[i].GetZoneName() == zoneName {
```
maastha, line 45: so essentially we are setting an "id" for the patch request specs based on the zone name irrespective of whether the spec exists or not. And we don't care about setting the id on the "correct" matching spec as well.
In case new specs are added, some specs won't have an id for the same zone name.
Is this correct?

internal/service/advancedcluster/resource_update_logic.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+	"go.mongodb.org/atlas-sdk/v20240530002/admin"
+)
+
+func noIDsPopulatedInReplicationSpecs(replicationSpecs *[]admin.ReplicationSpec20250101) bool {
```
maastha, line 12: Can we add comments about "why" these methods are used? maybe mention which scenarios need these?

Would be helpful to capture API behavior here for future reference

internal/service/advancedcluster/resource_update_logic.go
AgustinBettati: > we are setting an "id" for the patch request specs based on the zone name

Correct, keep in mind that this transition is going from the old shard schema (either sharded or geo-sharded) which can only define symmetric shards for a zone. This is why we can hold this statement: `we don't care about setting the id on the "correct" matching spec as well`. If new specs are added they would not have an id, which is accurate in terms of the update request.

internal/service/advancedcluster/resource_update_logic.go
AgustinBettati: left a comment in the single usage of this function providing context, this function by itself is trivial so thought a comment was more valuable for the place where it is called.

AgustinBettati:
Failing test is in a unrelated test `TestAccClusterAdvancedCluster_singleShardedMultiCloud` due to "OUT_OF_CAPACITY"

internal/service/advancedcluster/resource_update_logic.go
maastha: makes sense, thanks!

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -611,13 +661,28 @@ func checkExists(resourceName string) resource.TestCheckFunc {
...
+		err := getClusterHandlingRetry(ids["project_id"], ids["cluster_name"])
+		if err == nil {
 			return nil
 		}
-		return fmt.Errorf("cluster(%s:%s) does not exist", rs.Primary.Attributes["project_id"], rs.Primary.ID)
+		return fmt.Errorf("cluster(%s:%s) does not exist: %w", rs.Primary.Attributes["project_id"], rs.Primary.ID, err)
 	}
 }
 
+func getClusterHandlingRetry(projectID, clusterName string) error {
```
maastha, line 672: nice

maastha (approved):
LGTM

PR #2476, opened 2024-08-02
fix: Adjusts auto scaling configurations during `mongodbatlas_advanced_cluster` updates

maastha:
## Description

Link to any related issue(s): [CLOUDP-264633](https://jira.mongodb.org/browse/CLOUDP-264633)

## Type of change:

## Required Checklist:


internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -895,6 +897,41 @@ func updateRequest(ctx context.Context, d *schema.ResourceData, projectID, clust
 	return cluster, nil
 }
 
+func syncAutoScalingConfigs(replicationSpecs *[]admin.ReplicationSpec20250101) {
```
AgustinBettati: Would capture in a comment why we need to do this sync on our side. From my understanding this is due to:
- Existing replication specs can have the autoscaling values present in the state with default values even if not defined in the config, while a new replication spec would not have this value present. 
- API expects autoScaling and analyticsAutoScaling aligned cross all region configs in the PATCH request.

Correct me if the above statements are not correct.

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -895,6 +897,41 @@ func updateRequest(ctx context.Context, d *schema.ResourceData, projectID, clust
...
+			}
+			if regionConfig.AnalyticsAutoScaling != nil && defaultAnalyticsAutoScaling == nil {
+				defaultAnalyticsAutoScaling = regionConfig.AnalyticsAutoScaling
+			}
+		}
+	}
+	applyDefaultAutoScaling(replicationSpecs, defaultAutoScaling, defaultAnalyticsAutoScaling)
+}
+
+func applyDefaultAutoScaling(replicationSpecs *[]admin.ReplicationSpec20250101, defaultAutoScaling, defaultAnalyticsAutoScaling *admin.AdvancedAutoScalingSettings) {
```
AgustinBettati: Just to confirm, if a user effectively defines different auto scaling options in 2 separate regions the different values would be included in the patch and the user would face an expected error.
Only downside I see of this approach is that if users defines autoscaling only in one region config, the apply would be successful and might not be transparent that it has been applied to all. Maybe an adjustment in docs could help for this case.

maastha: Yes auto-scaling settings are supposed to be cluster-wide, so if they are different users will get an error OR if auto-scaling is configured for only selected regionConfigs, then also user will get an error. Will update the docs to reflect this.

internal/service/advancedcluster/resource_advanced_cluster.go
maastha: Correct.

AgustinBettati:
Followed up by including documentation adjustment, include clarifying comment in new function, and including unit testing.

I was not able to get an acceptance test to capture the scenario of importing the resource, and making use of that imported state for generating an update. For reference I made use of `ImportState` and `ImportStatePersist`, but this leads to an error as resource is already present in the current state of test (`Terraform is already managing a remote object for
        mongodbatlas_advanced_cluster.test`). Changing the name of the resource before importing seemed like an option, but there is not way to remove the state of the previous resource name.

docs/resources/advanced_cluster.md
```diff
@@ -540,8 +540,8 @@ If you are upgrading a replica set to a sharded cluster, you cannot increase the
...
-* `auto_scaling` - (Optional) Configuration for the Collection of settings that configures auto-scaling information for the cluster. The values for the `auto_scaling` parameter must be the same for every item in the `replication_specs` array. See [below](#auto_scaling)
-* `analytics_auto_scaling` - (Optional) Configuration for the Collection of settings that configures analytics-auto-scaling information for the cluster. The values for the `analytics_auto_scaling` parameter must be the same for every item in the `replication_specs` array. See [below](#analytics_auto_scaling)
+* `auto_scaling` - (Optional) Configuration for the Collection of settings that configures auto-scaling information for the cluster. The values for the `auto_scaling` parameter must be the same for all `region_configs` in all `replication_specs`. See [below](#auto_scaling)
+* `analytics_auto_scaling` - (Optional) Configuration for the Collection of settings that configures analytics-auto-scaling information for the cluster. The values for the `analytics_auto_scaling` parameter must be the same for all `region_configs` in all `replication_specs`. See [below](#analytics_auto_scaling)
```
maastha, line 544: nit: maybe add here that auto-scaling configs are required to be the same for the entire cluster

AgustinBettati: I believe the sentence `The values for the auto_scaling parameter must be the same for all region_configs in all replication_specs.` would cover this, but not sure if you refer to rephrasing this part.

kyuan-mongodb (approved):
LGTM!

docs/resources/advanced_cluster.md
maastha: yeah i basically wanted to call out that auto-scaling configurations are cluster-wide even if users need to set them individually for each rep_spec, but no strong opinion, feel free to leave as is

PR #2501, opened 2024-08-13
feat: Implements `mongodbatlas_stream_processor` resource

oarbusi:
## Description

Implements `mongodbatlas_stream_processor`
- Create operation
   - Can auto start the stream if the `state` attribute is set to `STARTED`. If attribute is not set or set to `CREATED`, the stream processor will only be created. Using any other `state` value will cause the creation of the Stream processor to fail 
- Update Operation
   - Can only Start or Stop the stream, setting the `state` attribute to `STARTED` or `STOPPED`. If any other attribute is modified, the operation will fail.
- Delete operation
- Import operation

Link to any related issue(s): CLOUDP-264839

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/streamprocessor/model.go
+++ b/internal/service/streamprocessor/model.go
@@ -1,0 +1,20 @@
+package streamprocessor
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"go.mongodb.org/atlas-sdk/v20240805001/admin"
+)
+
+func NewTFStreamProcessor(ctx context.Context, apiResp *admin.StreamsProcessor) (*TFStreamProcessorRSModel, diag.Diagnostics) {
+	return &TFStreamProcessorRSModel{}, nil
+}
+
+func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel) (*admin.StreamsProcessor, diag.Diagnostics) {
+	return &admin.StreamsProcessor{}, nil
+}
+
+func NewStreamProcessorWithStats(ctx context.Context, apiResp *admin.StreamsProcessorWithStats) (*TFStreamProcessorRSModel, diag.Diagnostics) {
+	return &TFStreamProcessorRSModel{}, nil
+}
--- a/internal/service/streamprocessor/model.go
+++ b/internal/service/streamprocessor/model.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 	"go.mongodb.org/atlas-sdk/v20240805001/admin"
 )
 
@@ -15,6 +16,14 @@ func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel) 
 	return &admin.StreamsProcessor{}, nil
 }
 
-func NewStreamProcessorWithStats(ctx context.Context, apiResp *admin.StreamsProcessorWithStats) (*TFStreamProcessorRSModel, diag.Diagnostics) {
-	return &TFStreamProcessorRSModel{}, nil
+func NewStreamProcessorWithStats(ctx context.Context, projectID, instanceName string, apiResp *admin.StreamsProcessorWithStats) (*TFStreamProcessorRSModel, diag.Diagnostics) {
+	tfModel := &TFStreamProcessorRSModel{
+		InstanceName:      types.StringValue(instanceName),
+		ProcessorName:     types.StringPointerValue(&apiResp.Name),
+		ProjectID:         types.StringValue(projectID),
+		State:             types.StringPointerValue(&apiResp.State),
+		ProcessorID:       types.StringNull(),
+		ChangeStreamToken: types.StringNull(),
+	}
+	return tfModel, nil
 }
--- a/internal/service/streamprocessor/model.go
+++ b/internal/service/streamprocessor/model.go
@@ -6,16 +6,47 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"go.mongodb.org/atlas-sdk/v20240805001/admin"
 )
 
 func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel) (*admin.StreamsProcessor, diag.Diagnostics) {
-	return &admin.StreamsProcessor{}, nil
+	pipeline, diags := convertPipelineToSdk(plan.Pipeline.ValueStringPointer())
+	if diags != nil {
+		return nil, diags
+	}
+	streamProcessor := &admin.StreamsProcessor{
+		Name:     plan.ProcessorName.ValueStringPointer(),
+		Pipeline: &pipeline,
+	}
+
+	if !plan.Options.IsNull() && !plan.Options.IsUnknown() {
+		optionsModel := &TFOptionsModel{}
+		if diags := plan.Options.As(ctx, optionsModel, basetypes.ObjectAsOptions{}); diags.HasError() {
+			return nil, diags
+		}
+		dlqModel := &TFDlqModel{}
+		if diags := optionsModel.Dlq.As(ctx, dlqModel, basetypes.ObjectAsOptions{}); diags.HasError() {
+			return nil, diags
+		}
+		streamProcessor.Options = &admin.StreamsOptions{
+			Dlq: &admin.StreamsDLQ{
+				Coll:           dlqModel.Coll.ValueStringPointer(),
+				ConnectionName: dlqModel.ConnectionName.ValueStringPointer(),
+				Db:             dlqModel.DB.ValueStringPointer(),
+			},
+		}
+	}
+
+	return streamProcessor, nil
 }
 
 func NewStreamProcessorWithStats(ctx context.Context, projectID, instanceName string, apiResp *admin.StreamsProcessorWithStats, stateOptions types.Object) (*TFStreamProcessorRSModel, diag.Diagnostics) {
 	if apiResp == nil {
 		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("streamProcessor API response is nil", "")}
+	}
+	if stateOptions.IsUnknown() {
+		stateOptions = types.ObjectNull(OptionsObjectType.AttrTypes)
 	}
 	pipelineTF, diags := convertPipelineToTF(apiResp.GetPipeline())
 	if diags.HasError() {
@@ -83,7 +114,7 @@ func convertStatsToTF(stats any) (types.String, diag.Diagnostics) {
 
 func extractChangeStreamTokenFromStats(stats any) (types.String, diag.Diagnostics) {
 	if stats == nil {
-		return types.StringValue("{}"), nil
+		return types.StringNull(), nil
 	}
 	var statsMap map[string]interface{}
 
@@ -97,7 +128,25 @@ func extractChangeStreamTokenFromStats(stats any) (types.String, diag.Diagnostic
 		return types.StringValue(""), diag.Diagnostics{diag.NewErrorDiagnostic("failed to unmarshal stats", err.Error())}
 	}
 
-	changeStreamToken := statsMap["data"].(map[string]interface{})["changeStreamToken"]
+	if data := statsMap["data"]; data != nil {
+		dataMap := data.(map[string]interface{})
+		changeStreamToken, ok := dataMap["changeStreamToken"]
+		if ok {
+			return types.StringValue(changeStreamToken.(string)), nil
+		}
+	}
 
-	return types.StringValue(changeStreamToken.(string)), nil
+	return types.StringNull(), nil
+}
+
+func convertPipelineToSdk(pipeline *string) ([]any, diag.Diagnostics) {
+	if pipeline == nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("Pipeline is required", "")}
+	}
+	var pipelineSliceOfMaps []any
+	err := json.Unmarshal([]byte(*pipeline), &pipelineSliceOfMaps)
+	if err != nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("failed to unmarshal pipeline", err.Error())}
+	}
+	return pipelineSliceOfMaps, nil
 }
--- a/internal/service/streamprocessor/model.go
+++ b/internal/service/streamprocessor/model.go
@@ -11,7 +11,7 @@ import (
 )
 
 func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel) (*admin.StreamsProcessor, diag.Diagnostics) {
-	pipeline, diags := convertPipelineToSdk(plan.Pipeline.ValueStringPointer())
+	pipeline, diags := convertPipelineToSdk(plan.Pipeline.ValueString())
 	if diags != nil {
 		return nil, diags
 	}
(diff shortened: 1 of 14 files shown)


internal/service/streamprocessor/model.go
```diff
@@ -51,3 +111,42 @@ func convertStatsToTF(stats any) (types.String, diag.Diagnostics) {
...
+	statsJSON, err := json.Marshal(stats)
+	if err != nil {
+		return types.StringValue(""), diag.Diagnostics{diag.NewErrorDiagnostic("failed to marshal stats", err.Error())}
+	}
+
+	err = json.Unmarshal(statsJSON, &statsMap)
+	if err != nil {
+		return types.StringValue(""), diag.Diagnostics{diag.NewErrorDiagnostic("failed to unmarshal stats", err.Error())}
+	}
+
```
EspenAlbert: Maybe we can instead cast directly to map[string]interface ? Not sure marshalling and unmarshalling is necessary

internal/service/streamprocessor/model.go
```diff
@@ -51,3 +111,42 @@ func convertStatsToTF(stats any) (types.String, diag.Diagnostics) {
...
+	if data := statsMap["data"]; data != nil {
+		dataMap := data.(map[string]interface{})
```
EspenAlbert: unsafe type cast, use dataMap, isOk?

internal/service/streamprocessor/model.go
```diff
@@ -51,3 +111,42 @@ func convertStatsToTF(stats any) (types.String, diag.Diagnostics) {
...
+			return types.StringValue(changeStreamToken.(string)), nil
+		}
+	}
+
+	return types.StringNull(), nil
+}
+
+func convertPipelineToSdk(pipeline *string) ([]any, diag.Diagnostics) {
+	if pipeline == nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("Pipeline is required", "")}
```
EspenAlbert: since `pipeline` is required in the schema, I don't think this is necesssary. I would use `pipeline.ValueString()` directly

internal/service/streamprocessor/model_test.go
```diff
@@ -38,6 +39,7 @@ var statsExample = `
 	"memoryTrackerBytes": 0.0,
 	"name": "processor1",
 	"ok": 1.0,
+	"data": { "changeStreamToken": "changeStreamTokenValue" },
```
EspenAlbert: It would be nice to have this as an optional parameter and test both with and without data

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,254 @@
...
+	"net/http"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/config"
+	"go.mongodb.org/atlas-sdk/v20240805001/admin"
+)
+
+const streamProcessorName = "stream_processor"
```
EspenAlbert: looks like we are defining this in two places (`data_source.go#StreamProcessorName`). Let's reuse it

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,254 @@
...
+	_, _, err := connV2.StreamsApi.CreateStreamProcessor(ctx, projectID, instanceName, streamProcessorReq).Execute()
+
+	if err != nil {
+		resp.Diagnostics.AddError("error creating resource", err.Error())
+		return
+	}
+
+	streamProcessorParams := &admin.GetStreamProcessorApiParams{
+		GroupId:       projectID,
+		TenantName:    plan.InstanceName.ValueString(),
```
EspenAlbert: re-use `instanceName`

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,254 @@
...
+	streamProcessorResp, err := WaitStateTransition(ctx, streamProcessorParams, connV2.StreamsApi, []string{InitiatingState, CreatingState}, []string{CreatedState, FailedState})
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating stream processor", err.Error())
+	}
+
+	if needsStarting {
+		_, _, err := connV2.StreamsApi.StartStreamProcessorWithParams(ctx,
+			&admin.StartStreamProcessorApiParams{
+				GroupId:       plan.ProjectID.ValueString(),
+				TenantName:    plan.InstanceName.ValueString(),
```
EspenAlbert: same.: re-use

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,254 @@
...
+	processorName = &parts[3]
+
+	return
+}
+
+func updatedStateOnly(plan, state *TFStreamProcessorRSModel) bool {
+	return plan.ProjectID.Equal(state.ProjectID) &&
+		plan.InstanceName.Equal(state.InstanceName) &&
+		plan.ProcessorName.Equal(state.ProcessorName) &&
+		plan.Pipeline.Equal(state.Pipeline) &&
```
EspenAlbert: missing `options`

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,254 @@
...
+		).Execute()
+		if err != nil {
+			resp.Diagnostics.AddError("Error stopping stream processor", err.Error())
+		}
+	default:
+		resp.Diagnostics.AddError("transitions to states other than STARTED or STOPPED are not supported", "")
+		return
+	}
+
+	projectID := plan.ProjectID.ValueString()
```
EspenAlbert: move to top and re-use

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,254 @@
...
+func (r *streamProcessorRS) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var streamProcessorState *TFStreamProcessorRSModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &streamProcessorState)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	connV2 := r.Client.AtlasV2
+	if _, err := connV2.StreamsApi.DeleteStreamProcessor(ctx, streamProcessorState.ProjectID.ValueString(), streamProcessorState.InstanceName.ValueString(), streamProcessorState.ProcessorName.ValueString()).Execute(); err != nil {
+		resp.Diagnostics.AddError("error deleting resource", err.Error())
```
EspenAlbert, line 217: missing 404 case when it has been deleted outside TF

internal/service/streamprocessor/resource_test.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil
+	}
+}
+
+func checkDestroyStreamProcessor(s *terraform.State) error {
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "mongodbatlas_stream_processor" {
+			continue
+		}
+		ids := conversion.DecodeStateID(rs.Primary.ID)
```
EspenAlbert: I don't think we should use `DecodeStateID` here, should use same style as in `checkExists`?

internal/service/streamprocessor/resource_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/acc"
+)
+
+var (
+	resourceName = "mongodbatlas_stream_processor.processor"
+)
+
+func TestAccStreamProcessorRS_basic(t *testing.T) {
```
EspenAlbert, line 19: nice. Can we have another `Acc` test using cluster source and testing the `change_stream_token`?

internal/service/streamprocessor/state_transition_test.go
```diff
@@ -0,0 +1,145 @@
...
+type testCase struct {
+	expectedState *string
+	name          string
+	mockResponses []response
+	desiredStates []string
+	pendingStates []string
+	expectedError bool
+}
+
+func TestStreamProcessorStateTransition(t *testing.T) {
```
EspenAlbert, line 47: Can we add a test case using the `DroppedState`?

internal/service/streamprocessor/state_transition_test.go
```diff
@@ -0,0 +1,145 @@
...
+	testCases := []testCase{
```
EspenAlbert, line 48: [nit] Normally prefer using a map[string]struct{}

EspenAlbert (commented):
Good work. Mainly comments about implementation details and more testing.

internal/service/streamprocessor/resource.go
oarbusi: But this should be in the Read right? In the read the 404 case is handled. I noticed that an error was not being added in case of 404 to the Diagnostics. I will add that.

internal/service/streamprocessor/model.go
oarbusi: done

internal/service/streamprocessor/model.go
oarbusi: Done

internal/service/streamprocessor/resource.go
oarbusi: Done, data_source was using a wrong name, so using this one instead fixes the name. Good catch

internal/service/streamprocessor/resource_test.go
oarbusi: Done!

internal/service/streamprocessor/state_transition_test.go
oarbusi: done

internal/service/streamprocessor/resource.go
EspenAlbert: Sorry, Yes handling it in the read should be enough. I don't think should include that diagnostic message. We don't do that in other resources

internal/service/streamprocessor/state_transition_test.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+		{
+			name: "Error when transitioning to FAILED state",
+			mockResponses: []response{
+				{state: &InitiatingState, statusCode: sc200},
+				{state: &FailedState, statusCode: sc200},
+			},
+			expectedState: nil,
+			expectedError: true,
+			desiredStates: []string{CreatedState, FailedState},
```
EspenAlbert: [nit] I don't think `FailedState` should be part of targets. From the docs of `WaitForStateContext`:

> If the Refresh function returns an error, exit immediately with that error.

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,258 @@
...
+		return
+	}
+
+	streamProcessorParams := &admin.GetStreamProcessorApiParams{
+		GroupId:       projectID,
+		TenantName:    instanceName,
+		ProcessorName: processorName,
+	}
+
+	streamProcessorResp, err := WaitStateTransition(ctx, streamProcessorParams, connV2.StreamsApi, []string{InitiatingState, CreatingState}, []string{CreatedState, FailedState})
```
EspenAlbert: [Same comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2501#discussion_r1722983368)

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,258 @@
...
+
+	connV2 := r.Client.AtlasV2
+
+	projectID := state.ProjectID.ValueString()
+	instanceName := state.InstanceName.ValueString()
+	streamProcessor, apiResp, err := connV2.StreamsApi.GetStreamProcessor(ctx, projectID, instanceName, state.ProcessorName.ValueString()).Execute()
+	if err != nil {
+		if apiResp != nil && apiResp.StatusCode == http.StatusNotFound {
+			resp.State.RemoveResource(ctx)
+			resp.Diagnostics.AddError("resource not found", err.Error())
```
EspenAlbert: This line should be removed. (sorry for earlier confusion)

oarbusi: Removed in [1e09f4e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2501/commits/1e09f4e3d4f54ad618499d2a7a71cb8c1237533c)

internal/service/streamprocessor/state_transition_test.go
oarbusi: Removed in [1e09f4e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2501/commits/1e09f4e3d4f54ad618499d2a7a71cb8c1237533c)

internal/service/streamprocessor/resource.go
oarbusi: Removed in [1e09f4e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2501/commits/1e09f4e3d4f54ad618499d2a7a71cb8c1237533c)

EspenAlbert (approved):
LGTM. Thanks for addressing all the comments!

internal/service/streamprocessor/state_transition.go
```diff
@@ -0,0 +1,61 @@
...
+	FailedState     = "FAILED"
+)
+
+func WaitStateTransition(ctx context.Context, requestParams *admin.GetStreamProcessorApiParams, client admin.StreamsApi, pendingStates, desiredStates []string) (*admin.StreamsProcessorWithStats, error) {
+	stateConf := &retry.StateChangeConf{
+		Pending:    pendingStates,
+		Target:     desiredStates,
+		Refresh:    refreshFunc(ctx, requestParams, client),
+		Timeout:    1 * time.Minute,
+		MinTimeout: 3 * time.Second,
```
AgustinBettati, line 30: curious why these 3 seconds are needed

internal/service/streamprocessor/resource.go
```diff
@@ -0,0 +1,257 @@
...
+		return
+	}
+
+	connV2 := r.Client.AtlasV2
+	pendingStates := []string{CreatedState}
+	desiredState := []string{}
+	projectID := plan.ProjectID.ValueString()
+	instanceName := plan.InstanceName.ValueString()
+	processorName := plan.ProcessorName.ValueString()
+	if plan.State.Equal(state.State) || !updatedStateOnly(&plan, &state) {
```
AgustinBettati: Is there an edge case that is not covered if we leave only the first condition? If update is being called by terraform and state value has not changed, would expect that the update is not valid.

internal/service/streamprocessor/resource_test.go
```diff
@@ -0,0 +1,350 @@
...
+		CheckDestroy:             checkDestroyStreamProcessor,
+		Steps: []resource.TestStep{
+			{
+				Config:      streamProcessorConfigWithSampleConnection(projectID, instanceName, processorName, "STOPPED"),
+				ExpectError: regexp.MustCompile("When creating a stream processor, the only valid states are CREATED and STARTED"),
+			},
+		}})
+}
+
+func streamProcessorConfigWithSampleConnection(projectID, instanceName, processorName, state string) string {
```
AgustinBettati: nit: In recent refactors I see we have gone with approach of removing verbosity in the name of these config and functions, for examples here would be `configWIthSampleConnection` (for reference you can see resources like alert_configuration, backup_compliance_policy, advanced_cluster). We could adjust the scaffolding template as well to avoid confusion moving forward.

AgustinBettati (approved):
LGTM, leaving some doubts and suggestions

kanchana-mongodb (approved):

internal/service/streamprocessor/resource.go
oarbusi: Very good catch, after some fixes in [one of the lasts commits](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2501/commits/0fbbcf8ea1ce4e172a0939f5f764e78fa8e9a20a) this is not needed anymore. Removing it

internal/service/streamprocessor/state_transition.go
oarbusi: Usually state transitions for stream processors are very quick, but I left the 3 seconds to be safe so that the smallest time to wait before refreshes would be enough for the transition to be complete. Let me know if you think this should be changed.

internal/service/streamprocessor/resource_test.go
oarbusi: Done!

PR #2514, opened 2024-08-20
fix: Handles FAILED status as acceptable target state in `mongodbatlas_network_peering` resource

maastha:
## Description

TestAccNetworkRSNetworkPeering_Azure fails often and as per the error in [the recent logs](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/10462943118/job/28974186455) we don't handle FAILED as an acceptable target status. The API is expected to populate `error_state` field with actual failure cause.

I was not able to reproduce/figure out the actual cause of the failure yet but this PR also updates the `TestAccNetworkRSNetworkPeering_Azure` which should help us see the actual error the next time this test fails.

To summarize,
- For users, this PR prevents them from running into `error creating MongoDB Network Peering Connection: unexpected state 'FAILED', wanted target 'AVAILABLE, PENDING_ACCEPTANCE'. `
- For us, updating `TestAccNetworkRSNetworkPeering_Azure` will give details of the error message that is received from the API so we may handle it appropriately.

Link to any related issue(s): [CLOUDP-268949](https://jira.mongodb.org/browse/CLOUDP-268949)

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/networkpeering/resource_network_peering.go
+++ b/internal/service/networkpeering/resource_network_peering.go
@@ -9,14 +9,16 @@ import (
 	"strings"
 	"time"
 
+	"go.mongodb.org/atlas-sdk/v20240805001/admin"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/config"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/networkcontainer"
-	"go.mongodb.org/atlas-sdk/v20240805001/admin"
 )
 
 const (
@@ -247,7 +249,7 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 	stateConf := &retry.StateChangeConf{
 		Pending:    []string{"INITIATING", "FINALIZING", "ADDING_PEER", "WAITING_FOR_USER"},
-		Target:     []string{"AVAILABLE", "PENDING_ACCEPTANCE"},
+		Target:     []string{"AVAILABLE", "PENDING_ACCEPTANCE", "FAILED"},
 		Refresh:    resourceRefreshFunc(ctx, peer.GetId(), projectID, peerRequest.GetContainerId(), conn.NetworkPeeringApi),
 		Timeout:    1 * time.Hour,
 		MinTimeout: 10 * time.Second,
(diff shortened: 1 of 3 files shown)


EspenAlbert:
Not sure I fully understand 😅.
Is `FAILED` state ok?
I would assume users want to get an error if the resource fails to update?

internal/service/networkpeering/resource_network_peering.go
```diff
@@ -470,7 +472,7 @@ func resourceDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 	stateConf := &retry.StateChangeConf{
 		Pending:    []string{"AVAILABLE", "INITIATING", "PENDING_ACCEPTANCE", "FINALIZING", "ADDING_PEER", "WAITING_FOR_USER", "TERMINATING", "DELETING"},
-		Target:     []string{"DELETED"},
+		Target:     []string{"DELETED", "DELETION_FAILED", "FAILED"},
```
maastha, line 475: @EspenAlbert  moving your comment here as it's easier to communicate in thread

>Not sure I fully understand 😅.
Is FAILED state ok?
I would assume users want to get an error if the resource fails to update?

This is something that I intend to bring up in the tech meeting tomorrow. 
But note that currently we do treat FAILED as a target status in existing resources ([resource_privatelink_endpoint_serverless](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/25e10d6bfca3b6ffde6a112475cc35bad7634779/internal/service/privatelinkendpointserverless/resource_privatelink_endpoint_serverless.go#L90), [resource_privatelink_endpoint_service](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/240bd44f4a9a94b20b3a1bd82aedb88327d5a529/internal/service/privatelinkendpointservice/resource_privatelink_endpoint_service.go#L174), push_based_log_export, etc)

To summarize, FAILED is a target status (i.e. we stop making GET/refreshFunc calls once API returns FAILED status) anyway so this PR updates that.
 
What we need to figure out then is how do we handle this failure, do we let users figure that out from fields like error_state/error_message etc OR do we actually throw an error and halt TF execution OR do we just give them a warning about the failure (which is what will be discussed tomorrow)

For more info you can have a look at the [API spec for network_peering](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Network-Peering/operation/createPeeringConnection) APIs. Based on the cloud provider the API may return error details in errorState (Azure) or errorMessage(GCP) fields.

erabil-mdb (approved):
LGTM!

internal/service/networkpeering/resource_network_peering.go
AgustinBettati: With the current changes users would not receive an error right? I believe this is not the best experience as @EspenAlbert mentions.
It is true that `error_state` would now be populated, but if we include failed as a target state I believe we should add additional changes to return the error.

maastha: >With the current changes users would not receive an error right? 

right, users today could receive `error creating MongoDB Network Peering Connection: unexpected state 'FAILED', wanted target 'AVAILABLE, PENDING_ACCEPTANCE'.` error if they run into this status, so this PR fixes that. Note that because of this error, currently it will not be possible to see the value/cause of failure returned in `error_state` which is my main concern.

>I believe this is not the best experience as @EspenAlbert mentions.
It is true that error_state would now be populated, but if we include failed as a target state I believe we should add additional changes to return the error.

yes agreed, as mentioned let's finalize an approach on this in the tech sync today.

I think let's sync in the meeting today and come to a decision, based on that we can close on this PR accordingly.

PR #2519, opened 2024-08-22
test: Adds error and migration tests

EspenAlbert:
## Description

Changes:
- Add migration tests
- Add and Support TestAccStreamProcessor_JSONWhiteSpaceFormat
  - See [Failed test](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/10504792980/job/29101199975?pr=2519)
 - Fix updates to `state` and other attribute simultaneously should raise error
   - See [Failed test](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/10505399331/job/29103069516)
- Fix exit/return when invalid state is set on create, see [e880821af2537f8de20cb05b349510738abc0aef](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2519/commits/e880821af2537f8de20cb05b349510738abc0aef)
- Fix invalid transition from CREATED -> STOPPED [1f06993](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2519/commits/1f06993b9bf1ff2045b83979a4c6263fe26bfbe2)

Link to any related issue(s): CLOUDP-269090

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/fwtypes/json_string.go
+++ b/internal/common/fwtypes/json_string.go
@@ -1,0 +1,141 @@
+package fwtypes
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/schemafunc"
+)
+
+var (
+	_ basetypes.StringTypable                    = (*jsonStringType)(nil)
+	_ basetypes.StringValuable                   = (*JSONString)(nil)
+	_ basetypes.StringValuableWithSemanticEquals = (*JSONString)(nil)
+)
+
+type jsonStringType struct {
+	basetypes.StringType
+}
+
+var (
+	JSONStringType = jsonStringType{}
+)
+
+func (t jsonStringType) Equal(o attr.Type) bool {
+	other, ok := o.(jsonStringType)
+	if !ok {
+		return false
+	}
+	return t.StringType.Equal(other.StringType)
+}
+
+func (t jsonStringType) String() string {
+	return "jsonStringType"
+}
+
+func (t jsonStringType) ValueFromString(_ context.Context, in types.String) (basetypes.StringValuable, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	if in.IsNull() {
+		return JSONStringNull(), diags
+	}
+	if in.IsUnknown() {
+		return JSONStringUnknown(), diags
+	}
+	return JSONString{StringValue: in}, diags
+}
+
+func (t jsonStringType) ValueFromTerraform(ctx context.Context, in tftypes.Value) (attr.Value, error) {
+	attrValue, err := t.StringType.ValueFromTerraform(ctx, in)
+	if err != nil {
+		return nil, err
+	}
+	stringValue, ok := attrValue.(basetypes.StringValue)
+	if !ok {
+		return nil, fmt.Errorf("unexpected value type of %T", attrValue)
+	}
+	stringValuable, diags := t.ValueFromString(ctx, stringValue)
+	if diags.HasError() {
+		return nil, fmt.Errorf("unexpected error converting StringValue to StringValuable: %v", diags)
+	}
+	return stringValuable, nil
+}
+
+func (t jsonStringType) ValueType(context.Context) attr.Value {
+	return JSONString{}
+}
+
+func (t jsonStringType) Validate(ctx context.Context, in tftypes.Value, attrPath path.Path) diag.Diagnostics {
+	var diags diag.Diagnostics
+	if !in.IsKnown() || in.IsNull() {
+		return diags
+	}
+	var value string
+	err := in.As(&value)
+	if err != nil {
+		diags.AddAttributeError(
+			attrPath,
+			"Invalid Terraform Value",
+			"An unexpected error occurred while attempting to convert a Terraform value to a string. "+
+				"This generally is an issue with the provider schema implementation. "+
+				"Please contact the provider developers.\n\n"+
+				"Path: "+attrPath.String()+"\n"+
+				"Error: "+err.Error(),
+		)
+		return diags
+	}
+	if !json.Valid([]byte(value)) {
+		diags.AddAttributeError(
+			attrPath,
+			"Invalid JSON String Value",
+			"A string value was provided that is not valid JSON string format (RFC 7159).\n\n"+
+				"Path: "+attrPath.String()+"\n"+
+				"Given Value: "+value+"\n",
+		)
+		return diags
+	}
+	return diags
+}
+
+func JSONStringNull() JSONString {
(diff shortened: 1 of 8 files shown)


internal/common/schemafunc/json.go
```diff
@@ -1,42 +1,11 @@
...
 import (
-	"context"
 	"encoding/json"
 	"log"
 	"reflect"
-
-	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 )
 
-func DiffSuppressJSON() planmodifier.String {
```
EspenAlbert, line 12: This didn't work as expected. I was trying to add whitespace to the `pipeline` attribute, but still, an update was shown.
Worked when I instead used a `func (r *streamProcessorRS) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse)` on the resource.
However, it is too hacky to include. Considering users adding additional whitespace to be a low chance.

internal/service/streamprocessor/resource_test.go
```diff
@@ -128,25 +154,58 @@ func TestAccStreamProcessor_clusterType(t *testing.T) {
...
+		projectID         = acc.ProjectIDExecution(t)
+		processorName     = "new-processor"
+		instanceName      = acc.RandomName()
+		invalidJSONConfig = connectionConfig{connectionType: connTypeSample, pipelineStepIsSource: true, invalidJSON: true}
 	)
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { acc.PreCheckBasic(t) },
 		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
-		CheckDestroy:             checkDestroyStreamProcessor,
```
EspenAlbert, line 141: no stream processor is created

internal/service/streamprocessor/resource_test.go
```diff
@@ -61,6 +68,25 @@ func TestAccStreamProcessor_basic(t *testing.T) {
...
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		CheckDestroy:             checkDestroyStreamProcessor,
+		Steps: []resource.TestStep{
+			{
+				Config: config(t, projectID, instanceName, processorName, streamprocessor.CreatedState, sampleSrcConfigExtraSpaces, testLogDestConfig),
+				Check:  composeStreamProcessorChecks(projectID, instanceName, processorName, streamprocessor.CreatedState, false, false),
+			},
```
oarbusi, line 89: I don't see the point of testing that a json string with whitespaces is valid. I think this is not provide any value in the test suite and can be distracting

internal/service/streamprocessor/model.go
```diff
@@ -52,14 +53,21 @@ func NewStreamProcessorWithStats(ctx context.Context, projectID, instanceName st
 	if diags.HasError() {
 		return nil, diags
 	}
+	// we need to use pipelinePlan as the state must match the configuration in case the json has extra whitespace or different ordering
```
oarbusi: I don't fully understand the reasoning here. What cases are you thinking here?

EspenAlbert: If we use only the `apiResp.GetPipeline()` it means we are only accepting stream_processors with "standard" JSON format.
E.g., suppose the api returns:
```json
{
    // other attributes removed from exampel
    "pipeline": [{"$source":{"connectionName" :"sample_stream_solar"}},{"$emit":{"connectionName": "__testLog"}}] // notice that this is not a string
}
```
We will always dump this to:
`"[{\"$source\":{\"connectionName\":\"sample_stream_solar\"}},{\"$emit\":{\"connectionName\": \"__testLog\"}}]"`

This means that if the user has any other format that doesn't match our `dump` format they will get an error:
```
╷
│ 
│ When applying changes to mongodbatlas_stream_processor.processor, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new value: .pipeline: was
│ cty.StringVal("[{\"$source\":{\"connectionName\" : \"sample_stream_solar\"}},{\"$emit\":{\"connectionName\": \"__testLog\"}}]"), but now
│ cty.StringVal("[{\"$source\":{\"connectionName\":\"sample_stream_solar\"}},{\"$emit\":{\"connectionName\":\"__testLog\"}}]").
│ 
```
- This example shows only an extra whitespace
- But I can imagine usage of the `templatefile()` in the `*.tf` file reading a human-readable JSON and changing some values

WDYT? Luckily `jsonencode()` does the same as our `dump` function but isn't it convenient to be less strict? 🤔

internal/service/streamprocessor/resource_test.go
EspenAlbert: Please see [comment below for more explanation](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2519#discussion_r1727034943)

internal/common/fwtypes/json_string.go
```diff
@@ -0,0 +1,141 @@
+package fwtypes
```
EspenAlbert, line 1: This is to allow different JSON formats (whitespace/no-whitespace/line breaks etc.)
- See [this comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2519#discussion_r1727034943) for a full example.
- See [the TPF docs](https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/crud#planned-value-does-not-match-config-value) for the `Planned Value does not match Config Value` workaround.

\* More background why `StateFunc` was removed from SDK v2. Similar example for [`iam_role`](https://github.com/hashicorp/terraform-provider-aws/blob/e2afe00d6207f95b837fc35518efba1572da7908/internal/service/iam/role.go#L61)
> as of Terraform 0.12, it’s required that:
> Resources should never set or change an attribute value without the schema Computed flag.
> Resources should always set an attribute state value to the exact configuration value or prior state value, if not null.
> Both of these rules are being violated with that usage of StateFunc and are the main reason a similar functionality cannot be achieved with Plugin Framework 2. If you try to change a configured value in Plugin Framework you’ll get an error like:

internal/common/schemafunc/json.go
EspenAlbert: Not necessary. See above: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2519/files#r1728618273

internal/service/streamprocessor/resource_schema.go
```diff
@@ -30,12 +28,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				MarkdownDescription: "Human-readable label that identifies the stream instance.",
 			},
 			"pipeline": schema.StringAttribute{
-				Validators: []validator.String{
-					validate.StringIsJSON(),
-				},
-				PlanModifiers: []planmodifier.String{
-					schemafunc.DiffSuppressJSON(),
-				},
+				CustomType:          fwtypes.JSONStringType,
```
EspenAlbert, line 31: Finally, a usage of `CustomType`!

internal/service/streamprocessor/resource_test.go
EspenAlbert: ok to resolve @oarbusi ?

internal/service/streamprocessor/model.go
EspenAlbert: A bit outdated, see comment below https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2519#discussion_r1728618273

oarbusi (approved):
LGTM

PR #2528, opened 2024-08-27
feat: Supports change_stream_options_pre_and_post_images_expire_after_seconds in `mongodbatlas_cluster` and `mongodbatlas_advanced_cluster`

oarbusi:
## Description

Supports change_stream_options_pre_and_post_images_expire_after_seconds in mongodbatlas_cluster
Link to any related issue(s): CLOUDP-262696

## Type of change:

## Required Checklist:


internal/service/cluster/model_cluster.go
```diff
@@ -85,6 +85,10 @@ func flattenProcessArgs(p *matlas.ProcessArgs) []map[string]any {
 			"transaction_lifetime_limit_seconds":   p.TransactionLifetimeLimitSeconds,
 		},
 	}
+	if p.ChangeStreamOptionsPreAndPostImagesExpireAfterSeconds != nil {
+		flattenedProcessArgs[0]["change_stream_options_pre_and_post_images_expire_after_seconds"] = p.ChangeStreamOptionsPreAndPostImagesExpireAfterSeconds
```
AgustinBettati, line 89: Curious does the API respond with a default value?

internal/service/cluster/model_cluster.go
```diff
@@ -272,6 +276,14 @@ func expandProcessArgs(d *schema.ResourceData, p map[string]any) *matlas.Process
 		}
 	}
 
+	if _, ok := d.GetOkExists("advanced_configuration.0.change_stream_options_pre_and_post_images_expire_after_seconds"); ok {
+		if changeStreamOptionsSeconds := cast.ToInt64(p["change_stream_options_pre_and_post_images_expire_after_seconds"]); changeStreamOptionsSeconds != 0 {
```
AgustinBettati: q: why do we need to check value is != 0 in this case? If value is not defined by users the attribute will have value -1 right?

internal/service/cluster/resource_cluster_test.go
```diff
@@ -206,6 +208,7 @@ func TestAccCluster_basic_DefaultWriteRead_AdvancedConf(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "advanced_configuration.0.oplog_size_mb", "1000"),
 					resource.TestCheckResourceAttr(resourceName, "advanced_configuration.0.sample_refresh_interval_bi_connector", "310"),
 					resource.TestCheckResourceAttr(resourceName, "advanced_configuration.0.sample_size_bi_connector", "110"),
+					resource.TestCheckResourceAttr(resourceName, "advanced_configuration.0.change_stream_options_pre_and_post_images_expire_after_seconds", "-1"),
```
AgustinBettati, line 211: nice test scenario

docs/resources/cluster.md
```diff
@@ -497,6 +497,7 @@ Include **desired options** within advanced_configuration:
...
 * `transaction_lifetime_limit_seconds` - (Optional) Lifetime, in seconds, of multi-document transactions. Defaults to 60 seconds.
+* `change_stream_options_pre_and_post_images_expire_after_seconds` - (Optional) The minimum pre- and post-image retention time in seconds. This option corresponds to the `changeStreamOptions.preAndPostImages.expireAfterSeconds` cluster parameter. This setting controls the retention policy of change stream pre- and post-images. Pre- and post-images are the versions of a document before and after document modification, respectively.`expireAfterSeconds` controls how long MongoDB retains pre- and post-images. When set to -1 (off), MongoDB uses the default retention policy: pre- and post-images are retained until the corresponding change stream events are removed from the oplog. To set the minimum pre- and post-image retention time, specify an integer value greater than zero. Setting this too low could increase the risk of interrupting Realm sync or triggers processing.
```
AgustinBettati: would make clear in our docs that if value is not defined it will have the value of -1

AgustinBettati (commented):
overall approach LGTM, had some small doubts

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -248,6 +252,11 @@ func SchemaAdvancedConfig() *schema.Schema {
 					Optional: true,
 					Computed: true,
 				},
+				"change_stream_options_pre_and_post_images_expire_after_seconds": {
+					Type:     schema.TypeInt,
+					Optional: true,
+					Default:  -1,
```
EspenAlbert, line 258: What is the story behind the -1 default? 😅

internal/service/cluster/model_cluster.go
oarbusi: As per the documentation, valid values are integer values greater than zero or -1. I used the documentation from the API(which is hidden currently, so not showed in the API docs) but available in the code

internal/service/advancedcluster/model_advanced_cluster.go
oarbusi: Basically it's the value that turns off this option

internal/service/cluster/model_cluster.go
oarbusi: No it doesn't, if it's not set the value returned by the API is null

docs/resources/cluster.md
oarbusi: Let me know if now is better

internal/service/cluster/model_cluster.go
AgustinBettati: Given its an invalid value, should we instead intentional include in the request so that an error is returned? With current implementation not sure how the user experience would be, likely an non empty plan after the apply.

oarbusi: Makes sense, I have changed it so that 0 value gets sent in the request letting the API fail

davidhou17 (approved):

internal/service/advancedcluster/model_advanced_cluster.go
```diff
@@ -446,25 +455,29 @@ func expandBiConnectorConfig(d *schema.ResourceData) *admin.BiConnector {
 	return nil
 }
 
-func flattenProcessArgs(p *admin20240530.ClusterDescriptionProcessArgs) []map[string]any {
-	if p == nil {
+func flattenProcessArgs(p20240530 *admin20240530.ClusterDescriptionProcessArgs, p *admin.ClusterDescriptionProcessArgs20240805) []map[string]any {
+	if p20240530 == nil {
```
EspenAlbert, line 459: names looks strange to me. I would go for a more generic: `pOld` and `pNew`

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -576,12 +580,16 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 		return diag.FromErr(fmt.Errorf(ErrorClusterAdvancedSetting, "replication_specs", clusterName, err))
 	}
 
-	processArgs, _, err := connV220240530.ClustersApi.GetClusterAdvancedConfiguration(ctx, projectID, clusterName).Execute()
+	processArgs20240530, _, err := connV220240530.ClustersApi.GetClusterAdvancedConfiguration(ctx, projectID, clusterName).Execute()
+	if err != nil {
+		return diag.FromErr(fmt.Errorf(errorConfigRead, clusterName, err))
+	}
+	processArgs, _, err := connV2.ClustersApi.GetClusterAdvancedConfiguration(ctx, projectID, clusterName).Execute()
```
EspenAlbert, line 587: Why do we need both `processArgs20240530` and `processArgs`? Will the `GetClusterAdvancedConfiguration` ever return nil?

internal/service/advancedcluster/model_advanced_cluster.go
AgustinBettati, line 258: Just to confirm, was the final agreement with the API team that we send the default of -1 if attribute is not defined?

AgustinBettati (approved):
LGTM, just a final doubt on my end

EspenAlbert (approved):
Nice. Some small nits. Also, got a bit confused about the schema for advanced configuration only lives in the advanced_cluster.

internal/service/advancedcluster/model_advanced_cluster.go
oarbusi: -1 and null(not set) is equivalent in this case for the API. Setting the default value to -1 in the schema is necessary for the case of the user wanting to go back to the default value(which might not always be the same) after having set a different value. This would means that the user would remove the attribute from the TF config, and if there's not default value, it wouldn't be updated in the server because if the value is null in a PATCH, it's not sent

internal/service/advancedcluster/model_advanced_cluster.go
oarbusi: I did it like this to be consistent on how we handle different SDK version and also thinking in the future where we might only use the "new", so the refactor would be cleaner and we would only keep the `p` (there wouldn't be a need to change from `pNew` to `p`)

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: ChangeStreamOptionsPreAndPostImagesExpireAfterSeconds is only available in `processArgs` which is from the latest SDK versrion

oarbusi:
Merging. I ran the local build of the provider and advanced_cluster and cluster works as expected

internal/service/advancedcluster/resource_advanced_cluster.go
EspenAlbert: Can we not remove the `processArgs20240530`?

PR #2575, opened 2024-09-10
chore: Adds `resource_policy` model conversions

EspenAlbert:
## Description

- `resource_policy` model conversion

Link to any related issue(s): CLOUDP-272606

## Type of change:

## Required Checklist:


internal/service/resourcepolicy/model.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	listType, diags := types.ListValueFrom(ctx, PolicyObjectType, tfModels)
+	if diags.HasError() {
+		return nil, diags
+	}
+	return &listType, nil
+}
+
+func NewTFPoliciesModelToSDK(ctx context.Context, input types.List) (*[]admin.ApiAtlasPolicyCreate, diag.Diagnostics) {
+	var tfPolicies []TFPolicyModel
```
EspenAlbert: Both CREATE/UPDATE models only need to convert the policies:

```golang
type ApiAtlasResourcePolicyCreate struct {
	// Human-readable label that describes the atlas resource policy.
	Name *string `json:"name,omitempty"`
	// List of policies that make up the atlas resource policy.
	Policies *[]ApiAtlasPolicyCreate `json:"policies,omitempty"`
}
type ApiAtlasResourcePolicyEdit struct {
	// Human-readable label that describes the atlas resource policy.
	Name *string `json:"name,omitempty"`
	// List of policies that make up the atlas resource policy.
	Policies *[]ApiAtlasPolicyCreate `json:"policies,omitempty"`
}
```

internal/service/resourcepolicy/model.go
```diff
@@ -0,0 +1,85 @@
...
+		LastUpdatedDate:   types.StringPointerValue(conversion.TimePtrToStringPtr(input.LastUpdatedDate)),
+		Name:              types.StringPointerValue(input.Name),
+		OrgID:             types.StringPointerValue(input.OrgId),
+		Policies:          *policies,
+		Version:           types.StringPointerValue(input.Version),
+	}, nil
+}
+
+func NewUserMetadataObjectType(ctx context.Context, input *admin.ApiAtlasUserMetadata) (*types.Object, diag.Diagnostics) {
+	if input == nil {
```
lantoli: is it ok to call NewUserMetadataObjectType with nil input or it should be considered a logical error? same for NewPolicyObjectType

internal/service/resourcepolicy/model.go
```diff
@@ -0,0 +1,85 @@
...
+		LastUpdatedByUser: *lastUpdatedByUser,
+		LastUpdatedDate:   types.StringPointerValue(conversion.TimePtrToStringPtr(input.LastUpdatedDate)),
+		Name:              types.StringPointerValue(input.Name),
+		OrgID:             types.StringPointerValue(input.OrgId),
+		Policies:          *policies,
+		Version:           types.StringPointerValue(input.Version),
+	}, nil
+}
+
+func NewUserMetadataObjectType(ctx context.Context, input *admin.ApiAtlasUserMetadata) (*types.Object, diag.Diagnostics) {
```
lantoli: an alternative would be to have *diag.Diagnostics as param and add errors there

internal/service/resourcepolicy/model_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	for testName, tc := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			SDKModel := parseSDKModel(t, tc.SDKRespJSON)
+			ctx := context.Background()
+			resultModel, diags := resourcepolicy.NewTFResourcePolicyModel(ctx, &SDKModel)
+			assertDiagsOK(t, diags)
+
+			asserter := assert.New(t)
```
oarbusi: why is this needed? wouldn't `assert.WhateverMethod()` work?

lantoli (approved):

internal/service/resourcepolicy/testdata/policy_clusterForbidCloudProvider.json
oarbusi, line 1: this is a nice idea!

oarbusi (approved):
LGTM overall

internal/service/resourcepolicy/model.go
lantoli: curious to see how this function will be used

internal/service/resourcepolicy/testdata/policy_clusterForbidCloudProvider.json
lantoli: I like it too. One small con is that data is not in the test so you need to look at both places to check what is going on in the test

oarbusi: Agreed, probably this approach is more useful in cases where the json object is quite big and has a lot of attributes. For smaller ones, creating directly the SDK object might be better.

internal/service/resourcepolicy/model_test.go
EspenAlbert: We use this in other places too [ex1](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-272606_rp_model_conversions/internal/testutil/acc/attribute_checks_test.go#L62), [ex2](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-272606_rp_model_conversions/internal/testutil/mig/test_case_test.go#L28).
Just makes each of the assert calls a bit simpler

internal/service/resourcepolicy/model.go
EspenAlbert: It is used on line 13+17.
Why is it public you may ask?
1. Makes it easier to unit test
2. When generating the code, we don't know what will be usages. 
3. 🤔 Don't see why not?

WDYT?

internal/service/resourcepolicy/model.go
EspenAlbert: It is ok. Added a test case in ac5d66c7237c050dedbc0c5cd4cb613ee791a19a

internal/service/resourcepolicy/testdata/policy_clusterForbidCloudProvider.json
EspenAlbert: IMO: I prefer this over typing it all out in the test. Especially, if you want to add more test cases.
Also, I think it is easier for auto-generation tools to work with an actual response, and then generate all the constants in the test.
Still, I'm struggling to find a nice way of initializing the TF models. As soon as there is some nesting, readability is terrible IMO

internal/service/resourcepolicy/model.go
```diff
@@ -52,7 +53,8 @@ func NewUserMetadataObjectType(ctx context.Context, input *admin.ApiAtlasUserMet
 }
 
 func NewPolicyObjectType(ctx context.Context, input *[]admin.ApiAtlasPolicy) (*types.List, diag.Diagnostics) {
-	if input == nil {
+	var nilPointer *[]admin.ApiAtlasPolicy
```
EspenAlbert, line 48: Thank you for spotting this @lantoli

internal/service/resourcepolicy/model.go
EspenAlbert: bfa7fa96fbac898a06cfc596d083d7cc75cc8ff1 @lantoli 
It looks cleaner now. Thank you for the suggestion ❤️
*Update: 013da09d2dae4d5e515d63beed1c1461f5d0ac3c even cleaner!
*Update2: d22eaa6ea9fbce9f69fc2855377875f798bc5918 last one?

internal/service/resourcepolicy/model_test.go
```diff
@@ -0,0 +1,189 @@
...
+			userIDUpdate:    userIDUpdate,
+			userNameCreate:  userNameCreate,
+			userNameUpdate:  userNameUpdate,
+			createdDate:     createdDate,
+			lastUpdatedDate: lastUpdatedDate,
+			orgID:           orgID,
+			policyID:        policyID,
+			policyBody:      policyBody,
+			policies0ID:     policies0ID,
+			version:         version,
```
AgustinBettati: q: is there a reason for not having a single expected TF model result here that can be asserted as we do in other unit tests? Currently each property is in the test case and then requires a particular assertion.

EspenAlbert: Good callout.
I started that way to create the test "iteratively" to avoid having to develop the full TFModel.
However, I agree it is cleaner to do a single equal operator.
I refactored in 574799e672009523245228daa67a73537b8609f9 and during the refactoring, I felt the need for helpers, so I created 0313baed184cf1887884d834130bd51285b987c3
WDYT?

AgustinBettati: Nice, thank you. 
I do still feel that `createTFModel` is doing very similar logic to the function we are actual testing `NewTFResourcePolicyModel`, which is why I like the idea of making the creation of the expected TF model as trivial as possible (ideally inlined in the test case directly). There might be some complications to what I am proposing so let me know your thoughts.

internal/service/resourcepolicy/resource_schema.go
```diff
@@ -107,22 +107,29 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
-	Version           types.String `tfsdk:"version"`
+	CreatedByUser     types.Object    `tfsdk:"created_by_user"`
+	CreatedDate       types.String    `tfsdk:"created_date"`
+	ID                types.String    `tfsdk:"id"`
+	LastUpdatedByUser types.Object    `tfsdk:"last_updated_by_user"`
+	LastUpdatedDate   types.String    `tfsdk:"last_updated_date"`
+	Name              types.String    `tfsdk:"name"`
+	OrgID             types.String    `tfsdk:"org_id"`
+	Version           types.String    `tfsdk:"version"`
+	Policies          []TFPolicyModel `tfsdk:"policies"`
```
EspenAlbert, line 118: @AgustinBettati thank you for showing how this is possible!

internal/service/resourcepolicy/model_test.go
EspenAlbert: Thanks for the conversation. Led to improvements.
I choose to keep the `createTFModel` as it simplifies distinguishing the test cases.

internal/service/resourcepolicy/model.go
lantoli: @EspenAlbert i was thinking more about this and I don't think this applies here, as the problem is only when using interfaces, and that's not the case here.  
I think nilPointer and nil will be the same in this case

internal/testutil/unit/fw_model.go
```diff
@@ -0,0 +1,32 @@
...
+	ctx := context.Background()
+	list, diags := types.ListValueFrom(ctx, elementType, tfList)
+	AssertDiagsOK(t, diags)
+	return list
+}
+
+func AssertDiagsOK(t *testing.T, diags diag.Diagnostics) {
+	t.Helper()
+	if diags.HasError() {
+		t.Errorf("unexpected errors found: %s", diags.Errors()[0].Summary())
```
lantoli, line 30: @EspenAlbert what about showing all errors with +v?  e.g.: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/provider/provider_test.go#L67

PR #2584, opened 2024-09-11
feat: Imports implementation and tests for `mongodbatlas_mongodb_employee_access_grant`

lantoli:
## Description

Imports implementation and tests for `mongodbatlas_mongodb_employee_access_grant`

Link to any related issue(s): CLOUDP-272254

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/conversion/import_state.go
+++ b/internal/common/conversion/import_state.go
@@ -1,0 +1,57 @@
+package conversion
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func IDWithProjectIDClusterName(projectID, clusterName string) (string, error) {
+	if err := ValidateProjectID(projectID); err != nil {
+		return "", err
+	}
+	if err := ValidateClusterName(clusterName); err != nil {
+		return "", err
+	}
+	return projectID + "-" + clusterName, nil
+}
+
+func ImportStateProjectIDClusterName(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	parts := strings.Split(req.ID, "-")
+	if len(parts) != 2 {
+		resp.Diagnostics.AddError("invalid import ID", "expected 2 parts with project_id and cluster_name: "+req.ID)
+		return
+	}
+	projectID, clusterName := parts[0], parts[1]
+	if err := ValidateProjectID(projectID); err != nil {
+		resp.Diagnostics.AddError("invalid project_id in import ID", err.Error())
+	}
+	if err := ValidateClusterName(clusterName); err != nil {
+		resp.Diagnostics.AddError("invalid cluster_name in import ID", err.Error())
+	}
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("project_id"), projectID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("cluster_name"), clusterName)...)
+}
+
+func ValidateProjectID(projectID string) error {
+	re := regexp.MustCompile("^([a-f0-9]{24})$")
+	if !re.MatchString(projectID) {
+		return fmt.Errorf("project_id must be a 24 character hex string: %s", projectID)
+	}
+	return nil
+}
+
+func ValidateClusterName(clusterName string) error {
+	re := regexp.MustCompile("^([a-zA-Z0-9][a-zA-Z0-9-]*)?[a-zA-Z0-9]+$")
+	if !re.MatchString(clusterName) || len(clusterName) < 1 || len(clusterName) > 64 {
+		return fmt.Errorf("cluster_name must be a string with length between 1 and 64, starting and ending with an alphanumeric character, and containing only alphanumeric characters and hyphens: %s", clusterName)
+	}
+	return nil
+}
--- a/internal/common/conversion/import_state.go
+++ b/internal/common/conversion/import_state.go
@@ -10,16 +10,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 )
 
-func IDWithProjectIDClusterName(projectID, clusterName string) (string, error) {
-	if err := ValidateProjectID(projectID); err != nil {
-		return "", err
-	}
-	if err := ValidateClusterName(clusterName); err != nil {
-		return "", err
-	}
-	return projectID + "-" + clusterName, nil
-}
-
 func ImportStateProjectIDClusterName(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
 	parts := strings.SplitN(req.ID, "-", 2)
 	if len(parts) != 2 {
--- a/internal/common/conversion/import_state.go
+++ b/internal/common/conversion/import_state.go
@@ -21,7 +21,7 @@ func IDWithProjectIDClusterName(projectID, clusterName string) (string, error) {
 }
 
 func ImportStateProjectIDClusterName(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
-	parts := strings.Split(req.ID, "-")
+	parts := strings.SplitN(req.ID, "-", 2)
 	if len(parts) != 2 {
 		resp.Diagnostics.AddError("invalid import ID", "expected 2 parts with project_id and cluster_name: "+req.ID)
 		return
--- a/internal/common/conversion/import_state_test.go
+++ b/internal/common/conversion/import_state_test.go
@@ -9,10 +9,11 @@ import (
 )
 
 const (
-	projectID          = "123456789012345678901234"
-	projectIDInvalid   = "invalid_project_id"
-	clusterName        = "clusterName"
-	clusterNameInvalid = "_invalidClusterName"
+	projectID           = "123456789012345678901234"
+	projectIDInvalid    = "invalid_project_id"
+	clusterName         = "clusterName"
+	clusterNameWithDash = "cluster-name"
+	clusterNameInvalid  = "_invalidClusterName"
 )
 
 func TestIDWithProjectIDClusterName(t *testing.T) {
@@ -27,6 +28,12 @@ func TestIDWithProjectIDClusterName(t *testing.T) {
 			clusterName: clusterName,
 			expectedID:  projectID + "-" + clusterName,
 		},
+		"valid cluster name with dash": {
+			projectID:   projectID,
+			clusterName: clusterNameWithDash,
+			expectedID:  projectID + "-" + clusterNameWithDash,
+		},
+
 		"invalid project_id": {
 			projectID:           projectIDInvalid,
 			clusterName:         clusterName,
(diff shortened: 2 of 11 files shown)


internal/common/conversion/import_state.go
```diff
@@ -0,0 +1,57 @@
...
+	if err := ValidateProjectID(projectID); err != nil {
+		return "", err
+	}
+	if err := ValidateClusterName(clusterName); err != nil {
+		return "", err
+	}
+	return projectID + "-" + clusterName, nil
+}
+
+func ImportStateProjectIDClusterName(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
lantoli, line 13: trying to make common functions to help with import

internal/common/conversion/import_state.go
```diff
@@ -0,0 +1,57 @@
...
+
+func ValidateProjectID(projectID string) error {
+	re := regexp.MustCompile("^([a-f0-9]{24})$")
+	if !re.MatchString(projectID) {
+		return fmt.Errorf("project_id must be a 24 character hex string: %s", projectID)
+	}
+	return nil
+}
+
+func ValidateClusterName(clusterName string) error {
```
lantoli, line 41: this is the current Atlas cluster name regexp, don't know if the regexp or message are too specific and we should rely on Atlas.  
I feel like it's ok in this case as project_id and cluster_names are broadly used and format is well-know, so we can fail early so user can have a better error message

internal/service/mongodbemployeeaccessgrant/resource_schema.go
```diff
@@ -6,18 +6,32 @@ import (
...
+			"id": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
 			"project_id": schema.StringAttribute{
-				Required:            true,
+				Required: true,
+				PlanModifiers: []planmodifier.String{
```
lantoli, line 18: no need to fail if project_id or cluster_name are changed as it's easy to revert the resource if needed, no data will be lost

internal/testutil/acc/import_state.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+)
+
+func ImportStateIDFuncProjectIDClusterName(resourceName string) resource.ImportStateIdFunc {
```
lantoli, line 11: common helper function so can be used in other resources

internal/common/conversion/import_state_test.go
```diff
@@ -0,0 +1,69 @@
...
+		"valid cluster name with dash": {
+			projectID:   projectID,
+			clusterName: clusterNameWithDash,
+			expectedID:  projectID + "-" + clusterNameWithDash,
+		},
+
+		"invalid project_id": {
+			projectID:           projectIDInvalid,
+			clusterName:         clusterName,
+			expectedErrContains: projectIDInvalid,
```
EspenAlbert: I would prefer to see `invalid project_id`, which is part of the message, instead of the wrong input value. WDYT?

internal/service/mongodbemployeeaccessgrant/data_source_schema.go
```diff
@@ -9,6 +9,9 @@ import (
 func DataSourceSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
 		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
EspenAlbert: Why do we need this?

internal/service/mongodbemployeeaccessgrant/resource.go
```diff
@@ -99,15 +107,21 @@ func (r *rs) createOrUpdate(ctx context.Context, tfModelFunc func(context.Contex
...
 	}
 	connV2 := r.Client.AtlasV2
 	projectID := tfModel.ProjectID.ValueString()
 	clusterName := tfModel.ClusterName.ValueString()
 	if _, _, err := connV2.ClustersApi.GrantMongoDBEmployeeAccess(ctx, projectID, clusterName, atlasReq).Execute(); err != nil {
-		diagnostics.AddError(errorCreate, err.Error())
+		diagnostics.AddError(errorCreateUpdate, err.Error())
+		return
+	}
+	id, err := conversion.IDWithProjectIDClusterName(projectID, clusterName)
```
EspenAlbert: Seems weird to add the `id` attribute for me. 😅
It doesn't align with the API and might be confusing to the user.
Also, it adds some computed code.

internal/common/conversion/import_state_test.go
lantoli: no a strong opinion, i favored to return an error, and having the invalid value inside the error message so you can get a clue about what value was wrong

internal/service/mongodbemployeeaccessgrant/resource.go
lantoli: i got errors in the import state tests when ID was not defined:
```

resource_test.go:26: ImportStateVerify: New resource missing identifier attribute "id", ensure attribute value is properly set or use ImportStateVerifyIdentifierAttribute to choose different attribute
```

so that's why added ID.

then I also added in data source as they share the model

internal/service/mongodbemployeeaccessgrant/data_source_schema.go
lantoli: see below, I added id to RS, and it shares tf model with DS

internal/common/conversion/import_state_test.go
lantoli: at the end I've checked both in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2584/commits/31efc698ab7d7a8ce9ad423950b903ca81c477b0

internal/service/mongodbemployeeaccessgrant/resource.go
EspenAlbert: https://developer.hashicorp.com/terraform/plugin/framework/acctests#no-id-found-in-attributes

> terraform-plugin-testing version 1.5.0 and later no longer require managed resources and data resources to implement the id attribute. 

I remember discussing this with @AgustinBettati not that long ago.

Let me see if I can find the workaround.

EspenAlbert: I think you should be able to use the `ImportStateVerifyIdentifierAttribute` and set it to something else, e.g., `cluster_name`

lantoli: is that ok even if cluster_name isn't necessarily unique? two clusters in different projects could have the same name.   it's just to know if this test fail is just related to the testing framework but not having ID won't affect ImportState functionality when using the resource

EspenAlbert: I believe it is only related to the testing framework. I won't affect ImportState

internal/common/conversion/import_state_test.go
EspenAlbert: Thanks. I like how the new tests reads. Clearly shows how the error message looks to a user :D

internal/service/mongodbemployeeaccessgrant/resource.go
lantoli: @EspenAlbert cc @AgustinBettati you were right, thanks!  removing ID in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2584/commits/d692acc59a8356760d1726e02c9aee3a04142a4d

internal/service/mongodbemployeeaccessgrant/resource.go
```diff
@@ -89,6 +90,7 @@ func (r *rs) Delete(ctx context.Context, req resource.DeleteRequest, resp *resou
 }
 
 func (r *rs) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	conversion.ImportStateProjectIDClusterName(ctx, req, resp)
```
lantoli, line 93: one-line import implementation, ImportStateProjectIDClusterName can be used by any other resource in TPF which depends on project_id and cluster_name

internal/service/mongodbemployeeaccessgrant/resource.go
AgustinBettati: Yes I agree with Espen, we should be able to use ImportStateVerifyIdentifierAttribute defining an attribute that will uniquely match the imported resource within the context of the test.

internal/common/conversion/import_state.go
```diff
@@ -0,0 +1,57 @@
...
+
+func IDWithProjectIDClusterName(projectID, clusterName string) (string, error) {
+	if err := ValidateProjectID(projectID); err != nil {
+		return "", err
+	}
+	if err := ValidateClusterName(clusterName); err != nil {
+		return "", err
+	}
+	return projectID + "-" + clusterName, nil
+}
```
AgustinBettati: Is this function being used? I see it is tested, but not used in the import function of the resource

internal/service/mongodbemployeeaccessgrant/resource.go
lantoli: done, please review again, thx!

internal/common/conversion/import_state.go
lantoli: it's used from acc.ImportStateIDFuncProjectIDClusterName , helper function for testing import

AgustinBettati: ohh I see, in this case could we move it to `testutil/acc/import_state.go` to avoid confusion? In resource implementation we would not need to construct an id string formed by project id + cluster name.

lantoli: it was to have the "direct" and "opposite" functions together so it's clear they go together, but you have a point, moved in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2584/commits/9d4d66f501c360dab4f7e337252bc80a5d1f8280

internal/service/mongodbemployeeaccessgrant/resource.go
EspenAlbert: Nice! Thank you for changing 😁

internal/service/mongodbemployeeaccessgrant/data_source_schema.go
EspenAlbert: Resolved in other conversation

internal/service/mongodbemployeeaccessgrant/resource_schema.go
```diff
@@ -6,18 +6,26 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 )
 
 func ResourceSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
 		Attributes: map[string]schema.Attribute{
 			"project_id": schema.StringAttribute{
-				Required:            true,
+				Required: true,
+				PlanModifiers: []planmodifier.String{
```
EspenAlbert, line 18: good catch

EspenAlbert (approved):
Thanks for addressing the comments and extracting the shared logic 😁

AgustinBettati (approved):
LGTM, thanks for followups

PR #2585, opened 2024-09-11
feat: Implements `resource_policy`

EspenAlbert:
## Description

-  Implements `resource_policy`
- Adds test group to CI
- Adds a new test client (test impacts the organization and can cause other tests to fail)
- Adds new variables to the test runner

Link to any related issue(s): CLOUDP-272019

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -83,9 +83,6 @@ on:
       mongodb_atlas_rp_org_id:
         type: string
         required: true
-      mongodb_atlas_rp_base_url:
-        type: string
-        required: true
       mongodb_atlas_gov_project_owner_id:
         type: string
         required: true
@@ -163,7 +160,6 @@ env:
   MONGODB_ATLAS_GOV_ORG_ID:  ${{ inputs.mongodb_atlas_gov_org_id }}
   MONGODB_ATLAS_RP_PUBLIC_KEY: ${{ secrets.mongodb_atlas_rp_public_key }}
   MONGODB_ATLAS_RP_PRIVATE_KEY: ${{ secrets.mongodb_atlas_rp_private_key }}
-  MONGODB_ATLAS_RP_BASE_URL:  ${{ inputs.mongodb_atlas_rp_base_url }}
   MONGODB_ATLAS_RP_ORG_ID:  ${{ inputs.mongodb_atlas_rp_org_id }}
 
 
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -80,6 +80,12 @@ on:
       mongodb_atlas_gov_base_url:
         type: string
         required: true
+      mongodb_atlas_rp_org_id:
+        type: string
+        required: true
+      mongodb_atlas_rp_base_url:
+        type: string
+        required: true
       mongodb_atlas_gov_project_owner_id:
         type: string
         required: true
@@ -125,6 +131,10 @@ on:
         required: true
       mongodb_atlas_gov_public_key:
         required: true
+      mongodb_atlas_rp_private_key:
+        required: true
+      mongodb_atlas_rp_public_key:
+        required: true
       azure_directory_id:
         required: true
       azure_resource_group_name:
@@ -151,6 +161,10 @@ env:
   MONGODB_ATLAS_GOV_PRIVATE_KEY: ${{ secrets.mongodb_atlas_gov_private_key }}
   MONGODB_ATLAS_GOV_BASE_URL:  ${{ inputs.mongodb_atlas_gov_base_url }}
   MONGODB_ATLAS_GOV_ORG_ID:  ${{ inputs.mongodb_atlas_gov_org_id }}
+  MONGODB_ATLAS_RP_PUBLIC_KEY: ${{ secrets.mongodb_atlas_rp_public_key }}
+  MONGODB_ATLAS_RP_PRIVATE_KEY: ${{ secrets.mongodb_atlas_rp_private_key }}
+  MONGODB_ATLAS_RP_BASE_URL:  ${{ inputs.mongodb_atlas_rp_base_url }}
+  MONGODB_ATLAS_RP_ORG_ID:  ${{ inputs.mongodb_atlas_rp_org_id }}
 
 
 jobs:  
@@ -192,6 +206,7 @@ jobs:
       network: ${{ steps.filter.outputs.network == 'true' || env.mustTrigger == 'true' }}
       project: ${{ steps.filter.outputs.project == 'true' || env.mustTrigger == 'true' }}
       push_based_log_export: ${{ steps.filter.outputs.push_based_log_export == 'true' || env.mustTrigger == 'true' }}
+      resource_policy: ${{ steps.filter.outputs.resource_policy == 'true' || env.mustTrigger == 'true' }}
       search_deployment: ${{ steps.filter.outputs.search_deployment == 'true' || env.mustTrigger == 'true' }}
       search_index: ${{ steps.filter.outputs.search_index == 'true' || env.mustTrigger == 'true' }}
       serverless: ${{ steps.filter.outputs.serverless == 'true' || env.mustTrigger == 'true' }}
@@ -268,7 +283,9 @@ jobs:
             - 'internal/service/projectipaccesslist/*.go'
           push_based_log_export:
             - 'internal/service/pushbasedlogexport/*.go'
-          search_deployment:
+          resource_policy:
+            - 'internal/service/resourcepolicy/*.go'
+            search_deployment:
             - 'internal/service/searchdeployment/*.go'
           search_index:
             - 'internal/service/searchindex/*.go'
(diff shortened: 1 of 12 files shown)


internal/service/resourcepolicy/model_test.go
```diff
@@ -151,3 +150,24 @@ func Test_NewUserMetadataObjectTypeWithNilArg(t *testing.T) {
 	unit.AssertDiagsOK(t, diags)
 	assert.Equal(t, types.ObjectNull(resourcepolicy.UserMetadataObjectType.AttrTypes), obj)
 }
+
+func Test_NewTFPoliciesModelToSDK(t *testing.T) {
```
oarbusi:
```suggestion
func TestNewTFPoliciesModelToSDK(t *testing.T) {
```

internal/service/resourcepolicy/model_test.go
```diff
@@ -151,3 +150,24 @@ func Test_NewUserMetadataObjectTypeWithNilArg(t *testing.T) {
...
+	ctx := context.Background()
+	policies := []resourcepolicy.TFPolicyModel{
+		{
+			Body: types.StringValue("policy1"),
+			ID:   types.StringValue("id1"),
```
oarbusi, line 159: is the ID necessary in the TF config to create a policy? isn't this returned by the API?

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -151,6 +158,9 @@ env:
   MONGODB_ATLAS_GOV_PRIVATE_KEY: ${{ secrets.mongodb_atlas_gov_private_key }}
   MONGODB_ATLAS_GOV_BASE_URL:  ${{ inputs.mongodb_atlas_gov_base_url }}
   MONGODB_ATLAS_GOV_ORG_ID:  ${{ inputs.mongodb_atlas_gov_org_id }}
+  MONGODB_ATLAS_RP_PUBLIC_KEY: ${{ secrets.mongodb_atlas_rp_public_key }}
```
lantoli: knit: if these env vars are only used in one resource tests, better to move it into its test job

internal/service/resourcepolicy/model_test.go
```diff
@@ -151,3 +150,24 @@ func Test_NewUserMetadataObjectTypeWithNilArg(t *testing.T) {
 	unit.AssertDiagsOK(t, diags)
 	assert.Equal(t, types.ObjectNull(resourcepolicy.UserMetadataObjectType.AttrTypes), obj)
 }
+
+func Test_NewTFPoliciesModelToSDK(t *testing.T) {
+	ctx := context.Background()
+	policies := []resourcepolicy.TFPolicyModel{
```
lantoli, line 156: knit: we're using an intermediate name here, not ResourcePolicy or remove it completely, I would favor TFModel, alternatively TFResourcePolicyModel would be "more correct" if you want to keep the resource name, same for other structs

internal/service/resourcepolicy/resource.go
```diff
@@ -0,0 +1,175 @@
...
+	"go.mongodb.org/atlas-sdk/v20240805003/admin"
+)
+
+const resourceName = "resource_policy"
+
+var _ resource.ResourceWithConfigure = &resourcePolicyRS{}
+var _ resource.ResourceWithImportState = &resourcePolicyRS{}
+
+var (
+	errorCreate = "error creating resource"
```
lantoli: can be const instead of var

recommend to include the resource name in the message, e.g.: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/mongodbemployeeaccessgrant/resource.go#L18

internal/service/resourcepolicy/resource.go
```diff
@@ -0,0 +1,175 @@
...
+	orgID := plan.OrgID.ValueString()
+	policies, diags := NewTFPoliciesModelToSDK(ctx, plan.Policies)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	connV2 := r.Client.AtlasV2
+	resourcePolicyAPI := connV2.AtlasResourcePoliciesApi
+	policySDK, _, err := resourcePolicyAPI.CreateAtlasResourcePolicy(ctx, orgID, &admin.ApiAtlasResourcePolicyCreate{
```
lantoli: knit: i suppose it's a matter of preference, I prefer to have the API call in one line, so would create the api request param outside

internal/service/resourcepolicy/resource.go
```diff
@@ -0,0 +1,175 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	orgID := state.OrgID.ValueString()
+	resourcePolicyID := state.ID.ValueString()
+	connV2 := r.Client.AtlasV2
+	policySDK, apiResp, err := connV2.AtlasResourcePoliciesApi.GetAtlasResourcePolicy(ctx, orgID, resourcePolicyID).Execute()
+
+	if err != nil {
+		if apiResp != nil && apiResp.StatusCode == http.StatusNotFound {
```
lantoli, line 84: apiResp is never nil, no need to check it, also you can avoid double nesting doing something like:

```
if  apiResp.StatusCode == http.StatusNotFound {
	resp.State.RemoveResource(ctx)
	return
}

if err != nil {
	resp.Diagnostics.AddError(errorRead, err.Error())
	return
}
```

internal/service/resourcepolicy/resource.go
```diff
@@ -0,0 +1,175 @@
...
+	newResourcePolicyModel, diags := NewTFResourcePolicyModel(ctx, policySDK)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, newResourcePolicyModel)...)
+}
+
+func (r *resourcePolicyRS) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var plan TFResourcePolicyModel
```
lantoli, line 101: in employee access I only used **Plan**: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/mongodbemployeeaccessgrant/resource.go#L73

what's the difference between Plan and State, and when you use each?  I see State is only for ID?

internal/testutil/acc/factory.go
```diff
@@ -61,6 +61,16 @@ func ConnV2UsingGov() *admin.APIClient {
 	return client.(*config.MongoDBClient).AtlasV2
 }
 
+func ConnV2UsingResourcePolicy() *admin.APIClient {
+	cfg := config.Config{
+		PublicKey:  os.Getenv("MONGODB_ATLAS_RP_PUBLIC_KEY"),
```
lantoli: knit alternatively you could change the  GH action to be like this, so it get the credentials in the normal env vars: 
```
resource_policy:
...
      - name: Acceptance Tests
        env:
        MONGODB_ATLAS_PUBLIC_KEY: ${{  ${{ secrets.mongodb_atlas_rp_public_key }} }}
         MONGODB_ATLAS_PRIVATE_KEY: ${{  ${{ secrets.mongodb_atlas_rp_private_key }} }}
          MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
          MONGODB_ATLAS_ENABLE_PREVIEW: "true"
          ACCTEST_PACKAGES: |
            ./internal/service/resourcepolicy
        run: make testacc

```

lantoli: env vars like MONGODB_ATLAS_PUBLIC_KEY defined in the job will override env vars defined for all the workflow

lantoli (approved):

internal/testutil/acc/factory.go
EspenAlbert: Great idea! It avoids any extra provider config or env-vars. You only need to be aware of not running more tests in that job. 
Changed in 77b1c43c3243c8a3e98d4cff79a6af19d7744139

.github/workflows/acceptance-tests-runner.yml
EspenAlbert: changed in 77b1c43c3243c8a3e98d4cff79a6af19d7744139

internal/service/resourcepolicy/resource.go
EspenAlbert: This is taken straight from the scaffolding. https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/tools/scaffold/template/resource.tmpl#L75
No strong preference...

internal/service/resourcepolicy/model_test.go
EspenAlbert: changed in c1b1adb5d07aad9d03b330d76b42d7ef7a5c77a0

internal/service/resourcepolicy/model_test.go
EspenAlbert: Correct, doesn't really matter, shouldn't fail, changed one of the cases to not include `ID` in c1b1adb5d07aad9d03b330d76b42d7ef7a5c77a0

internal/service/resourcepolicy/model_test.go
EspenAlbert: Changed in c1b1adb5d07aad9d03b330d76b42d7ef7a5c77a0 WDYT?

internal/service/resourcepolicy/resource.go
EspenAlbert: Thanks! Made the changes in c1b1adb5d07aad9d03b330d76b42d7ef7a5c77a0

internal/service/resourcepolicy/resource.go
EspenAlbert: inlined the `api`. Created the param inline too. I think it is readable. And another variable named `req` can be confusing

internal/service/resourcepolicy/resource.go
EspenAlbert: changed to Plan in c1b1adb5d07aad9d03b330d76b42d7ef7a5c77a0. Should use the value from state based on the schema anyway

oarbusi (approved):

pierwill (approved):

internal/service/resourcepolicy/resource.go
AgustinBettati: I believe there can be cases where the SDK returns an err without having an apiResp, specifically when failing to create a valid request

PR #2590, opened 2024-09-12
fix: Errors during create should only show one error message and required actions

EspenAlbert:
## Description

Error during create should only show one error message and required actions

Link to any related issue(s): CLOUDP-273149

New error message:
```
╷
│ Error: Error starting stream processor. You need to fix the processor and import the resource or delete it manually and re-run terraform apply.
│ 
│   with mongodbatlas_stream_processor.stream-processor-sample-example,
│   on main.tf line 48, in resource "mongodbatlas_stream_processor" "stream-processor-sample-example":
│   48: resource "mongodbatlas_stream_processor" "stream-processor-sample-example" {
│ 
│ https://cloud-dev.mongodb.com/api/atlas/v2/groups/664619d870c247237f4b86a6/streams/InstanceName/processor/sampleProcessorName:start POST: HTTP 400 Bad Request (Error code:
│ "STREAM_PROCESSOR_GENERIC_ERROR") Detail: Streams Processor with this name (sampleProcessorName) had a problem occur: Found a Time Series collection solar with a timeField ts that doesn't match the
│ $emit.timeField _ts. Reason: Bad Request. Params: [sampleProcessorName Found a Time Series collection solar with a timeField ts that doesn't match the $emit.timeField _ts]
╵
```

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/streamprocessor/resource.go
+++ b/internal/service/streamprocessor/resource.go
@@ -17,10 +17,11 @@ const StreamProcessorName = "stream_processor"
 
 var _ resource.ResourceWithConfigure = &streamProcessorRS{}
 var _ resource.ResourceWithImportState = &streamProcessorRS{}
-var (
-	errorDuringCreateStartActions    = "You need to fix the processor and import the resource or delete it manually and re-run terraform apply."
-	errorDuringCreateStart           = fmt.Sprintf("Error starting stream processor. %s", errorDuringCreateStartActions)
-	errorDuringCreateStartTransition = fmt.Sprintf("Error changing state of stream processor. %s", errorDuringCreateStartActions)
+
+const (
+	errorCreateStartActions    = "You need to fix the processor and import the resource or delete it manually and re-run terraform apply."
+	errorCreateStart           = "Error starting stream processor. " + errorCreateStartActions
+	errorCreateStartTransition = "Error changing state of stream processor. " + errorCreateStartActions
 )
 
 func Resource() resource.Resource {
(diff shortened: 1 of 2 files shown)


oarbusi (approved):

internal/service/streamprocessor/resource.go
```diff
@@ -17,6 +17,11 @@ const StreamProcessorName = "stream_processor"
 
 var _ resource.ResourceWithConfigure = &streamProcessorRS{}
 var _ resource.ResourceWithImportState = &streamProcessorRS{}
+var (
+	errorDuringCreateStartActions    = "You need to fix the processor and import the resource or delete it manually and re-run terraform apply."
```
lantoli: knit: I'd remove "During" to make it more similar to error const in other resources

also I'd use const instead of var, e.g.
```
errorDuringCreateStart           = fmt.Sprintf("Error starting stream processor. %s", errorDuringCreateStartActions)
```

can be changed to:

```
errorDuringCreateStart   =  "Error starting stream processor. " +  errorDuringCreateStartActions
```

internal/service/streamprocessor/resource.go
```diff
@@ -187,6 +195,7 @@ func (r *streamProcessorRS) Update(ctx context.Context, req resource.UpdateReque
 		).Execute()
 		if err != nil {
 			resp.Diagnostics.AddError("Error stopping stream processor", err.Error())
+			return
```
lantoli, line 199: would be nice to have some linter or something to detect when we're not returning after adding error

lantoli (approved):

internal/service/streamprocessor/resource.go
EspenAlbert: Thanks. Changed in 58707c352efb05185f548944ffab927eee535bf0

internal/service/streamprocessor/resource.go
EspenAlbert: An alternative is creating a helper function that could be called with

```golang
if err != nil {
  return add_error_message(resp, err, "Error stopping stream_processor")
  }
```
It wouldn't return anything but makes sure the error is added and can be called with `return` directly to avoid an extra `return` statement on the next line. WDYT?

lantoli: that could work.   one pro about having param diagnostics *diag.Diagnostics is that you can potentially add multiple errrors in the same function,  and it's quite direct from RS functions to do if diagnostics.HasError(). you can pass &resp.Diagnostics as the param from those functions

EspenAlbert: Good idea! I'll keep this PR as is, but I will add it to the model conversion PR!

pierwill (approved):

Issue #2662, opened 2024-10-07
[Bug]: mongodbatlas_team removes member either the change doesn't show it


lolejar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Provider Version

1.20.0

### Terraform Version

1.9.7

### Terraform Edition

Terraform Open Source (OSS)

### Current Behavior

This started happening since Thursday 03.10.2024 without anychanges to IaC.
I have a IaC then runs each 10 minutes (for more over than 2 years) that synchronizes access to MongoDB teams from external management system.

When I run terraform apply / plan it shows me change in usernames as expected. When I terraform apply, usernames are added to state, however they are not being created in MongoDB Team AND the execution of provider fails with message it was not possible to remove user as it CANNOT_REMOVE_LAST_USER_FROM_TEAM. 

That user it tries to remove is the one that I am aware cannot be removed and is included in default_team_members and is always member. To me it looks like Terraform wants to remove all usernames and add them again. Plan shows only additions, no removals.

### Terraform configuration to reproduce the issue

```hcl
resource "mongodbatlas_team" "managers" {
  for_each  = local.mongodb_projects__map
  org_id    = var.organization_id
  name      = "${each.value.project_name} - Managers"
  usernames = concat(sort(module.default_team_members.emails), sort(module.managers_emails["${each.value.resource}"].emails))

  depends_on = [
    mongodbatlas_org_invitation.organization_owners,
    mongodbatlas_org_invitation.organization_members,
    mongodbatlas_org_invitation.organization_readers
  ]
}
```

### Steps To Reproduce

1. I run apply
2. Users for teams are fetched from external system and their emails are being resolved
3. Terraform provider detect changes between state and MongoDB
4. It shows only additions to the team (1 team member always present - 1 unchanged element hidden )
5. It starts applying changes 
6. It fails that is cannon delete that 1 team member always present
7. Remaining users are added to state
8. Remaining users are not added to MongoDB Team
9. Only member of MongoDB Team is default team member

### Logs

```txt
# mongodbatlas_team.managers["team_key"] will be updated in-place
  ~ resource "mongodbatlas_team" "managers" {
        id        = "<redacted>"
        name      = "<redacted>"
      ~ usernames = [
          + "xxx@aaa.com",
          + "yyy@aaa.com",
          + "zzz@aaa.com",
            # (1 unchanged element hidden) 
        ]
        # (2 unchanged attributes hidden)
    }

[...]

PR #2669, opened 2024-10-09
fix: Support update of usernames in `mongodbatlas_team` resource by ensuring team is never empty

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-277323 (https://github.com/mongodb/terraform-provider-mongodbatlas/issues/2662), HELP-65735

**Context**: A new validation has been added in the API which verifies a team is never empty. Our existing logic for updating the `usernames` attribute removed all users and then added the new ones, but this now fails before removing the last user.

**Solution**: Logic is very similar to [CFN implementation](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/teams/cmd/resource/team-user/team_user.go#L26), which only add and removes the changed users. Some adjustments where made:
- ensures that additions are done before removals, if all usernames are different the team will not be empty at any point.
- for unit testing mocking is done directly with SDK mocks and no intermediate service interface.

Both of these adjustments will need to be moved to CFN, and also brings the question of how we could have this common code in a single repository and avoid duplicating logic/testing in both places.

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -38,9 +38,6 @@ on:
       mongodb_atlas_teams_ids:
         type: string
         required: true  
-      mongodb_atlas_username:
-        type: string
-        required: true 
       azure_atlas_app_id:
         type: string
         required: true
@@ -503,7 +500,8 @@ jobs:
       - name: Acceptance Tests
         env:
           MONGODB_ATLAS_PROJECT_OWNER_ID: ${{ inputs.mongodb_atlas_project_owner_id }}
-          MONGODB_ATLAS_USERNAME: ${{ inputs.mongodb_atlas_username }}
+          MONGODB_ATLAS_USERNAME: ${{ vars.MONGODB_ATLAS_USERNAME }}
+          MONGODB_ATLAS_USERNAME_2: ${{ vars.MONGODB_ATLAS_USERNAME_2 }}
           AZURE_ATLAS_APP_ID: ${{ inputs.azure_atlas_app_id }}
           AZURE_SERVICE_PRINCIPAL_ID: ${{ inputs.azure_service_principal_id }}
           AZURE_TENANT_ID: ${{ inputs.azure_tenant_id }}
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -97,7 +97,6 @@ jobs:
       mongodb_atlas_base_url: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_BASE_URL_QA || vars.MONGODB_ATLAS_BASE_URL }}
       mongodb_atlas_project_owner_id: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_PROJECT_OWNER_ID_QA || vars.MONGODB_ATLAS_PROJECT_OWNER_ID }}
       mongodb_atlas_teams_ids: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_TEAMS_IDS_QA || vars.MONGODB_ATLAS_TEAMS_IDS }}
-      mongodb_atlas_username: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_USERNAME_CLOUD_QA || vars.MONGODB_ATLAS_USERNAME_CLOUD_DEV }}
       azure_atlas_app_id: ${{ inputs.atlas_cloud_env == 'qa' && vars.AZURE_ATLAS_APP_ID_QA || vars.AZURE_ATLAS_APP_ID }}
       azure_service_principal_id: ${{ inputs.atlas_cloud_env == 'qa' && vars.AZURE_SERVICE_PRINCIPAL_ID_QA || vars.AZURE_SERVICE_PRINCIPAL_ID }}
       azure_tenant_id: ${{ inputs.atlas_cloud_env == 'qa' && vars.AZURE_TENANT_ID_QA || vars.AZURE_TENANT_ID }}
--- a/internal/service/team/update_user.go
+++ b/internal/service/team/update_user.go
@@ -58,10 +58,6 @@ func GetChangesForTeamUsers(currentUsers, newUsers []admin.CloudAppUser) (toAdd,
 	currentUsersSet := InitUserSet(currentUsers)
 	newUsersSet := InitUserSet(newUsers)
 
-	// Create two arrays to store the elements to be added and deleted
-	toAdd = []string{}
-	toDelete = []string{}
-
 	// Iterate over the elements in B and add them to the toAdd array if they are not in A
 	for elem := range newUsersSet {
 		if _, ok := currentUsersSet[elem]; !ok {
@@ -76,7 +72,6 @@ func GetChangesForTeamUsers(currentUsers, newUsers []admin.CloudAppUser) (toAdd,
 		}
 	}
 
-	// Return the two arrays
 	return toAdd, toDelete, nil
 }
 
--- a/internal/service/team/update_user.go
+++ b/internal/service/team/update_user.go
@@ -1,0 +1,89 @@
+package team
+
+import (
+	"context"
+
+	"go.mongodb.org/atlas-sdk/v20240805004/admin"
+)
+
+func UpdateTeamUsers(teamsAPI admin.TeamsApi, usersAPI admin.MongoDBCloudUsersApi, existingTeamUsers *admin.PaginatedApiAppUser, newUsernames []string, orgID, teamID string) error {
+	validNewUsers, err := ValidateUsernames(usersAPI, newUsernames)
+	if err != nil {
+		return err
+	}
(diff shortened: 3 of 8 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -97,7 +97,6 @@ jobs:
       mongodb_atlas_base_url: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_BASE_URL_QA || vars.MONGODB_ATLAS_BASE_URL }}
       mongodb_atlas_project_owner_id: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_PROJECT_OWNER_ID_QA || vars.MONGODB_ATLAS_PROJECT_OWNER_ID }}
       mongodb_atlas_teams_ids: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_TEAMS_IDS_QA || vars.MONGODB_ATLAS_TEAMS_IDS }}
-      mongodb_atlas_username: ${{ inputs.atlas_cloud_env == 'qa' && vars.MONGODB_ATLAS_USERNAME_CLOUD_QA || vars.MONGODB_ATLAS_USERNAME_CLOUD_DEV }}
```
AgustinBettati, line 100: noticed there is not need to have different values for dev/qa, unified to single GH variable MONGODB_ATLAS_USERNAME. Checked [tests are passing in qa](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/11255205368) as well.
Once pr is merged will remove `MONGODB_ATLAS_USERNAME_CLOUD_DEV` and `MONGODB_ATLAS_USERNAME_CLOUD_QA`

internal/service/team/update_user_test.go
```diff
@@ -0,0 +1,159 @@
...
+			mockFuncExpectations: func(mockTeamsApi *mockadmin.TeamsApi, mockUsersApi *mockadmin.MongoDBCloudUsersApi) {
+				mockValidUser2 := mockadmin.NewMongoDBCloudUsersApi(t)
+				mockUsersApi.EXPECT().GetUserByUsername(mock.Anything, validuser1).Return(admin.GetUserByUsernameApiRequest{ApiService: mockValidUser2})
+				mockValidUser2.EXPECT().GetUserByUsernameExecute(mock.Anything).Return(&admin.CloudAppUser{Id: &validuser1}, nil, nil)
+
+				addCall := mockTeamsApi.EXPECT().AddTeamUser(mock.Anything, mock.Anything, mock.Anything, &[]admin.AddUserToTeam{{Id: validuser1}}).Return(admin.AddTeamUserApiRequest{ApiService: mockTeamsApi})
+				mockTeamsApi.EXPECT().AddTeamUserExecute(mock.Anything).Return(nil, nil, nil)
+
+				removeCall := mockTeamsApi.EXPECT().RemoveTeamUser(mock.Anything, mock.Anything, mock.Anything, validuser2).Return(admin.RemoveTeamUserApiRequest{ApiService: mockTeamsApi})
+				removeCall.NotBefore(addCall.Call) // Ensures new additions are made before removing
```
AgustinBettati, line 142: ensures no regression in future changes (apart from the new acceptance test)

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -38,9 +38,6 @@ on:
       mongodb_atlas_teams_ids:
         type: string
         required: true  
-      mongodb_atlas_username:
```
lantoli, line 41: why aren't we using workflow params for these 2 vars?

internal/service/team/update_user.go
```diff
@@ -0,0 +1,89 @@
+package team
```
lantoli, line 1: knit: not sure about naming the file specific for update users, instead of something more general

internal/service/team/update_user.go
```diff
@@ -0,0 +1,89 @@
...
+	currentUsersSet := InitUserSet(currentUsers)
+	newUsersSet := InitUserSet(newUsers)
+
+	// Create two arrays to store the elements to be added and deleted
+	toAdd = []string{}
+	toDelete = []string{}
+
+	// Iterate over the elements in B and add them to the toAdd array if they are not in A
+	for elem := range newUsersSet {
+		if _, ok := currentUsersSet[elem]; !ok {
```
lantoli, line 63: if you do  map[string]bool rathen than map[string]interface{}  then the code would be simpler as you can just do: 
```
if  !currentUsersSet[elem] {
 toAdd = ...
}
```
as not found elements will return false (nil value for bool).

same for toDelete

there is an optimization to do  map[string]struct{} (note it's struct not interface) but I don't think it's worth it in this case

internal/service/team/update_user.go
```diff
@@ -0,0 +1,89 @@
...
+	return validUsers, nil
+}
+
+func GetChangesForTeamUsers(currentUsers, newUsers []admin.CloudAppUser) (toAdd, toDelete []string, err error) {
+	// Create two sets to store the elements in A and B
+	currentUsersSet := InitUserSet(currentUsers)
+	newUsersSet := InitUserSet(newUsers)
+
+	// Create two arrays to store the elements to be added and deleted
+	toAdd = []string{}
```
lantoli: i don't think you need to do this as you're using append later and will work with nil as first slice

internal/testutil/acc/pre_check.go
```diff
@@ -82,6 +82,14 @@ func PreCheckAtlasUsername(tb testing.TB) {
 	}
 }
 
+func PreCheckAtlasUsernames(tb testing.TB) {
+	tb.Helper()
+	PreCheckAtlasUsername(tb)
+	if os.Getenv("MONGODB_ATLAS_USERNAME_2") == "" {
```
lantoli, line 88: should we also check MONGODB_ATLAS_USERNAME ?

internal/service/team/update_user.go
lantoli: alternatively you can just use slices.Contains() instead of creating the maps, it's a bit less efficient but I don't think it's significant here.

lantoli (approved):

internal/service/team/update_user.go
```diff
@@ -0,0 +1,89 @@
...
+	toDelete = []string{}
+
+	// Iterate over the elements in B and add them to the toAdd array if they are not in A
+	for elem := range newUsersSet {
+		if _, ok := currentUsersSet[elem]; !ok {
+			toAdd = append(toAdd, elem)
+		}
+	}
+
+	// Iterate over the elements in A and add them to the toDelete array if they are not in B
```
lantoli, line 68: we can also say that we're doing a "set diff" (elements in a set that are not in the other set)

internal/service/team/update_user.go
```diff
@@ -0,0 +1,89 @@
+package team
+
+import (
+	"context"
+
+	"go.mongodb.org/atlas-sdk/v20240805004/admin"
+)
+
+func UpdateTeamUsers(teamsAPI admin.TeamsApi, usersAPI admin.MongoDBCloudUsersApi, existingTeamUsers *admin.PaginatedApiAppUser, newUsernames []string, orgID, teamID string) error {
```
lantoli, line 9: should we check that at least one user remains or we leave the error to Atlas?

marcosuma (approved):

kyuan-mongodb (approved):
LGTM!

internal/service/team/update_user.go
AgustinBettati: dont have a strong opinion, I believe this is fairly particular case where we have a lot of business logic for handling the update of 1 attributes given it consumes multiple endpoints

internal/service/team/update_user.go
AgustinBettati: Interesting, I actually thought !currentUsersSet[elem] would panic if not present but that is not the case. I will say that I see current approach more readable as you know it does not rely on the value of the map, and only its presence.

internal/service/team/update_user.go
AgustinBettati: nice, adjusted 👍

internal/testutil/acc/pre_check.go
AgustinBettati: yes it is check that env variable by calling `PreCheckAtlasUsername` in the line above

internal/service/team/update_user.go
AgustinBettati: would leave the error to atlas so implementation is flexible to future changes (e.g. API deciding to support empty teams again)

internal/service/team/update_user.go
lantoli: I think that map[string]bool and use if mymap[key] is more Go idiomatic but ok to leave it as it is

lantoli: knit: as you're not using the value in map, only key, the idiomatic way would be map[string]struct{} meaning that you're not storing any value as oppossed to interface{} (any) that means that you can store any value.

AgustinBettati: did some reading and agree that map[string]bool seems like the common approach. thanks for inputs, adjusting in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2675.

PR #2673, opened 2024-10-09
chore: PoC - Model generation - support primitive types at root level

maastha:
## Description

PoC - Model generation - support primitive types at root level

Link to any related issue(s): [CLOUDP-275829](https://jira.mongodb.org/browse/CLOUDP-275829)

## Type of change:

## Required Checklist:

### DIFF
--- a/tools/codegen/codespec/api_spec.go
+++ b/tools/codegen/codespec/api_spec.go
@@ -15,8 +15,8 @@ var (
 )
 
 // This function only builds the schema from a proxy and returns the basic type and format without handling oneOf, anyOf, allOf, or nullable types.
-func BuildSchema(proxy *base.SchemaProxy) (*OASSchema, error) {
-	resp := &OASSchema{}
+func BuildSchema(proxy *base.SchemaProxy) (*APISpecSchema, error) {
+	resp := &APISpecSchema{}
 
 	schema, err := proxy.BuildSchema()
 	if err != nil {
@@ -34,7 +34,7 @@ func BuildSchema(proxy *base.SchemaProxy) (*OASSchema, error) {
 	return resp, nil
 }
 
-func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType]) (*OASSchema, error) {
+func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType]) (*APISpecSchema, error) {
 	if mediaTypes == nil {
 		return nil, errSchemaNotFound
 	}
@@ -64,7 +64,7 @@ func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType])
 	return nil, errSchemaNotFound
 }
 
-func buildSchemaFromRequest(op *high.Operation) (*OASSchema, error) {
+func buildSchemaFromRequest(op *high.Operation) (*APISpecSchema, error) {
 	if op == nil || op.RequestBody == nil || op.RequestBody.Content == nil || op.RequestBody.Content.Len() == 0 {
 		return nil, errSchemaNotFound
 	}
@@ -72,7 +72,7 @@ func buildSchemaFromRequest(op *high.Operation) (*OASSchema, error) {
 	return getSchemaFromMediaType(op.RequestBody.Content)
 }
 
-func buildSchemaFromResponse(op *high.Operation) (*OASSchema, error) {
+func buildSchemaFromResponse(op *high.Operation) (*APISpecSchema, error) {
 	if op == nil || op.Responses == nil || op.Responses.Codes == nil || op.Responses.Codes.Len() == 0 {
 		return nil, errSchemaNotFound
 	}
--- a/tools/codegen/codespec/api_spec.go
+++ b/tools/codegen/codespec/api_spec.go
@@ -24,7 +24,7 @@ func BuildSchema(proxy *base.SchemaProxy) (*APISpecSchema, error) {
 	}
 
 	if len(schema.Type) == 0 {
-		return nil, errInvalidSchema
+		return nil, fmt.Errorf("invalid schema. no values for schema.Type found")
 	}
 
 	resp.Type = schema.Type[0]
--- a/tools/codegen/codespec/api_spec.go
+++ b/tools/codegen/codespec/api_spec.go
@@ -38,17 +38,6 @@ func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType])
 		return nil, errSchemaNotFound
 	}
 
-	// hard-coding a default API version for now, if "application/vnd.atlas.2023-01-01+json" version is not found
-	// then sortedMediaTypes below will read the available media type
-	jsonMediaType, ok := mediaTypes.Get("application/vnd.atlas.2023-01-01+json")
-	if ok && jsonMediaType.Schema != nil {
-		s, err := BuildSchema(jsonMediaType.Schema)
-		if err != nil {
-			return nil, err
-		}
-		return s, nil
-	}
-
 	sortedMediaTypes := orderedmap.SortAlpha(mediaTypes)
 	for pair := range orderedmap.Iterate(context.Background(), sortedMediaTypes) {
 		mediaType := pair.Value()
--- a/tools/codegen/codespec/api_spec.go
+++ b/tools/codegen/codespec/api_spec.go
@@ -29,7 +29,6 @@ func BuildSchema(proxy *base.SchemaProxy) (*APISpecSchema, error) {
 
 	resp.Type = schema.Type[0]
 	resp.Schema = schema
-	resp.Format = resp.Schema.Format
 
 	return resp, nil
 }
@@ -50,7 +49,7 @@ func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType])
 	}
 
 	sortedMediaTypes := orderedmap.SortAlpha(mediaTypes)
-	for pair := range orderedmap.Iterate(context.TODO(), sortedMediaTypes) {
+	for pair := range orderedmap.Iterate(context.Background(), sortedMediaTypes) {
 		mediaType := pair.Value()
 		if mediaType.Schema != nil {
 			s, err := BuildSchema(mediaType.Schema)
@@ -88,7 +87,7 @@ func buildSchemaFromResponse(op *high.Operation) (*APISpecSchema, error) {
 	}
 
 	sortedCodes := orderedmap.SortAlpha(op.Responses.Codes)
-	for pair := range orderedmap.Iterate(context.TODO(), sortedCodes) {
+	for pair := range orderedmap.Iterate(context.Background(), sortedCodes) {
 		responseCode := pair.Value()
 		statusCode, err := strconv.Atoi(pair.Key())
 		if err != nil {
--- a/tools/codegen/codespec/api_spec.go
+++ b/tools/codegen/codespec/api_spec.go
@@ -38,7 +38,8 @@ func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType])
 		return nil, errSchemaNotFound
 	}
 
-	// hard-coding API version for now, this should be dynamically handled to perhaps use the latest API version
+	// hard-coding a default API version for now, if "application/vnd.atlas.2023-01-01+json" version is not found
+	// then sortedMediaTypes below will read the available media type
 	jsonMediaType, ok := mediaTypes.Get("application/vnd.atlas.2023-01-01+json")
 	if ok && jsonMediaType.Schema != nil {
 		s, err := BuildSchema(jsonMediaType.Schema)
--- a/tools/codegen/codespec/api_spec_schema.go
+++ b/tools/codegen/codespec/api_spec_schema.go
@@ -12,16 +12,16 @@ var errInvalidSchema = fmt.Errorf("invalid schema")
 type APISpecSchema struct {
 	Schema *base.Schema
 	Type   string
-	Format string
 }
 
 type APISpecResource struct {
-	Description      *string
-	CreateOp         *high.Operation
-	ReadOp           *high.Operation
-	UpdateOp         *high.Operation
-	DeleteOp         *high.Operation
-	CommonParameters []*high.Parameter
+	Description        *string
+	DeprecationMessage *string
+	CreateOp           *high.Operation
+	ReadOp             *high.Operation
+	UpdateOp           *high.Operation
+	DeleteOp           *high.Operation
+	CommonParameters   []*high.Parameter
 }
 
 func (s *APISpecSchema) GetComputability(name string) ComputedOptionalRequired {
@@ -31,7 +31,7 @@ func (s *APISpecSchema) GetComputability(name string) ComputedOptionalRequired {
 		}
 	}
 
-	return ComputedOptional
+	return Optional
 }
 
 func (s *APISpecSchema) GetDeprecationMessage() *string {
@@ -53,7 +53,7 @@ func (s *APISpecSchema) GetDescription() *string {
 }
 
 func (s *APISpecSchema) IsSensitive() *bool {
-	isSensitive := s.Format == OASFormatPassword
+	isSensitive := s.Schema.Format == OASFormatPassword
 
 	if !isSensitive {
 		return nil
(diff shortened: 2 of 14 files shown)


tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,27 +2,233 @@
...
-	genConfig, _ := config.ParseGenConfigYAML(configPath)
+	// map OAS resource model to CodeSpecModel
+	_ = oasResourceToProviderSpecModel(oasResource, genConfig.Resources[resourceName], resourceName)
+
+	// return &Model{Resources: []Resource{*codeSpecResource}}, nil
+	return TestExampleCodeSpecification(), nil
+}
+
+/*
+Computability:
```
maastha: will remove

tools/codegen/codespec/api_spec_schema.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/pb33f/libopenapi/datamodel/high/base"
+	high "github.com/pb33f/libopenapi/datamodel/high/v3"
+)
+
+var errInvalidSchema = fmt.Errorf("invalid schema")
+
+type APISpecSchema struct {
+	Schema *base.Schema
+	Type   string
+	Format string
```
AgustinBettati: curious why format has to be defined here, istead of accessing Schema.Format value when needed

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,27 +2,233 @@
...
+	return pathItem.Parameters, nil
+}
 
-	// TODO: remove after ToProviderSpecModel() implemented
-	return TestExampleCodeSpecification()
+func applyConfigSchemaOptions(resourceConfig config.Resource, resource *Resource) {
+	// TODO: implement in follow-up PR
 }
 
 func TestExampleCodeSpecification() *Model {
```
AgustinBettati: I believe this was a left over right?

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,27 +2,233 @@
...
+		// DeprecationMessage: ,
+		Attributes: attributes,
+	}
+
+	resource := &Resource{
+		Name:   name,
+		Schema: schema,
+	}
+
+	applyConfigSchemaOptions(resourceConfig, resource)
```
AgustinBettati: like this clear separation 👍

tools/codegen/codespec/api_spec.go
```diff
@@ -0,0 +1,104 @@
...
+	return resp, nil
+}
+
+func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType]) (*APISpecSchema, error) {
+	if mediaTypes == nil {
+		return nil, errSchemaNotFound
+	}
+
+	// hard-coding API version for now, this should be dynamically handled to perhaps use the latest API version
+	jsonMediaType, ok := mediaTypes.Get("application/vnd.atlas.2023-01-01+json")
```
AgustinBettati: Agree we can grab the latest version. I believe in the transformed API spec only latest version is included right?

tools/codegen/codespec/api_spec.go
```diff
@@ -0,0 +1,104 @@
...
+	for pair := range orderedmap.Iterate(context.TODO(), sortedMediaTypes) {
+		mediaType := pair.Value()
+		if mediaType.Schema != nil {
+			s, err := BuildSchema(mediaType.Schema)
+			if err != nil {
+				return nil, err
+			}
+			return s, nil
+		}
+	}
```
AgustinBettati: would this be the logic for grabbing the latest version?

tools/codegen/codespec/api_spec.go
```diff
@@ -0,0 +1,104 @@
...
+		return getSchemaFromMediaType(okResponse.Content)
+	}
+
+	createdResponse, ok := op.Responses.Codes.Get(OASResponseCodeCreated)
+	if ok {
+		return getSchemaFromMediaType(createdResponse.Content)
+	}
+
+	sortedCodes := orderedmap.SortAlpha(op.Responses.Codes)
+	for pair := range orderedmap.Iterate(context.TODO(), sortedCodes) {
```
AgustinBettati: nit: I see usage of context.TODO(), if we wont propagate a ctx should we use context.Background()?

tools/codegen/codespec/api_spec.go
maastha: this would right now grab all available media type responses, we can filter to only use the latest. Thinking more on this each of CRUD may have different versions no? in that case we'll need to ensure same API version is used for each. WDYT?
I haven't handled this yet because was wondering if API version is something we want to define in the config?

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: yes, test is yet to be updated

tools/codegen/codespec/api_spec.go
maastha: makes sense, updated

tools/codegen/codespec/api_spec.go
maastha: yes

tools/codegen/codespec/api_spec.go
maastha: updated to grab the available version

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: updated, TestExampleCodeSpecification has now been removed

tools/codegen/codespec/api_spec_schema.go
maastha: good point, updated

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -19,16 +26,78 @@ func TestConvertToProviderSpec(t *testing.T) {
...
+				},
+				},
+			},
 		},
 	}
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
-			result := codespec.ToProviderSpecModel(tc.inputOpenAPISpecPath, tc.inputConfigPath, &tc.inputResourceName)
+			result, _ := codespec.ToProviderSpecModel(tc.inputOpenAPISpecPath, tc.inputConfigPath, tc.inputResourceName)
```
EspenAlbert: instead of ignoring error, maybe we can assert it is nil?

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,69 +2,218 @@
...
 	"github.com/mongodb/terraform-provider-mongodbatlas/tools/codegen/openapi"
 )
 
-// using blank identifiers for now, will be removed in follow-up PRs once logic for conversion is added
-func ToProviderSpecModel(atlasAdminAPISpecFilePath, configPath string, resourceName *string) *Model {
-	_, err := openapi.ParseAtlasAdminAPI(atlasAdminAPISpecFilePath)
+func ToProviderSpecModel(atlasAdminAPISpecFilePath, configPath string, resourceName string) (*Model, error) {
+	apiSpec, err := openapi.ParseAtlasAdminAPI(atlasAdminAPISpecFilePath)
 	if err != nil {
 		panic(err)
```
EspenAlbert: Do we need to panic here? Looks like we are doing it in `main.go` anyway?

tools/codegen/codespec/api_spec.go
```diff
@@ -0,0 +1,93 @@
...
+func BuildSchema(proxy *base.SchemaProxy) (*APISpecSchema, error) {
+	resp := &APISpecSchema{}
+
+	schema, err := proxy.BuildSchema()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build schema from proxy: %w", err)
+	}
+
+	if len(schema.Type) == 0 {
+		return nil, errInvalidSchema
```
EspenAlbert: what about a more specific error message?

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,69 +2,218 @@
...
+		}
+
+		s, err := BuildSchema(param.Schema)
+		if err != nil {
+			continue
+		}
+
+		paramName := param.Name
+		s.Schema.Description = param.Description
+		parameterAttribute, err := s.buildResourceAttr(paramName, ComputedOptional)
```
EspenAlbert: Why ComputedOptional?
If it is an `id` it is computed. If it is in the `create` then it is required?
*I can see this is handled elsewhere. Maybe a comment here to mention that `ComputedOptional` is only a placeholder?

tools/codegen/codespec/merge_attributes.go
```diff
@@ -0,0 +1,79 @@
+package codespec
+
+import "sort"
+
+func mergeAttributes(pathParams, createRequest, createResponse, readResponse Attributes) Attributes {
```
EspenAlbert, line 5: This is super powerful! Having all the attributes in one place and merging. I like it 👍

tools/codegen/codespec/merge_attributes.go
```diff
@@ -0,0 +1,79 @@
...
+			existingAttr.ComputedOptionalRequired = computability
+		} else {
+			newAttr := *attr
+			newAttr.ComputedOptionalRequired = computability
+			merged[attr.Name] = &newAttr
+		}
+	}
+
+	// Path parameters: all attributes will be "required"
+	for i := range pathParams {
```
EspenAlbert, line 27: This might not work. E.g., `projectId` is a path parameter but it shouldn't be required

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,69 +2,218 @@
...
+		DeprecationMessage: oasResource.DeprecationMessage,
+		Attributes:         attributes,
 	}
+
+	resource := &Resource{
+		Name:   name,
+		Schema: schema,
+	}
+
+	applyConfigSchemaOptions(resourceConfig, resource)
```
EspenAlbert, line 63: nice! I guess this is where you'll do all the overrides

EspenAlbert (approved):
Well done!
A few comments. But looks like a good structure 👍

tools/codegen/codespec/merge_attributes.go
maastha: not sure i understand, if it's a path parameter it will be required no?
Can you share a use-case?

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: right

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: lmc

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: Yes this should be Required, will update

maastha:
NOTE: Computability evaluation methods and structure may change when we handle nested attributes
cc @AgustinBettati

tools/codegen/codespec/testdata/config-no-schema-opts.yml
```diff
@@ -0,0 +1,14 @@
...
+      method: GET
+    create:
+      path: /api/atlas/v2/groups/{groupId}/testResource
+      method: POST
+    update:
+      path: /api/atlas/v2/groups/{groupId}/testResource
+      method: PATCH
+    delete:
+      path: /api/atlas/v2/groups/{groupId}/testResource
+      method: DELETE
```
AgustinBettati, line 14: Just thinking in terms of simplifying implementation and config, does DELETE method add value?

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -2,69 +2,218 @@
...
+func pathParamsToAttributes(createOp, readOp *high.Operation) Attributes {
+	pathParams := createOp.Parameters
+	pathParams = append(pathParams, readOp.Parameters...)
+
+	// TODO: fix path param description not mapping
+	pathAttributes := Attributes{}
+	for _, param := range pathParams {
+		if param.In != OASPathParam && param.In != OASQueryParam {
+			continue
+		}
```
AgustinBettati: one detail, while path params are required query params can be optional. Would say for now we can focus on parsing path params.

AgustinBettati (approved):
LGTM, some considerations to keep in mind

tools/codegen/codespec/api_spec_schema.go
```diff
@@ -0,0 +1,63 @@
...
+	DeprecationMessage *string
+	CreateOp           *high.Operation
+	ReadOp             *high.Operation
+	UpdateOp           *high.Operation
+	DeleteOp           *high.Operation
+	CommonParameters   []*high.Parameter
+}
+
+func (s *APISpecSchema) GetComputability(name string) ComputedOptionalRequired {
+	for _, prop := range s.Schema.Required {
```
lantoli: if s.Schema.Required is a []string,   slices.Contains can be used to simplify instead of looping yourself

tools/codegen/codespec/api_to_provider_spec_mapper.go
lantoli: panic is not bad for a tool, but it's a bit "ugly" as you'll get the execution stack trace and a lot of info that can confuse the tool user. 

In case we want to exit the program, a nicer alternative can be to use log.Fatal or log.Fatalf that shows a message and do os.Exit(1),  so it'll be flagged as failure for scripts or GHAs

tools/codegen/codespec/testdata/config-no-schema-opts.yml
maastha: at this point for automation, no, but in the long run, config would contain all of CRUD paths right?

AgustinBettati: yes eventually might be useful, understand it does not add much overhead so okay to leave as is.

tools/codegen/codespec/merge_attributes.go
EspenAlbert: https://github.com/mongodb/atlas-sdk-go/blob/main/openapi/atlas-api-transformed.yaml#L1150

maastha: hmmm, that's true. For example for cluster, groupId and clusterId can be path params in read/GET but in the schema only groupId will be a required attribute and clusterId would actually be computed.

I think then a general approach here can be only create/POST path params should be required.

 @AgustinBettati does that sound good? or I may be missing a case? lmk will update the spec then

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
maastha: updated to use require.NoError(t, err)

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: updated to use log.Fatalf wherever applicable

tools/codegen/codespec/api_spec.go
maastha: updated

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: updated

tools/codegen/codespec/api_to_provider_spec_mapper.go
maastha: updated to use only pathParams for now

tools/codegen/codespec/api_spec_schema.go
maastha: done, thanks for the tip!

tools/codegen/codespec/merge_attributes.go
AgustinBettati: I agree with the approach of defining as required only the path params present in the POST operation. If there are other path params in GET/UPDATE/DELETE the will very likely be present in the response of the post (computed).

PR #2674, opened 2024-10-10
chore: Adds sleep in tests to avoid out of capacity errors

lantoli:
## Description

Adds sleep in tests to avoid out of capacity errors. 

This PR is to test the hypothesis that out of capacity errors in tests are caused because clusters are created simultaneously in the same project, see [HELP-65223](https://jira.mongodb.org/browse/HELP-65223) for more info.

The previous PR didn't work because the way tests are executed in Go. They run in serial until `t.Parallel()` is called. In that PR sleep was before that call so only test preparation was serialized but then all parallel tests would start at the same time.

In this PR, the sleep is in PreCheck functions, that are called when the tests are already running in parallel.

## Type of change:

## Required Checklist:


internal/testutil/acc/shared_resource.go
```diff
@@ -48,8 +48,6 @@ func ProjectIDExecution(tb testing.TB) string {
 		sharedInfo.projectName = RandomProjectName()
 		tb.Logf("Creating execution project: %s\n", sharedInfo.projectName)
 		sharedInfo.projectID = createProject(tb, sharedInfo.projectName)
-	} else {
```
lantoli, line 51: revert previous PR

GNUmakefile
```diff
@@ -8,7 +8,7 @@ endif
 
 ACCTEST_REGEX_RUN?=^TestAcc
 ACCTEST_TIMEOUT?=300m
-PARALLEL_GO_TEST?=20
+PARALLEL_GO_TEST?=50
```
lantoli, line 11: don't limit parallel tests

internal/testutil/acc/shared_resource.go
```diff
@@ -87,10 +85,23 @@ func ClusterNameExecution(tb testing.TB) (projectID, clusterName string) {
...
+// SerialSleep waits a few seconds so clusters are not created concurrently in the execution project, see HELP-65223.
+// This must be called once the test is marked as parallel, e.g. in PreCheck inside Terraform tests.
+func SerialSleep(tb testing.TB) {
+	tb.Helper()
+	SkipInUnitTest(tb)
+	require.True(tb, sharedInfo.init, "SetupSharedResources must called from TestMain test package")
+
+	sharedInfo.muSleep.Lock()
+	defer sharedInfo.muSleep.Unlock()
+	time.Sleep(10 * time.Second)
```
EspenAlbert, line 97: Would it be good to have some jitter?

EspenAlbert, line 97: This means parallel test N will be delayed by 10s x N? Why 10s?

lantoli: correct. 10s to pick a number that is big enough to make sure clusters are not running very close. we're not still sure if this is the issue.

once we validate this is the problem (or not), then we can get the number lower

EspenAlbert: Okay. Thanks for explaining and investigating this 👍

EspenAlbert (approved):

marcosuma (approved):

PR #2693, opened 2024-10-15
chore: PoC - Model generation - Supporting nested schema (List, ListNested, Set & SetNested)

maastha:
## Description

PoC - Model generation - Supporting nested schema (List, ListNested, Set & SetNested)
Maps and SingleNestedObjects will be addressed in a follow-up PR.

Link to any related issue(s): [CLOUDP-278226](https://jira.mongodb.org/browse/CLOUDP-278226)

## Type of change:

## Required Checklist:

### DIFF
--- a/tools/codegen/codespec/api_to_provider_spec_mapper_test.go
+++ b/tools/codegen/codespec/api_to_provider_spec_mapper_test.go
@@ -12,7 +12,7 @@ import (
 
 const (
 	testFieldDesc     = "Test field description"
-	testResourceDesc  = "Configures the project level settings for the Test Resource feature."
+	testResourceDesc  = "POST API description"
 	testPathParamDesc = "Path param test description"
 )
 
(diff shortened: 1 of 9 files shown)


tools/codegen/codespec/element_type.go
```diff
@@ -0,0 +1,31 @@
...
+type ElemType int
+
+const (
+	Bool ElemType = iota
+	Float64
+	Int64
+	Number
+	String
+	Unknown
+)
```
AgustinBettati: Would leave the model together in `model.go`, and having specific logic for api spec -> model in a separate file. In terms of the unknown might be better to error out as we would be able to generate a "correct" model.

maastha: For Unknown, it does throw an error except if the element is an array or object in which case it is essentially silently ignored. Unknown is more like a place holder for nil ElemType.
How would you want to handle this scenario from the schema generation side? i.e. if the elem type is an object or array?

AgustinBettati: From code generation side it currently support existing model element types (Bool, Float64, Int64, Number, String), which brought me the question of what to do in case of the new Unknown.
Since this is currently a limitation, I see 2 options:
- Best effort and simply default to a string. This would not be a correct generation but can still be helpful as we see generation of all the rest of the schema.
- Error out directly, more transparent but we wont get value of a potential "partial generation".

More inclined towards first option.

maastha: I'm more inclined towards the first as well. Would like this to be best effort instead of completely breaking for unhandled cases. We can always continue improving as we come across more and more test cases.

Let me update to default to string in that case.

maastha: updated

tools/codegen/codespec/merge_attributes.go
```diff
@@ -2,6 +2,7 @@ package codespec
 
 import "sort"
 
+// TODO: update to infer computability of inner nested attributes
```
EspenAlbert, line 5: Looks like this might be done already?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -104,3 +104,104 @@ func TestConvertToProviderSpec(t *testing.T) {
...
+								ComputedOptionalRequired: codespec.Computed,
+								List: &codespec.ListAttribute{
+									ElementType: codespec.String,
+								},
+								Description: conversion.StringPtr(testFieldDesc),
+							},
+							{
+								Name:                     "nested_object_array_attr",
+								ComputedOptionalRequired: codespec.Required,
+								ListNested: &codespec.ListNestedAttribute{
```
AgustinBettati, line 149: should we include a set nest example in this unit test as well?

AgustinBettati (approved):
LGTM, syncing offline about how Unknown element type will handled on generation side

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
maastha: yes, will include in the follow-up PR

tools/codegen/codespec/merge_attributes.go
maastha: On nested levels the logic should be revisited

PR #2717, opened 2024-10-18
chore: PoC - Support configuration of timeout in schema

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-279256

Enables defining timeouts attribute in resulting schema by using the following configuration definition:
```
resources:
  push_based_log_export:
    ...
    schema:
      ...
      timeouts: ["create", "update", "delete"]
```

## Type of change:

## Required Checklist:


tools/codegen/config.yml
```diff
@@ -9,30 +9,27 @@ resources:
...
-            "github.com/hashicorp/terraform-plugin-framework/path"
-            ],
-            definition: "stringvalidator.ConflictsWith(path.MatchRoot(\"name\"))"
-          }]
-      
-        prefix_path:
-          computability:
-            optional: true
-            computed: true
+      # overrides:
```
AgustinBettati, line 15: commenting out config features that are not yet supported

tools/codegen/codespec/model.go
```diff
@@ -44,6 +44,7 @@ type Attribute struct {
 	Set          *SetAttribute
 	Int64        *Int64Attribute
 	SingleNested *SingleNestedAttribute
+	Timeouts     *TimeoutsAttribute
```
maastha, line 47: Should be on a resource level (in Schema might be more intuitive). Any reason we want to keep it at an attribute level? Is it only because it simplifies code generation?

tools/codegen/schema/schema_attribute.go
```diff
@@ -22,20 +22,32 @@ func GenerateSchemaAttributes(attrs codespec.Attributes) CodeStatement {
 	}
 }
 
-func generateAttribute(attr *codespec.Attribute) CodeStatement {
-	generator := typeGenerator(attr)
+type attributeGenerator interface {
+	AttributeCode() CodeStatement
+}
+
+type conventionalAttributeGenerator struct {
```
maastha: can you confirm what does "conventional" signify here?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -337,6 +337,12 @@ func TestConvertToProviderSpec_nested_schemaOverrides(t *testing.T) {
 									},
 								},
 							},
+							{
```
maastha, line 340: is specifying a default timeout value out of scope at this point?

tools/codegen/codespec/model.go
AgustinBettati: Was thinking at the attribute level given how it is configured in terraform plugin framework schema. You will see it is configure as an [attribute](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/searchdeployment/resource_search_deployment_schema.go#L61), not as a property at the schema level. Defining at the resource is definitely possible, but think an attribute type is more aligned with how it is defined.

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
AgustinBettati: Default timeout values are needed/configured in CRUD logic. Given that as of now they would not be used, would keep the model/config with the information necessary from schema generation aspect

tools/codegen/schema/schema_attribute.go
AgustinBettati: Can see how this can get confusing. Refactored to have single interface `attributeGenerator`, and then specific type implementation that share common code structure (all except timeout attribute type) call `commonAttrStructure` function. LMK if this makes it more clear/straightforward.

maastha: >(all except timeout attribute type)

yeah this is why I was inclining a bit towards keeping timeouts at resource level [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2717#discussion_r1808543660) instead of an attribute. I understand how it is represented as an attribute in the TPF schema but I can also see how we always have filter it out when writing "common" code for actual attributes.

maastha (approved):
LGTM

tools/codegen/schema/schema_attribute.go
AgustinBettati: Looked into the alternative, at the end there is some common code also in the typed model generation which would need adjust if placed at the schema level. Leaving as is for now but thanks for brining this up.

PR #2729, opened 2024-10-23
chore: Update operation improvements for `mongodbatlas_flex_cluster` resource

oarbusi:
## Description

- Avoids update when `project_id`, `name` or `provider_settings` attributes are modified.
- UseStateForUnknown plan modifier is used to avoid unnecessary plan changes `(known after apply)` when the resource is updated but we know that those values won't change. e.g. termination_protection_enabled

Link to any related issue(s): CLOUDP-279920

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/customplanmodifier/non_updatable.go
+++ b/internal/common/customplanmodifier/non_updatable.go
@@ -15,11 +15,11 @@ type nonUpdatableStringAttributePlanModifier struct {
 }
 
 func (d *nonUpdatableStringAttributePlanModifier) Description(ctx context.Context) string {
-	return "Ensures that update operations fails when updating an attribute."
+	return d.MarkdownDescription(ctx)
 }
 
 func (d *nonUpdatableStringAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
-	return d.Description(ctx)
+	return "Ensures that update operations fails when updating an attribute."
 }
 
 func (d *nonUpdatableStringAttributePlanModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
--- a/internal/common/customplanmodifier/non_updatable.go
+++ b/internal/common/customplanmodifier/non_updatable.go
@@ -10,25 +10,25 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-func NonUpdatableAttributePlanModifier(attribute string) planmodifier.String {
-	return &nonUpdatableAttributePlanModifier{
+func NonUpdatableStringAttributePlanModifier(attribute string) planmodifier.String {
+	return &nonUpdatableStringAttributePlanModifier{
 		Attribute: attribute,
 	}
 }
 
-type nonUpdatableAttributePlanModifier struct {
+type nonUpdatableStringAttributePlanModifier struct {
 	Attribute string
 }
 
-func (d *nonUpdatableAttributePlanModifier) Description(ctx context.Context) string {
+func (d *nonUpdatableStringAttributePlanModifier) Description(ctx context.Context) string {
 	return "Ensures that update operations fails when updating an attribute."
 }
 
-func (d *nonUpdatableAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
+func (d *nonUpdatableStringAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
 	return d.Description(ctx)
 }
 
-func (d *nonUpdatableAttributePlanModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
+func (d *nonUpdatableStringAttributePlanModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
 	var planAttributeValue types.String
 	planFullPath, diag := req.Plan.PathMatches(ctx, GetFullPathExpression(ctx, d.Attribute))
 	resp.Diagnostics.Append(diag...)
--- a/internal/common/customplanmodifier/non_updatable.go
+++ b/internal/common/customplanmodifier/non_updatable.go
@@ -1,0 +1,51 @@
+package customplanmodifier
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	planmodifier "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func NonUpdatableAttributePlanModifier(attribute string) planmodifier.String {
+	return &nonUpdatableAttributePlanModifier{
+		Attribute: attribute,
+	}
+}
+
+type nonUpdatableAttributePlanModifier struct {
+	Attribute string
+}
+
+func (d *nonUpdatableAttributePlanModifier) Description(ctx context.Context) string {
+	return "Ensures that update operations fails when updating an attribute."
+}
+
+func (d *nonUpdatableAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
+	return d.Description(ctx)
+}
+
+func (d *nonUpdatableAttributePlanModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
+	var planAttributeValue types.String
+	diags := req.Plan.GetAttribute(ctx, path.Root(d.Attribute), &planAttributeValue)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var stateAttributeValue types.String
+	req.State.GetAttribute(ctx, path.Root(d.Attribute), &stateAttributeValue)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !stateAttributeValue.IsNull() && stateAttributeValue.ValueString() != planAttributeValue.ValueString() {
+		resp.Diagnostics.AddError(
+			"attribute is not updatable",
+			"attribute is not updatable",
+		)
+		return
+	}
+}
--- a/internal/common/customplanmodifier/non_updatable.go
+++ b/internal/common/customplanmodifier/non_updatable.go
@@ -7,14 +7,11 @@ import (
 	planmodifier "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 )
 
-func NonUpdatableStringAttributePlanModifier(attribute string) planmodifier.String {
-	return &nonUpdatableStringAttributePlanModifier{
-		Attribute: attribute,
-	}
+func NonUpdatableStringAttributePlanModifier() planmodifier.String {
+	return &nonUpdatableStringAttributePlanModifier{}
 }
 
 type nonUpdatableStringAttributePlanModifier struct {
-	Attribute string
 }
 
 func (d *nonUpdatableStringAttributePlanModifier) Description(ctx context.Context) string {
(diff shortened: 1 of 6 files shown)


internal/service/flexcluster/resource.go
```diff
@@ -196,3 +205,27 @@ func splitFlexClusterImportID(id string) (projectID, clusterName *string, err er
 
 	return
 }
+
+func isUpdateAllowed(ctx context.Context, plan, state *TFModel) (bool, error) {
+	if !plan.ProjectId.Equal(state.ProjectId) {
```
marcosuma: curious why are we adding this validation here and not letting the API fail?

oarbusi: API would fail with errors like `GROUP_NOT_FOUND`, `CLUSTER_NOT_FOUND` because we would be trying to update a non-existing cluster or a cluster in a non-existing project, which I think can be confusing for the user

internal/service/flexcluster/resource.go
```diff
@@ -196,3 +205,27 @@ func splitFlexClusterImportID(id string) (projectID, clusterName *string, err er
...
+		return false, errors.New("project_id cannot be updated")
+	}
+	if !plan.Name.Equal(state.Name) {
+		return false, errors.New("name cannot be updated")
+	}
```
AgustinBettati: Given we are opting for throwing a validation over using RequiresReplace() plan modifier, could be nice to implement a plan modifier that returns a validation when there are changes. This way we can use these across all resource implementation and not have custom code.

internal/service/flexcluster/resource_schema.go
```diff
@@ -72,32 +89,53 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 				MarkdownDescription: "Version of MongoDB that the instance runs.",
 			},
 			"state_name": schema.StringAttribute{
-				Computed:            true,
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
AgustinBettati: From our best practices doc I see we use this plan modifier for `computed attributes that will not have changes in their value`. I believe for state_name we could have different values.

internal/service/flexcluster/resource.go
```diff
@@ -196,3 +205,27 @@ func splitFlexClusterImportID(id string) (projectID, clusterName *string, err er
...
+	stateProviderSettings := &TFProviderSettings{}
+	if diags := state.ProviderSettings.As(ctx, stateProviderSettings, basetypes.ObjectAsOptions{}); diags.HasError() {
+		return false, errors.New("provider_settings cannot be updated")
+	}
+	if !planProviderSettings.BackingProviderName.Equal(stateProviderSettings.BackingProviderName) {
+		return false, errors.New("backing_provider_name cannot be updated")
+	}
+	if !planProviderSettings.RegionName.Equal(stateProviderSettings.RegionName) {
+		return false, errors.New("region_name cannot be updated")
```
AgustinBettati: q: Was surprised to see that region_name cannot be updated, was curious if we tested the API just to confirm this. (cc @cveticm as we have spoke about this)

oarbusi: I tested it, region_name cannot be updated and was also confirmed by the upstream team

internal/service/flexcluster/resource_schema.go
oarbusi: You are right! Will change it and also will confirm with the upstream team if there is any other attribute which might change after update

internal/service/flexcluster/resource.go
AgustinBettati: thanks for confirming 👍

internal/service/flexcluster/resource.go
oarbusi: I have done this, let me know what you think

oarbusi:
flex clusters API is flaky in dev environment, I have successfully run the test using QA environment
```
Running tool: /Users/oriol.abadal/.asdf/shims/go test -timeout 2400s -run ^(TestAccFlexClusterRS_basic|TestAccFlexClusterRS_failedUpdate)$ github.com/mongodb/terraform-provider-mongodbatlas/internal/service/flexcluster

ok  	github.com/mongodb/terraform-provider-mongodbatlas/internal/service/flexcluster	69.006s
```

internal/common/customplanmodifier/non_updatable.go
```diff
@@ -0,0 +1,72 @@
...
+	return "Ensures that update operations fails when updating an attribute."
+}
+
+func (d *nonUpdatableAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
+	return d.Description(ctx)
+}
+
+func (d *nonUpdatableAttributePlanModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
+	var planAttributeValue types.String
+	planFullPath, diag := req.Plan.PathMatches(ctx, GetFullPathExpression(ctx, d.Attribute))
```
AgustinBettati: q: Instead of using req.Plan and req.State, would using req.PlanValue and req.StateValue simplify the implementation?

internal/common/customplanmodifier/non_updatable.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	planmodifier "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func NonUpdatableAttributePlanModifier(attribute string) planmodifier.String {
```
AgustinBettati: in the name would mention it is string specific, eventually we might have similar implementation for other types.

AgustinBettati (commented):
looks really good, just one question in terms of the implementation

internal/service/flexcluster/resource_schema.go
oarbusi: Done

internal/common/customplanmodifier/non_updatable.go
oarbusi: Agreed, done

internal/common/customplanmodifier/non_updatable.go
oarbusi: Thanks for the suggestion, using req.PlanValue and req.StateValue makes it a lot simpler.

internal/common/customplanmodifier/non_updatable.go
```diff
@@ -0,0 +1,50 @@
...
+	if !stateAttributeValue.IsNull() && stateAttributeValue.ValueString() != planAttributeValue.ValueString() {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("%s cannot be updated", d.Attribute),
+			fmt.Sprintf("%s cannot be updated", d.Attribute),
+		)
+		return
+	}
+}
+
+func GetFullPathExpression(ctx context.Context, attribute string) path.Expression {
```
AgustinBettati: this can also be removed

internal/common/customplanmodifier/non_updatable.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	planmodifier "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+func NonUpdatableStringAttributePlanModifier(attribute string) planmodifier.String {
+	return &nonUpdatableStringAttributePlanModifier{
+		Attribute: attribute,
+	}
+}
```
AgustinBettati:
```suggestion
func NonUpdatableStringPlanModifier() planmodifier.String {
	return &nonUpdatableStringAttributePlanModifier{}
}
```

no need to have the attribute name now, also suggestion for keeping name more concise

internal/service/flexcluster/resource.go
```diff
@@ -106,23 +107,25 @@ func (r *rs) Read(ctx context.Context, req resource.ReadRequest, resp *resource.
 }
 
 func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
-	var tfModel TFModel
-	resp.Diagnostics.Append(req.Plan.Get(ctx, &tfModel)...)
+	var plan TFModel
+	var state TFModel
```
AgustinBettati: state is not needed right?

AgustinBettati (commented):
small cleanup suggestions due to new implementation

internal/common/customplanmodifier/non_updatable.go
oarbusi: Not technically needed, but I kept it so that the error message is clear on which attribute can't be updated. No strong opinion here, so if you still think it should be removed I can remove it

internal/service/flexcluster/resource.go
oarbusi: Right, removed

internal/common/customplanmodifier/non_updatable.go
oarbusi: removed, thanks

internal/common/customplanmodifier/non_updatable.go
AgustinBettati: Ohh I see, could we use `req.Path` to get this information?

oarbusi: `req.Path` does have the info. THanks

internal/common/customplanmodifier/non_updatable.go
```diff
@@ -0,0 +1,36 @@
...
+
+type nonUpdatableStringAttributePlanModifier struct {
+}
+
+func (d *nonUpdatableStringAttributePlanModifier) Description(ctx context.Context) string {
+	return "Ensures that update operations fails when updating an attribute."
+}
+
+func (d *nonUpdatableStringAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
+	return d.Description(ctx)
```
lantoli: nit: we're giving preference to MarkdownDescription, i know it's the same behavior but to be consistent I would define the string in MarkdownDescription

internal/service/flexcluster/resource.go
```diff
@@ -106,23 +107,23 @@ func (r *rs) Read(ctx context.Context, req resource.ReadRequest, resp *resource.
...
 	}
 
-	projectID := tfModel.ProjectId.ValueString()
-	clusterName := tfModel.Name.ValueString()
+	projectID := plan.ProjectId.ValueString()
+	clusterName := plan.Name.ValueString()
 
 	connV2 := r.Client.AtlasV2
-	_, _, err := connV2.FlexClustersApi.UpdateFlexCluster(ctx, projectID, tfModel.Name.ValueString(), flexClusterReq).Execute()
+	_, _, err := connV2.FlexClustersApi.UpdateFlexCluster(ctx, projectID, plan.Name.ValueString(), flexClusterReq).Execute()
```
lantoli, line 126: it's not clear to me,  what can be updated in a flex cluster? maybe we can have an "if" in Update and don't need nonUpdatableStringAttributePlanModifier

oarbusi: All the attributes that have the custom plan modifier in the schema are the non-updatable attributes. Having nonUpdatableStringAttributePlanModifier instead of an if in the Update operation makes this more reusable and I think will help in the auto-generation (it's easier to add a plan modifier in an attribute rather than adding custom logic in the update operation)

lantoli: discussed offline, good to go

lantoli (approved):

internal/service/flexcluster/resource.go
AgustinBettati: +1 on the automation aspect

EspenAlbert: ~Curious why we don't use the `stringplanmodifier.RequiresReplace()`?~ I saw the tech document decision

PR #2755, opened 2024-10-30
chore: Automatically updates Terraform version used in repository and test-suite

maastha:
## Description

Automatically updates Terraform version used in repository and test-suite.

This job will create a PR like https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2756 when a change in supported TF versions is detected.

Link to any related issue(s): [CLOUDP-243394](https://jira.mongodb.org/browse/CLOUDP-243394)

## Type of change:

## Required Checklist:


.github/workflows/update-tf-versions.yml
```diff
@@ -30,4 +30,29 @@ jobs:
           commit-message: "doc: Updates Terraform Compatibility Matrix documentation"
           delete-branch: true
           branch: terraform-compatibility-matrix-update
-          body: "Automatic updates for Terraform Compatibility Matrix documentation. **Action Required**: Update test-suite.yml, .tools-version files, and TF_VERSION_LATEST GitHub environment variable if needed."
+          body: "Automatic updates for Terraform Compatibility Matrix documentation. **Action Required**: Update TF_VERSION_LATEST GitHub environment variable if needed."
```
lantoli: if this PR is being generated with this PR:  https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2756/files

I think it's better to keep:  schedule_terraform_matrix: '["${{ vars.TF_VERSION_LATEST }}"]'
so not changing it to:   schedule_terraform_matrix: '["1.9.x"]'

lantoli: also not sure about the change to .tool-version from: terraform 1.9.7
to: terraform 1.9.0

not sure why it's being downgraded, in fact it would be great if it upgrades to 1.9.8 that is the latest version

lantoli: can we also update these 2 files? 
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/scripts/generate-doc.sh#L35
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/scripts/generate-docs-all.sh#L29

maastha: >I think it's better to keep: schedule_terraform_matrix: '["${{ vars.TF_VERSION_LATEST }}"]'
so not changing it to: schedule_terraform_matrix: '["1.9.x"]'

Could you share more on why it shouldn't be updated? Currently TF_VERSION_LATEST is set to 1.9.x in the repo so it won't really make a difference so I think it's better to automate it to reduce one manual dependency.


>also not sure about the change to .tool-version from: terraform 1.9.7
to: terraform 1.9.0

It looks like a downgrade right now because we have been manually maintaining that. The idea is that these updates should happen automatically when a new major TF version is released. We're currently not keeping track of minor versions.

scripts/update-tf-version-in-repository.sh
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -euo pipefail
+
+# example value of TF_VERSIONS_ARR='["1.9.x", "1.8.x", "1.7.x"]'
+TF_VERSIONS_ARR=$(./scripts/get-terraform-supported-versions.sh "false")
```
EspenAlbert, line 19: [nit] would be nice to use `./scripts/get-terraform-supported-versions.sh "latest"` and do the LATEST_TF_VERSION logic in the script

EspenAlbert (approved):
Nice. I like the idea of fewer commits due to terraform minor upgrade, and not having to rely on the vars.TF_VERSION_LATEST 👍

lantoli:
hi @maastha cc @EspenAlbert , some context that can help:
Current value for TF_VERSION_LATEST is 1.9.x, so scripts/tools/GHA/features using it will pick the latest patch for 1.9, it's currently 1.9.8.

However there are some tools/features like .tool-versions, or generate-doc.sh and generate-docs-all.sh (for tfplugindocs generate) that need an explicit version like 1.9.8, 1.9.x can not be used in these cases. 
So in that case i think we should use the latest  patch to make it equivalent to tools using the .X notation.  E.g. I think it's better to have now 1.9.8 in .tool-versions instead of 1.9.0 so we can anticipate problems in the latest published TF version. Similarly for  tfplugindocs generate tool.

Additionally I think it's better to use TF_VERSION_LATEST directly when format 1.9.x can be used, instead of having explicitly 1.9.x in those tools, so we don't need to update many files when a major TF version is released.  ( note that TF_VERSION_LATEST doesn't need to be updated for minor releases)

edit: i see TF_VERSION_LATEST is used in 4 places, i'm ok if we want to delete the var and use 1.9.x directly, and update them when a major release comes.   But I still think we should update patches in tools not supporting .X notation.

tf_var

.github/workflows/update-tf-versions.yml
lantoli: added a comment to the PR with context

maastha: @lantoli updated to use the latest patch version wherever applicable. This is the new PR: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2759/files 

Yet to update the occurrences of TF_VERSION_LATEST in acceptance-tests and examples.yml, I'll create a ticket for that if I'm not able to get that to work by EOD.

lantoli: nice!

lantoli (approved):
LGTM, thanks for the follow-ups

scripts/update-tf-version-in-repository.sh
```diff
@@ -0,0 +1,40 @@
...
+DOC_ALL_SCRIPT="scripts/generate-docs-all.sh"
+
+LATEST_TF_VERSION=$(echo "$TF_VERSIONS_ARR" | sed -E 's/^\["([^"]+).*/\1/')
+
+# Update Terraform versions in test-suite.yml
+sed -i.bak -E "/^ *terraform_matrix:/,/^ *provider_matrix:/ s|(default: ')[^']*(')|\1$TF_VERSIONS_ARR\2|" "$TEST_SUITE_YAML_FILE"
+
+sed -i.bak -E "s|schedule_terraform_matrix: '.*'|schedule_terraform_matrix: '[\"$LATEST_TF_VERSION\"]'|" "$TEST_SUITE_YAML_FILE"
+
+# Update patch version occurences
```
lantoli:
```suggestion
# Update patch version occurrences
```

.github/workflows/update-tf-versions.yml
maastha: Created https://jira.mongodb.org/browse/CLOUDP-281749 to replace TF_VERSION_LATEST

PR #2795, opened 2024-11-13
chore: Initial implementation of Move State

lantoli:
## Description

Initial implementation of Move State.  Three approaches are shown so the best one can be chosen:
- Preferred: This is the recommended way by Hashicorp. Schema and model must be defined in TFP format. They can be reused if source resource is in TPF, if not, they must be created. It's more work but it has safer types so it can be a good option if full transformation logic is needed.
- Raw state: Only the model needs to be done (not full model, only attributes we want to read), but access to attributes is a bit harder as it has to be done individually. It can be a good alternative if we only need to extract project_id and cluster_name and leave the state creation to Read operation.
- JSON: Similar to previous approach, we only need to define model for project_id and cluster_name fields if we only need those. It's a bit more convenient to access the fields but problem is according to JSON attribute doc:  "it is not safe to assume that all states written by 0.12.0 or higher will be in JSON format; future versions may switch to an alternate encoding for states."

Link to any related issue(s): CLOUDP-283855

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedclustertpf/move_state.go
+++ b/internal/service/advancedclustertpf/move_state.go
@@ -2,36 +2,123 @@ package advancedclustertpf
 
 import (
 	"context"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
+// TODO: We temporarily use mongodbatlas_database_user instead of mongodbatlas_cluster to set up the initial environment
 func (r *rs) MoveState(context.Context) []resource.StateMover {
 	return []resource.StateMover{
 		{
-			StateMover: func(ctx context.Context, req resource.MoveStateRequest, resp *resource.MoveStateResponse) {
-				timeout := timeouts.Value{
-					Object: types.ObjectValueMust(
-						map[string]attr.Type{
-							"create": types.StringType,
-							"update": types.StringType,
-							"delete": types.StringType,
-						},
-						map[string]attr.Value{
-							"create": types.StringValue("30m"),
-							"update": types.StringValue("30m"),
-							"delete": types.StringValue("30m"),
-						}),
-				}
-				tfNewModel, shouldReturn := mockedSDK(ctx, &resp.Diagnostics, timeout)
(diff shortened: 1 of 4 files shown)


internal/service/advancedclustertpf/move_state.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// TODO: We temporarily use mongodbatlas_database_user instead of mongodbatlas_cluster to set up the initial environment
```
lantoli, line 26: so we don't need to wait until a cluster is created every time, of course this will be changed to cluster once all the initial setup is done

internal/service/advancedclustertpf/move_state_test.go
```diff
@@ -0,0 +1,127 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "name", clusterName),
+					resource.TestCheckResourceAttr(resourceName, "project_id", projectID),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAdvancedCluster_move_rawstate(t *testing.T) {
```
lantoli, line 37: test code is duplicated but is not refactored because we'll use only one approach, this is to show the different possible approaches

EspenAlbert (approved):
Interesting! Thank you for presenting the three methods! Really nice to see acutal implementations :clap:
I guess the `preferred` method we could import the `TFModel` using the `from` resource package directly and parse as usual.

Otherwise, it seems like we have to defined a new model representing `from` resource. 
IMO: `JSON` model is the simplest to define and I don't see any value in using `req.SourceRawState` over `req.JSON`. But maybe I'm missing something?

lantoli:
@EspenAlbert I updated the PR description about JSON problem: "it is not safe to assume that all states written by 0.12.0 or higher will be in JSON format; future versions may switch to an alternate encoding for states."

also found that we can pass in rawstate options `tfprotov6.UnmarshalOpts{ValueFromJSONOpts: tftypes.ValueFromJSONOpts{IgnoreUndefinedAttributes: true}})` to ignore undefined fields so similarly to JSON we don't need to define the full model, PR description updated

lantoli:
@EspenAlbert about `preferred` model we don't have it for cluster, we set attributes individually, and I assume we won't have models for any resource in SDKv2

PR #1735, opened 2023-12-11
refactor: Migrates advanced_cluster resource to TPF - DO NOT DELETE BRANCH

maastha:
## Description

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s):

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have read the [contribution guidelines](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CONTRIBUTING.md)
- [ ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals, I defined an isolated PR with a relevant title as it will be used in the auto-generated changelog.

PR #2798, opened 2024-11-13
chore: Adds Update Request Dumping for adv_cluster TPF

EspenAlbert:
## Description

- Adds Update Request Dumping ClusterDescription
  - Solved with a generic approach described in `patch_payload.go`: [func PatchPayloadNoChanges[T any](state, plan, reqPatch *T) (bool, error) {](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/dfded2d4eb549f7e45f395a85e8ee04019ae08d5/internal/common/conversion/patch_payload.go#L120)
  - And used by `patch_framework.go` with [func PatchPayloadHasChangesTpf[T any, R any](ctx context.Context, diags *diag.Diagnostics, state, plan *T, converter func(ctx context.Context, input *T, diags *diag.Diagnostics) *R, reqPatch *R) bool {](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/dfded2d4eb549f7e45f395a85e8ee04019ae08d5/internal/common/conversion/patch_framework.go#L10)
  - Invoked by `resource.go` with: 
    - [patchReq := &admin.ClusterDescription20240805{}](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/dfded2d4eb549f7e45f395a85e8ee04019ae08d5/internal/service/advancedclustertpf/resource.go#L103-L104)
    - [clusterChanges := conversion.PatchPayloadHasChangesTpf(ctx, diags, &state, &plan, NewAtlasReq, patchReq)](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/dfded2d4eb549f7e45f395a85e8ee04019ae08d5/internal/service/advancedclustertpf/resource.go#L104)
- Adds support for using both Legacy ProcessArgs and current ProcessArgs
- Some of the `testdata/*.json` payloads needed to be manually updated due to [Data Consistency Errors](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/data-consistency-errors#finding-data-consistency-errors), see more [details in Outcomes doc](https://docs.google.com/document/d/1QxrB1BOiMCvUjwAsllG0M3rHKl6ymKJOEF6L5aNYJmE/edit?tab=t.0#heading=h.7vh9rm5y86nc) (follow up PR will automate this process)
- Adds test case for sharded cluster
- New 3rd parties:
  - `go get github.com/evanphx/json-patch/v5`
    - [repo](https://github.com/evanphx/json-patch)
  - `go get github.com/wI2L/jsondiff@latest`
    - [repo](https://github.com/wI2L/jsondiff)
  - Both libraries works with [RFC6902 JavaScript Object Notation (JSON) Patch operations](https://datatracker.ietf.org/doc/html/rfc6902#section-4.1)
- [Next step](https://jira.mongodb.org/browse/CLOUDP-284793) is doing "proper" mocking with the SDK and cleaning up all these `*.json` files and add more tests cases

Link to any related issue(s): CLOUDP-284479

## Type of change:

## Required Checklist:


oarbusi:
nit: naming and package of `PatchPayloadHasChangesTpf` and `PatchPayloadNoChanges` methods are misleading IMO. methods are returning booleans (and not converting objects from what I understand) but are in the `conversion` package. Maybe creating a new patchpayload package and renaming the methods would help to make this clearer

internal/common/conversion/patch_framework.go
```diff
@@ -0,0 +1,22 @@
+package conversion
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+func PatchPayloadHasChangesTpf[T any, R any](ctx context.Context, diags *diag.Diagnostics, state, plan *T, converter func(ctx context.Context, input *T, diags *diag.Diagnostics) *R, reqPatch *R) bool {
```
AgustinBettati: might be good to document what T and R capture respectively. I understand T would be the TPF model, while R would be the SDK PATCH model?
Maybe the converter function can also be extracted to a type alias just from better readability.

internal/common/conversion/patch_payload.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, err
+	}
+	decodedPatch, err := jsonpatch.DecodePatch(patchKeyBytes)
+	if err != nil {
+		return nil, err
+	}
+	return decodedPatch, nil
+}
+
+// PatchPayloadNoChanges uses the state and plan to changes to find the patch request, including changes only when:
```
EspenAlbert: This might be a good place to start reading after the PR Description
(Only docstring, and then having a look at the test cases)

internal/common/conversion/patch_payload.go
```diff
@@ -0,0 +1,166 @@
...
+// - --> Except specific computed attributes in nested_attributes (e.g., `replication_spec[*].(id|zone_id`)
+// - The state and plan can be dumped to json
+// How it works:
+// 1. Use `jsondiff` to find the patch, aka. operations to go from state to plan
+// 2. Groups the operations by attribute name
+// 3. Filters the operations to only include replaced or added values
+// 4. Adds nested "removed" values from the state to the request
+// 5. Use `jsonpatch` to apply each attribute plan & state patch to an empty JSON object
+// 6. Update `reqPatch` pointer with the final JSON object marshaled to `T` or return true if no changes (`{}`)
+func PatchPayloadNoChanges[T any](state, plan, reqPatch *T) (bool, error) {
```
AgustinBettati: In terms of PATCH with array properties that have nested elements, curious how this scenario is handled. Asking as from latest conversation with upstream team Cluster API if there is any change within replicationSpecs/regionConfigs the expectation is to send all elements and each element providing its properties.

internal/service/advancedclustertpf/resource.go
```diff
@@ -91,19 +93,50 @@ func (r *rs) Read(ctx context.Context, req resource.ReadRequest, resp *resource.
...
+	var state, plan TFModel
+	diags := &resp.Diagnostics
+	diags.Append(req.Plan.Get(ctx, &plan)...)
+	diags.Append(req.State.Get(ctx, &state)...)
+	if diags.HasError() {
 		return
 	}
-	tfNewModel, shouldReturn := mockedSDK(ctx, &resp.Diagnostics, plan.Timeouts)
+	patchReq := &admin.ClusterDescription20240805{}
+	clusterChanges := conversion.PatchPayloadHasChangesTpf(ctx, diags, &state, &plan, NewAtlasReq, patchReq)
```
AgustinBettati: As a more general doubt, I am curious what are the downsides of going for an approach of sending all information in the PATCH request as done in other (if not all) TPF resources implementations. Automation is also a factor to keep in mind.
I understand cluster schema is much more complex so there might be more considerations, but we have to weigh in the additional complexity from our side if we go for the approach of forming partial patch requests.

AgustinBettati (commented):
some doubts and suggestions, definitely an interesting idea of creating the same SDK model from the state and config to capture actual differences.

EspenAlbert:
> nit: naming and package of `PatchPayloadHasChangesTpf` and `PatchPayloadNoChanges` methods are misleading IMO. methods are returning booleans (and not converting objects from what I understand) but are in the `conversion` package. Maybe creating a new patchpayload package and renaming the methods would help to make this clearer

The `converter` argument is doing the conversion.
However, I agree that another package can be beneficial. 

@oarbusi 
WDYT `common/patch`? then I can rename the function to be: `patch.PayloadHasChangesTpf(...)`

internal/common/conversion/patch_framework.go
EspenAlbert: Good point, changing in 8b827e86519db

EspenAlbert: For the converter, looks like a generic type alias is expected to come [early in 2025](https://medium.com/@okoanton/generic-type-aliases-in-go-1-24-what-they-are-and-why-we-need-them-07ca05539500)

internal/common/conversion/patch_payload.go
EspenAlbert: Added test in 790d5322aa0865b2032d16c443c670fef5c5298f

EspenAlbert: Added fix in 24a0f30c648075ff704cbc537898a050f3edff2a

internal/service/advancedclustertpf/resource.go
EspenAlbert: Good point, some thoughts:
1. Avoid sending updates if they are not needed to reduce change time
  - This is especially true for resources with multiple endpoints, e.g., if only adv_config has changed we shouldn't send a cluster PATCH (this is done now)
  - Uncertain how much partial updates requests reduce backend time but whenever we can skip a request I think it has a high value and will greatly reduce `apply` time
  - Might also decrease the chance of sending a deprecated field
2. Performance due to smaller payloads (minor importance)
3. Debugging help (minor importance) (if there are any errors it can be easier to see what exactly changed)

The downside is the "magic" and the risk of an incomplete solution. But I prefer it to [100s of lines checking for changes](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1735/files#diff-518b3a200e851d92e5dbe0e7882bc66180011a09895ca7b2370d5fd0a3ee5834R154)

internal/common/conversion/patch_framework.go
```diff
@@ -0,0 +1,22 @@
+package conversion
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+func PatchPayloadHasChangesTpf[TFModel any, SDKRequest any](ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel, converter func(ctx context.Context, input *TFModel, diags *diag.Diagnostics) *SDKRequest, reqPatch *SDKRequest) bool {
```
lantoli: what about reqPatch *SDKRequest being the returned value instead of bool, nil meaning there are no changes or an  errors was added to diagnostics? or it's because you can't create in in the func using like reqPatch := new(SDKRequest) ?

internal/common/conversion/patch_payload.go
```diff
@@ -0,0 +1,180 @@
...
+
+	jsonpatch "github.com/evanphx/json-patch/v5"
+	"github.com/wI2L/jsondiff"
+)
+
+type attrPatchOperations struct {
+	data map[string][]jsondiff.Operation
+}
+
+const rootAttributeName = "not|valid|identifier"
```
lantoli: what are the | in "not|valid|identifier" ?

internal/service/advancedclustertpf/resource.go
lantoli: It's true that this is more complex, but I think it's more complex this generic "engine", but resources will be simpler to do, so the complexity is narrowed to this generic implementation, that we need to test it thoughtfully.

lantoli (approved):
LGTM, as this is a new resource, I think it's ok to merge and continue to validate this is a good approach to follow.

lantoli:
> WDYT common/patch? then I can rename the function to be: patch.PayloadHasChangesTpf(...)

@EspenAlbert  or maybe common/update to say where it can be useful

internal/common/conversion/patch_framework.go
EspenAlbert: Thanks! Worked! 10529bede4db38b0bd967d303e3972b1852af8f3

internal/common/conversion/patch_payload.go
EspenAlbert: Initially, to ensure no collision. But can simply use `""`. Changed to 9aea7077af0f28573e7391c1fa3795b19fea7001

EspenAlbert:
> > WDYT common/patch? then I can rename the function to be: patch.PayloadHasChangesTpf(...)
> 
> @EspenAlbert or maybe common/update to say where it can be useful

Thanks for the feedback <3 
Went for this approach: `update.PatchPayloadTpf`

internal/common/conversion/patch_payload.go
AgustinBettati: thanks! One additional scenario you may want to add is having an additional replication spec [here](https://github.com/mongodb/terraform-provider-mongodbatlas/commit/24a0f30c648075ff704cbc537898a050f3edff2a#diff-781277f1db7d7d3121a7d2235442c297324ddd774a9b7d209456f3e891a412c4R135) (in both the state and plan) and ensure we are sending it in the PATCH.

internal/service/advancedclustertpf/resource.go
AgustinBettati: Yes its definitely great that we get a generic implementation for other resources. I do also like that it relies on conversion functions (to created the update SDK models), so effectively using this function can be an opt-in for each resource.
The point about "This is especially true for resources with multiple endpoints" is the point that has most value IMO, so for resources that consume single update method we might not have this need.

AgustinBettati (approved):
LGTM, left a follow up suggestion for another test case but feel free to include in follow PR

internal/common/conversion/patch_payload.go
EspenAlbert: Is it this scenario or something different? @AgustinBettati 
["New list entry added should be included": {](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/c3b6c4f4754998176e337daffcc65836de68d7c2/internal/common/update/patch_payload_test.go#L83)

AgustinBettati: might be slightly different, was thinking for that same test case adding another replication spec element in both the state and plan (that has no changes) and will need to be included in the PATCH.

PR #2813, opened 2024-11-21
chore: Adds initial TPF implementation for advanced_cluster

EspenAlbert:
## Description

- Note: `CLOUDP-284793_unit_test_sdk_mocking` branch name is missleading but PR changed, so prefer keeping it to have all discussions relevant.
- Adds inital TPF implementation for advanced_cluster
  - Limited to ISS schema and some missing logic (see technical design for details)
- Adds 1 "live" acceptance tests: `TestAccClusterAdvancedCluster_basicTenant`
- Fixes some issues with `move_state.go`
- More tests can be found in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2820

Link to any related issue(s): CLOUDP-284793

## Type of change:

## Required Checklist:


.github/workflows/acceptance-tests-runner.yml
```diff
@@ -359,6 +359,13 @@ jobs:
           terraform_wrapper: false  
       - name: Prepare new advanced_cluster
         run: make tools enable-advancedclustertpf  
+      - name: Mocked Acceptance Tests
+        env:
+          MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
```
lantoli: no needed, MONGODB_ATLAS_LAST_VERSION is only used for mig tests

internal/common/constant/error_codes.go
```diff
@@ -0,0 +1,5 @@
+package constant
+
+const (
+	ErrorCodeClusterNotFound = "CLUSTER_NOT_FOUND"
```
lantoli: nit: keep it at the moment in tpf package if not needed outside

internal/service/advancedclustertpf/resource.go
```diff
@@ -55,40 +67,24 @@ func (r *rs) Schema(ctx context.Context, req resource.SchemaRequest, resp *resou
...
-	if err != nil {
-		resp.Diagnostics.AddError("errorCreate", fmt.Sprintf(errorCreate, err.Error()))
+	diags := &resp.Diagnostics
+	if diags.HasError() {
 		return
 	}
-	tfNewModel, shouldReturn := mockedSDK(ctx, &resp.Diagnostics, plan.Timeouts)
-	if shouldReturn {
-		return
+	summary, detail := r.createCluster(ctx, &plan, diags, resp)
```
lantoli: why not createCluster adds errors to diags (passing pointer) and doesn't need to return anything, and we check like previously: if diags.HasError() {...

in that way it could also potentially add warnings

.github/workflows/acceptance-tests-runner.yml
EspenAlbert: Thanks. Deleting

internal/common/constant/error_codes.go
EspenAlbert: Ok. Removing

internal/service/advancedclustertpf/resource.go
EspenAlbert: The problem is that you cannot do 
```go
return diags.AddError(summary, detail)
```
You will have to do it on two lines:
```go
diags.AddError(summary, detail)
return
```

As these functions have ~ 8 different return positions, I thought it was more readable to return summary, detail in one line.

WDYT?

Remembering to do the `return` is an easy mistake that we don't have any linting for yet

lantoli: i see your point but I would go for removing  summary and detail from the response. 

createCluster can add logs, e.g. from the call to NewAtlasReq. 

I think it's confusing and can be error prone that we return empty summary and detail but we might have filled diags so there is actually an error. There are different ways to raise an error.

Also it's more "elegant" as main operations will look similar:

```
func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
	var plan TFModel
	diags := &resp.Diagnostics
	if diags.HasError() {
		return
	}
        r.createCluster(ctx, &plan, diags, resp)
}
```

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+func (r *rs) readCluster(ctx context.Context, model *TFModel, state *tfsdk.State, diags *diag.Diagnostics) {
+	clusterName := model.Name.ValueString()
+	projectID := model.ProjectID.ValueString()
+	api := r.Client.AtlasV220240805.ClustersApi
+	readResp, _, err := api.GetCluster(ctx, projectID, clusterName).Execute()
 	if err != nil {
-		diags.AddError("errorCreate", fmt.Sprintf(errorCreate, err.Error()))
-		return nil, true
+		if admin.IsErrorCode(err, ErrorCodeClusterNotFound) {
+			state.RemoveResource(ctx)
```
lantoli, line 194: RemoveResource should only be called from Read,  readCluster is also called from Update: https://developer.hashicorp.com/terraform/plugin/framework/resources/update

```
Caveats:
An error is returned if the response state has the RemoveResource() method called. This method is not valid during update. Return an error if the resource is no longer exists.
```

internal/service/advancedclustertpf/resource.go
```diff
@@ -55,40 +68,24 @@ func (r *rs) Schema(ctx context.Context, req resource.SchemaRequest, resp *resou
...
-			return
-		}
-		resp.Diagnostics.Append(resp.State.Set(ctx, tfModel)...)
-	} else {
-		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
-	}
+	r.readCluster(ctx, &state, &resp.State, diags)
 }
 
 func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
lantoli: I love how Create and Read are, few lines and high level, delegating in the function doing the hard work.   We might want to use a similar approach for Update, and even Delete.

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	if diags.HasError() {
 		return
 	}
-	name := state.Name.ValueString()
-	clusterID := state.ClusterID.ValueString()
-	if clusterID == "" {
-		resp.Diagnostics.AddError("errorDelete", fmt.Sprintf(errorDelete, name, "clusterID is empty"))
+	clusterName := state.Name.ValueString()
+	projectID := state.ProjectID.ValueString()
+	if clusterName == "" {
```
lantoli: when can clusterName be empty in Delete? can projectID also be empty?

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+		resp.Diagnostics.AddError("errorDelete", fmt.Sprintf(errorDelete, clusterName, "clusterName is empty"))
 		return
 	}
+	api := r.Client.AtlasV2.ClustersApi
```
lantoli, line 132: i like this approach of api var

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	_, err := api.DeleteCluster(ctx, projectID, clusterName).Execute()
+	if err != nil {
+		diags.AddError("errorDelete", fmt.Sprintf(errorDelete, clusterName, err.Error()))
```
lantoli, line 135: do we want literal "errorDelete" as the summary? we might want a more "human" text.  similar for others

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	}
+	_, summary, detail := r.awaitChanges(ctx, &state.Timeouts, diags, projectID, clusterName, "delete")
```
lantoli: similar to previous case I would also remove summary/detail,    we're also adding errors to diags in awaitChanges, so it's confusing to have 2 ways to raise errors

internal/service/advancedclustertpf/resource.go
EspenAlbert: I tried to make the change, and I agree it looks better and is more error safe.
3d3ebe5e2f42206b4d660685807e703761f05a22

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
 }
 
-func mockedSDK(ctx context.Context, diags *diag.Diagnostics, timeout timeouts.Value) (*TFModel, bool) {
-	sdkResp, err := ReadClusterResponse()
+func (r *rs) createCluster(ctx context.Context, plan *TFModel, diags *diag.Diagnostics, resp *resource.CreateResponse) (summary, detail string) {
+	sdkReq := NewAtlasReq(ctx, plan, diags)
+	if diags.HasError() {
+		return "", ""
+	}
+	api := r.Client.AtlasV220240805.ClustersApi
```
lantoli, line 150: I think it's more clear to use api only for AtlasV2, i would use apiSomething for others similar to apiLegacy.  What's special in AtlasV220240805 that we can't use AtlasV2? so we can name it with that

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
 	if diags.HasError() {
 		return
 	}
+	api := r.Client.AtlasV220240805.ClustersApi
 	patchReq := update.PatchPayloadTpf(ctx, diags, &state, &plan, NewAtlasReq)
+	projectID := plan.ProjectID.ValueString()
+	clusterName := plan.Name.ValueString()
+	var cluster *admin20240805.ClusterDescription20240805
+	var advConfig *admin20240805.ClusterDescriptionProcessArgs20240805
```
lantoli: similar comment as in api, if we using "legacy" for admin20240530, i would use something for admin20240805

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
 	if diags.HasError() {
 		return
 	}
+	api := r.Client.AtlasV220240805.ClustersApi
 	patchReq := update.PatchPayloadTpf(ctx, diags, &state, &plan, NewAtlasReq)
+	projectID := plan.ProjectID.ValueString()
+	clusterName := plan.Name.ValueString()
```
lantoli: nit: use var (...) block when declaring multiple variables, also valid for vars with :=

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
 	if diags.HasError() {
 		return
 	}
+	api := r.Client.AtlasV220240805.ClustersApi
 	patchReq := update.PatchPayloadTpf(ctx, diags, &state, &plan, NewAtlasReq)
+	projectID := plan.ProjectID.ValueString()
```
lantoli: Update must fail if projectID or clusterName change, we don't want to delete the current cluster and create a new one.

we don't want stringplanmodifier.RequiresReplace(),  in name in schema.

I think we should check it here in Update operation

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	var advConfig *admin20240805.ClusterDescriptionProcessArgs20240805
+	var legacyAdvConfig *admin20240530.ClusterDescriptionProcessArgs
+	var err error
 	if patchReq != nil {
-		err := StoreUpdatePayload(patchReq)
-		if err != nil {
-			diags.AddError("error storing update payload", fmt.Sprintf("error storing update payload: %s", err.Error()))
+		pauseAfter := false
+		if patchReq.Paused != nil && patchReq.GetPaused() {
+			// More changes than pause, need to pause after
```
lantoli, line 107: is first part of the comment correct in the 2 places?  // More changes than pause

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	}
+	summary, detail := r.convertClusterAddAdvConfig(ctx, nil, nil, readResp, model.Timeouts, diags, state)
+	if summary != "" || detail != "" {
+		diags.AddError(summary, detail)
 	}
-	tfNewModel := NewTFModel(ctx, sdkResp, timeout, diags)
-	sdkAdvConfig, err := ReadClusterProcessArgsResponse()
+}
+func (r *rs) updateAndWait(ctx context.Context, patchReq *admin20240805.ClusterDescription20240805, diags *diag.Diagnostics, tfModel *TFModel) *admin20240805.ClusterDescription20240805 {
+	api := r.Client.AtlasV220240805.ClustersApi
```
lantoli: similar to other comments, use apiSomething for the relevant thing about AtlasV220240805

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
-	sdkAdvConfigLegacy, err := ReadClusterProcessArgsResponseLegacy()
+	return cluster
+}
+
+func (r *rs) awaitChanges(ctx context.Context, t *timeouts.Value, diags *diag.Diagnostics, projectID, clusterName, changeReason string) (cluster *admin20240805.ClusterDescription20240805, summary, detail string) {
+	var timeoutDuration time.Duration
+	var localDiags diag.Diagnostics
+	var targetState = retrystrategy.RetryStrategyIdleState
+	var extraPending = []string{}
+	switch changeReason {
```
lantoli: we might want to use consts for changeReason values, also when calling awaitChange like:
```
// use const instead of "update"
cluster, summary, detail := r.awaitChanges(ctx, &tfModel.Timeouts, diags, projectID, clusterName, "update")
```

lantoli: although if it's only to get the timeout, ok to leave as it is at the moment. we might also want to pass the specific timeout instead of all timeouts, although i see changeReason is also used in an error text, and also in a condition

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	case "create":
+		timeoutDuration, localDiags = t.Create(ctx, defaultTimeout)
+		diags.Append(localDiags...)
```
lantoli: don't we need to check diags.HasError in each case?

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
-	}
-
-	// TODO: Use requests to do actual updates with Admin API
-	tfNewModel, shouldReturn := mockedSDK(ctx, diags, plan.Timeouts)
-	// TODO: keep project_id and name from plan to avoid overwriting for move_state tests. We should probably do the same with the rest of attributes
-	tfNewModel.Name = plan.Name
-	tfNewModel.ProjectID = plan.ProjectID
-	if shouldReturn {
-		return
+	if cluster != nil {
```
lantoli: i think it's a bit hard to follow what happens to cluster var, maybe we could use different cluster vars instead of the same one across the full Update func.

for instance not sure when the else happens here and we  call readCluster

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+		return nil, "", ""
+	}
+	cluster, ok := clusterAny.(*admin20240805.ClusterDescription20240805)
+	if !ok {
+		return nil, "errorAwaitingCluster", fmt.Sprintf(errorCreate, "unexpected type from WaitForStateContext")
+	}
+	return cluster, "", ""
+}
+
+func (r *rs) convertClusterAddAdvConfig(ctx context.Context, legacyAdvConfig *admin20240530.ClusterDescriptionProcessArgs, advConfig *admin20240805.ClusterDescriptionProcessArgs20240805, cluster *admin20240805.ClusterDescription20240805, resourceTimeouts timeouts.Value, diags *diag.Diagnostics, state *tfsdk.State) (summary, detail string) {
```
lantoli: same as other comments, i wouldn't use summary/detail

lantoli: does convertClusterAddAdvConfig do more than doing the advanced config call? why does it have "convertCluster" in the name?

internal/common/retrystrategy/retry_state.go
```diff
@@ -1,18 +1,20 @@
...
-	RetryStrategyPausedState     = "PAUSED"
-	RetryStrategyUpdatingState   = "UPDATING"
-	RetryStrategyDeletingState   = "DELETING"
-	RetryStrategyInitiatingState = "INITIATING"
-	RetryStrategyIdleState       = "IDLE"
-	RetryStrategyFailedState     = "FAILED"
-	RetryStrategyActiveState     = "ACTIVE"
-	RetryStrategyDeletedState    = "DELETED"
-
+	RetryStrategyActiveState            = "ACTIVE"
```
lantoli, line 4: nit: we might want to use enums, e.g.: https://www.sohamkamani.com/golang/enums/

internal/service/advancedclustertpf/model_ClusterDescription20240805.go
```diff
@@ -120,7 +120,8 @@ func NewReplicationSpecsObjType(ctx context.Context, input *[]admin.ReplicationS
 
 func NewTagsObjType(ctx context.Context, input *[]admin.ResourceTag, diags *diag.Diagnostics) types.Set {
 	if input == nil {
-		return types.SetNull(TagsObjType)
+		// API Response not consistent, using empty array instead of null
```
lantoli: can you elaborate the inconsistencies here?

internal/service/advancedclustertpf/resource_unit_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+const (
+	resourceName        = "mongodbatlas_advanced_cluster.test"
+	processResponseOnly = "processResponseOnly"
+	projectID           = "111111111111111111111111"
+	clusterName         = "test"
+)
+
+func TestMockAdvancedCluster_replicaset(t *testing.T) {
+	var (
```
lantoli: nit: const instead of var, except fullUpdateResumed and vars

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +96,230 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
+	cluster, summary, detail := r.awaitChanges(ctx, &tfModel.Timeouts, diags, projectID, clusterName, "update")
+	if summary != "" || detail != "" {
+		diags.AddError(summary, detail)
+		return nil
 	}
-	sdkAdvConfigLegacy, err := ReadClusterProcessArgsResponseLegacy()
+	return cluster
+}
+
+func (r *rs) awaitChanges(ctx context.Context, t *timeouts.Value, diags *diag.Diagnostics, projectID, clusterName, changeReason string) (cluster *admin20240805.ClusterDescription20240805, summary, detail string) {
```
lantoli: nit: does awaitChanges belong better to resource.go or state_transition.go?

internal/service/advancedclustertpf/resource_unit_test.go
```diff
@@ -0,0 +1,332 @@
...
+				ImportStateVerify:                    true,
+				ImportStateVerifyIdentifierAttribute: "name",
+			},
+		},
+	}
+	unit.MockTestCaseAndRun(t, vars, &unit.MockHTTPDataConfig{AllowMissingRequests: true, AllowReReadGet: true}, &testCase)
+}
+
+func shortenRetries() {
+	advancedclustertpf.RetryMinTimeout = 1 * time.Second
```
lantoli: changing global vars 😱    i see how it helps here, but somebody could changed them from Prod code by mistake.  let's keep at the moment if there is not a safer way and we don't need public global vars.

also tests running in parallel might modify them concurrently

internal/service/advancedclustertpf/resource.go
```diff
@@ -55,40 +68,21 @@ func (r *rs) Schema(ctx context.Context, req resource.SchemaRequest, resp *resou
 func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	var plan TFModel
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
```
lantoli: swap lines like in the other operations:
```
diags := &resp.Diagnostics
diags.Append(req.Plan.Get(ctx, &plan)...)
if diags.HasError() {
	return
}
```

internal/service/advancedclustertpf/resource.go
lantoli: nice

internal/service/advancedclustertpf/testdata/TestMockAdvancedCluster_configSharded.yaml
```diff
@@ -0,0 +1,474 @@
+step_count: 3
```
lantoli: is step_count really needed? it seems to be the number of diff_requests

internal/testutil/unit/http_mocker.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	err = tracker.initStep()
+	require.NoError(t, err)
+	for _, method := range []string{"GET", "POST", "PUT", "DELETE", "PATCH"} {
+		myTransport.RegisterRegexpResponder(method, regexp.MustCompile(".*"), tracker.receiveRequest(method))
+	}
+	return mockTransport, tracker.checkStepRequests
+}
+
+var _versionDatePattern = regexp.MustCompile(`(\d{4}-\d{2}-\d{2})`)
```
lantoli: why does it start with underscore?

lantoli (approved):
LGTM, great job, some non-blocking comments

internal/service/advancedclustertpf/resource.go
EspenAlbert: It is done imediately after the switch

internal/service/advancedclustertpf/resource.go
EspenAlbert: Updated

internal/service/advancedclustertpf/resource.go
EspenAlbert: Good catch. I believe it is a low chance since read always should be called before update... Fixed with allowNotFOund

internal/service/advancedclustertpf/resource.go
EspenAlbert: It is required, so shouldn't be empty, think it was a check left from the mocked api

internal/service/advancedclustertpf/resource.go
EspenAlbert: Good point.

internal/service/advancedclustertpf/resource.go
EspenAlbert: Let's discuss this offline

internal/service/advancedclustertpf/resource_unit_test.go
EspenAlbert: Would love to know if there is a better way.
Not worried about concurrent modifications as all of them want the same values so the end result is consistent.
Also, I don't believe we will run `Mock` and `Acc` tests together

internal/service/advancedclustertpf/testdata/TestMockAdvancedCluster_configSharded.yaml
EspenAlbert: Not really needed, however, I like it for readability

internal/testutil/unit/http_mocker.go
EspenAlbert: Thanks. Fixed

internal/service/advancedclustertpf/resource.go
EspenAlbert: Update changed in 426f1b19bbeb918db24d8f0249ddf05547d6f97e
WDYT?

internal/service/advancedclustertpf/resource.go
EspenAlbert: Updated in c8f4ac2954f5e9bf45b455e214c7630dcfe51b2a
WDYT?
Cluster == `nil` when there hasn't been any updates to cluster. Should I add a comment?

internal/service/advancedclustertpf/resource.go
EspenAlbert: Yes! Changed in c49418c555785d788f4577e65a9a7d53cef46e16

internal/service/advancedclustertpf/resource.go
EspenAlbert: Keeping this unresolved. Adding to https://jira.mongodb.org/browse/CLOUDP-286481

internal/service/advancedclustertpf/resource.go
EspenAlbert: Will be resolved in https://jira.mongodb.org/browse/CLOUDP-286479

internal/service/advancedclustertpf/resource.go
EspenAlbert: Will be handled in https://jira.mongodb.org/browse/CLOUDP-286481

internal/service/advancedclustertpf/resource.go
EspenAlbert: Hopefully more clear now after refactoring

internal/common/retrystrategy/retry_state.go
EspenAlbert: Keeping as is to avoid more changes

internal/service/advancedclustertpf/model_ClusterDescription20240805.go
EspenAlbert: Updated. Please reopen if it is still unclear

internal/service/advancedclustertpf/resource_unit_test.go
EspenAlbert: prefer keeping it as is instead of change between `var` and `const` but happy to change if you insist :D

internal/service/advancedclustertpf/resource.go
lantoli: i wouldn't do the check then, and let API fails if it's empty

internal/service/advancedclustertpf/resource.go
lantoli: i like it better, thanks

internal/service/advancedclustertpf/testdata/TestMockAdvancedCluster_configSharded.yaml
lantoli: fair enough, in that case I would check that step_count and the counts are the same or raise an error

internal/service/advancedclustertpf/resource.go
lantoli: I like the the latest code more.   I would invert the condition (i think it's clearer) and add the comment, so something like:

```
if cluster == nil { // no cluster updates
	model = r.readCluster(ctx, &plan, &resp.State, diags, false)
diags)
} else {
	model = r.convertClusterAddAdvConfig(ctx, legacyAdvConfig, advConfig, cluster, plan.Timeouts, 	
}
```

Now that I understand better this code, do we really need to call readCluster  if there are no updates?

EspenAlbert: Changed in 5d9c638645429e4ee35ec1822f201f1f63d41df5
I think so, to ensure any computed attribute has their latest value

internal/service/advancedclustertpf/testdata/TestMockAdvancedCluster_configSharded.yaml
EspenAlbert: Added in 34c76516c152c91aa4c2382cbe60fd514392e510

internal/service/advancedclustertpf/resource.go
lantoli: @EspenAlbert not sure it's needed, cc @AgustinBettati @maastha  in case you know.

looking at: https://developer.hashicorp.com/terraform/plugin/framework/resources/update
```
Only successfully modified parts of the resource should be return updated data in the state response.
```

I'm not sure we should update the state if there is not any change.

(I suppose Read was called before Update so computed attributes should be up-to-date)

lantoli: I see we do Read as the last step in Update in the current implementation, but not sure if it's just a precaution or it's really needed.

internal/service/advancedclustertpf/resource.go
```diff
@@ -99,80 +101,200 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
 }
 
 func (r *rs) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
 	conversion.ImportStateProjectIDClusterName(ctx, req, resp, "project_id", "name")
 }
 
-func mockedSDK(ctx context.Context, diags *diag.Diagnostics, timeout timeouts.Value) (*TFModel, bool) {
-	sdkResp, err := ReadClusterResponse()
+func (r *rs) createCluster(ctx context.Context, plan *TFModel, diags *diag.Diagnostics) *TFModel {
+	sdkReq := NewAtlasReq(ctx, plan, diags)
```
AgustinBettati, line 146: in existing implementation during create operation we have 2 validations for attributes that can only be defined after creation (`accept_data_risks_and_force_replica_set_reconfig`) or require specific values (`advanced_configuration.0.oplog_size_mb`). No concern in leaving out atm, but do we have a ticket to revisit existing logic making sure we dont lose these validations or additional business logic?

internal/service/advancedclustertpf/resource.go
```diff
@@ -54,40 +69,27 @@ func (r *rs) Schema(ctx context.Context, req resource.SchemaRequest, resp *resou
...
-		resp.Diagnostics.AddError("errorCreate", fmt.Sprintf(errorCreate, err.Error()))
+	diags := &resp.Diagnostics
+	diags.Append(req.Plan.Get(ctx, &plan)...)
+	if diags.HasError() {
 		return
 	}
-	tfNewModel, shouldReturn := mockedSDK(ctx, &resp.Diagnostics, plan.Timeouts)
-	if shouldReturn {
-		return
+	model := r.createCluster(ctx, &plan, diags)
```
AgustinBettati, line 77: very nit, but why not have the create logic in the root create function? Applies for read/update as well

internal/service/advancedclustertpf/resource_schema.go
```diff
@@ -186,9 +186,12 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "Version of MongoDB that the cluster runs.",
 			},
-			"name": schema.StringAttribute{ // TODO: fail if trying to update
+			"name": schema.StringAttribute{
 				Required:            true,
 				MarkdownDescription: "Human-readable label that identifies this cluster.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
```
AgustinBettati, line 194: nit: we could use NonUpdatableStringAttributePlanModifier to force user to delete and create a new cluster

internal/service/advancedclustertpf/model_ClusterDescription20240805.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
-	"go.mongodb.org/atlas-sdk/v20241113001/admin"
+	"go.mongodb.org/atlas-sdk/v20240805005/admin"
```
AgustinBettati, line 10: Related to this file, would keep in mind we are simplifying a lot of logic in `NewReplicationSpecsObjType` with regards to how we read replication specs in SDK v2. Right now we rely exclusively on order when there are multiple replication specs. No action to take now but more to keep in mind as we progress to a more stable implementation

internal/service/advancedclustertpf/model_ClusterDescription20240805.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
```
AgustinBettati, line 7: nit: in terms of file name can we fix string case? maybe `model_cluster` and `model_advanced_config` for now?

AgustinBettati (commented):
Leaving non-blocking comments, struggled a bit with the review but great to see we are getting close to a first implementation

internal/service/advancedclustertpf/model_ClusterDescription20240805.go
EspenAlbert: The advantage of having them identical to SDK model is the simplified auto-generation and using a single admin version per model_xxx file.
Added it as a point to: [the next Jira task](https://jira.mongodb.org/browse/CLOUDP-286479)

internal/service/advancedclustertpf/model_ClusterDescription20240805.go
EspenAlbert: Thanks. Noted for next Jira issue.

internal/service/advancedclustertpf/resource.go
EspenAlbert: `accept_data_risks_and_force_replica_set_reconfig` is solved with a plan modifier in the schema.
Adding `oplog_size_mb` to [jira task](https://jira.mongodb.org/browse/CLOUDP-286481)

internal/service/advancedclustertpf/resource.go
EspenAlbert: Trying to keep a consistent higher abstraction in CRUD methods to enhance re-use and make navigation simpler.
(Quite subjective, not sured it has the desired effect 😅)

internal/service/advancedclustertpf/resource_schema.go
EspenAlbert: [Discussion happening, tl;dr: We might keep existing behavior](https://docs.google.com/document/d/1QxrB1BOiMCvUjwAsllG0M3rHKl6ymKJOEF6L5aNYJmE/edit?tab=t.0#heading=h.lrcstpu2gdyw)

internal/service/advancedclustertpf/resource.go
AgustinBettati: > Read as the last step in Update in the current implementation

This is how we handle all our SDK v2 implementations, but dont believe we do this in TPF. We tend to populate the resulting state in update operation using the response of the API. Could we use the resulting model obtained from `AwaitChanges`? This effectively does multiple read calls.

lantoli: cc @EspenAlbert

EspenAlbert: We do this in this implementation. It is only when exclusively advanced_configuration has been updated we are doing an extra read.
I'll experiment with not doing the read and only update adv_config and see if it also works 👍

PR #2890, opened 2024-12-13
feat: Implements `mongodbatlas_stream_privatelink_endpoint` resource

oarbusi:
## Description

 Implements `mongodbatlas_stream_privatelink_endpoint` resource.

Acceptance and migration tests are not included in this PR. They will be added as part of CLOUDP-289961. 
Link to any related issue(s): CLOUDP-289662

## Type of change:

## Required Checklist:


jwilliams-mongo (approved):

internal/service/streamprivatelinkendpoint/model.go
```diff
@@ -4,28 +4,55 @@ package streamprivatelinkendpoint
...
 func NewAtlasReq(ctx context.Context, plan *TFModel) (*admin.StreamsPrivateLinkConnection, diag.Diagnostics) {
-	// var tfList []complexArgumentData
-	// resp.Diagnostics.Append(plan.ComplexArgument.ElementsAs(ctx, &tfList, false)...)
-	// if resp.Diagnostics.HasError() {
-	// 	return nil, diagnostics
-	// }
-	return &admin.StreamsPrivateLinkConnection{}, nil
+	result := &admin.StreamsPrivateLinkConnection{
+		DnsDomain:         plan.DnsDomain.ValueStringPointer(),
+		DnsSubDomain:      &[]string{},
```
AgustinBettati, line 39:
```suggestion
```

internal/service/streamprivatelinkendpoint/resource.go
```diff
@@ -27,125 +39,137 @@ type rs struct {
...
-	// TODO: process response into new terraform state
-	// newStreamPrivatelinkEndpointModel, diags := NewTFModel(ctx, apiResp)
-	// if diags.HasError() {
-	// 	resp.Diagnostics.Append(diags...)
-	// 	return
-	// }
-	// resp.Diagnostics.Append(resp.State.Set(ctx, newStreamPrivatelinkEndpointModel)...)
+	if plan.DnsSubDomain.IsNull() && len(newStreamPrivatelinkEndpointModel.DnsSubDomain.Elements()) == 0 {
+		newStreamPrivatelinkEndpointModel.DnsSubDomain = types.ListNull(types.StringType)
+	}
```
AgustinBettati, line 82: q: Do we need to rely on the plan value? Could we potentially always store null if api response is empty?

internal/service/streamprivatelinkendpoint/resource.go
```diff
@@ -27,125 +39,137 @@ type rs struct {
...
+		return
+	}
 
-	// TODO: make PATCH request to Atlas API and handle error in response
-	// connV2 := r.Client.AtlasV2
-	// if err != nil {
-	//	resp.Diagnostics.AddError("error updating resource", err.Error())
-	//	return
-	//}
+	if state.DnsSubDomain.IsNull() && len(newStreamPrivatelinkEndpointModel.DnsSubDomain.Elements()) == 0 {
```
AgustinBettati, line 114: same question as above

internal/service/streamprivatelinkendpoint/resource_schema.go
```diff
@@ -31,7 +31,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "Interface endpoint ID that is created from the service endpoint ID provided.",
 			},
-			"provider": schema.StringAttribute{
+			"provider_name": schema.StringAttribute{
```
AgustinBettati, line 34: curious about this naming change given the API defines it as `provider`

internal/service/streamprivatelinkendpoint/state_transition.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if privateEndpointResp, ok := result.(*admin.StreamsPrivateLinkConnection); ok {
+		return privateEndpointResp, nil
+	}
+	return nil, errors.New("did not obtain valid result when waiting for state transition")
+}
+
+func refreshFunc(ctx context.Context, projectID, connectionID string, client admin.StreamsApi) retry.StateRefreshFunc {
```
AgustinBettati, line 63: we do have this logic duplicated in various resources, a very nice follow up could be extracting a more generic function with common logic.

internal/service/streamprivatelinkendpoint/resource_schema.go
oarbusi: provider is a reserved word which can not be used as an attribute name, that is why I changed it

internal/service/streamprivatelinkendpoint/model.go
oarbusi: currently the API has an incorrect behavior where the API call fails if dnssubdomain is not sent, that is why I am explicitly setting it to an empty slice. Will change it back once API is fixed

internal/service/streamprivatelinkendpoint/state_transition.go
oarbusi: Agreed, will do it further along the project

internal/service/streamprivatelinkendpoint/resource.go
oarbusi: API returns an empty list instead of null when there are no dnsSubDomains defined, causing the following error in the provider: 
```
Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to mongodbatlas_stream_privatelink_endpoint.test, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new
│ value: .dns_sub_domain: was null, but now cty.ListValEmpty(cty.String).
│ 
```
this logic I added is similar to the `tags` attribute in other resources where we handle equally the null and empty list. Let me know if it makes sense

AgustinBettati: Makes sense. I was more curious what are the cons of not relying on the `plan.DnsSubDomain` value, so if apiResp is nil or empty array we store nil in the state. The benefit I see from this is we remove this custom logic that lives outside of conversion logic and relies on the terraform plan value.

internal/service/streamprivatelinkendpoint/resource_schema.go
AgustinBettati: Was not aware, thanks for clarifying. Worth mentioning in technical design for future reference, definitely an edge case to keep in mind for automation.

internal/service/streamprivatelinkendpoint/resource.go
oarbusi: if we rely solely in the api reponse and we store nil in the state when apiResp is nil or empty array there will be inconsistent results, for example, if we set `dns_sub_domain = []`, and we set nil in the state, this will happen:
```
Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to mongodbatlas_stream_privatelink_endpoint.test, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new
│ value: .dns_sub_domain: was cty.ListValEmpty(cty.String), but now null.
│ 
```

AgustinBettati: I see, I guess we will have to continue supporting this additional logic. Only other alternative is that the API is capable for returning nil or [] depending on what is sent in the request, but I doubt this is feasible.

AgustinBettati (approved):
LGTM

PR #2960, opened 2025-01-13
ci: Adds clean-atlas-org action and related tests for project cleanup

EspenAlbert:
## Description

Adds clean-atlas-org action and related tests for project cleanup

Link to any related issue(s): CLOUDP-288937
Example runs with DRY_RUN=true flag: 
- [run1](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/12751798850/job/35539713618)
- [run2](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/12751975695)

## Type of change:

## Required Checklist:


.github/templates/clean-atlas-org/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: clean-atlas-org
```
EspenAlbert, line 1: Using a template to also call this workflow from the CFN repo

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+func TestCleanProjectAndClusters(t *testing.T) {
+	client := acc.ConnV2()
+	ctx := context.Background()
+	cleanOrg, _ := strconv.ParseBool(os.Getenv("MONGODB_ATLAS_CLEAN_ORG"))
+	if !cleanOrg {
+		t.Skip("skipping test; set MONGODB_ATLAS_CLEAN_ORG=true to run")
+	}
+	dryRun, _ := strconv.ParseBool(os.Getenv("DRY_RUN"))
+	onlyZeroClusters, _ := strconv.ParseBool(os.Getenv("MONGODB_ATLAS_CLEAN_ONLY_0_CLUSTERS"))
```
EspenAlbert, line 43: This 5 hours ago is a bit arbitrary, but I expect no test will keep on running for more than 5 hours.
This is to avoid in-progress tests to be accidentally killed

.github/templates/clean-atlas-org/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+description: 'Cleans up an Atlas organization bot projects'
+
+inputs:
+  repo-path:
+    description: 'Path to the terraform-provider-mongodbatlas repository'
+    default: ''
+  dry-run:
+    description: 'Dry run the cleanup'
+    default: 'true'
```
EspenAlbert: should change to `false` before merging

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+	skipProjectsAfter := time.Now().Add(-5 * time.Hour)
```
EspenAlbert: This 5 hours ago is a bit arbitrary, but I expect no test will keep on running for more than 5 hours.
This is to avoid in-progress tests to be accidentally killed

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+	maxDeleteCount := 250
```
EspenAlbert: this is the maximum group_count for an org before hitting a `MAX_GROUPS_PER_ORG_EXCEEDED` limit

.github/templates/clean-atlas-org/action.yaml
lantoli: will you also change Test Suite to use this cleanup?

Makefile
```diff
@@ -32,6 +32,12 @@ help:
 build: fmt fmtcheck ## Generate the binary in ./bin
 	go build -ldflags "$(LINKER_FLAGS)" -o $(DESTINATION)
 
+.PHONY: clean-atlas-org
+clean-atlas-org: clean-atlas-org ## Run a test to clean all projects and pending resources in an Atlas org, supports export DRY_RUN=false (default=true)
+	@$(eval export MONGODB_ATLAS_CLEAN_ORG?=true)
```
lantoli, line 37: nit (not strong opinion): to have these default values at the beginning of the file as other default values.   although also I understand that these vars are very specific and only useful for this goal.

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/atlas-sdk/v20241113004/admin"
+)
+
+var (
+	botProjectPrefixes = []string{
+		"cfn-test-bot-",
+		"test-acc-tf-p-",
+	}
+	blessedPrefixes = []string{
```
lantoli: nit: ignoredPrefixes / skippedPrefixes / keptPrefixes, bless sounds a bit weird in this context

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+		if skipReason != "" {
+			t.Logf("skip project %s, reason: %s", projectName, skipReason)
+			continue
+		}
+		projectID := p.GetId()
+		projectsToDelete[projectName] = projectID
+	}
+	for name, projectID := range projectsToDelete {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
lantoli, line 102: interesting approach to use test, alternatively you could use a normal go program an use goroutines for concurrency, but for this use case I think this is fine

.github/templates/clean-atlas-org/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+  clean-only-0-clusters:
```
marcosuma: nit
```suggestion
  clean-only-when-no-clusters:
```

lantoli (approved):
LGTM

internal/testutil/clean/org_clean_test.go
marcosuma: should the 5 hours be something that can be set as env variable? otherwise might be better to simply define it as a constant

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+	cleanOrg, _ := strconv.ParseBool(os.Getenv("MONGODB_ATLAS_CLEAN_ORG"))
+	if !cleanOrg {
+		t.Skip("skipping test; set MONGODB_ATLAS_CLEAN_ORG=true to run")
+	}
+	dryRun, _ := strconv.ParseBool(os.Getenv("DRY_RUN"))
+	onlyZeroClusters, _ := strconv.ParseBool(os.Getenv("MONGODB_ATLAS_CLEAN_ONLY_0_CLUSTERS"))
+	skipProjectsAfter := time.Now().Add(-5 * time.Hour)
+	maxDeleteCount := 250
+	projects, err := dsschema.AllPages(ctx, func(ctx context.Context, pageNum int) (dsschema.PaginateResponse[admin.Group], *http.Response, error) {
+		return client.ProjectsApi.ListProjects(ctx).ItemsPerPage(100).PageNum(pageNum).Execute()
```
marcosuma: nit: make 100 a const

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
+package clean_test
```
lantoli, line 1: this seems more a tool than part of the provider and belongs to `tools` folder, but I suppose it's harder to use it as tests then

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+				t.Logf("peering %s already requested deletion", peerID)
+				continue
+			}
+			require.NoError(t, err)
+		}
+	}
+	return len(peeringResults)
+}
+
+func removeDataLakePipelines(ctx context.Context, t *testing.T, dryRun bool, client *admin.APIClient, projectID string) int {
```
lantoli, line 293: nit: remove*** seem very similar so there might be some refactor opportunities to extract common logic

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+			t.Logf("skip project %s, reason: %s", projectName, skipReason)
+			continue
+		}
+		projectID := p.GetId()
+		projectsToDelete[projectName] = projectID
+	}
+	for name, projectID := range projectsToDelete {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			changes := removeProjectResources(ctx, t, dryRun, client, projectID)
```
marcosuma, line 103: should you skip this call if `onlyZeroClusters` is true?

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+			if changes != "" {
+				t.Logf("project %s %s", name, changes)
+			}
+			for i := range 10 {
```
marcosuma: not a Go expert, but I think there should be a more standard way to write this retry strategy :) 

at minimum, 10 should be a const

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+				if attempt > 1 {
+					time.Sleep(60 * time.Second)
+				}
+				t.Logf("attempt %d to delete project %s", attempt, name)
+				if dryRun {
+					return
+				}
+				err = deleteProject(ctx, client, projectID)
+				if err == nil {
+					return
```
marcosuma, line 119: in the previous for cycle, you used `break` for something similar. Am I right? don't you want to just end this internal for cycle and continue with other projects to delete?

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	return ""
+}
+func deleteProject(ctx context.Context, client *admin.APIClient, projectID string) error {
+	_, _, err := client.ProjectsApi.DeleteProject(ctx, projectID).Execute()
+	if err == nil || admin.IsErrorCode(err, "PROJECT_NOT_FOUND") {
+		return nil
+	}
+	_, _, err = client.ProjectsApi.DeleteProject(ctx, projectID).Execute()
+	return err
```
marcosuma: not really understanding why this is written two times?

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func removeProjectResources(ctx context.Context, t *testing.T, dryRun bool, client *admin.APIClient, projectID string) string {
```
marcosuma, line 165: this method risk to become very big with the assumption that a lot of things are associated to a project. Why is this even happening in the first place? Why those resources are not being eliminated?

.github/workflows/cleanup-test-env.yml
```diff
@@ -12,33 +12,23 @@ jobs:
...
-      - name: Cleanup test env network
-        shell: bash
+      - uses: ./.github/templates/clean-atlas-org
+        name: Clean Atlas Org QA
         env:
           MONGODB_ATLAS_PUBLIC_KEY: ${{ secrets.MONGODB_ATLAS_PUBLIC_KEY_CLOUD_QA }}
           MONGODB_ATLAS_PRIVATE_KEY: ${{ secrets.MONGODB_ATLAS_PRIVATE_KEY_CLOUD_QA }}
           MONGODB_ATLAS_ORG_ID: ${{ vars.MONGODB_ATLAS_ORG_ID_CLOUD_QA }}
-          MONGODB_ATLAS_OPS_MANAGER_URL: ${{ vars.MONGODB_ATLAS_BASE_URL_QA }}
-        run: ./scripts/cleanup-test-env.sh
```
AgustinBettati, line 44: should we remove the cleanup shell script that was being used before?

internal/testutil/clean/org_clean_test.go
AgustinBettati: I agree that seeing this within a test file was confusing. Is there a reason we need to define it as a test?

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,238 @@
...
+			}
+			for i := range 10 {
+				attempt := i + 1
+				if attempt > 1 {
+					time.Sleep(60 * time.Second)
+				}
+				t.Logf("attempt %d to delete project %s", attempt, name)
+				if dryRun {
+					return
+				}
```
AgustinBettati, line 116: Is leaving the dry run option valuable? Thinking we could remove this option altogether to simplify

Makefile
EspenAlbert: I don't see the value in having these defaults at the beginning when they are only used for this goal, but maybe I'm missing something? 😅

internal/testutil/clean/org_clean_test.go
EspenAlbert: Simplifies parallelization and we can use other internal packages within the repo.

internal/testutil/clean/org_clean_test.go
EspenAlbert: That is done in `projectSkipReason`

internal/testutil/clean/org_clean_test.go
EspenAlbert: I think it is valuable to double-check you are really sure what you are doing

internal/testutil/clean/org_clean_test.go
lantoli: we're already using internal packages from tools, e.g. [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-294337_state_upgrader/tools/jira-release-version/main.go#L12)

goroutines (and maybe channels) can be used, it's harder than tests but is not so bad, e.g.: [here](https://github.com/lantoli/AdventOfCode/blob/master/2024_go/13_claw.go#L37)

I agree that this test way is easier to do, so not sure about priority to spend effort on moving it as a regular tool

internal/testutil/clean/org_clean_test.go
EspenAlbert: `return` finish the test for the "current" project. Not sure what you mean

internal/testutil/clean/org_clean_test.go
EspenAlbert: True. Removed

internal/testutil/clean/org_clean_test.go
EspenAlbert: I think there can be many reasons, and not something we can 100% avoid.
If we had more retry mechanisms in our provider it would probably happen with fewer intervals.

internal/testutil/clean/org_clean_test.go
EspenAlbert: Changed to constant in 98527571f8dc59aa539d521bd92484084a174b3e

internal/testutil/clean/org_clean_test.go
EspenAlbert: Thanks! Changed to keptPrefixes in 98527571f8dc59aa539d521bd92484084a174b3e

internal/testutil/clean/org_clean_test.go
EspenAlbert: Thanks changed in 98527571f8dc59aa539d521bd92484084a174b3e

internal/testutil/clean/org_clean_test.go
EspenAlbert: true, not sure if it is worth doing now though

internal/testutil/clean/org_clean_test.go
EspenAlbert: true, would love to see a better approach. Changed to const in 98527571f8dc59aa539d521bd92484084a174b3e

.github/workflows/cleanup-test-env.yml
EspenAlbert: Thanks. Removed in 37c891b457cebd5215d931e91d5c33480f6122fd

internal/testutil/clean/org_clean_test.go
maastha: Can we time box trying to use goroutines? If it takes longer then we can keep it like this. 
At the least, I'd leave comments explaining why this is defined as test or at least clarify this is not an actual test to worry about

internal/testutil/clean/org_clean_test.go
```diff
@@ -0,0 +1,244 @@
...
+	skipProjectsAfter := time.Now().Add(-keepProjectsCreatedWithinHours * time.Hour)
+	maxDeleteCount := 250
+	projects, err := dsschema.AllPages(ctx, func(ctx context.Context, pageNum int) (dsschema.PaginateResponse[admin.Group], *http.Response, error) {
+		return client.ProjectsApi.ListProjects(ctx).ItemsPerPage(itemsPerPage).PageNum(pageNum).Execute()
+	})
+	require.NoError(t, err)
+	t.Logf("found %d projects (DRY_RUN=%t)", len(projects), dryRun)
+	projectsToDelete := map[string]string{}
+	for _, p := range projects {
+		if len(projectsToDelete) > maxDeleteCount {
```
maastha: [q] why do we want to limit delete count?

internal/testutil/clean/org_clean_test.go
maastha: Could you share an example when you think this might help to be more cautious? I'm thinking we use "keep" for anything relevant that shouldn't be removed, other than that may be I'm missing a use-case?

internal/testutil/clean/org_clean_test.go
maastha: a lot of times we see `Error running post-test destroy, there may be dangling resources:` in our tests resulting from flaky backends. Some of them could be eliminated with more retry mechanisms like @EspenAlbert  said but some may not. 
I think we can start by tracking this error in our test suite failure file more closely.
added an item to next tech meeting.

internal/testutil/clean/org_clean_test.go
AgustinBettati: I would first question if we have the need of executing in parallel, does it take considerable time when executed sequentially?

lantoli: I would merge this as it is (with @maastha idea about explaining why it is a test), and would consider goroutines in a follow-up PR as a future improvement, if we give priority to this.
@AgustinBettati I think so, we don't want the cleanup process to take 10 minutes x clusters to delete (plus endpoints, etc.)

internal/testutil/clean/org_clean_test.go
marcosuma: Are you trying here to run these deletion in parallel?

internal/testutil/clean/org_clean_test.go
marcosuma: right, I actually didn't see the `t.Run(name, func(t *testing.T) {` before.. and actually don't understand why we're using it (but left same question before, so ok not to answer here).

internal/testutil/clean/org_clean_test.go
EspenAlbert: Yes, I would prefer keeping it as tests.
Then the logs for each project is grouped by the test too.

It can take more than 10 minutes to delete a project with active clusters, since the cluster shutdown can be slow.
I'll work on adding some documentation to this.
Updated in 927e4b9e3ebdaf3adc3ef436b9869caa9769b92c to avoid failure when the test fails to remove a project. 
(I can imagine this happening often due to the long cluster shutdown time, but that doesn't mean we don't want to run our test-suite.yml workflow)

internal/testutil/clean/org_clean_test.go
EspenAlbert: Not really necessary. Removing it.

internal/testutil/clean/org_clean_test.go
EspenAlbert: Yes, each project cleanup runs in its own test that runs in parallel.

internal/testutil/clean/org_clean_test.go
EspenAlbert: Imagine you are cleaning up your own private ORG. Then you might want to delete more than the `bot` projects or at least see which bot projects you have...

Makefile
lantoli: that was my point, that in this case it's ok

lantoli (approved):

maastha (approved):
LGTM

AgustinBettati (approved):
LGTM

lantoli (approved):

PR #2972, opened 2025-01-15
chore: Adds cipher config & default_max_time_ms to advanced_cluster_tpf

maastha:
## Description

Add cipher config & default_max_time_ms to advanced_cluster_tpf

Link to any related issue(s): 
[CLOUDP-291163](https://jira.mongodb.org/browse/CLOUDP-291163)
[CLOUDP-291164](https://jira.mongodb.org/browse/CLOUDP-291164)

## Type of change:

## Required Checklist:


internal/service/advancedclustertpf/schema.go
```diff
@@ -528,6 +529,23 @@ func AdvancedConfigurationSchema(ctx context.Context) schema.SingleNestedAttribu
 				Optional:            true,
 				MarkdownDescription: "fail_index_key_too_long", // TODO: add description
 			},
+			"default_max_time_ms": schema.Int64Attribute{
+				Optional:            true,
```
EspenAlbert, line 533: Are you sure about `Optional` here? If so, the `AddAdvancedConfig` must be modified to support types.Int64PointerValue() for when this is null. Otherwise we get non empty refresh plans

internal/service/advancedclustertpf/schema.go
```diff
@@ -528,6 +529,23 @@ func AdvancedConfigurationSchema(ctx context.Context) schema.SingleNestedAttribu
...
+				MarkdownDescription: "Default time limit in milliseconds for individual read operations to complete. This parameter is supported only for MongoDB version 8.0 and above.",
+				PlanModifiers: []planmodifier.Int64{
+					PlanMustUseMongoDBVersion(8.0, EqualOrHigher),
+				},
+			},
+			"custom_openssl_cipher_config_tls12": schema.SetAttribute{
```
EspenAlbert, line 539: Looks like backend returns an empty list by default. We would either have to force it to `null` based on config value or change this to computed

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -351,10 +347,6 @@ func TestAccClusterAdvancedCluster_advancedConfig_oldMongoDBVersion(t *testing.T
 		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
 		CheckDestroy:             acc.CheckDestroyCluster,
 		Steps: []resource.TestStep{
-			{
```
maastha: moved test case to schema_test.go

internal/service/advancedclustertpf/schema.go
maastha: modified AddAdvancedConfig to use types.Int64PointerValue()

internal/service/advancedclustertpf/schema.go
maastha: forcing it to be null if no elements

internal/service/advancedclustertpf/resource.go
```diff
@@ -356,25 +359,25 @@ func (r *rs) applyAdvancedConfigurationChanges(ctx context.Context, diags *diag.
 		advConfig       *admin.ClusterDescriptionProcessArgs20240805
 		legacyAdvConfig *admin20240530.ClusterDescriptionProcessArgs
 	)
-	patchReqProcessArgs := update.PatchPayloadTpf(ctx, diags, &state.AdvancedConfiguration, &plan.AdvancedConfiguration, NewAtlasReqAdvancedConfiguration)
```
maastha: swapped the order in which calls are made to the old/new processArgs API. This is currently a workaround to ensure the new API call is made later. This is because there is a bug where defaultMaxTimeMS (in the new API version) is reset to `null` if the PATCH call to any of the API versions doesn't contain a value for defaultMaxTimeMS.
This bug is being addressed in https://jira.mongodb.org/browse/CLOUDP-295775 and should be merged this week (slack thread in the ticket https://jira.mongodb.org/browse/CLOUDP-291164).

After the bug is addressed this swap is no longer required but I think keeping it as is isn't a problem and should not be a blocker for this PR

marcosuma: should we leave this comment in the code?

maastha: They're planning to address the ticket today itself, I'll add a comment if by the time this is ready to merge the fix is still not in. Otherwise shouldn't be required as no behavior change.

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -404,7 +394,7 @@ func TestAccClusterAdvancedCluster_advancedConfig(t *testing.T) {
 			CustomOpensslCipherConfigTls12: &[]string{"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"},
 		}
 		processArgsUpdatedCipherConfig = &admin.ClusterDescriptionProcessArgs20240805{
-			DefaultMaxTimeMS: conversion.IntPtr(65),
+			DefaultMaxTimeMS: conversion.IntPtr(67),
```
lantoli: I'm curious why we're changing this value

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
```diff
@@ -44,9 +46,25 @@ func AddAdvancedConfig(ctx context.Context, tfModel *TFModel, input *admin.Clust
...
+	} else {
+		advancedConfig.CustomOpensslCipherConfigTls12 = types.SetNull(types.StringType) // required to handle move state
 	}
+
 	objType, diagsLocal := types.ObjectValueFrom(ctx, AdvancedConfigurationObjType.AttrTypes, advancedConfig)
 	diags.Append(diagsLocal...)
 	tfModel.AdvancedConfiguration = objType
 }
+
+func customOpensslCipherConfigTLS12(ctx context.Context, processArgs *admin.ClusterDescriptionProcessArgs20240805, diags *diag.Diagnostics) types.Set {
```
lantoli: nit: in new funcs, we're favoring diags as second arg, after ctx

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
```diff
@@ -44,9 +46,25 @@ func AddAdvancedConfig(ctx context.Context, tfModel *TFModel, input *admin.Clust
...
+	if processArgs != nil && len(*processArgs.CustomOpensslCipherConfigTls12) == 0 {
+		return types.SetNull(types.StringType)
+	}
+
+	customOpensslCipherConfigTLS12, d := types.SetValueFrom(ctx, types.StringType, processArgs.CustomOpensslCipherConfigTls12)
```
lantoli, line 65: i think processArgs can be nil at this point, then it'll panic

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
```diff
@@ -44,9 +46,25 @@ func AddAdvancedConfig(ctx context.Context, tfModel *TFModel, input *admin.Clust
...
+		advancedConfig.CustomOpensslCipherConfigTls12 = types.SetNull(types.StringType) // required to handle move state
 	}
+
 	objType, diagsLocal := types.ObjectValueFrom(ctx, AdvancedConfigurationObjType.AttrTypes, advancedConfig)
 	diags.Append(diagsLocal...)
 	tfModel.AdvancedConfiguration = objType
 }
+
+func customOpensslCipherConfigTLS12(ctx context.Context, processArgs *admin.ClusterDescriptionProcessArgs20240805, diags *diag.Diagnostics) types.Set {
+	if processArgs != nil && len(*processArgs.CustomOpensslCipherConfigTls12) == 0 {
```
lantoli:
```suggestion
	if processArgs == nil || len(*processArgs.CustomOpensslCipherConfigTls12) == 0 {
```

I think we want this condition here, or not sure what we want to do when processArgs is/is not nil

internal/service/advancedclustertpf/resource.go
```diff
@@ -347,6 +349,7 @@ func (r *rs) applyPinnedFCVChanges(ctx context.Context, diags *diag.Diagnostics,
 	return nil
 }
 
+// create - empty state
```
AgustinBettati: was this a left over?

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
maastha: updated

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
maastha: you're right, updated the condition

internal/service/advancedcluster/resource_advanced_cluster_test.go
AgustinBettati: q: are tests in schema_test.go (advancedclustertpf) run for the SDK v2 implementation? Because if not we are actually removing this check for the SDK v2 implementation.

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
lantoli: can we have a test that passes now, but wouldn't have passed before this change?   to make sure the implementation logic is correct now, and we don't change it later by mistake while doing some code changes

marcosuma: Reg tests: in general I went to check and it looks like we only added one test check in this PR, do we need more tests or are we actually leveraging the already existing ones you created before?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -41,7 +41,7 @@ const (
 	ErrorClusterAdvancedSetting    = "error setting `%s` for MongoDB ClusterAdvanced (%s): %s"
 	ErrorAdvancedClusterListStatus = "error awaiting MongoDB ClusterAdvanced List IDLE: %s"
 	ErrorOperationNotPermitted     = "error operation not permitted"
-	ErrorDefaultMaxTimeMinVersion  = "default_max_time_ms can not be set for mongo_db_major_version lower than 8.0"
+	ErrorDefaultMaxTimeMinVersion  = "`advanced_configuration.default_max_time_ms` can only be configured if the mongo_db_major_version is 8.0 or higher"
```
maastha, line 44: ensuring the error messaging is same in both implementations. Required for common tests.

internal/service/advancedclustertpf/resource.go
maastha: yes removed

internal/service/advancedcluster/resource_advanced_cluster_test.go
maastha: you're right, I missed that, updated. Resolving this comment to avoid confusion with my other comment. Please feel free to reopen.

internal/service/advancedclustertpf/resource.go
maastha: This bug has been addressed by upstream, so I've reverted call swap change.

internal/service/advancedcluster/resource_advanced_cluster_test.go
maastha: Reverted this change. I had to add it for the test because in case DefaultMaxTimeMS didn't change value it wouldn't be included in the PATCH request which caused issues as mentioned in my comment above. Now that has been addressed by upstream in https://jira.mongodb.org/browse/CLOUDP-295775 so we don't need this change.

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
maastha: @lantoli Due to this above condition you provided I was able to refactor the original logic a bit. Previously I had to add an additional `else` block such as [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2972/commits/76212a3f7abd0351b30490c3831c3b319d209809#diff-bef1833adcf6508cb5d2efe463fffb92a315ded67bf2743550d5f1bfe178ab9c). Now due to the updated condition we don't need that anymore. If the condition is changed for some reason the move block tests will fail so we should have that covered.

@marcosuma As for tests in general, I think the current tests `TestAccClusterAdvancedCluster_advancedConfig_oldMongoDBVersion`, `TestAccClusterAdvancedCluster_advancedConfig` and `TestAccClusterAdvancedCluster_defaultWrite` provide quite decent coverage for the attributes added in this PR already for null, default or setting/updating values. Please let me know if there's any additional test case I can handle to improve.

EspenAlbert (approved):
Nice! [nit] Remove `WIP` from title. And address [Leo's comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2972#discussion_r1922385772)

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
```diff
@@ -44,9 +46,23 @@ func AddAdvancedConfig(ctx context.Context, tfModel *TFModel, input *admin.Clust
...
 	}
+
+	advancedConfig.CustomOpensslCipherConfigTls12 = customOpensslCipherConfigTLS12(ctx, diags, input) // required to handle move state
 	objType, diagsLocal := types.ObjectValueFrom(ctx, AdvancedConfigurationObjType.AttrTypes, advancedConfig)
 	diags.Append(diagsLocal...)
 	tfModel.AdvancedConfiguration = objType
 }
+
+func customOpensslCipherConfigTLS12(ctx context.Context, diags *diag.Diagnostics, processArgs *admin.ClusterDescriptionProcessArgs20240805) types.Set {
+	if processArgs == nil || len(*processArgs.CustomOpensslCipherConfigTls12) == 0 {
```
EspenAlbert, line 61: I suspect this might fail with inconsistent error if the user has specified `custom_openssl_cipher_config_tls12 = []` but might never be an issue?

marcosuma (approved):

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
maastha: This has been handled now let's continue the conversation in the other thread

internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go
maastha: As discussed offline, I don't think that's a valid case

maastha:
Unrelated TestAccClusterAdvancedClusterConfig_replicationSpecsAutoScaling test is failing. Merging this PR.

lantoli:
@maastha as in yesterday's meeting, I recommend to do a manual migration from sdkv2 to tpf to check that attributes in this feature are migrated correctly,  also speciale cases for arrays and objects for nil, {}, []

PR #2973, opened 2025-01-16
chore: Enable ISS autoscaling support in advancedclustertpf implementation

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-291158

Summary of changes that are included in order of priority:

- Adjust logic in `applyClusterChanges` to ensure that when new sharding configuration is used latest API is called (this was not the case). This enables asymmetric autoscaling to be enabled.
- Simplify `applyClusterChanges` to only rely on one old API (2023-02-01 as in SDK v2), before it had logic calling both 2023-02-01 and 2024-08-05.
- Simplify `applyClusterChanges` to not handle change in paused configuration together with other changes. This aligns with the limitation our [docs mentions](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/advanced_cluster#paused-1) and how SDK v2 behaves.
- Move all sdk related conversion into `advancedclustertpf` package removing duplicate code which was also present in `advancedcluster`, this causes the majority of line diffs.

## Type of change:

## Required Checklist:


internal/service/advancedclustertpf/resource.go
```diff
@@ -136,15 +133,15 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
 		return
 	}
 
-	stateUsingLegacy := usingLegacySchema(ctx, state.ReplicationSpecs, diags)
```
AgustinBettati, line 139: aligning terminology of `ShardingConfig` as mentioned in our docs and sdk v2 implementation

internal/service/advancedclustertpf/resource.go
```diff
@@ -386,83 +383,50 @@ func (r *rs) applyAdvancedConfigurationChanges(ctx context.Context, diags *diag.
...
-			pauseAfter = true
-			patchReq.Paused = nil
-		}
-	} else if patchReq.Paused != nil && !patchReq.GetPaused() {
-		// More changes than pause, need to resume before applying changes
-		if !reflect.DeepEqual(resumeRequest, *patchReq) {
-			patchReq.Paused = nil
-			_ = r.updateAndWait(ctx, &resumeRequest, diags, plan)
-		}
-	}
```
AgustinBettati, line 417: Not finding a reason of handling the paused change in a separate request, and can see it was leading to more complexity.
For the case of `More changes than pause, need to resume before applying changes` we are clear in our docs that paused cluster do not support updates ([reference](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/advanced_cluster#paused-1)).

internal/service/advancedclustertpf/model_sdk_version_conversion.go
```diff
@@ -158,3 +163,81 @@ func convertTag20241023to20240805(tags *[]admin.ResourceTag) *[]admin20240805.Re
...
 	return &result
 }
+
+func ConvertRegionConfigSlice20241023to20240530(slice *[]admin.CloudRegionConfig20240805) *[]admin20240530.CloudRegionConfig {
+	if slice == nil {
+		return nil
+	}
+	cloudRegionSlice := *slice
+	results := make([]admin20240530.CloudRegionConfig, len(cloudRegionSlice))
+	for i := range len(cloudRegionSlice) {
```
lantoli:
```suggestion
	for i := range cloudRegionSlice {
```

it's the same

internal/service/advancedclustertpf/model_sdk_version_conversion.go
```diff
@@ -1,11 +1,16 @@
...
 	admin20240805 "go.mongodb.org/atlas-sdk/v20240805005/admin"
 	"go.mongodb.org/atlas-sdk/v20241113004/admin"
 )
 
-func newLegacyModel(clusterDescription *admin.ClusterDescription20240805) *admin20240805.ClusterDescription20240805 {
+// Conversions from one SDK model version to another are used to avoid duplicating our flatten/expand conversion functions.
+// - These functions must not contain any business logic.
+// - All will be removed once we rely on a single API version.
+
+func ConvertClusterDescription20241023to20240805(clusterDescription *admin.ClusterDescription20240805) *admin20240805.ClusterDescription20240805 {
```
lantoli: can common funcs ConvertRegionConfigSlice20241023to20240530 and ConvertClusterDescription20241023to20240805 be in common.go file or in a file with prefix `common_`  to make it easier to know what is being shared?

helper private funcs can be anywhere though

internal/service/advancedclustertpf/model_sdk_version_conversion.go
```diff
@@ -158,3 +163,81 @@ func convertTag20241023to20240805(tags *[]admin.ResourceTag) *[]admin20240805.Re
 	}
 	return &result
 }
+
+func ConvertRegionConfigSlice20241023to20240530(slice *[]admin.CloudRegionConfig20240805) *[]admin20240530.CloudRegionConfig {
```
lantoli: consider using [CopyModel](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/c84ef81fb8fa223a1ce08db616f72d1536a8974a/internal/common/conversion/model_generation.go#L10C6-L10C15) instead of these convert funcs.  

I can help if CopyModel needs to be adpated / improved.

internal/service/advancedclustertpf/resource.go
lantoli: nice simplification.  mentioning @EspenAlbert in case he sees some problem

lantoli (approved):
LGTM

internal/service/advancedclustertpf/model_sdk_version_conversion.go
AgustinBettati: like the idea of adding the common_ prefix so we dont overload a single file

internal/service/advancedclustertpf/model_sdk_version_conversion.go
AgustinBettati: I am seeing that the current implementation does not support copying nested structs that have a different type: `panic: field has different type: BiConnector`. I am think we might need some recursive logic in CopyModel for the case of nested structs that have different types.

lantoli: yes, that's on purpose, but we can improve it if it fits this use case

AgustinBettati:
running failed advanced_cluster TestAccClusterAdvancedCluster_pausedToUnpaused test just to confirm it is CI related https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/12868995884

internal/service/advancedclustertpf/model_sdk_version_conversion.go
lantoli: thanks!  do you mind to change in acc test the matching from:

```
- 'internal/service/advancedclustertpf/common.go'
```

to 

```
- 'internal/service/advancedclustertpf/common*.go'
```

marcosuma:
@AgustinBettati not sure it's related but TestAccClusterAdvancedCluster_pausedToUnpaused is failing and you were proposing a change in `paused` [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2973#discussion_r1920221097)

AgustinBettati:
@marcosuma confirmed this was unrelated, ran the failing test of the advancedcluster group and it passed ([reference](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2973#issuecomment-2602421632)). Note the paused refactor was in the advanced cluster tpf implementation, for which the test group was successful.

internal/service/advancedclustertpf/resource.go
EspenAlbert: I thought this was an `artifical` limitation in our old implementation.
Removing this logic will not make it possible to:
1. Resume a cluster that has been paused and accidentally has some extra changes.
2. Change a cluster and then pause it in the same apply step

As the user probably can workaround these by ensuring only one change at a time, it seems we are ok with this?

internal/service/advancedclustertpf/resource.go
```diff
@@ -386,83 +383,50 @@ func (r *rs) applyAdvancedConfigurationChanges(ctx context.Context, diags *diag.
...
-	if diags.HasError() {
-		return nil
-	}
-	replicationSpecsUpdated := patchReq.ReplicationSpecs != nil
-	if replicationSpecsUpdated {
-		// Cannot call latest API (2024-10-23 or newer) as it can enable ISS autoscaling
```
EspenAlbert, line 423: Just to confirm. Calling the latest API has no side-effects? Or we are ok with those side effects?

EspenAlbert (approved):
A few questions. But looks good. Great work! Thanks for the improvements :clap:

internal/service/advancedclustertpf/resource.go
AgustinBettati: Correct, I would consider that use case 1 is a current limitation in SDK v2 implementation as well. For case 2 we would be sending sending all request info in a single PATCH (paused `true` + additional changes) for the case of new sharding config, so if API does not support this I believe we can simply return that same error. Again as you mention if paused change is performed in isolated manner there is no risk of errors.

internal/service/advancedclustertpf/resource.go
AgustinBettati: Only side effect 2024-10-23 has is that if POST/PATCH have replication_specs (with autoscaling property) the autoscaling mode will change to SHARD. This is why the current logic ensures if old sharding config is used replication spec related changes are done with old API, and then everything else can be done safely with latest API (similar to SDK v2 implementation).

internal/service/advancedclustertpf/model_sdk_version_conversion.go
AgustinBettati: good point, adjusted

internal/service/advancedclustertpf/resource.go
lantoli: what happens in step 1, we'll send an error or cluster could get in a inconsistent state?

@AgustinBettati @EspenAlbert just to double check, also cc @marcosuma , as it looks like we have the opportunity to allow users to make pause with other changes at the same time (managing this in TF side even if Atlas doesn't support it),  do we want to allow that capability, or we prefer to keep the SDKv2 limitation and don't allow extra changes when pause is updated?

marcosuma: My feeling is that we are removing business logic and adhering to the the API behaviour. This would force users to two operations what they were doing with one. Am I right? What is the strong reason we are making this change? Is it about simplifying our code base?

lantoli: @marcosuma No, right now they need  2 operations,  the improvement in TPF is to allow both changes in 1 operation. 
In general we're maintaining parity with SDKv2 and keep improvements for later, as in other cases like failing if cluster name changes instead of deleting and creating a new one.

I suppose we want to follow the same tenant here and simplify the code, and have the user to need 2 operations as in SDKv2.   

but as the code was already done, we might want to keep the functionality, but we can also do it in the future.


So I think that I probably agree with @AgustinBettati to keep SDKv2 behavior and code simplification, and leave this improvement for the future cc @EspenAlbert

AgustinBettati: Sorry, had overlooked some details in SDK v2 implementation which do simplify one of the use cases.

**Resume a paused cluster and also add additional changes** (no difference):
- SDK v2: Done in same PATCH request, this returns error `Cannot update cluster while it is paused or being paused.`
- TPF current PR: Both paused and other changes in same PATCH request, would have same error as SDK v2.

**Pause a cluster and also add additional changes** (difference):
- SDK v2: Sends pausing request as a follow up request after doing other changes. Avoids error `Cannot update and pause cluster at the same time.`.
- TPF current PR: Both paused and other changes in same PATCH request, has error `Cannot update and pause cluster at the same time.`.

Will follow up with change/revert in TPF so `Pause a cluster and also add additional changes` is supported as in SDK v2 and we ensure no regression. If `Resume a paused cluster and also add additional changes` is also simple will factor in as well (however this not supported in SDK v2).

marcosuma: @AgustinBettati  I'd prioritize for now the "feature parity" so based on your last comment it should really be about making "Pause a cluster and also add additional changes" behave the same

AgustinBettati: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2973/commits/94c90c01f639301491fa18841d93441d401f296b adds capability of `Pause a cluster and also add additional changes`, also modified the existing acceptance test (that runs in SDK v2 and TPF) to capture this. `Resume a paused cluster and also add additional changes` was trivial to include so added as well.

EspenAlbert: Nice @AgustinBettati LGTM

AgustinBettati:
Failing tests are unrelated: `TestAccMockableAdvancedCluster_tenantUpgrade` yet to be fixed in master, and `TestAccClusterAdvancedClusterConfig_symmetricGeoShardedOldSchema`/`TestAccMockableAdvancedCluster_symmetricShardedOldSchema ` due to last state: 'CREATING', timeout: 3h0m0s

PR #2975, opened 2025-01-16
chore: Enable migration tests for clusters in legacy schema

lantoli:
## Description

Enable migration tests for clusters in legacy schema. With this PR, all mig tests for TPF are passing.

Link to any related issue(s): CLOUDP-295165

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedcluster/resource_advanced_cluster_migration_test.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster_migration_test.go
@@ -29,7 +29,6 @@ func TestMigAdvancedCluster_singleShardedMultiCloud(t *testing.T) {
 }
 
 func TestMigAdvancedCluster_symmetricGeoShardedOldSchema(t *testing.T) {
-	acc.SkipIfAdvancedClusterV2Schema(t) // unexpected update and then: error operation not permitted, nums_shards from 1 -> > 1
 	testCase := symmetricGeoShardedOldSchemaTestCase(t, false)
 	mig.CreateAndRunTest(t, &testCase)
 }
--- a/internal/service/advancedclustertpf/move_upgrade_state.go
+++ b/internal/service/advancedclustertpf/move_upgrade_state.go
@@ -3,6 +3,7 @@ package advancedclustertpf
 import (
 	"context"
 	"fmt"
+	"math/big"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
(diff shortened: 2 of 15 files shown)


internal/service/advancedcluster/resource_advanced_cluster_migration_test.go
```diff
@@ -29,7 +29,6 @@ func TestMigAdvancedCluster_singleShardedMultiCloud(t *testing.T) {
 }
 
 func TestMigAdvancedCluster_symmetricGeoShardedOldSchema(t *testing.T) {
-	acc.SkipIfAdvancedClusterV2Schema(t) // unexpected update and then: error operation not permitted, nums_shards from 1 -> > 1
```
lantoli, line 32: the last mig test to enable for TPF

internal/service/advancedclustertpf/move_upgrade_state.go
```diff
@@ -137,3 +146,52 @@ func getAttrTimeout(diags *diag.Diagnostics, rawState map[string]tftypes.Value)
...
+		if numShards.Cmp(one) > 0 { // legacy schema if numShards > 1
+			return true
+		}
+	}
+	return false
+}
+
+// sendLegacySchemaRequestToRead sets ClusterID to a special value so Read can know whether it must use legacy schema.
+// private state can't be used here because it's not available in Move Upgrader.
+// ClusterID is computed (not optional) so the value will be overridden in Read and the special value won't ever appear in the state file.
```
lantoli: let me know if there is any question about why we use ClusterID as a side channel to communicate between State Move / Upgrader and Read.   Also if you have a better idea, please let me know

internal/service/advancedclustertpf/move_upgrade_state.go
```diff
@@ -137,3 +146,52 @@ func getAttrTimeout(diags *diag.Diagnostics, rawState map[string]tftypes.Value)
...
+func sendLegacySchemaRequestToRead(model *TFModel) {
+	model.ClusterID = types.StringValue("forceLegacySchema")
+}
```
AgustinBettati: As an alternative, is it too complex to populate the replication spec list only with the bare minimum so our existing logic detects the legacy sharding config (objects with num_shards)? With this approach we would avoid `receivedLegacySchemaRequestInRead` using cluster_id which looks more hacky

internal/service/advancedclustertpf/move_upgrade_state.go
```diff
@@ -39,101 +41,149 @@ func stateUpgraderFromV1(ctx context.Context, req resource.UpgradeStateRequest,
 	setStateResponse(ctx, &resp.Diagnostics, req.RawState, &resp.State)
 }
 
-func setStateResponse(ctx context.Context, diags *diag.Diagnostics, stateIn *tfprotov6.RawState, stateOut *tfsdk.State) {
-	rawStateValue, err := stateIn.UnmarshalWithOpts(tftypes.Object{
+// Minimum attributes needed from source schema. Read will fill in the rest
```
marcosuma: non-blocking comment: in the future where new attributes will be added, will there be some compile-time or test failure if that field needs to be specified here?

internal/service/advancedclustertpf/move_upgrade_state.go
marcosuma: if we go with this option, do you think we can add a check of ClusterID != `forceLegacySchema` in one of the tests that are testing the upgrade?

internal/service/advancedclustertpf/move_upgrade_state.go
lantoli: in line 69 , we're using: IgnoreUndefinedAttributes: true}
that means that we're flexible in the schema, if you see the code below the real mandatory ones are project_id and cluster, the ones we try to used them in a best-effort basis,  it's ok if they don't come (e.g. later when moving from flex cluster)

so the schema doesn't make it fail if they some attribute doesn't exist,  later when the value is tried to be read it will be null

marcosuma: > it's ok if they don't come
ok then why are we even populating them? What is the advantage of populating? Answering this question should also help me with understanding the consequence of:
> later when the value is tried to be read it will be null

lantoli: if the previous version/resource have them and we don't set them, then there will a plan change, e.g. if timeouts or retain_backups_enabled is defined in SDKv2 and want to migrate to TPF, users will get a plan change saying that timeouts/retain_backups_enabled will be deleted.  
(we don't want plan changes when upgrading from sdkv2 to tpf or moved block from cluster to tpf adv_cluster)

in the case that will come with moving from flex cluster to adv_cluster for example, they don't exist so it's ok not to send them.

lantoli: it's to avoid plan changes by filling attributes that Read can't do

marcosuma: got it!

internal/service/advancedclustertpf/move_upgrade_state.go
lantoli: not using ClusterID any more

internal/service/advancedclustertpf/move_upgrade_state.go
lantoli: I was a bit reluctant in case we needed to pass a lot of structure but at the end we only need to pass the num_shards in the replications_configs, changed

internal/service/advancedclustertpf/move_upgrade_state.go
lantoli: clarified here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2975/commits/b67243fab155f0f51f69b614bcded7e74c293a94

internal/service/advancedclustertpf/move_upgrade_state.go
AgustinBettati: correct, looks good. If we can leave a comment give context of why we add replication specs that would be good for future reference

AgustinBettati (approved):
LGTM

internal/service/advancedclustertpf/move_upgrade_state.go
lantoli: i did here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2975/commits/8fe7cd68035b5904a77ebf373590a7520c7711e0

would you add something else to the latest PR doc?

marcosuma (approved):

PR #2977, opened 2025-01-17
test: Updates importIgnoredFields function to handle dynamic replication specs

EspenAlbert:
## Description

Updates importIgnoredFields function to handle dynamic replication specs

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedcluster/resource_advanced_cluster_test.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster_test.go
@@ -224,7 +224,7 @@ func replicaSetMultiCloudTestCase(t *testing.T, isAcc bool) resource.TestCase {
 				Config: configReplicaSetMultiCloud(t, isAcc, orgID, projectName, clusterNameUpdated),
 				Check:  checkReplicaSetMultiCloud(isAcc, clusterNameUpdated, 3),
 			},
-			acc.TestStepImportCluster(resourceName, "replication_specs", "retain_backups_enabled"),
+			acc.TestStepImportCluster(resourceName),
 		},
 	}
 }
@@ -253,7 +253,7 @@ func singleShardedMultiCloudTestCase(t *testing.T, isAcc bool) resource.TestCase
 				Config: configShardedOldSchemaMultiCloud(t, isAcc, projectID, clusterNameUpdated, 1, "M10", nil),
 				Check:  checkShardedOldSchemaMultiCloud(isAcc, clusterNameUpdated, 1, "M10", true, nil),
 			},
-			acc.TestStepImportCluster(resourceName, "replication_specs"),
+			acc.TestStepImportCluster(resourceName),
 		},
 	}
 }
@@ -282,7 +282,7 @@ func TestAccClusterAdvancedCluster_unpausedToPaused(t *testing.T) {
 				Config:      configSingleProviderPaused(t, true, projectID, clusterName, true, anotherInstanceSize),
 				ExpectError: regexp.MustCompile("CANNOT_UPDATE_PAUSED_CLUSTER"),
 			},
-			acc.TestStepImportCluster(resourceName, "replication_specs"),
+			acc.TestStepImportCluster(resourceName),
 		},
 	})
 }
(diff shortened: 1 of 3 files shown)


internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2782,13 +2782,17 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
 	`, orgID, projectName, clusterName, mongoDBMajorVersion, pinnedFCVAttr)) + dataSourcesTFNewSchema
 }
 
-func importIgnoredFields() []string {
+func importIgnoredFields(replicationSpecs int) []string {
 	if config.AdvancedClusterV2Schema() {
 		return []string{}
```
lantoli: nit: i think you can just return nil here

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2782,13 +2782,17 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
 	}
-	return []string{
-		"replication_specs.0.region_configs.0.read_only_specs",
-		"replication_specs.0.region_configs.0.analytics_specs",
-		"replication_specs.0.region_configs.0.electable_specs.0.ebs_volume_type",
+	ignored := []string{}
```
lantoli: nit: similarly, you don't need to initialize, can just do:
```
var ignored []string
```

nil slices are in general equivalent to empy (non-nil) ones

lantoli (approved):

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2782,13 +2782,17 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
 	`, orgID, projectName, clusterName, mongoDBMajorVersion, pinnedFCVAttr)) + dataSourcesTFNewSchema
 }
 
-func importIgnoredFields() []string {
+func importIgnoredFields(replicationSpecs int) []string {
```
marcosuma: why these ignored methods and why adding now the int?

I am not seeing a lot of documentation and this code is going to be very hard to follow and understand why in few months

lantoli: note that it has a `AdvancedClusterV2Schema` clause.  It's because SDKv2 doesn't fill those fields (we talked about this in yesterday's meeting) so we don't want to check those attributes only for TPF.  Maybe a better param name would be replicationSpecCount, to specify how many replicationSpecs are, so we want to ignore those attribute changes in import tests, as we know they happen because SDKv2 and TPF fill them differently.

these attribute checks are new to TPF, were not being done in SDKv2

lantoli: doc added and refactor, please see the latest code

marcosuma (approved):

internal/testutil/acc/advanced_cluster.go
```diff
@@ -42,14 +43,24 @@ var (
...
+func TestStepImportCluster(resourceName string, ignorePrefixFields ...string) resource.TestStep {
+	ignorePrefixFields = append(ignorePrefixFields,
+		"retain_backups_enabled", // This field is TF specific and not returned by Atlas, so Import can't fill it in.
+	)
+
+	// auto_scaling & specs (electable_specs, read_only_specs, etc.) are only set in state in SDKv2 if present in the definition.
+	// However, as import doesn't have a previous state to compare with, import will always fill them.
+	// This will make these fields differ in the state, although the plan change won't be shown to the user as they're computed values.
+	if !config.AdvancedClusterV2Schema() {
+		ignorePrefixFields = append(ignorePrefixFields, "replication_specs")
```
EspenAlbert, line 55: This is quite broad. But I like that you could use the `TestStepImportCluster` instead of the extra helper

lantoli: @EspenAlbert  it is for SDKv2 tests, but reality is that any spec that is not in the definition file must be included, so at then it's like playing to exclude the elements that are not in the definition file. 

however it's more restrict for TPF now, because some tests had "replication_specs" for SDKv2 and TPF, not this is only for SDKv2 tests.

PR #2994, opened 2025-01-22
fix: Doesn't send `analyzers` field in search_index if absent in the definition

lantoli:
## Description

Doesn't send `analyzers` field in search_index if absent in the definition. It fixes HELP-70041.

In this way Atlas UI won't show an error when editing the index in JSON as happening now:
index3

Link to any related issue(s): CLOUDP-296191

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/searchindex/model_search_index.go
+++ b/internal/service/searchindex/model_search_index.go
@@ -72,7 +72,7 @@ func unmarshalSearchIndexFields(str string) ([]map[string]any, diag.Diagnostics)
 	return fields, nil
 }
 
-func unmarshalSearchIndexAnalyzersFields(str string) ([]admin.AtlasSearchAnalyzer, diag.Diagnostics) {
+func UnmarshalSearchIndexAnalyzersFields(str string) ([]admin.AtlasSearchAnalyzer, diag.Diagnostics) {
 	fields := []admin.AtlasSearchAnalyzer{}
 	if str == "" {
 		return nil, nil // don't send analyzers field to Atlas if empty
--- a/internal/service/searchindex/model_search_index.go
+++ b/internal/service/searchindex/model_search_index.go
@@ -75,7 +75,7 @@ func unmarshalSearchIndexFields(str string) ([]map[string]any, diag.Diagnostics)
 func unmarshalSearchIndexAnalyzersFields(str string) ([]admin.AtlasSearchAnalyzer, diag.Diagnostics) {
 	fields := []admin.AtlasSearchAnalyzer{}
 	if str == "" {
-		return fields, nil
+		return nil, nil // don't send analyzers field to Atlas if empty
 	}
 	dec := json.NewDecoder(bytes.NewReader([]byte(str)))
 	dec.DisallowUnknownFields()
--- a/.changelog/2994.txt
+++ b/.changelog/2994.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/mongodbatlas_search_index: Don't send empty `analyzers` attribute to Atlas
+```
--- a/internal/service/searchindex/model_search_index_test.go
+++ b/internal/service/searchindex/model_search_index_test.go
@@ -1,0 +1,81 @@
+package searchindex_test
+
+import (
+	"testing"
+
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/searchindex"
+	"github.com/stretchr/testify/assert"
+	"go.mongodb.org/atlas-sdk/v20241113004/admin"
+)
+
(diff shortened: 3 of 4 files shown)


jwilliams-mongo (approved):

oarbusi (approved):
LGTM, could we capture this case in the unit tests of unmarshalSearchIndexAnalyzersFields ?

lantoli:
@oarbusi search_index doesn't have any unit test yet, I think that would be a bigger planned effort to add them including for this func. And I think this behavior change can't be captured in an acc test because everything works fine and nothing changes in the TF side, the only problem is that "analyzers" is shown in the Atlas UI JSON editor

internal/service/searchindex/model_search_index.go
```diff
@@ -75,7 +75,7 @@ func unmarshalSearchIndexFields(str string) ([]map[string]any, diag.Diagnostics)
 func unmarshalSearchIndexAnalyzersFields(str string) ([]admin.AtlasSearchAnalyzer, diag.Diagnostics) {
```
AgustinBettati: I agree with @oarbusi a simple unit test of this function capturing the empty string case will ensure no regressions in the future

internal/service/searchindex/model_search_index.go
```diff
@@ -75,7 +75,7 @@ func unmarshalSearchIndexFields(str string) ([]map[string]any, diag.Diagnostics)
...
 	fields := []admin.AtlasSearchAnalyzer{}
 	if str == "" {
-		return fields, nil
+		return nil, nil // don't send analyzers field to Atlas if empty
```
AgustinBettati, line 78: is this change fixing the issue? Since the function signature is returning a slice and not pointer to slice I believe a pointer to empty slice is still being sent.

lantoli: yes I tested manually Create and Update, and it worked. 
we sent a zero-len slice before, now we send an nil slice

internal/service/searchindex/model_search_index.go
marcosuma: @lantoli assuming you are 100% sure that this fixes the issue, and you're talking about bigger planned effort, could we solve this by creating a separate issue to create the unit tests like @oarbusi and @AgustinBettati are also saying? A test here is very important to make sure this stays like that in the future.

internal/service/searchindex/model_search_index.go
AgustinBettati: Okay, I am still confused as I would expect the fix to be [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/searchindex/resource_search_index.go#L424) for create and [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/searchindex/resource_search_index.go#L228) for update where we are placing a pointer to the nil slice (instead of nil pointer).

internal/service/searchindex/model_search_index.go
lantoli: unit tests added here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2994/commits/89f1ad32bfa425ff8b5289f3052ce9bf0b116ae2

internal/service/searchindex/model_search_index.go
lantoli: exactly, that was what I did first in Create, then I extracted it to the func that both lines call before to get the analyzers so it works for both Create and Update

internal/service/searchindex/model_search_index.go
marcosuma: you are awesome, @lantoli !

marcosuma (approved):

internal/service/searchindex/model_search_index.go
AgustinBettati: Sorry for being insistent but could you verify again that it is not sent in the request? With my knowledge of the SDK/json marshaling an empty array would be sent in the request when defining a code like the following (similar to code currently in place):
```
fields := nil
searchIndexRequest.Definition.Analyzers = &analyzers
```

oarbusi (approved):
LGTM, thanks for creating the test

internal/service/searchindex/model_search_index.go
lantoli: discussed offline, both approaches behave the same

AgustinBettati: analyzer property is sent but with value of nil, so API interprets this the same as not including the property alltogether

AgustinBettati (approved):
LGTM, thanks for sharing details!

PR #3001, opened 2025-01-24
feat: Supports Flex Clusters in `mongodbatlas_advanced_cluster` resource and data sources (SDKv2)

oarbusi:
## Description

Supports Flex Clusters in `mongodbatlas_advanced_cluster` resource and data sources (SDKv2)

Link to any related issue(s): CLOUDP-295164

## Type of change:

## Required Checklist:


internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -605,14 +630,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
-
+	isFlex := false
 	cluster, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
 	if err != nil {
 		if resp != nil && resp.StatusCode == http.StatusNotFound {
 			d.SetId("")
 			return nil
 		}
-		return diag.FromErr(fmt.Errorf(errorRead, clusterName, err))
+		if isFlex = admin.IsErrorCode(err, "CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API"); !isFlex { // if cluster is flex we need to call different API
```
marcosuma: I see in DS we have no other way than making two calls, but could we avoid it here by looking at the state and inferring if it's flex or not? I really want to avoid making two calls for flex cluster every time

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1232,9 +1287,19 @@ func resourceDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		params.RetainBackups = conversion.Pointer(v.(bool))
 	}
 
+	isFlex := false
 	_, err := connV2.ClustersApi.DeleteClusterWithParams(ctx, params).Execute()
 	if err != nil {
-		return diag.FromErr(fmt.Errorf(errorDelete, clusterName, err))
+		if isFlex = admin.IsErrorCode(err, "CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API"); !isFlex { // if cluster is flex we need to call different API
```
marcosuma: same here

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1487,96 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
 	_, err := stateConf.WaitForStateContext(ctx)
 	return err
 }
+
+func isFlex(clusterType string) bool {
+	return clusterType == flexcluster.FlexClusterType
+}
+
+func getFlexClusterUpdateRequest(d *schema.ResourceData) *admin.FlexClusterDescriptionUpdate20241113 {
```
marcosuma: I am not sure I follow.. if we are in the update method, shouldn't it be enough to only check if it's an upgrade and if not it's definitely only an update? I am wondering if we can avoid this "allow-list" kind if statement

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2798,3 +2798,121 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+				}
+			}
+			%[5]s
+			termination_protection_enabled = false
+		}
+	`, projectID, clusterName, providerName, region, tags) +
+		dataSourcesTFOldSchema + strings.ReplaceAll(acc.FlexDataSource, "mongodbatlas_flex_cluster.", "mongodbatlas_advanced_cluster.")
+}
+
+func TestAccClusterFlexCluster_basic(t *testing.T) {
```
marcosuma, line 2866: should this test also try to upgrade to a dedicated cluster?

marcosuma:
This is a very long PR. I believe it could have been split in some of its part, if so please for the next time let's try to be intentional on having smaller PRs

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: Not implemented yet, will be done later

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: When the read is run when doing an import operation the state is not available, so it is not possible to do without doing 2 calls

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: There are 3 possibilities when we enter the Update operation in advanced_cluster when it's a flex cluster:
- Update of tags/termination_protection_enabled
- Update of non-updatable attributes
- Upgrade from M0 to Flex

If it's not an upgrade, we shouldn't just do the update, and instead we have to check if only the updatable attributes are being updated. If we don't do this, terraform will want to keep updating the resource because there will be a difference between the config/plan and the state even though we have run the update

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: In the delete, doing 2 calls can be avoided. Changed

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -605,14 +630,33 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
+		clusterName := d.Get("name").(string)
+		flexClusterResp, err := flexcluster.GetFlexCluster(ctx, projectID, clusterName, connV2.FlexClustersApi)
+		if err != nil {
+			return diag.FromErr(fmt.Errorf(errorReadFlex, clusterName, err))
+		}
+		diags := setFlexFields(d, flexClusterResp)
+
+		if err := d.Set("cluster_id", flexClusterResp.GetId()); err != nil {
+			return diag.FromErr(fmt.Errorf(ErrorFlexClusterSetting, "cluster_id", clusterName, err))
+		}
```
AgustinBettati, line 643: any reason why this is not contained within `setFlexFields`?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -813,6 +857,17 @@ func isUsingOldShardingConfiguration(d *schema.ResourceData) bool {
 }
 
 func resourceUpdateOrUpgrade(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	replicationSpecs := expandAdvancedReplicationSpecs(d.Get("replication_specs").([]any), nil)
+
+	if isFlex((*replicationSpecs)[0].GetRegionConfigs()[0].GetProviderName()) {
```
AgustinBettati: could we have isFlex receive `*schema.ResourceData` and contain logic to safely access into the nested provider_name attribute?

internal/service/flexcluster/model.go
```diff
@@ -131,3 +132,68 @@ func ConvertProviderSettingsToTF(ctx context.Context, providerSettings admin.Fle
 	}
 	return &providerSettingsObject, nil
 }
+
+func NewAtlasCreateReqSDKv2(d *schema.ResourceData, replicationSpecs *[]admin.ReplicationSpec20240805) *admin.FlexClusterDescriptionCreate20241113 {
```
AgustinBettati: given `flexcluster` is the package that contains the TPF implementation, find it confusing to define this function here which is related to an SDK v2 implementation. Would a separate file prefixed with `flex_` in the advancedcluster package work?

internal/service/advancedcluster/resource_advanced_cluster.go
marcosuma: got it - so should the logic return error for non-updatable attributes?

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2798,3 +2798,121 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+			}`
+	}
+	return fmt.Sprintf(`
+		resource "mongodbatlas_advanced_cluster" "test" {
+			project_id   = %[1]q
+			name         = %[2]q
+			cluster_type = "REPLICASET"
+			replication_specs {
+				region_configs {
+					provider_name = "FLEX"
```
AgustinBettati, line 2853: Is there a follow up ticket to support `cluster_type="FLEX"` and `effective_cluster_type`?

internal/service/advancedcluster/resource_advanced_cluster.go
marcosuma: right, but don't we have a way to know if the read is coming from an import?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1486,96 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
...
 	return err
 }
+
+func isFlex(clusterType string) bool {
+	return clusterType == flexcluster.FlexClusterType
+}
+
+func getFlexClusterUpdateRequest(d *schema.ResourceData) *admin.FlexClusterDescriptionUpdate20241113 {
+	if d.HasChange("tags") || d.HasChange("termination_protection_enabled") &&
+		(!d.HasChange("cluster_type") && !d.HasChange("replication_specs") && !d.HasChange("project_id") && !d.HasChange("name")) {
```
AgustinBettati: would extract to variables so it is clear the scenario that is captured here. Also believe the current condition of `tags change` || ( `termination change` && `other changes`) is not the intended condition

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: it already does return an error [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/files#diff-317cbb5ab8946cb906d1e81141844344600ded4bb8b6620e3a617b94ff30aaa4R869)

internal/service/flexcluster/resource_test.go
```diff
@@ -1,22 +1,20 @@
...
-	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/acc"
 )
 
 var (
-	resourceType         = "mongodbatlas_flex_cluster"
 	resourceName         = "mongodbatlas_flex_cluster.test"
 	dataSourceName       = "data.mongodbatlas_flex_cluster.test"
 	dataSourcePluralName = "data.mongodbatlas_flex_clusters.test"
+	DataSources          = ``
```
AgustinBettati: I believe this is a left over

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: We could try to look if there are values in the state in the read, but I don't think that is a good idea. It would add complexity to the code and not be reliable (state can have outdated values, for example when cluster has just been upgraded from M0 to Flex, state will have the cluster defined as M0, but it actually will be a Flex and the state needs to be updated(in the read operation), so API is a better source of truth for the read

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: This attribute does not exists in the data source

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: I have tried to reduce the amount of new code. `expandAdvancedReplicationSpecs` already existed before and was being used in other places. Apart from that, provider_name is required so it will always have a value so I think this is safe

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: I think that the condition is correct. An updatable attribute has a change (`d.HasChange("tags") || d.HasChange("termination_protection_enabled")`) **AND** the non-updatable attributes have not changed (`(!d.HasChange("cluster_type") && !d.HasChange("replication_specs") && !d.HasChange("project_id") && !d.HasChange("name"))`)
If no changes to updatable attributes, result would be false
If changes to the non-updatabale attributes, result would be false

internal/service/flexcluster/model.go
oarbusi: For now I am trying to put everything related to Flex clusters in `flexcluster`, both SDKv2 and TPF. `advancedcluster` and `advancedclustertpf` packages are big already, so I think keeping this here for now is a ok. 
I am open to moving this elsewhere when implementation for TPF is done, and we are able to more clearly see where everything fits better. Let me know what you think

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: After talking to the upstream team, they have told me that cluster_type for flex clusters will be always REPLICASET and that LOADBALANCED was an idea that was scrapped. They are now working on removing thata possible value from the backend and API docs so that there is no confusion. This allows us to not change schema, not add effective_cluster_type. Technical design has also been updated accordingly

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: Extracted the conditions to variables to improve clarity. Also, a parenthesis was missing. Thanks!

internal/service/advancedcluster/resource_advanced_cluster.go
EspenAlbert: What about refactoring out this logic?
I argue it should be in the `common_admin_sdk.go` and reused once both from resource and data-sources

internal/service/flexcluster/model.go
```diff
@@ -131,3 +132,68 @@ func ConvertProviderSettingsToTF(ctx context.Context, providerSettings admin.Fle
...
+		{
+			"standard":     str.GetStandard(),
+			"standard_srv": str.GetStandardSrv(),
+		},
+	}
+}
+
+func FlattenFlexProviderSettingsIntoReplicationSpecs(providerSettings admin.FlexProviderSettings20241113, priority *int) []map[string]any {
+	tfMaps := []map[string]any{{}}
+	tfMaps[0]["num_shards"] = 1                              // default value
```
EspenAlbert, line 146: Are all these values really necessary?

oarbusi: Yes, thay have default values and if not set here after creation there will be a non-empty plan

internal/service/flexcluster/model.go
```diff
@@ -131,3 +132,68 @@ func ConvertProviderSettingsToTF(ctx context.Context, providerSettings admin.Fle
...
+	tfMaps[0]["zone_name"] = "ZoneName managed by Terraform" // default value
```
EspenAlbert, line 147: We might consider using `Zone 1` which is the backend default.
[See this issue, maybe you have already checked that it works as expected?](https://jira.mongodb.org/browse/CLOUDP-294339)

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2798,3 +2798,121 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+				Config:      configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_2", true),
+				ExpectError: regexp.MustCompile("flex cluster update is not supported except for tags and termination_protection_enabled fields"),
+			},
+		},
+	})
+}
+
+func checkFlexClusterConfig(projectID, clusterName, providerName, region string, tagsCheck bool) resource.TestCheckFunc {
+	checks := []resource.TestCheckFunc{acc.CheckExistsFlexCluster()}
+	attrMapAdvCluster := map[string]string{
```
EspenAlbert, line 2895: Only `computed` attributes need to be checked (Leo's point on technical discussion)

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -429,6 +436,24 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	connV220240805 := meta.(*config.MongoDBClient).AtlasV220240805
 	connV2 := meta.(*config.MongoDBClient).AtlasV2
 	projectID := d.Get("project_id").(string)
+	replicationSpecs := expandAdvancedReplicationSpecs(d.Get("replication_specs").([]any), nil)
+
+	if isFlex((*replicationSpecs)[0].GetRegionConfigs()[0].GetProviderName()) {
```
EspenAlbert: Can you not keep everything as is and instead check for `isFlex` on the `params` object on L516? Not sure why f`lexcluster.NewAtlasCreateReqSDKv2` is necessary

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2798,3 +2798,121 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+		resource "mongodbatlas_advanced_cluster" "test" {
+			project_id   = %[1]q
+			name         = %[2]q
+			cluster_type = "REPLICASET"
+			replication_specs {
+				region_configs {
+					provider_name = "FLEX"
+					backing_provider_name = %[3]q
+					region_name = %[4]q
+					priority      = 0
```
EspenAlbert: priority 0?

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2798,3 +2798,121 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+			{
+				Config: configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_1", false),
+				Check:  checkFlexClusterConfig(projectID, clusterName, "AWS", "US_EAST_1", false),
+			},
+			{
+				Config: configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_1", true),
+				Check:  checkFlexClusterConfig(projectID, clusterName, "AWS", "US_EAST_1", true),
+			},
+			{
+				Config:            configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_1", true),
```
EspenAlbert: Config not necessary in import step.
Maybe re-use `acc.TestStepImportCluster(resourceName)` instead

internal/service/advancedcluster/resource_advanced_cluster.go
lantoli: I think it would be more clear to encapsulate the flex logic in funcs instead of making current funcs longer, similar idea to @EspenAlbert about joining resource and ds logic

internal/service/advancedcluster/resource_advanced_cluster_test.go
lantoli: nice!

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: Priority is a required attribute but not needed for flex cluster, so it can have any value, and it will have no effect. In the state we set the value put in the config and that is only to avoid plan changes

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: Good point, although this is only for resources migrated to TPF, so it does not apply for the checks of advanced_cluster because it's for SDKv2. For flex cluster checks (which is using. TPF), it could make sense to remove the checks for project_id and name but since the checks map is being used for both the singular data source (where the attributes are non-computed) and plural data source (where attributes are computed), I would prefer to keep it so we check all computed attributes of the plural data source. And we stay with a couple of unnecessary checks being done to keep this method simpler

internal/service/flexcluster/model.go
oarbusi: "ZoneName managed by Terraform" is the Default value of the attribute in TF. If we change this, here, we should then change the `zone_name` default value in TF as well.

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: Flex Cluster API uses a different object (`FlexClusterDescriptionCreate20241113`) than the Cluster API uses, so we can't use the params object (of type `ClusterDescription20240805`). I changed a bit the order so that `expandAdvancedReplicationSpecs` is only called once

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: Could we create a ticket to add this attribute in SDK v2 data source implementation? For reference in TPF implementation I see it was included in both resource and data sources (which makes sense to me).

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: Okay, I understand it will likely be a safe operation but I think the code smell of `(*replicationSpecs)[0].GetRegionConfigs()[0].GetProviderName()` accessing indexes and values directly can be remediated quickly by accessing the values safely within `isFlex`. Will also make calling isFlex funtion more simple by just passing all replication specs.

internal/service/flexcluster/model.go
AgustinBettati: all good if you prefer to revisit later on, however would consider that this function using `d *schema.ResourceData` will not be used in any of the tpf implementations `flexcluster` or `advancedclustertpf`

internal/service/advancedcluster/resource_advanced_cluster_test.go
AgustinBettati: great 👍

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: created CLOUDP-296904

internal/service/advancedcluster/resource_advanced_cluster.go
AgustinBettati: great, this does make it easier to follow

internal/service/advancedcluster/resource_advanced_cluster_test.go
AgustinBettati: Should we use priority `7` in our acceptance tests and examples? This way when users transition to dedicated no changes are needed in this attribute (similar to [tenant](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/advanced_cluster#example-tenant-cluster))

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: Makes sense, I refactored the method here [4683708](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/4683708d7755b03d18620a7baf1d12860e0d13b9)

internal/service/advancedcluster/resource_advanced_cluster.go
marcosuma: @EspenAlbert what logic are you referring to? 
@oarbusi you have a point on the risk of outdated state

internal/service/advancedcluster/resource_advanced_cluster.go
marcosuma: yes but the part where I disagree is on the separation of concerns. 

IMO you should have an update method (line 863) that:
```
if isFlex(replicationSpecs) {
     if <anyOfNonUpdatableIsChanged> {
          return diag.Error(...)
     }

     flexUpdateRequest := ...
     flexUpgradeRequest := ...

}
```

and the `getFlexClusterUpdateRequest` only focuses of creating the update request, instead of also caring about returning the error. That's how the code can be a bit more intuitive.

In addition, a separate doubt I have is: if user updates `termination_protection_enabled` and also `goes to M10` it looks like you are ignoring the upgrade and only do the update?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1487,105 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
...
+
+	_, err := flexcluster.UpdateFlexCluster(ctx, projectID, clusterName, flexUpdateRequest, connV2.FlexClustersApi)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf(errorUpdateFlex, err))
+	}
+
+	return resourceRead(ctx, d, meta)
+}
+
+func setFlexFields(d *schema.ResourceData, flexCluster *admin.FlexClusterDescription20241113) diag.Diagnostics {
```
marcosuma, line 1488: is this method only usable from this implementation? or also some chance for TPF?

oarbusi: SDKv2 only. Like `setRootFields`

internal/service/advancedcluster/resource_advanced_cluster_test.go
marcosuma: +1 to use 7
> Priority is a required attribute but not needed for flex cluster
@oarbusi can you please have a task to update our docs about this behaviour?

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: I understand your point. I refactored it to separate logic to detect valid update and valid upgrade from creating the update upgrade object

On the second point, currently the provider supports either updating or upgrading in one TF apply. Because they use different endpoints and is an important operation, I wouldn't advise on doing upgrade & update in the same update operation behind the scenes.

internal/service/flexcluster/model.go
oarbusi: Good point. Moved this into a new model_flex.go file inside advancedcluster.

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: Changed

internal/service/flexcluster/model.go
AgustinBettati: thanks!

internal/service/advancedcluster/model_flex.go
```diff
@@ -0,0 +1,63 @@
...
+	"go.mongodb.org/atlas-sdk/v20241113004/admin"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/constant"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/flexcluster"
+)
+
+func NewFlexCreateReq(d *schema.ResourceData, replicationSpecs *[]admin.ReplicationSpec20240805) *admin.FlexClusterDescriptionCreate20241113 {
```
maastha, line 13: I think some of the methods in this file can be moved to common files (I believe common_admin_sdk.go) in advancedclustertpf no?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1268,23 +1333,36 @@ func resourceImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*s
...
+		flexCluster, err := flexcluster.GetFlexCluster(ctx, *projectID, *name, connV2.FlexClustersApi)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't import flex cluster %s in project %s, error: %s", *name, *projectID, err)
+		}
+		clusterID = flexCluster.GetId()
+		clusterName = flexCluster.GetName()
 	}
 
-	if err := d.Set("project_id", cluster.GetGroupId()); err != nil {
+	if err := d.Set("project_id", clusterID); err != nil {
```
maastha: project_id is being set here, typo?

internal/service/advancedcluster/model_flex.go
oarbusi: I think it will make more sense to move this to the common_admin_sdk whenever it's needed in the advancedclustertpf package. For now here there are SDKv2 specific methods which I don't think fit in tpf for now

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: good catch, changed!

internal/service/advancedcluster/resource_advanced_cluster.go
EspenAlbert: In `advancedcluster/resource_advanced_cluster.go` the `connV2.ClustersApi.GetCluster` call is now wrapped with isFlex and custom error handling and re-reading from `flexCluster, err := flexcluster.GetFlexCluster(ctx, *projectID, *name, connV2.FlexClustersApi)`. 

This is done in `data_source_advanced_cluster.go` in `dataSourceRead`

When we implement this in TPF, we would have to redo it again, therefore, starting in `advancedclustertpf` would simplify the job later and keep the code DRY, it would also separate what is admin.SDK logic and what is TF provider/schema logic

internal/service/advancedcluster/resource_advanced_cluster.go
EspenAlbert: You could use it if `NewFlexCreateReq` accepts `ClusterDescription20240805` instead of `*[]admin.ReplicationSpec20240805`

oarbusi: I think now its fine to keep as is. `expandAdvancedReplicationSpecs` is called once and `replicationSpecs` is used in `NewFlexCreateReq(d, replicationSpecs)` in the case of flex and for dedicated is used to create the `ClusterDescription20240805` object. I prefer not to use dedicated cluster objects for flex logic/conversion methods

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: I agree that the resource read and data source read are very similar, although they have some slight differences(cluster_id & id being set). 
I prefer to not yet consider TPF implementation here and prematurely optimize. I think it'll be better to refactor and unify this logic when working on TPF implementation

oarbusi: Done the refactor, resolving

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: done

internal/service/advancedclustertpf/common_admin_sdk.go
```diff
@@ -3,11 +3,15 @@ package advancedclustertpf
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
+	sdkv2Diag "github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
EspenAlbert: IMO: `v2/diag` should not be used here. Instead a `tpf_adapter.go` in SDKv2 packages to convert the diags. See https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2998 for example

internal/service/advancedclustertpf/common_admin_sdk.go
```diff
@@ -170,3 +173,22 @@ func ReadCluster(ctx context.Context, diags *diag.Diagnostics, client *config.Mo
 	}
 	return readResp
 }
+
+func GetClusterDetails(ctx context.Context, projectID, clusterName string, client *admin.APIClient) (isFlex bool, clusterDesc *admin.ClusterDescription20240805, flexClusterResp *admin.FlexClusterDescription20241113, diags sdkv2Diag.Diagnostics, err error) {
```
EspenAlbert: please see how other signatures receives diags instead. Too many return values here IMO

oarbusi: [a8437cd](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/a8437cd2e5b8f045febc2887bb20e4a2960bde78)

internal/service/advancedclustertpf/common_admin_sdk.go
oarbusi: [a8437cd](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/a8437cd2e5b8f045febc2887bb20e4a2960bde78)

internal/service/advancedclustertpf/common_admin_sdk.go
```diff
@@ -170,3 +172,24 @@ func ReadCluster(ctx context.Context, diags *diag.Diagnostics, client *config.Mo
 	}
 	return readResp
 }
+
+func GetClusterDetails(ctx context.Context, diags *diag.Diagnostics, projectID, clusterName string, client *admin.APIClient) (isFlex bool, clusterDesc *admin.ClusterDescription20240805, flexClusterResp *admin.FlexClusterDescription20241113, err error) {
```
EspenAlbert: Nice, but you don't need to return `err error` though. Instead you can use `diags.HasError()`

EspenAlbert: Also, you don't need to return `isFlex`, that can be inferred when `flexClusterResp != nil`

oarbusi: done. Personally I don't like the end result of all these refactors and don't see much benefit from it after having done them. Might make more sense later but not sure

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -292,13 +292,27 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			return nil
+	diags := diag.Diagnostics{}
+	diagsFramework := ConvertV2DiagsToFrameworkDiags(diags)
+	clusterDesc, flexClusterResp := advancedclustertpf.GetClusterDetails(ctx, diagsFramework, projectID, clusterName, connV2)
+	diags = ConvertFrameworkDiagsToV2Diags(*diagsFramework)
+	if diags.HasError() {
+		return diags
+	}
+
```
EspenAlbert: [nit] too many empty lines?

internal/service/advancedcluster/tpf_adapter.go
```diff
@@ -0,0 +1,26 @@
+package advancedcluster
+
+import (
+	frameworkdiag "github.com/hashicorp/terraform-plugin-framework/diag"
+	v2diag "github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+)
+
+func ConvertV2DiagsToFrameworkDiags(v2Diags v2diag.Diagnostics) *frameworkdiag.Diagnostics {
```
EspenAlbert: These already exist in [common/conversion/diags](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/common/conversion/diags.go)

EspenAlbert: This file is meant to be used from `resource_advanced_cluster` and `data_source_advanced_cluster`. And do the conversion from TPF diags to SDKv2 diags

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -292,13 +292,27 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
 	}
 
-	clusterDesc, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
-	if err != nil {
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			return nil
+	diags := diag.Diagnostics{}
+	diagsFramework := ConvertV2DiagsToFrameworkDiags(diags)
+	clusterDesc, flexClusterResp := advancedclustertpf.GetClusterDetails(ctx, diagsFramework, projectID, clusterName, connV2)
+	diags = ConvertFrameworkDiagsToV2Diags(*diagsFramework)
```
EspenAlbert: move diags creation and conversion logic to `tpf_adapter.go`.
Can wrap the call `GetClusterDetails(ctx, projectID, clusterName, connV2) (clusterDesc, flexClusterResp, diags)`

.changelog/3001.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:enhancement
+resource/mongodbatlas_advanced_cluster: Supports flex clusters
```
JuliaMongo: Please capitalize Flex in all cases, even in internal comments, so that we don't perpetuate lower case name.
Flex is a marketing term.

Please clarify what you mean by "Supports".
Do you mean that certain API endpoints allow users to create, delete and update MongoDB Flex clusters?
If so, which endpoints are those -- please list them.
If so, what what the state of affairs before this change? For example, you could state:
Previously, you could do X. With this change, when you do Y, these things (A, B, C) happen. 

What are the things that happen? What was happening before and why did we change this? 
Is the changelog a publicly facing artifact? If so, where do user can learn more about Flex clusters?
You could add "Learn <more>", where "more" is a link that explains stuff.

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1268,23 +1323,36 @@ func resourceImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*s
 		return nil, err
 	}
 
+	isFlex := false
 	cluster, _, err := connV2.ClustersApi.GetCluster(ctx, *projectID, *name).Execute()
 	if err != nil {
-		return nil, fmt.Errorf("couldn't import cluster %s in project %s, error: %s", *name, *projectID, err)
+		if isFlex = admin.IsErrorCode(err, "CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API"); !isFlex { // if cluster is flex we need to call different API
```
JuliaMongo: is a Flex cluster,

JuliaMongo: Also:
CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API -- I have concerns about this choice of naming for an internal construct in code. Reasons:
The customer-facing change is the opposite of what this variable's name states:
The customer-facing change is that when user use cluster API to create previously known clusters (serverless or m2/m5), they will instead create Flex clusters.
This variable's name states the opposite. I know the variable is internal. But it sounds confusing nevertheless, and may confuse those who see this code later and are not familiar with the context.

If I am wrong on any of this, please ignore. Just wanted to point out what I see, from my non-dev perspective. Sorry for the intrusion.

internal/service/flexcluster/resource.go
```diff
@@ -134,20 +132,10 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
-	flexClusterParams := &admin.GetFlexClusterApiParams{
-		GroupId: projectID,
-		Name:    clusterName,
-	}
 
-	flexClusterResp, err := WaitStateTransition(ctx, flexClusterParams, connV2.FlexClustersApi, []string{retrystrategy.RetryStrategyUpdatingState}, []string{retrystrategy.RetryStrategyIdleState})
+	flexClusterResp, err := UpdateFlexCluster(ctx, projectID, clusterName, flexClusterReq, connV2.FlexClustersApi)
 	if err != nil {
-		resp.Diagnostics.AddError("error waiting for resource to be updated", err.Error())
+		resp.Diagnostics.AddError("error updating resource", err.Error())
```
JuliaMongo: the resource

internal/service/flexcluster/resource.go
```diff
@@ -172,18 +160,11 @@ func (r *rs) Delete(ctx context.Context, req resource.DeleteRequest, resp *resou
...
-	flexClusterParams := &admin.GetFlexClusterApiParams{
-		GroupId: flexClusterState.ProjectId.ValueString(),
-		Name:    flexClusterState.Name.ValueString(),
-	}
+	err := DeleteFlexCluster(ctx, flexClusterState.ProjectId.ValueString(), flexClusterState.Name.ValueString(), connV2.FlexClustersApi)
 
-	if err := WaitStateTransitionDelete(ctx, flexClusterParams, connV2.FlexClustersApi); err != nil {
-		resp.Diagnostics.AddError("error waiting for resource to be deleted", err.Error())
+	if err != nil {
+		resp.Diagnostics.AddError("error deleting resource", err.Error())
```
JuliaMongo: the resource

JuliaMongo (changes requested):
I can't get the information I need from this changelog, as a user. If I am an internal user, I don't have enough context to see what this changelog tells me. If I am an external user, I also don't know what this changelog is telling me and why it matters.

internal/service/advancedcluster/data_source_advanced_cluster.go
EspenAlbert: resolved in 7bdec723c99c5ea1fbe2fd094c4783e2fe826b72

internal/service/advancedcluster/tpf_adapter.go
EspenAlbert: @oarbusi refactored as per your request in 7bdec723c99c5ea1fbe2fd094c4783e2fe826b72 please have a look

internal/service/advancedcluster/data_source_advanced_cluster.go
EspenAlbert: @oarbusi refactored as per your request in 7bdec723c99c5ea1fbe2fd094c4783e2fe826b72 please have a look

.changelog/3001.txt
oarbusi: Will change the capitalization. 
The [CHANGELOG](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CHANGELOG.md) is an artifact that lives in the repository. This is automatically generated when we merge PRs into master using the files in .changelog directory. 
Supports means that now the mongodbatlas_advanced_cluster resource and data source are able to create, read,update and delete Flex clusters. Before in mongodbatlas_advanced_cluster only dedicated, shared and free clusters were being supported.
In the changelog we don't usually put all the links and extra information for important features. We do this in the upgrade guides of the release(e.g. https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/docs/guides/1.26.0-upgrade-guide.md ) and we will create a upgrade guide with more details on Flex clusters in mongodbatlas_advanced_cluster resource & data source further down the line

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: `The customer-facing change is that when user use cluster API to create previously known clusters (serverless or m2/m5), they will instead create Flex clusters.` this is not the case for the changes here. If I am not mistaken, you are referring to the shimming logic that will, under the hood, create Flex clusters when shared/serverless are created. 
For context this will not be noticeable for terraform users because APIs will still work and have no changes, although we will mention and link to relevant docs in the TF documentation an upgrade guides

internal/service/flexcluster/resource.go
oarbusi: changed

internal/service/flexcluster/resource.go
oarbusi: changed

.changelog/3001.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:enhancement
+resource/mongodbatlas_advanced_cluster: Supports Flex clusters
+```
+
+```release-note:enhancement
+data-source/mongodbatlas_advanced_cluster: Supports Flex clusters
+```
+
+```release-note:enhancement
+data-source/mongodbatlas_advanced_clusters: Supports Flex clusters
```
lantoli: nit: next time I think it's better to do smaller PRs, example do data sources in a different PR

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -291,14 +291,22 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			return nil
+	clusterDesc, flexClusterResp, diags := GetClusterDetails(ctx, client, projectID, clusterName)
+	if diags.HasError() {
+		return diags
+	}
+	if flexClusterResp == nil && clusterDesc == nil {
+		return nil
+	}
+	if flexClusterResp != nil {
```
lantoli, line 301: if flexClusterResp are not nil we're giving preference to flex cluster, is that correct?

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -291,14 +291,22 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
-
-	clusterDesc, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
-	if err != nil {
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			return nil
+	clusterDesc, flexClusterResp, diags := GetClusterDetails(ctx, client, projectID, clusterName)
+	if diags.HasError() {
+		return diags
+	}
+	if flexClusterResp == nil && clusterDesc == nil {
```
lantoli, line 298: can it happen that both are nil without having an error? just wondering if this check is needed

internal/service/advancedcluster/data_source_advanced_cluster.go
```diff
@@ -291,14 +291,22 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	if v, ok := d.GetOk("use_replication_spec_per_shard"); ok {
 		useReplicationSpecPerShard = v.(bool)
 	}
-
-	clusterDesc, resp, err := connV2.ClustersApi.GetCluster(ctx, projectID, clusterName).Execute()
-	if err != nil {
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			return nil
+	clusterDesc, flexClusterResp, diags := GetClusterDetails(ctx, client, projectID, clusterName)
```
lantoli, line 294: are we keeping 404 not found logic?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -435,10 +437,29 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		rootDiskSizeGB = conversion.Pointer(v.(float64))
 	}
 
+	replicationSpecs := expandAdvancedReplicationSpecs(d.Get("replication_specs").([]any), rootDiskSizeGB)
+
+	if isFlex(replicationSpecs) {
```
lantoli: can this block be extracted to a func?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -598,21 +619,32 @@ func CreateStateChangeConfig(ctx context.Context, connV2 *admin.APIClient, proje
...
 	ids := conversion.DecodeStateID(d.Id())
 	projectID := ids["project_id"]
 	clusterName := ids["cluster_name"]
 
 	var replicationSpecs []map[string]any
+	cluster, flexClusterResp, diags := GetClusterDetails(ctx, client, projectID, clusterName)
+	if diags.HasError() {
+		return diags
+	}
+	if flexClusterResp == nil && cluster == nil {
```
lantoli, line 634: similar comment to a previous, can both be nil without errors?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1474,81 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
...
+	flexClusterName := flexCluster.GetName()
+	if err := d.Set("cluster_type", flexCluster.GetClusterType()); err != nil {
+		return diag.FromErr(fmt.Errorf(ErrorFlexClusterSetting, "cluster_type", flexClusterName, err))
+	}
+
+	if err := d.Set("backup_enabled", flexCluster.BackupSettings.GetEnabled()); err != nil {
+		return diag.FromErr(fmt.Errorf(ErrorFlexClusterSetting, "backup_enabled", flexClusterName, err))
+	}
+
+	if err := d.Set("connection_strings", flexcluster.FlattenFlexConnectionStrings(flexCluster.ConnectionStrings)); err != nil {
```
lantoli, line 1498: i don't miss SDKv2 :-)

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1474,81 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
...
+
+	if err := d.Set("name", flexCluster.GetName()); err != nil {
+		return diag.FromErr(fmt.Errorf(ErrorFlexClusterSetting, "name", flexClusterName, err))
+	}
+
+	if err := d.Set("project_id", flexCluster.GetGroupId()); err != nil {
+		return diag.FromErr(fmt.Errorf(ErrorFlexClusterSetting, "project_id", flexClusterName, err))
+	}
+
+	if err := d.Set("state_name", flexCluster.GetStateName()); err != nil {
```
lantoli, line 1522: I guess there are some common fields done for flex and regular clusters, is it worth it to extract common sets to a func?

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2832,3 +2832,115 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+	if tagsCheck {
+		attrMapFlex["tags.testKey"] = "testValue"
+		attrMapAdvCluster["tags.0.key"] = "testKey"
+		attrMapAdvCluster["tags.0.value"] = "testValue"
+	}
+	pluralMap := map[string]string{
+		"project_id": projectID,
+		"results.#":  "1",
+	}
+	checks = acc.AddAttrChecks(acc.FlexDataSourceName, checks, attrMapFlex)
```
lantoli, line 2939: consider CheckRSAndDS

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2832,3 +2832,115 @@ func configFCVPinning(t *testing.T, orgID, projectName, clusterName string, pinn
...
+	checks = acc.AddAttrSetChecks(acc.FlexDataSourceName, checks, attrSetFlex...)
+	checks = acc.AddAttrChecks(acc.FlexDataSourcePluralName, checks, pluralMap)
```
lantoli, line 2941: why are we using CheckRSAndDS but still adding some checks to all resources and data sources?

internal/service/flexcluster/model.go
lantoli: @EspenAlbert would you recommend to change our default zone_name?    A problem can be that users get unexpected plan changes to update their zone name and didn't explicitly set them?

internal/service/flexcluster/resource.go
```diff
@@ -56,20 +65,9 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
...
-	flexClusterParams := &admin.GetFlexClusterApiParams{
-		GroupId: projectID,
-		Name:    clusterName,
-	}
-
-	flexClusterResp, err := WaitStateTransition(ctx, flexClusterParams, connV2.FlexClustersApi, []string{retrystrategy.RetryStrategyCreatingState}, []string{retrystrategy.RetryStrategyIdleState})
+	flexClusterResp, err := CreateFlexCluster(ctx, projectID, clusterName, flexClusterReq, connV2.FlexClustersApi)
 	if err != nil {
-		resp.Diagnostics.AddError("error waiting for resource to be created", err.Error())
+		resp.Diagnostics.AddError("error creating the resource", err.Error())
```
lantoli: can we use consts, like ErrorCreateFlex?

internal/testutil/acc/flex_cluster.go
```diff
@@ -0,0 +1,54 @@
...
+		name := rs.Primary.Attributes["name"]
+		_, _, err := ConnV2().FlexClustersApi.GetFlexCluster(context.Background(), projectID, name).Execute()
+		if err == nil {
+			return fmt.Errorf("flex cluster (%s:%s) still exists", projectID, name)
+		}
+	}
+	return nil
+}
+
+func CheckExistsFlexCluster() resource.TestCheckFunc {
```
lantoli, line 39: do we need a different check func for flex clusters or the regular cluster one can be used?

internal/testutil/acc/flex_cluster.go
```diff
@@ -0,0 +1,54 @@
...
+	data "mongodbatlas_flex_cluster" "test" {
+		project_id = mongodbatlas_flex_cluster.test.project_id
+		name       = mongodbatlas_flex_cluster.test.name
+	}
+	data "mongodbatlas_flex_clusters" "test" {
+		project_id = mongodbatlas_flex_cluster.test.project_id
+	}`
+)
+
+func CheckDestroyFlexCluster(state *terraform.State) error {
```
lantoli, line 24: similar question for check exists, can we use the one for regular clusters?

lantoli (approved):

internal/service/advancedcluster/data_source_advanced_cluster.go
oarbusi: yes it can happen. It is the case of 404 from the cluster API

internal/testutil/acc/flex_cluster.go
oarbusi: Different API are used and a flex cluster can be 2 different resources. In this case it does not make sense to reuse IMO

internal/testutil/acc/flex_cluster.go
oarbusi: See other reply

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: In this test we check the `mongodbatlas_advanced_cluster` resource and data sources (L2944 & L2944) and `mongodbatlas_flex_cluster` data sources (L2939 to 2943)

internal/service/advancedcluster/data_source_advanced_cluster.go
oarbusi: yes, that is why we do the check `if flexClusterResp == nil && clusterDesc == nil {`

lantoli: when not found, err is nil here?

oarbusi: Because of the refactors here, diags will not have errors, flexClusterResp will be nil and clusterDesc as well when cluster api returns 404

.changelog/3001.txt
JuliaMongo: Sounds good. Thank you! Then could you please add this to the changelog entry in this PR:

```
resource/mongodbatlas_advanced_cluster: This resource can now create, read, update, and delete Flex clusters. Previously, this resource only supported dedicated, shared, and free clusters.

```

internal/service/advancedcluster/resource_advanced_cluster.go
JuliaMongo: Thank you for the clarification. I suggest, to avoid any future confusion by an dev reading this in the future, to add a comment to the code about the fact that this variable is NOT related to the shim logic that will create Flex clusters under the hood.

JuliaMongo (approved):
LGTM with a new suggested longer description for the changelog. Thank you, Oriol!

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1268,23 +1320,36 @@ func resourceImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*s
 		return nil, err
 	}
 
+	isFlex := false
 	cluster, _, err := connV2.ClustersApi.GetCluster(ctx, *projectID, *name).Execute()
 	if err != nil {
-		return nil, fmt.Errorf("couldn't import cluster %s in project %s, error: %s", *name, *projectID, err)
+		if isFlex = admin.IsErrorCode(err, "CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API"); !isFlex { // if cluster is flex we need to call different API
```
marcosuma: I'd suggest 
```
admin.IsErrorCode(err, "CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API")
```

to be a method in the common class. It helps encapsulating the business logic and avoid repeating the same one in multiple separate places (in this PR I found two with the same if)

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1268,23 +1320,36 @@ func resourceImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*s
...
+	}
+	clusterID := cluster.GetId()
+	clusterName := cluster.GetName()
+	if isFlex {
+		flexCluster, err := flexcluster.GetFlexCluster(ctx, *projectID, *name, connV2.FlexClustersApi)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't import flex cluster %s in project %s, error: %s", *name, *projectID, err)
+		}
+		clusterID = flexCluster.GetId()
+		clusterName = flexCluster.GetName()
```
marcosuma: with a bit of refactoring, you can actually merge this one and the logic in internal/service/advancedclustertpf/common_admin_sdk.go instead of duplicating code. Right?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1474,81 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
 	_, err := stateConf.WaitForStateContext(ctx)
 	return err
 }
+
+func isFlex(replicationSpecs *[]admin.ReplicationSpec20240805) bool {
```
marcosuma: this method and the other `isFlex` that I see should be centralised in a common place. They are giving you the same information but starting from different inputs.

Also this isFlex seems unrelated to the SDKv2 implementation, it could be reused for tpf, no?

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/unit"
 )
 
-const (
+var (
```
marcosuma: why is this changing?

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1409,3 +1474,81 @@ func waitForUpdateToFinish(ctx context.Context, connV2 *admin.APIClient, project
...
+
+	_, err := flexcluster.UpdateFlexCluster(ctx, projectID, clusterName, flexUpdateRequest, connV2.FlexClustersApi)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf(flexcluster.ErrorUpdateFlex, err))
+	}
+
+	return resourceRead(ctx, d, meta)
+}
+
+func setFlexFields(d *schema.ResourceData, flexCluster *admin.FlexClusterDescription20241113) diag.Diagnostics {
```
marcosuma, line 1488: aren't you going to have this method very similar for the TPF implementation? There is a more elegant way IMO to do this.

You can create a key,value map that is centralized and TF-schema agnostic. Let's say:

```
<"cluster_type", flexCluster.GetClusterType()>,
<"connection_strings", flexcluster.FlattenFlexConnectionStrings(flexCluster.ConnectionStrings),
<"create_date", conversion.TimePtrToStringPtr(flexCluster.CreateDate)>

```

The main pros of this is that you are separating the business logic from the TF schema implementation, and you'll be able to reuse it with the two schemas. It also has a main pro of maintenance: if this has to change in the future for a new field, you only have to do it in one place.

internal/service/advancedcluster/resource_advanced_cluster_test.go
marcosuma: to make sure we don't forget, can you leave a comment to this test with the CLOUDP that will do it?

internal/service/flexcluster/resource.go
```diff
@@ -213,3 +194,78 @@ func splitFlexClusterImportID(id string) (projectID, clusterName *string, err er
 
 	return
 }
+
+func CreateFlexCluster(ctx context.Context, projectID, clusterName string, flexClusterReq *admin.FlexClusterDescriptionCreate20241113, client admin.FlexClustersApi) (*admin.FlexClusterDescription20241113, error) {
```
marcosuma, line 198: nice to see you refactored it to re-use it across resources!

internal/service/flexcluster/resource.go
oarbusi: done [a06cadf](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/a06cadf1257c040df9cb8b1aa3dfc7b98553517a)

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: Initially changed to be able to use pointer. Rolled back the change and instead converting the string constants to variables so we can take their address [8ecb4e3](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/8ecb4e3f136d2c5e1007728253b455fb00cd01a1)

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: I don't think having a key value map would help. It would add an extra object layer inbetween the SDK object and the schema. 
If we have a key/value map, we would still need to do all the d.Set operations for each attribute of the schema in SDKv2 and for TPF we would still need to convert to the TFModel to store it in the state. I think adding a map would add unnecessary complexity but I will consider it when implementing this in TPF and if it makes sense I will do it

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: Moved it to advancedclustertpf package in [8c1625b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/8c1625b057cadd4003d183aeb8c85b3d2f1502e8)

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: done in [6b3ae8e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/6b3ae8ecb0a0f3c635dba2f27d03e6e196bb9d67)

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: I don't this it's necessary after having done the refactor here: [6b3ae8e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001/commits/6b3ae8ecb0a0f3c635dba2f27d03e6e196bb9d67)
`admin.IsErrorCode(err, "CANNOT_USE_FLEX_CLUSTER_IN_CLUSTER_API")` is only done once in the code now

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: I don't think there's a need to put a comment in the test, we have [CLOUDP-297666](https://jira.mongodb.org/browse/CLOUDP-297666) to implement the upgrade to dedicated and this will include adding tests

oarbusi:
Merging, failing tests are new flex cluster tests on `advanced_cluster_tpf` (expected to fail since it's not implemented yet. Ran `TestAccClusterAdvancedClusterConfig_singleShardedTransitionToOldSchemaExpectsError` and `TestAccMockableAdvancedCluster_tenantUpgrade` locally successfully

internal/service/flexcluster/model.go
EspenAlbert: I don't think flex cluster users would be impacted as this is a new case

internal/service/advancedcluster/resource_advanced_cluster.go
marcosuma: @oarbusi make sure you answer questions, I see the PR was merged but this is left unanswered
**update**: I think it's because you answered above already, all good!

oarbusi: replied this question in another thread asking same question https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3001#discussion_r1934053374

PR #3039, opened 2025-02-06
chore: Adds import checks to tests

EspenAlbert:
## Description

Adds import checks to tests

Link to any related issue(s): CLOUDP-298071

## Type of change:

## Required Checklist:


internal/testutil/acc/advanced_cluster.go
```diff
@@ -53,6 +53,7 @@ var (
 func TestStepImportCluster(resourceName string, ignorePrefixFields ...string) resource.TestStep {
 	ignorePrefixFields = append(ignorePrefixFields,
 		"retain_backups_enabled", // This field is TF specific and not returned by Atlas, so Import can't fill it in.
+		"mongo_db_major_version", // Risks plan change of 8 --> 8.0 (always normalized to `major.minor`)
```
EspenAlbert, line 56: Could potentially use a custom type to avoid plan changes, but seems overkill.
Both the apply for `retain_backups_enabled` and `mongo_db_major_version` are instant

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -965,6 +984,7 @@ func TestAccAdvancedCluster_replicaSetScalingStrategyAndRedactClientLogDataOldSc
 				Config: configReplicaSetScalingStrategyAndRedactClientLogDataOldSchema(t, true, orgID, projectName, clusterName, "NODE_TYPE", false),
 				Check:  checkReplicaSetScalingStrategyAndRedactClientLogData(true, "NODE_TYPE", false),
 			},
+			acc.TestStepImportCluster(resourceName, "replication_specs"), // Import with old schema will NOT use `num_shards`
```
EspenAlbert, line 987: Example of explicit `replication_specs` ignore when the old sharding configuration is used

internal/service/advancedclustertpf/move_upgrade_state.go
```diff
@@ -151,7 +151,7 @@ func getTimeoutFromStateObj(stateObj map[string]tftypes.Value) timeouts.Value {
 	return timeouts.Value{Object: obj}
 }
 
-func setOptionalModelAttrs(ctx context.Context, stateObj map[string]tftypes.Value, model *TFModel) {
+func setOptionalModelAttrs(stateObj map[string]tftypes.Value, model *TFModel) {
```
EspenAlbert, line 154: opportunistic removal of unused param

EspenAlbert:
Ran all tests for TPF locally (a few failures due to `replication_specs` and `mongo_db_major_version`, fix in [12ef64d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3039/commits/12ef64de031544d9c4f5319a5e2c42347bbcd560)

Running for SDKv2 [here](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/13188263486). A few failures.
Fixes in 17835c96183210948eee3c60db29bff8fd6a9c77 and [aafc86a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3039/commits/aafc86aa7f6c8875c3497dd381a0b543e678b434)

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -166,6 +166,7 @@ func TestAccMockableAdvancedCluster_tenantUpgrade(t *testing.T) {
 				Config: acc.ConvertAdvancedClusterToSchemaV2(t, true, configTenantUpgraded(projectID, clusterName, defaultZoneName)),
 				Check:  checksTenantUpgraded(projectID, clusterName),
 			},
+			acc.TestStepImportCluster(resourceName),
```
lantoli, line 169: nice to add more test checks/steps

internal/service/advancedcluster/resource_advanced_cluster_test.go
lantoli: nice to add an explanation

internal/testutil/acc/advanced_cluster.go
lantoli: although it's instant, I think it can be confusing for the user to see these plan changes, and think for example that cluster version will change and could lose data / affect cluster config.

we might do something like in Move/Upgrade `setOptionalModelAttrs` :
```
	if mongoDBMajorVersion := getAttrFromStateObj[string](stateObj, "mongo_db_major_version"); mongoDBMajorVersion != nil {
		model.MongoDBMajorVersion = types.StringPointerValue(mongoDBMajorVersion)
	}
```

in Import we don't have previous state but we can get from Atlas.
Or if we shouldn't call Atlas from Import, then we can set a dummy value in mongo_db_major_version so it's not null and the good value will be set in Read.

lantoli (approved):

internal/testutil/acc/advanced_cluster.go
EspenAlbert: Interesting approach. But I don't see how this will solve the problem.
The problem is the support of two different formats in the config:
1. `8`
2. `8.0`
We cannot know at import time which format they used, so we use what is returned by the API (`8.0`)

lantoli: ah, ok, the problem is only when the user specifies mongo_db_major_version in the config?  there are no plan changes if mongo_db_major_version is not in the config?

in that case ok to leave it as it is.

EspenAlbert: Exactly 👍

lantoli: I'm good then, thanks!

PR #3044, opened 2025-02-07
chore: Set default priority (7) for imported flex clusters in mongodbatlas_advanced_cluster

oarbusi:
## Description

Set default priority (7) for imported flex clusters in mongodbatlas_advanced_cluster

Link to any related issue(s): CLOUDP-296377

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedcluster/resource_advanced_cluster_test.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster_test.go
@@ -2905,6 +2905,7 @@ func TestAccClusterFlexCluster_basic(t *testing.T) {
 				Config: configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_1", true),
 				Check:  checkFlexClusterConfig(projectID, clusterName, "AWS", "US_EAST_1", true),
 			},
+			acc.TestStepImportCluster(resourceName),
 			{
 				Config:      configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_2", true),
 				ExpectError: regexp.MustCompile("flex cluster update is not supported except for tags and termination_protection_enabled fields"),
--- a/internal/service/advancedclustertpf/model_flex.go
+++ b/internal/service/advancedclustertpf/model_flex.go
@@ -13,6 +13,8 @@ import (
 	"go.mongodb.org/atlas-sdk/v20241113004/admin"
 )
 
+const defaultPriority int = 7
+
 func NewFlexCreateReq(clusterName string, terminationProtectionEnabled bool, tags *[]admin.ResourceTag, replicationSpecs *[]admin.ReplicationSpec20240805) *admin.FlexClusterDescriptionCreate20241113 {
 	if replicationSpecs == nil || len(*replicationSpecs) == 0 {
 		return nil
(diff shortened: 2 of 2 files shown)


internal/service/advancedclustertpf/model_flex.go
```diff
@@ -124,6 +126,9 @@ func NewTFModelFlexResource(ctx context.Context, diags *diag.Diagnostics, flexCl
 }
 
 func NewTFModelFlex(ctx context.Context, diags *diag.Diagnostics, flexCluster *admin.FlexClusterDescription20241113, priority *int) *TFModel {
+	if priority == nil {
```
lantoli, line 129: can we have a test that would have failed before this change?

internal/service/advancedclustertpf/model_flex.go
```diff
@@ -13,6 +13,8 @@ import (
 	"go.mongodb.org/atlas-sdk/v20241113004/admin"
 )
 
+const defaultPriority int = 7
```
lantoli, line 16:
```suggestion
const defaultPriority = 7
```

no need to say `int` .  or did you try and was failing because conversion.Pointer?

lantoli (approved):

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -2905,6 +2905,7 @@ func TestAccClusterFlexCluster_basic(t *testing.T) {
 				Config: configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_1", true),
 				Check:  checkFlexClusterConfig(projectID, clusterName, "AWS", "US_EAST_1", true),
 			},
+			acc.TestStepImportCluster(resourceName),
```
maastha, line 2908: do we want to update checkFlexClusterConfig to check for this default priority?

internal/service/advancedclustertpf/model_flex.go
oarbusi: New step that is doing the import in the tests would have failed without this

internal/service/advancedclustertpf/model_flex.go
oarbusi: Yes exactly, I needed to put the type to avoid issues when using conversion.Pointer()

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: IMO making sure the plan is empty after import is enough in this specific case (which is being done)

maastha (approved):

EspenAlbert (approved):
Don't remember if we decided to add a check in the plan modifier to error when `priority != 7` for a flex cluster?

- Pro: Can always use 7 for flex clusters and don't have to pass the param. User will not get plan changes if they choose to import cluster later
- Con: Possibly an extra step for the user to change priority before creating cluster

oarbusi:
> Don't remember if we decided to add a check in the plan modifier to error when `priority != 7` for a flex cluster?
> 
> * Pro: Can always use 7 for flex clusters and don't have to pass the param. User will not get plan changes if they choose to import cluster later
> * Con: Possibly an extra step for the user to change priority before creating cluster

@EspenAlbert I don't think that is possible to do in the planmodifier because we need to check that priority is 7 **when it's a flex cluster**. As far as I know in the plan modifier we don't have access to other attributes to do this check

EspenAlbert:
> > Don't remember if we decided to add a check in the plan modifier to error when `priority != 7` for a flex cluster?
> > 
> > * Pro: Can always use 7 for flex clusters and don't have to pass the param. User will not get plan changes if they choose to import cluster later
> > * Con: Possibly an extra step for the user to change priority before creating cluster
> 
> @EspenAlbert I don't think that is possible to do in the planmodifier because we need to check that priority is 7 **when it's a flex cluster**. As far as I know in the plan modifier we don't have access to other attributes to do this check

You have full access in the resource plan modifier:
[useStateForUnknowns(ctx, diags, &plan, &state)](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/advancedclustertpf/resource.go#L109)

internal/service/advancedclustertpf/model_flex.go
lantoli: great, thanks

PR #3083, opened 2025-02-18
chore: Adds support for plan modifier on replication_specs changed

EspenAlbert:
## Description

Adds support for plan modifier on replication_specs changed

Link to any related issue(s): CLOUDP-300282

Full Runs in TPF:
- [old implementation using Plan in PATCH](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/13400040306)
- [new implementation using Config in PATCH](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/13438837247)

## Type of change:

## Required Checklist:


internal/service/advancedclustertpf/model_to_ClusterDescription20240805.go
```diff
@@ -95,8 +95,8 @@ func newReplicationSpec20240805(ctx context.Context, input types.List, diags *di
 	for i := range elements {
 		item := &elements[i]
 		resp[i] = admin.ReplicationSpec20240805{
-			Id:            conversion.NilForUnknown(item.ExternalId, item.ExternalId.ValueStringPointer()),
-			ZoneId:        conversion.NilForUnknown(item.ZoneId, item.ZoneId.ValueStringPointer()),
+			Id:            conversion.NilForUnknownOrEmptyString(item.ExternalId),
```
EspenAlbert, line 98: Doesn't make sense to send an empty string for the `id`

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -2,38 +2,123 @@ package advancedclustertpf
...
-	"go.mongodb.org/atlas-sdk/v20241113005/admin"
 )
 
-func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, plan, state *TFModel) {
+func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
 	if !schemafunc.HasUnknowns(plan) {
 		return
 	}
-	patchReq, upgradeRequest := findClusterDiff(ctx, state, plan, diags, &update.PatchOptions{})
+	_, upgradeRequest := findClusterDiff(ctx, state, plan, diags, &update.PatchOptions{})
```
EspenAlbert: As findClusterDiff is only used in Update and here, we can probably simplify, but leaving that to a follow-up PR

internal/common/schemafunc/find_changes.go
```diff
@@ -0,0 +1,201 @@
...
+			leaf = strings.Split(leaf, "[")[0]
+		}
+		leafChanges[leaf] = true
+	}
+	return leafChanges
+}
+
+// FindAttributeChanges: Iterates through TFModel of state+plan and returns AttributeChanges for querying changed attributes
+// The implementation is similar to KeepUnknown, no support for types.Set or types.Tuple yet
+func FindAttributeChanges(ctx context.Context, src, dest any) AttributeChanges {
```
EspenAlbert, line 85: Why another reflect function?
- Use the same names as TF and therefore simplify call to CopyUnknowns.
- Could re-use the major parts of KeepUnknown
- Could have tried to modify `PatchPayload` to send back attributes changed, but it wasn't a good fit. PatchPayload does 2 things:
  - (1) Checking for changes on an attribute and for changed attributes
  - (2) Copies values from state
  - (3) Require Structs with `json` annotations to work
- Had an attempt at using the [path](https://github.com/hashicorp/terraform-plugin-framework/tree/main/path) package, but I was unsuccessful.

internal/common/schemafunc/find_changes.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (a *AttributeChanges) AttributeChanged(name string) bool {
+	changes := a.LeafChanges()
+	changed := changes[name]
+	return changed
+}
+
+func (a *AttributeChanges) KeepUnknown(attributeEffectedMapping map[string][]string) []string {
+	keepUnknown := []string{}
```
lantoli:
```suggestion
	var keepUnknown []string
```

nit: you can keep the var nil, and append will create it

internal/common/schemafunc/find_changes.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return false
+}
+
+func (a *AttributeChanges) leafChanges(removeIndex bool) map[string]bool {
+	leafChanges := map[string]bool{}
+	for _, change := range a.Changes {
+		var leaf string
+		parts := strings.Split(change, ".")
+		if len(parts) == 1 {
```
lantoli, line 74: just to confirm, can len(parts) be 0 sometimes?

internal/common/schemafunc/find_changes.go
```diff
@@ -0,0 +1,201 @@
...
+			leaf = parts[0]
```
lantoli: I think this `if` clause isn't needed, if len(parts) == 1 then  leaf = parts[len(parts)-1] is equivalent to parts[0]

can len(parts) be 0?

internal/common/schemafunc/find_changes.go
```diff
@@ -0,0 +1,201 @@
...
+// The implementation is similar to KeepUnknown, no support for types.Set or types.Tuple yet
+func FindAttributeChanges(ctx context.Context, src, dest any) AttributeChanges {
+	changes := FindChanges(ctx, src, dest)
+	return AttributeChanges{changes}
+}
+
+func FindChanges(ctx context.Context, src, dest any) []string {
+	valSrc, valDest := validateStructPointers(src, dest)
+	typeDest := valDest.Type()
+	changes := []string{} // Always return an empty list, as nested attributes might be added and then removed
```
lantoli:
```suggestion
	var changes []string // Always return an empty list, as nested attributes might be added and then removed
```

similar to a previous comment, can we return nil? it's normally preferred to an explicit empty slice

internal/common/schemafunc/find_changes.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+			changes = append(changes, moreChanges...)
+			continue
+		}
+	}
+	return changes
+}
+
+func findChangesInObject(ctx context.Context, src, dest types.Object, parentPath []string) []string {
+	changes := []string{}
```
lantoli: similar comment, can we init to nil?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -120,30 +131,50 @@ func validateKeepUnknown(keepUnknown []string) {
 }
 
 func copyUnknownsFromObject(ctx context.Context, src, dest types.Object, keepUnknown []string) types.Object {
+	// if something is null in the state and unknown in plan, we expect it to remain null
+	if src.IsNull() && dest.IsUnknown() {
+		return src
+	}
 	if src.IsNull() || dest.IsNull() {
```
lantoli, line 139: can we add a comment about this condition?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -168,3 +208,36 @@ func copyUnknownsFromList(ctx context.Context, src, dest types.List, keepUnknown
 	}
 	return types.ListValueMust(dest.ElementType(ctx), merged)
 }
+
+// https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types@v1.13.0
+func asUnknownValue(ctx context.Context, value attr.Value) attr.Value {
+	switch v := value.(type) {
+	case types.List:
```
lantoli, line 216: nit: this can probably be converted to a map from type to unknown object

lantoli (approved):

internal/common/schemafunc/plan_modifier_state_for_unknown.go
EspenAlbert: Not so sure, some values (e.g., types.Object, use `object.AttributeTypes(ctx)`) requires knowing the specific type and using it. Also, the `ctx` var, but maybe I'm missing something?

internal/common/schemafunc/find_changes.go
EspenAlbert: Thanks. Changed in 7dc03d07da78dda5733625fb913e89d5dfd67f82

internal/common/schemafunc/find_changes.go
EspenAlbert: Thanks. Changed in 7dc03d07da78dda5733625fb913e89d5dfd67f82

internal/common/schemafunc/find_changes.go
EspenAlbert: Updated the comment. Does it answer the question?

internal/common/schemafunc/find_changes.go
EspenAlbert: Thanks. Changed in 7dc03d07da78dda5733625fb913e89d5dfd67f82

internal/common/schemafunc/plan_modifier_state_for_unknown.go
EspenAlbert: Thanks. Added in 7dc03d07da78dda5733625fb913e89d5dfd67f82

internal/common/schemafunc/plan_modifier_state_for_unknown.go
lantoli: it's ok as it is

internal/common/schemafunc/find_changes.go
lantoli: do you mean this comment?:  , which make the test cases fail on nil vs []

I'm not sure yet, what's the difference between nil and empty slice string here?

EspenAlbert: Yes, that is the one I mean. It was simpler to keep it always `string[]{}` to avoid the failing test assertion.
More details:
The code works by adding a field to changes when it compares two fields.
However, even though the top level field is different, that could be because some values are unknown, therefore, it might be removed after, see L145 and L152:

```go
			moreChanges := findChangesInList(ctx, attributesSrc[name].(types.List), tfListDest, path)
			if len(moreChanges) == 0 {
				changes = slices.Delete(changes, len(changes)-1, len(changes))
			}
```

lantoli: ah, ok, i get it.  We could assert that if len is 0 in expected and got then that's ok else assertEqual, but ok to leave it as it is

EspenAlbert:
Merging now, with the caveat of changing the env-var before releasing, and making the decision for plan modifier after more data

PR #3089, opened 2025-02-19
chore: Fixes inconsistent result errors

EspenAlbert:
## Description

Fixes inconsistent result errors

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -20,7 +20,7 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, plan, sta
 	if diags.HasError() {
 		return
 	}
-	flexUpgrade, _ := flexChanges(planReq, stateReq, diags)
+	flexUpgrade, _ := flexUpgradedUpdated(planReq, stateReq, diags)
 	if diags.HasError() {
 		return
 	}
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -25,6 +25,7 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, plan, sta
 		return
 	}
 	if flexUpgrade {
+		// The flex cluster API doesn't return the same fields as the tenant API; therefore, computed fields will be `null` after the upgrade
 		keepUnknown := []string{"connection_strings", "state_name", "advanced_configuration", "encryption_at_rest_provider", "root_cert_type", "bi_connector_config"}
 		keepUnknown = append(keepUnknown, keepUnknownTenantUpgrade...)
 		schemafunc.CopyUnknowns(ctx, state, plan, keepUnknown)
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -9,6 +9,8 @@ import (
 	"go.mongodb.org/atlas-sdk/v20241113005/admin"
 )
 
+var keepUnknownTenantUpgrade = []string{"disk_size_gb", "cluster_id", "replication_specs", "backup_enabled", "create_date"}
+
 func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, plan, state *TFModel) {
 	if !schemafunc.HasUnknowns(plan) {
 		return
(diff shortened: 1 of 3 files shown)


oarbusi (approved):
LGMT! Thanks

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -9,10 +9,28 @@ import (
...
+	planReq := normalizeFromTFModel(ctx, plan, diags, false)
+	if diags.HasError() {
+		return
+	}
+	flexUpgrade, _ := flexChanges(planReq, stateReq, diags)
+	if diags.HasError() {
+		return
+	}
+	if flexUpgrade {
+		keepUnknown := []string{"connection_strings", "state_name", "advanced_configuration", "encryption_at_rest_provider", "root_cert_type", "bi_connector_config"}
```
lantoli, line 29: maybe a small comment about why we're keeping these attributes

lantoli (approved):

internal/service/advancedclustertpf/resource.go
```diff
@@ -560,3 +555,17 @@ func handleFlexUpdate(ctx context.Context, diags *diag.Diagnostics, client *conf
 	}
 	return newFlexModel
 }
+
+func flexChanges(planReq, stateReq *admin.ClusterDescription20240805, diags *diag.Diagnostics) (isUpgrade, isUpdate bool) {
```
lantoli: nit: returning 2 bool is difficult to see which is which (without looking at the func signature), maybe change the func name such as flexUpgradedUpdated

lantoli (approved):

internal/service/advancedclustertpf/resource.go
```diff
@@ -232,25 +232,20 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
-
-	if IsFlex(planReq.ReplicationSpecs) {
-		if isValidUpgradeToFlex(stateReq, planReq) {
-			upgradeModel := handleFlexUpgrade(ctx, diags, r.Client, waitParams, planReq, &plan)
-			if diags.HasError() {
-				return
-			}
-			diags.Append(resp.State.Set(ctx, upgradeModel)...)
-			return
+	flexUpgrade, flexUpdate := flexChanges(planReq, stateReq, diags)
```
marcosuma: could we take all this code out and extract in separate method?

marcosuma (approved):

internal/service/advancedclustertpf/resource.go
EspenAlbert: It would require 6 params and it is only used here. Not a strong opinion, but prefer keeping it for now

internal/service/advancedclustertpf/plan_modifier.go
EspenAlbert: Thanks. Added: f2827aacc0e43e34a2669babb432257dddfbfc34

internal/service/advancedclustertpf/resource.go
EspenAlbert: Thanks. Changed in f2827aacc0e43e34a2669babb432257dddfbfc34

internal/service/advancedclustertpf/resource.go
```diff
@@ -560,3 +555,17 @@ func handleFlexUpdate(ctx context.Context, diags *diag.Diagnostics, client *conf
 	}
 	return newFlexModel
 }
+
+func flexUpgradedUpdated(planReq, stateReq *admin.ClusterDescription20240805, diags *diag.Diagnostics) (isUpgrade, isUpdate bool) {
+	if !IsFlex(planReq.ReplicationSpecs) {
+		return false, false
+	}
+	if isValidUpgradeToFlex(stateReq, planReq) {
```
lantoli, line 563: @EspenAlbert I guess upgrade takes precedence so we don't allow both upgrade and update at the same time?

oarbusi: Not allowed to do upgrade & update in the same apply. Upgrade will be applied and then update needs to be done seperately

lantoli: @oarbusi thanks. should we send an error if both things happen?  
are we doing now  to ignore the update, do only the upgrade when changes are applied? 
and then they'll see plan changes and will need to apply again?   

If this is the case, it can be a bit confusing for customers,  as they will normally expect to apply all the changes and then have empty plans.

Issue #3103, opened 2025-02-25
[Bug]: imported project has a wrong configuration and unusable


VladimirUsov:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Provider Version

1.27.0

### Terraform Version

1.10.5

### Terraform Edition

Terraform Open Source (OSS)

### Current Behavior

Importing an existing Atlas project works but results in the wrong attribute value ("with_default_alerts_settings": null), which in turn leads to the inability to use the imported configuration—it just times out during a Terraform apply. Manually editing the state file and correcting the attribute value results in no errors when applying the configuration. Tested with several different provider versions, specifying the latest for the bug report.

### Terraform configuration to reproduce the issue

```hcl
resource "mongodbatlas_project" "project" {
  name                          = "testAtlasImport"
  org_id                        = var.org_id
  with_default_alerts_settings  = true
}
```

### Steps To Reproduce

1. Create a Mongo Atlas project via the provided Terraform template. (it actually does not matter if you provide the attribute with_default_alerts_settings or not)
2. Rename/move away your current terraform state file
3. Import project by running:  _terraform import mongodbatlas_project.project "${project_id}"_
4. Run terraform plan - it will show you that the _with_default_alerts_settings_ is going to be changed even though it's the same.
5. Run _terraform apply_ - this will just halt and **timeout** after 40m+

Optional (workaround):
6. Manually edit the state file created during the import. Change the "with_default_alerts_settings": null to "with_default_alerts_settings": true
7. Re-apply terraform - this time it would run successfully and report that configuration is matching. 

Expectation: the attribute would receive its proper value during the state during the import "with_default_alerts_settings": true and NOT "with_default_alerts_settings": null, and so applying the project after importing wouldn't fail. 

### Logs

```txt
PS C:\RESEARCH\Atlas\terraform-mongodbatlas-atlas-basic\mini> terraform apply

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # mongodbatlas_project.project will be created
  + resource "mongodbatlas_project" "project" {
      + cluster_count                                    = (known after apply)
      + created                                          = (known after apply)
      + id                                               = (known after apply)
      + ip_addresses                                     = (known after apply)
      + is_collect_database_specifics_statistics_enabled = (known after apply)
      + is_data_explorer_enabled                         = (known after apply)
[...]

PR #3105, opened 2025-02-25
test: Ensures project `withDefaultAlertsSettings` works with import and introduce create_only plan modifier

EspenAlbert:
## Description

Link to any related issue(s): CLOUDP-302586 & #3103

Updated context:
- The API now returns `with_default_alert_settings`, before this was only set on create and never returned.
- This can impact users:

### During Plan
- [x] If the user did not define the attribute, the default value of `true` should match what the API returns,  covered by: `TestMigProject_basic`
- [x] If the user defined the attribute in their config during create and never made changes, there should be no plan changes, covered by: `TestMigProject_withTrueDefaultAlertSettings`
- [x] same as previous but with `false` value, covered by: `TestMigProject_withFalseDefaultAlertSettings`
- [x] If the user created the project and later changed the value, the API and the config will differ and they will see an error when they run a `terraform plan`, see the [error message](#error-message) below.

The user will be forced to update their config to match the API value, otherwise they will get inconsistent result value.

### During Update

Before this change, Terraform allowed updates to `with_default_alert_settings`, but now we will raise an error:

#### Error message:
Example when the user has modified the attribute value after creation (created with `with_default_alerts_settings= true`)
```hcl
resource "mongodbatlas_project" "this" {
  org_id = var.org_id
  name   = "analytics_test"
  with_default_alerts_settings = false
}
```

```log
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: with_default_alerts_settings cannot be updated or set after import, remove it from the configuration or use the state value (see below).
│ 
│   with mongodbatlas_project.this,
│   on main.tf line 21, in resource "mongodbatlas_project" "this":
│   21: resource "mongodbatlas_project" "this" {
│ 
│ The current state value is true
```

- As soon as the user removes/updates the value, they will get No Plan Changes

### During Import

Import works now due to the API change, and the experience will be the same as described in `During Plan` above

## Type of change:

## Required Checklist:


internal/service/project/resource_project.go
oarbusi, line 1: Is it possible to test this?

.changelog/3105.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/mongodbatlas_project: Sets default value for `with_default_alerts_settings` during state import
```
EspenAlbert: I wouldn't consider this a **critical** bug, as it is only affecting `import` and the fix is running a normal `terraform apply`

oarbusi: What difference does it make if it's critical or not? It's still a bug

EspenAlbert: I was thinking about our discussion about the Migration guide content from yesterday and wanted to add some "color" to the change

internal/service/project/resource_project.go
EspenAlbert: Yup: Changed in 5e67e149912ecfc6a40e71124d6abe61ddbec26c

oarbusi (approved):

internal/service/project/resource_project.go
```diff
@@ -333,9 +333,13 @@ func (r *projectRS) ImportState(ctx context.Context, req resource.ImportStateReq
 }
 
 func updatePlanFromConfig(projectPlanNewPtr, projectPlan *TFProjectRSModel) {
-	// we need to reset defaults from what was previously in the state:
-	// https://discuss.hashicorp.com/t/boolean-optional-default-value-migration-to-framework/55932
-	projectPlanNewPtr.WithDefaultAlertsSettings = projectPlan.WithDefaultAlertsSettings
+	// After import the state&plan will be null so we set to true (default value)
```
EspenAlbert: The user will still get a plan change after import if they have set `with_default_alert_settings=false` in their config. But we don't have access to that information in the import

.changelog/3105.txt
lantoli: maybe in the changelog we can say WHAT is fixed instead of HOW is fixed, e.g.:

```
Fixes import 
```

or 
```
Fixes import  when `with_default_alerts_settings` is set
```

etc.

in the current message it's not clear what is the impact/benefit of this change for the customer

lantoli (approved):

internal/service/project/resource_project.go
maastha: Since this is still a possibility, I'm thinking if we should even implement this fix. This sounds like it should be expected behavior that WithDefaultAlertsSettings will not be set during import and maybe something we can call out in our documentation instead.

I'd find it harder to explain to users why import shows an update if `with_default_alert_settings=false` in their config but works when it's true.

internal/service/project/resource_project.go
```diff
@@ -333,9 +333,13 @@ func (r *projectRS) ImportState(ctx context.Context, req resource.ImportStateReq
 }
 
 func updatePlanFromConfig(projectPlanNewPtr, projectPlan *TFProjectRSModel) {
-	// we need to reset defaults from what was previously in the state:
-	// https://discuss.hashicorp.com/t/boolean-optional-default-value-migration-to-framework/55932
```
marcosuma, line 337: @EspenAlbert is it worth it to leave that reference?

internal/service/project/resource_project.go
```diff
@@ -333,9 +333,13 @@ func (r *projectRS) ImportState(ctx context.Context, req resource.ImportStateReq
...
-	projectPlanNewPtr.WithDefaultAlertsSettings = projectPlan.WithDefaultAlertsSettings
+	// After import the state&plan will be null so we set to true (default value)
+	if projectPlanNewPtr.WithDefaultAlertsSettings.IsNull() && projectPlan.WithDefaultAlertsSettings.IsNull() {
+		projectPlanNewPtr.WithDefaultAlertsSettings = types.BoolValue(true)
+	} else {
+		// Force value from plan as this is not returned from the API to avoid inconsistent result errors
```
marcosuma: it just feels a workaround from an API limitation, have we reached out to the team owning this feature?

marcosuma:
@EspenAlbert nit: can we have a why in the summary section to explain the change? also let's hold this merge until we have a conversation with the upstream team (if we did, we should connect the CLOUDPs and understand the why on their side)

kanchana-mongodb (approved):
LGTM

internal/service/project/resource_project.go
EspenAlbert: I checked the thread, and I'm not sure it adds much, but curious what others think.
We have a few places where this is done, for example we do the same for null vs [] tags

.changelog/3105.txt
EspenAlbert: Thanks. Updated in 57ddae6009bcba5fef6d37ddede76f2c111ea2d3

internal/service/project/resource_project.go
EspenAlbert: I understand is a create-only parameter that has a side effect of creating alert settings, but I'll check with upstream. [Api Docs](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Projects/operation/createProject)

internal/service/project/resource_project.go
maastha: I'd leave that as is, this was one of the first resources migrated to TPF, I think the reference may be helpful for any new folks not completely aware of TPF migration issues.

EspenAlbert: Added back in 7bc763ed6481b25a311faa31eae7befc3708260d

EspenAlbert:
Will get back to this on Wednesday

internal/common/customplanmodifier/create_only.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	planmodifier "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+type Modifier interface {
+	planmodifier.String
+	planmodifier.Bool
+}
+
+func CreateOnlyAttributePlanModifier() Modifier {
```
maastha: Can we add a comment on what this plan modifier does & why it's needed?

EspenAlbert:
> Will get back to this on Wednesday

Hope to get to this by the end of this week (after 1.29.0) release

internal/service/project/resource_project_test.go
```diff
@@ -647,9 +648,20 @@ func TestAccGovProject_withProjectOwner(t *testing.T) {
 
 func TestAccProject_withFalseDefaultSettings(t *testing.T) {
 	var (
-		orgID          = os.Getenv("MONGODB_ATLAS_ORG_ID")
-		projectOwnerID = os.Getenv("MONGODB_ATLAS_PROJECT_OWNER_ID")
-		projectName    = acc.RandomProjectName()
+		orgID              = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectOwnerID     = os.Getenv("MONGODB_ATLAS_PROJECT_OWNER_ID")
+		projectName        = acc.RandomProjectName()
+		importResourceName = resourceName + "2"
```
EspenAlbert: This test case is a bit different than usual, I tried to add comments, but let me know if anything is unclear.

marcosuma (changes requested):
@EspenAlbert let's hold on this one. I am not too convinced we are doing the right thing in adding more complexity to handle something like this. Let's chat in today's tech sync.

internal/service/project/resource_project.go
EspenAlbert: Outdated. Now returned by the API

internal/service/project/resource_project.go
EspenAlbert: Changed by API team

internal/common/customplanmodifier/create_only.go
```diff
@@ -0,0 +1,91 @@
...
+	return d.defaultBool != nil
+}
+
+func (d *createOnlyAttributePlanModifier) PlanModifyBool(ctx context.Context, req planmodifier.BoolRequest, resp *planmodifier.BoolResponse) {
+	if isCreate(&req.State) {
+		if !IsKnown(req.PlanValue) && d.UseDefault() {
+			resp.PlanValue = types.BoolPointerValue(d.defaultBool)
+		}
+		return
+	}
```
AgustinBettati, line 63: q: Is there a reason we have logic for setting the default value as part of the plan modifier, and not in the read operation?
I am suspecting is has to do with handling create vs import paths different so curious if we can be explicit and have that logic in the import function vs in the plan modifier.

EspenAlbert: Yes, by setting this in the plan modifier we make sure the default is visible in the plan and only applied during the `create` operation.

AFAIK: The import operation doesn't use the plan modifier, it only calls read and stores it in the state.

AgustinBettati: Makes sense. Would suggest adding additional clarification (as comment in `CreateOnlyAttributePlanModifierWithBoolDefault`) on why during create we set the default value (for improving the plan), and we can also clarify that given this logic the create operation does not need to populate this attribute in any way.

EspenAlbert: Good point. Added in 841840527c54133ee1600ff6383a6fb2b80d8e41

marcosuma (approved):

PR #3110, opened 2025-02-25
chore: Refactors unknown logic

lantoli:
## Description

Refactors unknown logic to simplify it, including:
- Don't do update logic with upgrades
- Remove env. var.

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/advancedclustertpf/common_admin_sdk.go
+++ b/internal/service/advancedclustertpf/common_admin_sdk.go
@@ -122,7 +122,7 @@ func ReadIfUnsetAdvancedConfiguration(ctx context.Context, diags *diag.Diagnosti
 	return configLegacy, configNew
 }
 
-func TenantUpgrade(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, req *admin.LegacyAtlasTenantClusterUpgradeRequest) *admin.ClusterDescription20240805 {
+func UpgradeTenant(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, req *admin.LegacyAtlasTenantClusterUpgradeRequest) *admin.ClusterDescription20240805 {
 	_, _, err := client.AtlasV2.ClustersApi.UpgradeSharedCluster(ctx, waitParams.ProjectID, req).Execute()
 	if err != nil {
 		addErrorDiag(diags, operationTenantUpgrade, defaultAPIErrorDetails(waitParams.ClusterName, err))
@@ -131,7 +131,7 @@ func TenantUpgrade(ctx context.Context, diags *diag.Diagnostics, client *config.
 	return AwaitChanges(ctx, client, waitParams, operationTenantUpgrade, diags)
 }
 
-func FlexToDedicatedUpgrade(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, req *admin.AtlasTenantClusterUpgradeRequest20240805) *admin.ClusterDescription20240805 {
+func UpgradeFlexToDedicated(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, req *admin.AtlasTenantClusterUpgradeRequest20240805) *admin.ClusterDescription20240805 {
 	_, _, err := client.AtlasV2.FlexClustersApi.UpgradeFlexCluster(ctx, waitParams.ProjectID, req).Execute()
 	if err != nil {
 		addErrorDiag(diags, operationFlexUpgrade, defaultAPIErrorDetails(waitParams.ClusterName, err))
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -13,13 +13,6 @@ import (
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/update"
 )
 
-const (
-	minimizeLevelNever   = "never"
-	minimizeLevelDefault = "default"
-	minimizeLevelAlways  = "always"
-	envVarNameMinimize   = "MONGODB_ATLAS_PLAN_MINIMIZE"
-)
-
 var (
 	// The flex cluster API doesn't return the same fields as the tenant API; therefore, computed fields will be `null` after the upgrade
 	keepUnknownTenantToFlex      = []string{"connection_strings", "state_name", "advanced_configuration", "encryption_at_rest_provider", "root_cert_type", "bi_connector_config", "mongo_db_major_version"}
@@ -44,22 +37,6 @@ var (
 		"zone_name":       {"zone_id"},         // zone_id copy from state is not safe when
 	}
 )
-
-func getMinimizeLevel() string {
-	envValue := strings.ToLower(os.Getenv(envVarNameMinimize))
-	if envValue == "" {
-		return minimizeLevelAlways // Experimenting with always to try to find bugs
-	}
-	return envValue
-}
-
-func minimizeNever() bool {
-	return getMinimizeLevel() == minimizeLevelNever
-}
-
-func minimizeAlways() bool {
-	return getMinimizeLevel() == minimizeLevelAlways
-}
 
 func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
 	if !schemafunc.HasUnknowns(plan) {
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -39,9 +39,6 @@ var (
 )
 
 func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
-	if !schemafunc.HasUnknowns(plan) {
-		return
-	}
 	stateReq := normalizeFromTFModel(ctx, state, diags, false)
 	planReq := normalizeFromTFModel(ctx, plan, diags, false)
 	if diags.HasError() {
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -28,6 +28,7 @@ var (
 	}
 )
 
+// useStateForUnknowns should be called only in Update, because of findClusterDiff
 func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
 	diff := findClusterDiff(ctx, state, plan, diags)
 	if diags.HasError() {
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -29,18 +29,12 @@ var (
 )
 
 func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
-	stateReq := normalizeFromTFModel(ctx, state, diags, false)
-	planReq := normalizeFromTFModel(ctx, plan, diags, false)
-	if diags.HasError() {
-		return
-	}
-	isUpgradeTenantToFlex, _ := flexUpgradedUpdated(planReq, stateReq, diags)
 	diff := findClusterDiff(ctx, state, plan, diags)
 	if diags.HasError() {
 		return
 	}
-	if isUpgradeTenantToFlex || diff.isUpgradeTenant() || diff.isUpgradeFlexToDedicated() {
-		return // Don't do anything in upgrades
+	if diff.isAnyUpgrade() { // Don't do anything in upgrades
+		return
 	}
 	attributeChanges := schemafunc.FindAttributeChanges(ctx, state, plan)
 	keepUnknown := []string{"connection_strings", "state_name"} // Volatile attributes, should not be copied from state
@@ -51,7 +45,6 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, pl
 	}
 }
 
-// TODO: last change to use instead of sdk model
 func useStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel, attrChanges *schemafunc.AttributeChanges) {
 	stateRepSpecsTF := TFModelList[TFReplicationSpecsModel](ctx, diags, state.ReplicationSpecs)
 	planRepSpecsTF := TFModelList[TFReplicationSpecsModel](ctx, diags, plan.ReplicationSpecs)
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -8,7 +8,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/schemafunc"
-	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/update"
 )
 
 var (
@@ -35,12 +34,12 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, pl
 	if diags.HasError() {
 		return
 	}
-	isTenantToFlex, _ := flexUpgradedUpdated(planReq, stateReq, diags)
-	diff := findClusterDiff(ctx, state, plan, diags, &update.PatchOptions{})
+	isUpgradeTenantToFlex, _ := flexUpgradedUpdated(planReq, stateReq, diags)
+	diff := findClusterDiff(ctx, state, plan, diags)
 	if diags.HasError() {
 		return
 	}
-	if isTenantToFlex || diff.isUpgradeTenant() || diff.isUpgradeFlexToDedicated() {
+	if isUpgradeTenantToFlex || diff.isUpgradeTenant() || diff.isUpgradeFlexToDedicated() {
 		return // Don't do anything in upgrades
 	}
 	attributeChanges := schemafunc.FindAttributeChanges(ctx, state, plan)
--- a/internal/service/advancedclustertpf/plan_modifier.go
+++ b/internal/service/advancedclustertpf/plan_modifier.go
@@ -70,44 +70,46 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, pl
 	if diags.HasError() {
 		return
 	}
-	flexUpgrade, _ := flexUpgradedUpdated(planReq, stateReq, diags)
-	if diags.HasError() {
-		return
-	}
-	if flexUpgrade {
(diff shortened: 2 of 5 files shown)


internal/service/advancedclustertpf/resource.go
```diff
@@ -505,15 +504,34 @@ func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string
...
+func (c *clusterDiff) isUpgradeTenant() bool {
+	return c.upgradeTenantReq != nil
+}
+
+func (c *clusterDiff) isUpgradeFlexToDedicated() bool {
+	return c.upgradeFlexToDedicatedReq != nil
+}
+
+// TODO: Struct for the response, including tenantToFlex and isUpdate, also payloads?
+func findClusterDiff(ctx context.Context, state, plan *TFModel, diags *diag.Diagnostics, options *update.PatchOptions) clusterDiff {
```
EspenAlbert: nice! Next step is possibly:
- Integrate `flexUpgradedUpdated`
- Handle `update.PatchOptions` instead of as an argument

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -45,69 +29,31 @@ var (
...
+	diff := findClusterDiff(ctx, state, plan, diags, &update.PatchOptions{})
 	if diags.HasError() {
 		return
 	}
-	if flexUpgrade {
-		keepUnknownTenantToFlex = append(keepUnknownTenantToFlex, tenantUpgradeRootKeepUnknown...)
-		schemafunc.CopyUnknowns(ctx, state, plan, keepUnknownTenantToFlex)
-		return
+	if isTenantToFlex || diff.isUpgradeTenant() || diff.isUpgradeFlexToDedicated() {
+		return // Don't do anything in upgrades
```
EspenAlbert: I like this simplification. Haven't seen the user experience difference, but might be worth returning early after `replication_specs` instead

lantoli: do you mean to update root attributes (although that would be returning later)? can you clarify?

EspenAlbert: Yes, the previous `determineKeepUnknownsRoot` and possibly combining the `keepUnknownFlexUpgrade` and `tenantUpgradeRootKeepUnknown`

internal/service/advancedclustertpf/resource.go
lantoli: flexUpgradedUpdated and findClusterDiff use SDK vs TF model, so I think it's better to keep them separate, joining them will just put everything together but without any simplification or duplicate reduction.

lantoli: can you elaborate about handling PatchOptions?

EspenAlbert: Exactly why I think we can re-use.
Notice how the stateReq and planReq below is created, these can be used for the flexUpgradedUpdated call

internal/service/advancedclustertpf/common_admin_sdk.go
```diff
@@ -122,7 +122,7 @@ func ReadIfUnsetAdvancedConfiguration(ctx context.Context, diags *diag.Diagnosti
 	return configLegacy, configNew
 }
 
-func TenantUpgrade(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, req *admin.LegacyAtlasTenantClusterUpgradeRequest) *admin.ClusterDescription20240805 {
+func UpgradeTenant(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, req *admin.LegacyAtlasTenantClusterUpgradeRequest) *admin.ClusterDescription20240805 {
```
lantoli, line 125: @EspenAlbert nit: do you know why these 2 funcs are in common file when they're not being used out of tpf? should they be moved to a non-common file?

internal/service/advancedclustertpf/plan_modifier.go
lantoli: my rationale to keep it as it is, is to simplify and avoid unexpected errors while proving a good user experience in plan outputs.   In this case where upgrade is one-time activity, I think it's ok just to show anything in the plan as the user will expect changes in an upgrade.     I think  upgrade logic can be tricky to be tested in all edge cases and I think it doesn't help so much as changes in a regular cluster (without upgrading)

internal/service/advancedclustertpf/resource.go
EspenAlbert: PatchOptions, used in the update:

```go
patchOptions := update.PatchOptions{
    IgnoreInStatePrefix: []string{"replicationSpecs"}, // only use config values for replicationSpecs, state values might come from the UseStateForUnknowns and shouldn't be used, `id` is added in updateLegacyReplicationSpecs
}
if usingNewShardingConfig(ctx, configModel.ReplicationSpecs, diags) {
    patchOptions.IgnoreInStateSuffix = append(patchOptions.IgnoreInStateSuffix, "id") // Not safe to send replication_spec.*.id when using the new schema: replicationSpecs.java.util.ArrayList[0].id attribute does not match expected format
}
```

used in the plan_modifier: ` &update.PatchOptions{}`
As the plan_modifier handles `replication_spec` changeson its own, we can avoid providing it as an argument, and use the logic from update here instead

internal/service/advancedclustertpf/plan_modifier.go
EspenAlbert: Ok, I agree :+1:

internal/service/advancedclustertpf/resource.go
```diff
@@ -505,15 +504,34 @@ func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string
...
+
+func (c *clusterDiff) isUpgradeTenant() bool {
+	return c.upgradeTenantReq != nil
+}
+
+func (c *clusterDiff) isUpgradeFlexToDedicated() bool {
+	return c.upgradeFlexToDedicatedReq != nil
+}
+
+// TODO: Struct for the response, including tenantToFlex and isUpdate, also payloads?
```
AgustinBettati: this can be removed right?

internal/service/advancedclustertpf/resource.go
```diff
@@ -505,15 +504,34 @@ func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string
 	return timeoutDuration
 }
 
-func findClusterDiff(ctx context.Context, state, plan *TFModel, diags *diag.Diagnostics, options *update.PatchOptions) (*admin.ClusterDescription20240805, *admin.LegacyAtlasTenantClusterUpgradeRequest, *admin.AtlasTenantClusterUpgradeRequest20240805) {
+type clusterDiff struct {
```
AgustinBettati, line 491:
```suggestion
type clusterPatchReq struct {
```
can help with readability

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -141,11 +83,8 @@ func useStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
 }
 
 // determineKeepUnknownsChangedReplicationSpec: These fields must be kept unknown in the replication_specs[index_of_changes]
-func determineKeepUnknownsChangedReplicationSpec(keepUnknownsAlways []string, isTenantUpgrade bool, attributeChanges *schemafunc.AttributeChanges, parentPath string) []string {
+func determineKeepUnknownsChangedReplicationSpec(keepUnknownsAlways []string, attributeChanges *schemafunc.AttributeChanges, parentPath string) []string {
 	var keepUnknowns = slices.Clone(keepUnknownsAlways)
-	if isTenantUpgrade {
-		keepUnknowns = append(keepUnknowns, tenantUpgradeReplicationSpecKeepUnknown...)
-	}
```
AgustinBettati, line 148: is there a reason this can be removed? (I see this is related to https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3110/files#r1971121687)

internal/service/advancedclustertpf/resource.go
```diff
@@ -558,7 +577,7 @@ func handleFlexUpdate(ctx context.Context, diags *diag.Diagnostics, client *conf
 	return newFlexModel
 }
 
-func flexUpgradedUpdated(planReq, stateReq *admin.ClusterDescription20240805, diags *diag.Diagnostics) (isUpgrade, isUpdate bool) {
+func flexUpgradedUpdated(planReq, stateReq *admin.ClusterDescription20240805, diags *diag.Diagnostics) (isTenantToFlex, isUpdate bool) {
```
AgustinBettati: nit: Is there room for improving naming here? Was hard to read when see the function call.

AgustinBettati (approved):
LGTM with some doubts

internal/service/advancedclustertpf/common_admin_sdk.go
lantoli: discussed offline, will keep them so they can be reused in the future

internal/service/advancedclustertpf/plan_modifier.go
lantoli: I'm simplifying the logic, so we only do the "unknown magic" when it's a normal update, we don't do anything when it's an upgrade (e.g. from tenant to dedicated or flex to dedicated, etc.) as there might be edges case and anyway this is a one-time change and customers expect changes

internal/service/advancedclustertpf/resource.go
lantoli: clusterDiff is really the response from findClusterDiff so I think the name is ok

internal/service/advancedclustertpf/resource.go
lantoli: yes, sorry, will do, it was to make sure i address all changes i want to do

internal/service/advancedclustertpf/resource.go
lantoli: yes, i'm trying to improve names but great if you have any suggestion

internal/service/advancedclustertpf/plan_modifier.go
AgustinBettati: Agree with this mindset so we don't make the implementation too complex/risky

marcosuma: good point Leo!

marcosuma (approved):

internal/service/advancedclustertpf/resource.go
lantoli: thanks @EspenAlbert for the offline catch-up.    PatchOptions is now in findClusterDiff: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3110/commits/b2e2f0628eeaf0d5bb268d5687bb5b79dc500e49

internal/service/advancedclustertpf/resource.go
lantoli: flexUpgradedUpdated doesn't exist anymore and its logic is in findClusterDiff: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3110/commits/ef0a6703a6833f6e41f59a841df14b3489030d4c

internal/service/advancedclustertpf/resource.go
lantoli: flexUpgradedUpdated doesn't exist anymore and its logic is in findClusterDiff: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3110/commits/ef0a6703a6833f6e41f59a841df14b3489030d4c

lantoli: @EspenAlbert is there any other improvement to do re: this comment or it's ok to resolve?

internal/service/advancedclustertpf/resource.go
```diff
@@ -225,61 +229,40 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
...
-			diags.Append(resp.State.Set(ctx, flexOut)...)
-		}
-		return
-	}
-
-	var clusterResp *admin.ClusterDescription20240805
 
 	// FCV update is intentionally handled before any other cluster updates, and will wait for cluster to reach IDLE state before continuing
-	clusterResp = r.applyPinnedFCVChanges(ctx, diags, &state, &configModel, waitParams)
+	clusterResp := r.applyPinnedFCVChanges(ctx, diags, &state, &configModel, waitParams)
```
lantoli, line 238: @EspenAlbert @AgustinBettati can you confirm it's ok to do the pinned FCV Update logic before the Flex upgrade / update ?   (I assume that flex can't be pinned fcv so this order is not important).   In this way we can have the switch with all upgrade cases including flex

AgustinBettati: Not sure if flex supports pinning FCV, but I would preserve having pinned FCV as the first operation as called out in our docs: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/advanced_cluster#pinned_fcv-1

> If a plan contains multiple changes, the FCV change will be applied first

oarbusi: Flex does not support pinning FCV

EspenAlbert: I see three alternative solutions (probably more):
1. Add validator to pinned_fcv field or in resource plan_modifier that raise error if pin fcv is set for a flex cluster
2. Move the flex switch cases before and exit early
3. Do nothing, and let the user face the API error (my preference)

internal/service/advancedclustertpf/resource.go
```diff
@@ -505,71 +488,104 @@ func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string
...
-func findClusterDiff(ctx context.Context, state, plan *TFModel, diags *diag.Diagnostics, options *update.PatchOptions) (*admin.ClusterDescription20240805, *admin.LegacyAtlasTenantClusterUpgradeRequest, *admin.AtlasTenantClusterUpgradeRequest20240805) {
-	isShardingUpgrade := isShardingConfigUpgrade(ctx, state, plan, diags)
-	if diags.HasError() {
-		return nil, nil, nil
-	} // old sharding config  (num_shards > 1) to new one
+type clusterDiff struct {
+	clusterPatchOnlyReq       *admin.ClusterDescription20240805
+	upgradeTenantReq          *admin.LegacyAtlasTenantClusterUpgradeRequest
+	upgradeFlexToDedicatedReq *admin.AtlasTenantClusterUpgradeRequest20240805
+	isUpgradeTenantToFlex     bool
```
EspenAlbert, line 495: I think `isUpgradeTenantToFlex` and `isUpdateOfFlex` should be `*admin.LegacyAtlasTenantClusterUpgradeRequest` and `*admin.FlexClusterDescriptionUpdate20241113` instead.
And the `isMethods` can be similar to existing methods.
But if it makes it less readable, I'm ok keeping as is too

EspenAlbert (approved):
Thank you!

internal/service/advancedclustertpf/resource.go
AgustinBettati: +1 to your preference, API error > custom tf logic

internal/service/advancedclustertpf/resource.go
EspenAlbert: Awesome job <3 You can consider https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3110#discussion_r1971723253

internal/service/advancedclustertpf/resource.go
```diff
@@ -98,7 +99,7 @@ type rs struct {
 // 1. UseStateForUnknown always copies the state for unknown values. However, that leads to `Error: Provider produced inconsistent result after apply` in some cases (see implementation below).
 // 2. Adding the different UseStateForUnknown is very verbose.
 func (r *rs) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
-	if req.Plan.Raw.IsNull() || req.State.Raw.IsNull() { // Can be null in case of destroy
+	if req.Plan.Raw.IsNull() || req.State.Raw.IsNull() { // Don't do anything if plan or state is null, this happens in Create, Delete or Import
```
EspenAlbert, line 102: [nit] Suggestion to add what we discussed offline:
`planModifier.IsUpdate(req)`

internal/service/advancedclustertpf/resource.go
oarbusi: should the name be reconsidered now that findClusterDiff does more things and clusterDiff returns 2 new boolean attributes?

lantoli:
@EspenAlbert @AgustinBettati can you confirm this is the flaky tests that sometimes fails in Test Suite? or i might have introduce the problem in this PR? thanks https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/13544105335/job/37851920586?pr=3110

internal/service/advancedclustertpf/resource.go
lantoli: the problem is that it also needs the line below `!schemafunc.HasUnknowns(&plan)` to check that is not Read, and it needs the plan.

internal/service/advancedclustertpf/resource.go
lantoli: I tried to change it to findClusterChanges but i'm not convinced, and also there is some attributeChanges that make it confusing.    I think diff is ok as all info is derived from the difference between the state and the plan, but let me know if you have a suggestion and can try it

internal/service/advancedclustertpf/resource.go
lantoli: sounds good, thank you all

internal/service/advancedclustertpf/resource.go
lantoli: you're right, but I see that implies quite a few refactors, e.g. GetFlexClusterUpdateRequest is returning admin.FlexClusterDescriptionUpdate20241113. 

so I prefer not do it in this PR as it will have many extra changes that were not the main purpose if this PR.

internal/service/advancedclustertpf/resource.go
lantoli: > Not sure if flex supports pinning FCV, but I would preserve having pinned FCV as the first operation 

@AgustinBettati before my change, flex logic was before FCV, my change makes FCV first

internal/service/advancedclustertpf/resource.go
```diff
@@ -505,71 +488,104 @@ func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string
...
-	if isValidUpgradeToFlex(stateReq, planReq) {
-		return true, false
+func handleFlexUpdate(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, configModel *TFModel) *TFModel {
+	configReq := normalizeFromTFModel(ctx, configModel, diags, false)
+	if diags.HasError() {
+		return nil
 	}
-	if isValidUpdateOfFlex(stateReq, planReq) {
-		return false, true
+	flexCluster, err := flexcluster.UpdateFlexCluster(ctx, configModel.ProjectID.ValueString(), configModel.Name.ValueString(),
```
lantoli, line 581: @oarbusi can you confirm that flexcluster.UpdateFlexCluster waits until changes are done, or we should do an explicit wait as we do in Upgrade?

oarbusi: flexcluster.UpdateFlexCluster waits on the changes being done

PR #3127, opened 2025-03-03
chore: Adds support for conditionally keepUnknown based on state value (node_count and auto_scaling fix)

EspenAlbert:
## Description

Adds support for conditionally keepUnknown based on state value (node_count and auto_scaling fix)

Link to any related issue(s): CLOUDP-303920

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/schemafunc/plan_modifier_state_for_unknown.go
+++ b/internal/common/schemafunc/plan_modifier_state_for_unknown.go
@@ -34,6 +34,55 @@ func HasUnknowns(obj any) bool {
 	}
 	return false
 }
+func CopyUnknownsWithCall(ctx context.Context, src, dest any, keepUnknownCall func(string, attr.Value) bool) {
+
+	valSrc, valDest := validateStructPointers(src, dest)
+	typeSrc := valSrc.Type()
+	typeDest := valDest.Type()
+	for i := range typeDest.NumField() {
+		fieldDest := typeDest.Field(i)
+		name, tfName := fieldNameTFName(&fieldDest)
+		srcValue := valSrc.FieldByName(name).Interface()
+		if keepUnknownCall(tfName, srcValue.(attr.Value)) {
+			continue
+		}
+		_, found := typeSrc.FieldByName(name)
+		if !found || !valDest.Field(i).CanSet() {
+			continue
+		}
+		nestedDest := valDest.FieldByName(name).Interface()
+		objValueSrc, okSrc := srcValue.(types.Object)
+		objValueDest, okDest := nestedDest.(types.Object)
+		if okSrc && okDest {
+			objValueNew := copyUnknownsFromObject(ctx, objValueSrc, objValueDest, keepUnknownCall)
+			valDest.Field(i).Set(reflect.ValueOf(objValueNew))
+			continue
+		}
+		listValueSrc, okSrc := srcValue.(types.List)
+		listValueDest, okDest := nestedDest.(types.List)
+		if okSrc && okDest {
+			listValueNew := copyUnknownsFromList(ctx, listValueSrc, listValueDest, keepUnknownCall)
+			valDest.Field(i).Set(reflect.ValueOf(listValueNew))
+			continue
+		}
+		if isUnknown(valDest.Field(i)) && !keepUnknownCall(tfName, srcValue.(attr.Value)) {
+			tflog.Info(ctx, fmt.Sprintf("Copying unknown field: %s\n", name))
+			valDest.Field(i).Set(valSrc.FieldByName(name))
+			continue
+		}
+	}
+}
+
+func CombineKeepUnknownCalls(calls ...func(string, attr.Value) bool) func(string, attr.Value) bool {
+	return func(name string, value attr.Value) bool {
+		for _, call := range calls {
+			if call(name, value) {
+				return true
+			}
+		}
+		return false
+	}
+}
 
 // CopyUnknowns use reflection to copy unknown fields from src to dest.
 // The implementation is similar to internal/common/conversion/model_generation.go#CopyModel
@@ -41,41 +90,9 @@ func HasUnknowns(obj any) bool {
 // nestedStructMapping is a map of field names to their type: object, list. (`set` not implemented yet)
 func CopyUnknowns(ctx context.Context, src, dest any, keepUnknown []string) {
 	validateKeepUnknown(keepUnknown)
-	valSrc, valDest := validateStructPointers(src, dest)
-	typeSrc := valSrc.Type()
-	typeDest := valDest.Type()
-	for i := range typeDest.NumField() {
-		fieldDest := typeDest.Field(i)
-		name, tfName := fieldNameTFName(&fieldDest)
-		if slices.Contains(keepUnknown, tfName) {
-			continue
-		}
-		_, found := typeSrc.FieldByName(name)
-		if !found || !valDest.Field(i).CanSet() {
-			continue
-		}
-		nestedSrc := valSrc.FieldByName(name).Interface()
-		nestedDest := valDest.FieldByName(name).Interface()
-		objValueSrc, okSrc := nestedSrc.(types.Object)
-		objValueDest, okDest := nestedDest.(types.Object)
-		if okSrc && okDest {
-			objValueNew := copyUnknownsFromObject(ctx, objValueSrc, objValueDest, keepUnknown)
-			valDest.Field(i).Set(reflect.ValueOf(objValueNew))
-			continue
-		}
-		listValueSrc, okSrc := nestedSrc.(types.List)
-		listValueDest, okDest := nestedDest.(types.List)
-		if okSrc && okDest {
-			listValueNew := copyUnknownsFromList(ctx, listValueSrc, listValueDest, keepUnknown)
-			valDest.Field(i).Set(reflect.ValueOf(listValueNew))
-			continue
-		}
-		if isUnknown(valDest.Field(i)) {
-			tflog.Info(ctx, fmt.Sprintf("Copying unknown field: %s\n", name))
-			valDest.Field(i).Set(valSrc.FieldByName(name))
-			continue
-		}
-	}
+	CopyUnknownsWithCall(ctx, src, dest, func(name string, value attr.Value) bool {
+		return slices.Contains(keepUnknown, name)
+	})
 }
 
 func fieldNameTFName(fieldDest *reflect.StructField) (name, tfName string) {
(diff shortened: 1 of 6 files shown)


internal/common/schemafunc/plan_modifier_state_for_unknown_test.go
```diff
@@ -85,6 +87,20 @@ func TestHasUnknown(t *testing.T) {
 	}
 }
 
+type TFAutoScalingModel struct {
```
lantoli, line 90: can't TFAutoScalingModel and AutoScalingObjType in schema.go be used? or you're having a subset in the test file?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -39,38 +39,60 @@ func HasUnknowns(obj any) bool {
 // The implementation is similar to internal/common/conversion/model_generation.go#CopyModel
 // keepUnknown is a list of fields that should not be copied, should always use the TF config name (snake_case)
 // nestedStructMapping is a map of field names to their type: object, list. (`set` not implemented yet)
-func CopyUnknowns(ctx context.Context, src, dest any, keepUnknown []string) {
+func CopyUnknowns(ctx context.Context, src, dest any, keepUnknown []string, keepUnknownCall ...func(string, attr.Value) bool) {
```
AgustinBettati: Would suggest defining an interface for func(string, attr.Value) bool, this way you can receive all criteria in a single list. Example:

```
type KeepUnknownCriteria interface {
    ShouldKeepUnknown(name string, value attr.Value) bool
}

func NewKeepUnknownCriteria(f func(string, attr.Value) bool) KeepUnknownCriteria {
    return KeepUnknownFunc(f)
}

func KeepAttributesInList(names []string) KeepUnknownCriteria {
    return NewKeepUnknownCriteria(func(name string, _ attr.Value) bool {
        return slices.Contains(names, name)
    })
}
```

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -27,6 +28,22 @@ var (
...
+	autoScalingStringValues = []string{"compute_min_instance_size", "compute_max_instance_size"}
+	keepUnknownsCalls       = []func(string, attr.Value) bool{
+		// node_count should only be copied from state when it is 0
+		func(name string, value attr.Value) bool {
+			return name == "node_count" && !value.Equal(types.Int64Value(0))
+		},
+		// auto_scaling bool attributes should only be copied from state when they are false
+		func(name string, value attr.Value) bool {
+			return slices.Contains(autoScalingBoolValues, name) && value.Equal(types.BoolValue(true))
+		},
```
AgustinBettati: description mentions that value should be copied when false (zero value), however the assertion is checking for true. Checking for false would be correct right?

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -39,7 +56,7 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, pl
 	keepUnknown := []string{"connection_strings", "state_name"} // Volatile attributes, should not be copied from state
 	keepUnknown = append(keepUnknown, attributeChanges.KeepUnknown(attributeRootChangeMapping)...)
 	keepUnknown = append(keepUnknown, determineKeepUnknownsAutoScaling(ctx, diags, state, plan)...)
-	schemafunc.CopyUnknowns(ctx, state, plan, keepUnknown)
+	schemafunc.CopyUnknowns(ctx, state, plan, keepUnknown, keepUnknownsCalls...)
```
lantoli: do we need to use keepUnknownsCalls here in the root, or just in the replication_specs loop?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -39,38 +39,60 @@ func HasUnknowns(obj any) bool {
...
+		listValueSrc, okSrc := srcValue.(types.List)
 		listValueDest, okDest := nestedDest.(types.List)
 		if okSrc && okDest {
-			listValueNew := copyUnknownsFromList(ctx, listValueSrc, listValueDest, keepUnknown)
+			listValueNew := copyUnknownsFromList(ctx, listValueSrc, listValueDest, keepUnknownCall)
 			valDest.Field(i).Set(reflect.ValueOf(listValueNew))
 			continue
 		}
-		if isUnknown(valDest.Field(i)) {
+		if isUnknown(valDest.Field(i)) && !keepUnknownCall(tfName, srcValue.(attr.Value)) {
```
lantoli: aren't we already doing the keepUnknownCall check in line 73?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
AgustinBettati: you could also leverage something like
```
func CombineCriteria(criteria ...KeepUnknownCriteria) KeepUnknownCriteria {
    return NewKeepUnknownCriteria(func(name string, value attr.Value) bool {
        for _, c := range criteria {
            if c.ShouldKeepUnknown(name, value) {
                return true
            }
        }
        return false
    })
}
```
So `CopyUnknowns` only receives a single `KeepUnknownCriteria`

AgustinBettati: One aspect to consider maybe as a follow up refactor. Would it be possible to keep criteria of which attribute can have unknown removed (known to be stable)? vs current approach of defining which attribute needs unknown preserved (known to be volatile).
This might be on the safer side because if a new attribute is added, by default right now it will have the unknown removed unless stated otherwise.

internal/common/schemafunc/plan_modifier_state_for_unknown_test.go
EspenAlbert: I like to keep a simplified self contained model here, but can change it

internal/common/schemafunc/plan_modifier_state_for_unknown.go
EspenAlbert: Looks interesting, but feels like more work for the caller

internal/service/advancedclustertpf/plan_modifier.go
EspenAlbert: tried to update the description. WDYT?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
EspenAlbert: Thanks, changed in c3e82cbf4356b322535cc47f8f16368b70f40771

internal/common/schemafunc/plan_modifier_state_for_unknown.go
EspenAlbert: Interesting 🤔 
A problem is when an unrelated attribute is changed.
For example, changing `tags`.
In that case, you want to mark all attributes in replication_specs.* as Known to avoid any Known After Apply in the state to hide the `replication_specs` altogether.

The current keepUnknown also depends on which attribute changes.

We would need to refactor:
If instance_size is changed, mark disk_iops as keepUnknown.
to
disk_iops is stable unless instance_size is changed

internal/common/schemafunc/plan_modifier_state_for_unknown_test.go
lantoli: that's fair, thanks.   maybe I would change the names then to make it clearer that they're not the real ones but test ones

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -147,7 +169,8 @@ func copyUnknownsFromObject(ctx context.Context, src, dest types.Object, keepUnk
 		attributesDest = fillUnknowns(ctx, attributesSrc)
 	}
 	for name, attr := range attributesDest {
-		if slices.Contains(keepUnknown, name) {
+		replacement := attributesSrc[name]
+		if keepUnknownCall(name, replacement) {
 			attributesMerged[name] = attr
 			continue
 		}
```
lantoli, line 174:
```suggestion
		
		if replacement := attributesSrc[name]; keepUnknownCall(name, replacement) {
			attributesMerged[name] = attr
			continue
		}
```

nit

internal/common/schemafunc/plan_modifier_state_for_unknown.go
lantoli: @EspenAlbert @AgustinBettati WDYT about this commit? https://github.com/mongodb/terraform-provider-mongodbatlas/commit/a038ef09613b8890a408c660a257241f5e58d198

AgustinBettati: @lantoli yeah that commit is more of a middle ground. I do like the idea of having a single interface to avoid the multiple arguments in the function call, additionally from the caller point of view it will be explicit you are defining different criteria for preserving unknown values. Mot a strong difference so will leave it up to you.

internal/service/advancedclustertpf/plan_modifier.go
AgustinBettati: Thanks! So this would be to cover the case the auto-scaling block is removed right?

internal/common/schemafunc/plan_modifier_state_for_unknown.go
AgustinBettati: Good point, so we with this approach we would need to defined criteria for essentially all attributes which can become unfeasible. I think we can discard this option, thanks for the inputs.

internal/service/advancedclustertpf/plan_modifier.go
EspenAlbert: discussed offline. Ok to remove

internal/common/schemafunc/plan_modifier_state_for_unknown.go
lantoli: thanks @AgustinBettati , I'll pick that commit then

lantoli (approved):

internal/common/schemafunc/plan_modifier_state_for_unknown.go
EspenAlbert: SGTM. Looks like it is merged in [e43d5c0](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3127/commits/e43d5c05d59b48df582cc91ff642e97498965672). Resolving

internal/service/advancedclustertpf/plan_modifier.go
EspenAlbert: Yes

PR #3139, opened 2025-03-07
feat: Add `error_message`, `provider_account_id`, and `interface_endpoint_name` to Stream PrivateLinkEndpoint models

tcannon91:
## Description
Add fields that exist in the Atlas Admin API for Streams PrivateLink Endpoints (error_message, provider_account_id, interface_endpoint_name). Also improve validation around the vendor and provider_name fields to fail faster (during planning) rather than waiting until getting a response from the API that is inconsistent with potential passed in values.

Link to any related issue(s):
https://jira.mongodb.org/browse/CLOUDP-301851 

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/streamprivatelinkendpoint/model.go
+++ b/internal/service/streamprivatelinkendpoint/model.go
@@ -18,7 +18,6 @@ func NewTFModel(ctx context.Context, projectID string, apiResp *admin.StreamsPri
 	result := &TFModel{
 		Id:                    types.StringPointerValue(apiResp.Id),
 		DnsDomain:             types.StringPointerValue(apiResp.DnsDomain),
-		DnsSubDomain:          types.ListNull(types.StringType),
 		ErrorMessage:          types.StringPointerValue(apiResp.ErrorMessage),
 		ProjectId:             types.StringPointerValue(&projectID),
 		InterfaceEndpointId:   types.StringPointerValue(apiResp.InterfaceEndpointId),
@@ -31,13 +30,11 @@ func NewTFModel(ctx context.Context, projectID string, apiResp *admin.StreamsPri
 		Vendor:                types.StringPointerValue(apiResp.Vendor),
 	}
 
-	if apiResp.DnsSubDomain != nil {
-		subdomain, diags := types.ListValueFrom(ctx, types.StringType, apiResp.GetDnsSubDomain())
-		if diags.HasError() {
-			return nil, diags
-		}
-		result.DnsSubDomain = subdomain
+	subdomain, diags := types.ListValueFrom(ctx, types.StringType, apiResp.GetDnsSubDomain())
+	if diags.HasError() {
+		return nil, diags
 	}
+	result.DnsSubDomain = subdomain
 
 	return result, nil
 }
--- a/internal/service/streamprivatelinkendpoint/model.go
+++ b/internal/service/streamprivatelinkendpoint/model.go
@@ -16,16 +16,21 @@ const (
 
 func NewTFModel(ctx context.Context, projectID string, apiResp *admin.StreamsPrivateLinkConnection) (*TFModel, diag.Diagnostics) {
 	result := &TFModel{
-		Id:                  types.StringPointerValue(apiResp.Id),
-		DnsDomain:           types.StringPointerValue(apiResp.DnsDomain),
-		ProjectId:           types.StringPointerValue(&projectID),
-		InterfaceEndpointId: types.StringPointerValue(apiResp.InterfaceEndpointId),
-		Provider:            types.StringPointerValue(apiResp.Provider),
-		Region:              types.StringPointerValue(apiResp.Region),
-		ServiceEndpointId:   types.StringPointerValue(apiResp.ServiceEndpointId),
-		State:               types.StringPointerValue(apiResp.State),
-		Vendor:              types.StringPointerValue(apiResp.Vendor),
+		Id:                    types.StringPointerValue(apiResp.Id),
+		DnsDomain:             types.StringPointerValue(apiResp.DnsDomain),
+		DnsSubDomain:          types.ListNull(types.StringType),
+		ErrorMessage:          types.StringPointerValue(apiResp.ErrorMessage),
+		ProjectId:             types.StringPointerValue(&projectID),
+		InterfaceEndpointId:   types.StringPointerValue(apiResp.InterfaceEndpointId),
+		InterfaceEndpointName: types.StringPointerValue(apiResp.InterfaceEndpointName),
+		Provider:              types.StringPointerValue(apiResp.Provider),
+		ProviderAccountId:     types.StringPointerValue(apiResp.ProviderAccountId),
+		Region:                types.StringPointerValue(apiResp.Region),
+		ServiceEndpointId:     types.StringPointerValue(apiResp.ServiceEndpointId),
+		State:                 types.StringPointerValue(apiResp.State),
+		Vendor:                types.StringPointerValue(apiResp.Vendor),
 	}
+
 	if apiResp.DnsSubDomain != nil {
 		subdomain, diags := types.ListValueFrom(ctx, types.StringType, apiResp.GetDnsSubDomain())
 		if diags.HasError() {
--- a/internal/service/streamprivatelinkendpoint/resource_schema.go
+++ b/internal/service/streamprivatelinkendpoint/resource_schema.go
@@ -2,9 +2,17 @@ package streamprivatelinkendpoint
 
 import (
 	"context"
+	"regexp"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var uppercaseValidator = stringvalidator.RegexMatches(
+	regexp.MustCompile("^[A-Z]+$"),
+	"must contain only uppercase characters",
 )
 
 func ResourceSchema(ctx context.Context) schema.Schema {
@@ -23,6 +31,10 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				MarkdownDescription: "Sub-Domain name of Confluent cluster. These are typically your availability zones.",
 				ElementType:         types.StringType,
 			},
+			"error_message": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Error message if the connection is in a failed state.",
+			},
 			"project_id": schema.StringAttribute{
 				Required:            true,
 				MarkdownDescription: "Unique 24-hexadecimal digit string that identifies your project. Use the [/groups](#tag/Projects/operation/listProjects) endpoint to retrieve all projects to which the authenticated user has access.\n\n**NOTE**: Groups and projects are synonymous terms. Your group id is the same as your project id. For existing groups, your group or project id remains the same. The resource and corresponding endpoints use the term groups.",
@@ -31,9 +43,20 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "Interface endpoint ID that is created from the specified service endpoint ID.",
 			},
+			"interface_endpoint_name": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Interface endpoint name that is created from the specified service endpoint ID.",
+			},
+			"provider_account_id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Account ID from the cloud provider.",
+			},
 			"provider_name": schema.StringAttribute{
 				Required:            true,
 				MarkdownDescription: "Provider where the Kafka cluster is deployed.",
+				Validators: []validator.String{
+					uppercaseValidator,
+				},
 			},
 			"region": schema.StringAttribute{
 				Optional:            true,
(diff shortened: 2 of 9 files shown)


internal/service/streamprivatelinkendpoint/resource_schema.go
```diff
@@ -31,9 +43,20 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+				MarkdownDescription: "Interface endpoint name that is created from the specified service endpoint ID.",
+			},
+			"provider_account_id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Account ID from the cloud provider.",
+			},
 			"provider_name": schema.StringAttribute{
 				Required:            true,
 				MarkdownDescription: "Provider where the Kafka cluster is deployed.",
+				Validators: []validator.String{
```
marcosuma: @tcannon91 are these the breaking changes you are talking about by putting the `feat!`? We want to be extremely careful in not releasing breaking changes that aren't strictly needed.

maastha: Additionally, we usually let the API handle validation so as to not replicate logic in Terraform, does the API return a meaningful error message if the input value for this is not uppercase?

tcannon91: Yes, these are the breaking changes I was talking about. The API does not return a meaningful error message. What happens is everything will actually work, but the payload returned by the API will capitalize some resources (the API types are enums) when the payload is returned. Terraform has validation where the passed in structure does not match the returned structure so it will throw an error:
```
╷
│ 
│ When applying changes to mongodbatlas_stream_privatelink_endpoint.test, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new value: .provider_name: was cty.StringVal("Azure"), but now cty.StringVal("AZURE").
│ 
```

There actually is no error from the API -- it will compare case insensitively to correctly match a cloud provider, but then return our normalized enum type (which is all caps). In this case, the terraform call returns an error but the requested resource actually was created and exists within atlas.

If we typically handle cases like this in a different way I can accommodate that.

marcosuma: @tcannon91 thanks for the clarifications. The best way to fix this is that the API request should fail if the input is not capitalized (see [IPA-111](https://docs.devprod.prod.corp.mongodb.com/ipa/111) on field ownership). 
Changing a field value that was set by the user (which is what the API currently does) should always be avoided. Instead a validation error should be returned saying that field should be capitalized.
We have a best practice of minimizing validations on Terraform and make them happen on the API. Besides Terraform, in fact, there are many other IaCs that would need this same logic (CFN, K8s, ...)

tcannon91: Sounds good. I will work on that. Just to lay out the process you are looking for here, it will go something like this, right?
1. Make changes at the api layer, needing a new major version
2. Propagate changes to a new major version of the sdk
3. Incorporate that latest sdk version into the terraform provider, requiring a new major version for the provider

Then this can be released.

Just wanted to tag my project lead so he can project that timeline out if it is acceptable to him. CC @jwongmongodb

jwongmongodb: Thanks for the update, I think it makes sense for this API change. We can update the project channel with the details of the extra work needed..

marcosuma: I think the API that accepts lowercase and returns UPPERCASE is a bug and so should be fixed. But your team knows this probably better.

tcannon91: Since the behavior from the API is existing today, I will likely revert this validation so we can push through some of these updates and my team can follow up with the bug fixes at the api layer and propagate them out to the consumer touchpoints (cli, terraform, etc)

jwongmongodb: I agree, a separate ticket to fix this endpoint feels like it can be done outside of this current ticket.

tcannon91: I went ahead and removed these changes and will remove the breaking changes indicators from the PR

internal/service/streamprivatelinkendpoint/model.go
```diff
@@ -16,16 +16,21 @@ const (
...
-		ProjectId:           types.StringPointerValue(&projectID),
-		InterfaceEndpointId: types.StringPointerValue(apiResp.InterfaceEndpointId),
-		Provider:            types.StringPointerValue(apiResp.Provider),
-		Region:              types.StringPointerValue(apiResp.Region),
-		ServiceEndpointId:   types.StringPointerValue(apiResp.ServiceEndpointId),
-		State:               types.StringPointerValue(apiResp.State),
-		Vendor:              types.StringPointerValue(apiResp.Vendor),
+		Id:                    types.StringPointerValue(apiResp.Id),
+		DnsDomain:             types.StringPointerValue(apiResp.DnsDomain),
+		DnsSubDomain:          types.ListNull(types.StringType),
```
marcosuma: I see you added this but it's not documented in the docs (unless I am missing it) - is that expected?

marcosuma (approved):
LGTM - good to have another pair of eyes cc @oarbusi

internal/service/streamprivatelinkendpoint/model.go
oarbusi: Why is dnsSubDomain being set to a null list here? is there a reason for this change now?

tcannon91: I was having failures when interacting with the models when I don't provide `DnsSubDomains`.

```
╷
│ Error: Value Conversion Error
│ 
│   with data.mongodbatlas_stream_privatelink_endpoints.plural_datasource,
│   on main.tf line 15, in data "mongodbatlas_stream_privatelink_endpoints" "plural_datasource":
│   15: data "mongodbatlas_stream_privatelink_endpoints" "plural_datasource" {
│ 
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Expected framework type from provider logic: types.ListType[basetypes.StringType] / underlying type: tftypes.List[tftypes.String]
│ Received framework type from provider logic: types.ListType[!!! MISSING TYPE !!!] / underlying type: tftypes.List[tftypes.DynamicPseudoType]
│ Path: results[0].dns_sub_domain
```

I think that this is required because the API respond with a null value for this list when it is not provided by the user. I am not familiar with terraform typings so if this error should be resolved in a different way, let me know :) 


As for the documentation change, I am not entirely sure what is expected here. I don't think this should change anything user facing.

marcosuma: Regarding the documentation, my bad: I thought it wasn't added in the documentation. But double checked now and it's there.

Regarding why this is needed now, I am not sure - was there a change on the API side? or maybe this was a bug? I am also not clear why this is needed now.

tcannon91: I believe this was a bug that always existed. I am hypothesizing that minimal testing was done without providing a dnsSubDomain previously. I can see we had unit tests covering it [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/f9f644d35ad4f72234c76af4557c842d6ea60078/internal/service/streamprivatelinkendpoint/model_test.go#L66) but I believe that is "upstream" of where the error I encountered was occurring. I believe the error message above is a terraform error, not from our provider logic so it would only be caught with acceptance tests or manual testing. 

Our acceptance tests pass in with dns subdomains (the second true parameter [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/f9f644d35ad4f72234c76af4557c842d6ea60078/internal/service/streamprivatelinkendpoint/resource_test.go#L53)) so we may have only caught this previously if manual testing was done without providing dns subdomains.

marcosuma: ok that's possible. thank you

tcannon91: Will wait for @oarbusi 's thoughts since I think he worked on a lot of this. If this is covering up something wrong or there is a different way to solve this, let me know :)

marcosuma:
@tcannon91  today we'll release 1.29 ([ref](https://mongodb.slack.com/archives/C04K3F1EGGK/p1741690177262579)). This change won't go in there but we're planning to release next week as well. For reference, this is the [changelog file](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CHANGELOG.md) to see what it'll be released next

internal/service/streamprivatelinkendpoint/model.go
EspenAlbert: You need to do it on a separate line and use `types.ListValueFrom(ctx, types.StringType, apiResp.GetDnsSubDomain())`
Similar to the code [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/advancedclustertpf/model_ClusterDescriptionProcessArgs20240805.go#L60-L67)

oarbusi: @tcannon91  Thanks for the details, it could be a miss on our side. Just to confirm, from the error message you shared, I see is happening for the plural data source. Does it happen also for singular data source & resource?

tcannon91: @oarbusi no, I only saw this with the plural data source. But it is possible I created a private link instance without the singular data source in my terraform files and then had the plural when I had multiple resources. 

I can try to play with this a bit more. 

Can someone explain their concern with my approach here? I am happy to adjust but I also want to learn what the problem is so I can be sure to understand and make better decisions going forward.

oarbusi: The approach itself is not concerning, it's the fact that it's now needed. We might have missed it when developing the resource, or something might have changed in the API. That's why we are trying to understand the change.
I think this change is okay, thanks for checking this and providing the fix. Appreciate it

tcannon91: I encountered the same error this morning with singular data source

```
│ Error: Value Conversion Error
│ 
│   with data.mongodbatlas_stream_privatelink_endpoint.singular_datasource,
│   on main.tf line 10, in data "mongodbatlas_stream_privatelink_endpoint" "singular_datasource":
│   10: data "mongodbatlas_stream_privatelink_endpoint" "singular_datasource" {
│ 
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please report the following
│ to the provider developer:
│ 
│ Expected framework type from provider logic: types.ListType[basetypes.StringType] / underlying type: tftypes.List[tftypes.String]
│ Received framework type from provider logic: types.ListType[!!! MISSING TYPE !!!] / underlying type: tftypes.List[tftypes.DynamicPseudoType]
│ Path: dns_sub_domain
```

And while running a `terraform plan`:
```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Value Conversion Error
│ 
│   with data.mongodbatlas_stream_privatelink_endpoint.singular_datasource,
│   on main.tf line 10, in data "mongodbatlas_stream_privatelink_endpoint" "singular_datasource":
│   10: data "mongodbatlas_stream_privatelink_endpoint" "singular_datasource" {
│ 
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please report the following
│ to the provider developer:
│ 
│ Expected framework type from provider logic: types.ListType[basetypes.StringType] / underlying type: tftypes.List[tftypes.String]
│ Received framework type from provider logic: types.ListType[!!! MISSING TYPE !!!] / underlying type: tftypes.List[tftypes.DynamicPseudoType]
│ Path: dns_sub_domain
╵
╷
│ Error: Value Conversion Error
│ 
│   with data.mongodbatlas_stream_privatelink_endpoints.plural_datasource,
│   on main.tf line 15, in data "mongodbatlas_stream_privatelink_endpoints" "plural_datasource":
│   15: data "mongodbatlas_stream_privatelink_endpoints" "plural_datasource" {
│ 
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please report the following
│ to the provider developer:
│ 
│ Expected framework type from provider logic: types.ListType[basetypes.StringType] / underlying type: tftypes.List[tftypes.String]
│ Received framework type from provider logic: types.ListType[!!! MISSING TYPE !!!] / underlying type: tftypes.List[tftypes.DynamicPseudoType]
│ Path: results[0].dns_sub_domain
╵
```

FWIW kind of taking @EspenAlbert's suggestion and modifying the existing code here to not be in a conditional, also solves the problem: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3139/files#diff-36d96f42b6febe3b4ed58e3c70bdea579d63292e1d9eb569988494c943402b45R35

Let me know if you want me to make that change rather than initializing the list to null.

tcannon91: I have added some acceptance tests around this, but I believe they will fail until I move this over to a branch within the repo rather than my forked repo here. Once we agree on a direction I will change the destination to a branch within the the mongodb repo and we can validate/iterate on final steps there

tcannon91:
Created this PR within the mongodb repo rather than my fork: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3161

internal/service/streamprivatelinkendpoint/model.go
EspenAlbert: > Let me know if you want me to make that change rather than initializing the list to null.

I would vote for this change. @oarbusi, do you have any concerns? I think setting it explicitly to null must have been a workaround or bug when waiting for the API to return the field.

oarbusi: LGTM, agreed on @EspenAlbert suggestion.

tcannon91: Sweet. I am going to resolve this thread as that suggestion is committed. We can take any further issues to the PR with a branch in the mongodb repo. https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3161

tcannon91:
Closing in favor of: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3161

PR #3140, opened 2025-03-07
chore: Updates logic in TPF to behave similarly to SDKv2 when blocks are deleted from the config

lantoli:
## Description

Updates logic in TPF to behave similarly to SDKv2 following O/C semantics when blocks are deleted from the config.

- If `analytics_specs`, `auto_scaling` or `analytics_auto_scaling` blocks were in the config and are deleted, value from the state will be used so it behaves like SDKv2. E.g. if `auto_scaling` was enabled for compute and disk, it'll continue enabled.
  - However this won't happen with child attributes. For example, if `disk_gb_enabled` in `auto_scaling` is `true` in the config and the attribute is deleted (but `auto_scaling` block is kept), then  `disk_gb_enabled` will be considered unknown so it won't be sent to Atlas. That will set the value to false, so disk auto-scaling will be disabled.
- If `read_only_specs` was in the config and is deleted, `node_count` will be got from the state. However the other attributes will be got from `electable_specs` to avoid inconsistency errors, e.g. if `instance_size` is changed in `electable_specs`, we want `read_only_specs` to use the same value.

Link to any related issue(s): CLOUDP-305352

## Type of change:

## Required Checklist:


internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -114,6 +190,8 @@ func determineKeepUnknownsUnchangedReplicationSpecs(ctx context.Context, diags *
 	return keepUnknowns
 }
 
+// TODO: check if it's needed or it's ok if just included in ignored_changes in lifeclcle
+// TODO: add disk_iops doc to include it in ignored_changes when instance_size changes
 func determineKeepUnknownsAutoScaling(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) []string {
```
AgustinBettati: Just to confirm, currently this function is not being used, this would be a consideration if we reincoporate logic for marking instance_size as a known value right?

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -45,6 +45,15 @@ func keepUnkownFuncWithNonEmptyAutoScaling(name string, replacement attr.Value)
 
 // useStateForUnknowns should be called only in Update, because of findClusterDiff
 func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
+	shardingConfigUpgrade := isShardingConfigUpgrade(ctx, state, plan, diags)
+	if diags.HasError() {
+		return
+	}
+	// Don't adjust region_configs upgrades if it's a sharding config upgrade because it will be done only in the first shard, because state only has the first shard with num_shards > 1.
+	// This avoid errors like AUTO_SCALINGS_MUST_BE_IN_EVERY_REGION_CONFIG.
+	if !shardingConfigUpgrade {
```
AgustinBettati, line 54: I understand the current implementation supports adding a new replication spec, so I am curious why that case passes but the sharding config upgrade scenario causes failures

internal/service/advancedclustertpf/plan_modifier.go
lantoli: correct, not being used, and we'll need to reincorporate

internal/service/advancedclustertpf/plan_modifier.go
lantoli: will discuss offline

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
...
+	for i := range minLen(planRepSpecsTF, stateRepSpecsTF) {
+		stateRegionConfigsTF := TFModelList[TFRegionConfigsModel](ctx, diags, stateRepSpecsTF[i].RegionConfigs)
+		planRegionConfigsTF := TFModelList[TFRegionConfigsModel](ctx, diags, planRepSpecsTF[i].RegionConfigs)
+		if diags.HasError() {
+			return
+		}
+		for j := range minLen(planRegionConfigsTF, stateRegionConfigsTF) {
+			stateReadOnlySpecs := TFModelObject[TFSpecsModel](ctx, stateRegionConfigsTF[j].ReadOnlySpecs)
+			planReadOnlySpecs := TFModelObject[TFSpecsModel](ctx, planRegionConfigsTF[j].ReadOnlySpecs)
+			planElectableSpecs := TFModelObject[TFSpecsModel](ctx, planRegionConfigsTF[j].ElectableSpecs)
```
marcosuma, line 120: q: how is it able to make sure `stateReadOnlySpecs` and `planElectableSpecs` correspond to the same block? eg in the plan a whole region_config was removed or added.

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
...
+				newPlanReadOnlySpecs := planReadOnlySpecs
+				if newPlanReadOnlySpecs == nil {
+					newPlanReadOnlySpecs = new(TFSpecsModel) // start with null attributes if not present plan
+				}
+				// unknown node_count is got from state, all other unknowns are got from electable_specs plan
+				copyAttrIfDestNotKnown(&planElectableSpecs.DiskSizeGb, &newPlanReadOnlySpecs.DiskSizeGb)
+				copyAttrIfDestNotKnown(&planElectableSpecs.EbsVolumeType, &newPlanReadOnlySpecs.EbsVolumeType)
+				copyAttrIfDestNotKnown(&planElectableSpecs.InstanceSize, &newPlanReadOnlySpecs.InstanceSize)
+				copyAttrIfDestNotKnown(&planElectableSpecs.DiskIops, &newPlanReadOnlySpecs.DiskIops)
+				copyAttrIfDestNotKnown(&stateReadOnlySpecs.NodeCount, &newPlanReadOnlySpecs.NodeCount)
```
marcosuma, line 131: for later: we'll need to find a way in the future to remember to update this if a new field in `_specs` is added. Perhaps some test assert that checks the # of fields matches the expected one (in this case it's 5)

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
...
+		return
+	}
+	for i := range minLen(planRepSpecsTF, stateRepSpecsTF) {
+		stateRegionConfigsTF := TFModelList[TFRegionConfigsModel](ctx, diags, stateRepSpecsTF[i].RegionConfigs)
+		planRegionConfigsTF := TFModelList[TFRegionConfigsModel](ctx, diags, planRepSpecsTF[i].RegionConfigs)
+		if diags.HasError() {
+			return
+		}
+		for j := range minLen(planRegionConfigsTF, stateRegionConfigsTF) {
+			stateReadOnlySpecs := TFModelObject[TFSpecsModel](ctx, stateRegionConfigsTF[j].ReadOnlySpecs)
```
marcosuma, line 118: so this logic could be described as:

if there is a `read_only_spec` in the state I use the `electable_specs` to "fill" the missing fields in the plan of `read_only_specs` with `copyAttrIfDestNotKnown` which also avoids copying if there's a value already (eg the user is actually updating read_only)

We use electable_specs because is the only spec that is surely going to be there

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
...
+				diags.Append(diagsLocal...)
+				if diags.HasError() {
+					return
+				}
+				planRegionConfigsTF[j].ReadOnlySpecs = objType
+			}
+			stateAnalyticsSpecs := TFModelObject[TFSpecsModel](ctx, stateRegionConfigsTF[j].AnalyticsSpecs)
+			planAnalyticsSpecs := TFModelObject[TFSpecsModel](ctx, planRegionConfigsTF[j].AnalyticsSpecs)
+			// don't get analytics_specs from state if node_count is 0 to avoid possible ANALYTICS_INSTANCE_SIZE_MUST_MATCH errors
+			if planAnalyticsSpecs == nil && stateAnalyticsSpecs != nil && stateAnalyticsSpecs.NodeCount.ValueInt64() > 0 {
```
marcosuma, line 143: if node_count is > 0, why aren't we doing the same thing of copying unknown values from electable as we did for read_only_specs ?

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
 	plan.ReplicationSpecs = listType
 }
 
+func AdjustRegionConfigsChildren(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel) {
```
AgustinBettati, line 105: We will need a comment here for clarification of what is then end goal

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
...
+		stateRegionConfigsTF := TFModelList[TFRegionConfigsModel](ctx, diags, stateRepSpecsTF[i].RegionConfigs)
+		planRegionConfigsTF := TFModelList[TFRegionConfigsModel](ctx, diags, planRepSpecsTF[i].RegionConfigs)
+		if diags.HasError() {
+			return
+		}
+		for j := range minLen(planRegionConfigsTF, stateRegionConfigsTF) {
+			stateReadOnlySpecs := TFModelObject[TFSpecsModel](ctx, stateRegionConfigsTF[j].ReadOnlySpecs)
+			planReadOnlySpecs := TFModelObject[TFSpecsModel](ctx, planRegionConfigsTF[j].ReadOnlySpecs)
+			planElectableSpecs := TFModelObject[TFSpecsModel](ctx, planRegionConfigsTF[j].ElectableSpecs)
+			if stateReadOnlySpecs != nil && planElectableSpecs != nil { // read_only_specs is present in state and electable_specs in the plan
```
AgustinBettati, line 121: We would only need to do this logic if planReadOnlySpecs is unknown right? Might be redudant but helps understand the specific scenario

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -91,6 +100,73 @@ func UseStateForUnknownsReplicationSpecs(ctx context.Context, diags *diag.Diagno
...
+				if newPlanReadOnlySpecs == nil {
+					newPlanReadOnlySpecs = new(TFSpecsModel) // start with null attributes if not present plan
+				}
+				// unknown node_count is got from state, all other unknowns are got from electable_specs plan
+				copyAttrIfDestNotKnown(&planElectableSpecs.DiskSizeGb, &newPlanReadOnlySpecs.DiskSizeGb)
+				copyAttrIfDestNotKnown(&planElectableSpecs.EbsVolumeType, &newPlanReadOnlySpecs.EbsVolumeType)
+				copyAttrIfDestNotKnown(&planElectableSpecs.InstanceSize, &newPlanReadOnlySpecs.InstanceSize)
+				copyAttrIfDestNotKnown(&planElectableSpecs.DiskIops, &newPlanReadOnlySpecs.DiskIops)
+				copyAttrIfDestNotKnown(&stateReadOnlySpecs.NodeCount, &newPlanReadOnlySpecs.NodeCount)
+				objType, diagsLocal := types.ObjectValueFrom(ctx, SpecsObjType.AttrTypes, newPlanReadOnlySpecs)
```
AgustinBettati, line 132: for preserving readability in main function we can extract this to separate function and clarify why we have this specific logic for read_only_specs and not analytics_specs

internal/service/advancedclustertpf/plan_modifier.go
lantoli: this logic improves over that in the case where the read_only_specs are kept, but some attribute is removed like instance_size,  in that case we continue getting from electable_specs so it doesn't work in Atlas

internal/service/advancedclustertpf/plan_modifier.go
lantoli: it's by position like in SDKv2 so it will have the same issues as in the current implementation. 

if a region_configs is added, that's ok because it didn't exist before and no need to see if blocks existed in config or not before.

if region_configs is removed, we might compare one region_config in plan with a different one in the state, so in that case there might be some error from Atlas if read_only_specs was in the config, and was deleted from the config, and then removed a region_configs, and the had different attribute values, e.g. a region_config with M10 and another one with M20

internal/service/advancedclustertpf/plan_modifier.go
lantoli: it's really if it's in the state but not in the plan:
```
newPlanReadOnlySpecs := planReadOnlySpecs
```

if it's in the plan, we use the plan, and then try go get values from electable_specs if some is unknown (imagine e.g. instance_size is deleted from read_only_specs block but block itself is kept).

if it's in the state and not in the plan, we get it from the electable_specs except node_count

internal/service/advancedclustertpf/plan_modifier.go
lantoli: because read_only_specs must match electable_specs attributes (except node_count). For example, if you change instance_size in electable_specs from M10 to M20, instance size changes also in read_only_specs.

However analytics_specs are independent from electable_specs, they can have different attributes like instance_size.

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -158,18 +236,36 @@ func autoScalingUsed(ctx context.Context, diags *diag.Diagnostics, state, plan *
...
 	return item
 }
+
+func copyAttrIfDestNotKnown[T attr.Value](src, dest *T) {
+	if !isKnown(*dest) {
+		*dest = *src
+	}
+}
+
+func isKnown(attribute attr.Value) bool {
```
lantoli, line 271: nit: note that  !isKnown != IsUnknown,   null would be !isKnown but not IsUnknown

internal/service/advancedclustertpf/plan_modifier.go
marcosuma: tl;dr: I think we are rushing this decision which might compromise the future this resource - While this logic looks clear to me, I am thinking to keep the existing behaviour and add warnings to our documentation on **remove blocks only after you've set values to zero**. From there, we can release improvements in the future after talking with Hashicorp. On the other side, if we release this now we are somehow forcing ourselves in never letting a block removal to trigger a deletion unless we release a big breaking change.

Longer thoughts:

From your comment, this logic has no guarantee to be exactly matching state and plan when a region config or a whole replication_spec is removed (I don't follow "the why" you said about adding, but I'll assume that) and it feels like we can not find a way unless we build the whole plan ourselves (and it gets wildly complex and discouraged). 
So we can say this logic works if the only removed block is an inner block of region_configs. If region_config or replication_specs are also removed at the same time, this logic doesn't work.

> it's by position like in SDKv2 so it will have the same issues as in the current implementation.

As I mentioned in the meeting last Thursday, I am unsure about this principle of following what the SDKv2 implementation does, especially if we know it has issues.

1. There is no guarantee our existing resource is implemented the right way and not a lot of customers are using it yet.
2. They are built on different frameworks and TPF is the most recent one
3. If I read your comment correctly we are saying that this won't work for some cases (maybe rare?) and there will be plan inconsistencies. But since this is our future resource, what happens when in future it might be possible to have independent values for individual auto-scaling or specs?
4. [for completeness] With this change we are basically agreeing on a sticky-state behaviour. In reality it's already the API that is doing this for us by always returning these blocks, we are "adapting" to it.

lantoli: @marcosuma this feature only happens when the block was in the config and it's not anymore. I agree that it's not perfect but I think it's better than the current behavior.  Currently a block removal won't trigger a plan change, but when there are some change elsewhere, the read_only_specs / auto-scaling blocks that are removed will be deleted / disabled.

With that in mind, I think it's better to push this PR, and improve it later if we think of a better way to do it. 

>  I am thinking to keep the existing behaviour and add warnings to our documentation on remove blocks only after you've set values to zero.

I'll also add in the doc that it's recommended not to remove blocks from the config. But some users won't see this doc so I think it's better to err on the cautious side if they remove the block.

cc @AgustinBettati @EspenAlbert if you want to add something else

marcosuma: > Currently a block removal won't trigger a plan change, but when there are some change elsewhere, the read_only_specs / auto-scaling blocks that are removed will be deleted / disabled.

I get that and I was the first one to say this is risky. But what is the point in pushing a code change that we know it's not perfect? In addition, not only it's not perfect but it's making a design decision that is permanent in the future. 
This is the scenario that made me change my mind: if we push this, and Terraform tells us that the recommendation we should follow is that a block removal should indeed trigger a removal, how are we going to handle that since we've now "educated" our customers that a block removal does nothing? It'll be a tougher change because it's a big breaking change in the behaviour. 
On the other side, if we don't do this now, we can call it out that it is discouraged to remove blocks and why (and we have a good "why" related to how the API behaves) and come back with a solid improvement in the future without having compromised anything.

marcosuma: **Update**: we synced and decided to still go ahead with this option. We will ignore block removals from `region_configs` and we will still update that deleting blocks is discouraged (with a why related to how the API works).
Balancing pros/cons, we think it's safer to release this code change to protect customers from undesired behaviour. As a note: we'll likely rework this decision in future after we meet with Hashicorp. In fact, the cons of going ahead with this one is that future changes will likely need to be interpreted as breaking changes.

All of this challenges are happening for mostly two reasons:
1. cluster PATCH API behaves like a PUT for the `replication_specs` block
2. cluster API does not follow IPA-111, meaning that a lot of fields need to be treated as Optional+Computed

While we could expose our resource and customers to this behaviour, we opt to release this change as protection.

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -45,14 +45,23 @@ func keepUnkownFuncWithNonEmptyAutoScaling(name string, replacement attr.Value)
...
+	}
 	diff := findClusterDiff(ctx, state, plan, diags)
 	if diags.HasError() || diff.isAnyUpgrade() { // Don't do anything in upgrades
 		return
 	}
 	attributeChanges := schemafunc.NewAttributeChanges(ctx, state, plan)
 	keepUnknown := []string{"connection_strings", "state_name"} // Volatile attributes, should not be copied from state
 	keepUnknown = append(keepUnknown, attributeChanges.KeepUnknown(attributeRootChangeMapping)...)
-	// pending revision if logic can be reincorporated safely: keepUnknown = append(keepUnknown, determineKeepUnknownsAutoScaling(ctx, diags, state, plan)...)
+	keepUnknown = append(keepUnknown, determineKeepUnknownsAutoScaling(ctx, diags, state, plan)...)
```
lantoli, line 64: we need to keep the auto-scaling logic at root level to keep unknowns in auto-scale related fields if there are changes in the cluster but no changes in rep_specs (e.g. tags), so we keep unknowns and don't set known values in instance_size, etc.

internal/service/advancedclustertpf/plan_modifier.go
marcosuma: @lantoli let's keep a CLOUDP for this (non-blocking)

internal/service/advancedclustertpf/plan_modifier.go
marcosuma: can you document in the code? better than a PR comment

internal/service/advancedclustertpf/plan_modifier.go
AgustinBettati: added a comment here

marcosuma (commented):
LGTM overall- just left two minor comments that are non-blocking but it'd be great to address (either now or as follow-up)

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -1419,6 +1474,113 @@ func configSharded(t *testing.T, projectID, clusterName string, withUpdate bool)
...
+						node_count      = 3
+					}
+					%[5]s
+				}
+			}
+		}
+	`, projectID, clusterName, instanceSize1, extraConfig0, extraConfig1))
+}
+
+func checkBlocks(projectID, clusterName string, firstStep bool) resource.TestCheckFunc {
```
EspenAlbert, line 1553: [nit] unused params projectID & clusterName

AgustinBettati (approved):
LGTM - All tests have passed with execution in https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/13790491504

internal/service/advancedclustertpf/plan_modifier.go
lantoli: CLOUDP-305915

internal/service/advancedclustertpf/plan_modifier.go
lantoli: thanks, added here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3154

internal/service/advancedcluster/resource_advanced_cluster_test.go
lantoli: @EspenAlbert thanks, changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3154

PR #3150, opened 2025-03-11
feat: Supports new Https connection type in streams processing

jwongmongodb:
## Description
We added a new connection type Https in streams processing. The configuration of the connection needs the `url` field and takes in an optional `headers` field.

Link to any related issue(s):
[Ticket](https://jira.mongodb.org/browse/CLOUDP-301845)

NOTE: this PR is the same as the previous [PR](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3138) in my fork to pass the acceptance test.

## Type of change:

## Required Checklist:


marcosuma (approved):

marcosuma:
@jwongmongodb today we'll release 1.29 ([ref](https://mongodb.slack.com/archives/C04K3F1EGGK/p1741690177262579)). This change won't go in there but we're planning to release next week as well. For reference, this is the [changelog file](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/CHANGELOG.md) to see what it'll be released next

EspenAlbert (approved):
Great job!. LGTM.

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -191,6 +191,31 @@ func TestAccStreamRSStreamConnection_sample(t *testing.T) {
...
+		projectID    = acc.ProjectIDExecution(t)
+		instanceName = acc.RandomName()
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             CheckDestroyStreamConnection,
+		Steps: []resource.TestStep{
+			{
+				Config: httpsStreamConnectionConfig(projectID, instanceName),
```
EspenAlbert: [nit] You might want to consider adding a step with different headers/url to verify Update works as expected

lmkerbey-mdb (approved):
LGTM.

internal/service/streamconnection/resource_stream_connection_test.go
jwongmongodb: So I went a step further and tested the case where I blank out the headers as a third step. This is where it got tricky and I needed more changes. 

Here's the high level summary:
In the update flow, it calls the `PATCH` endpoint for the streams connection. If there is an existing header and the config is missing the header field, the connection will not remove the header field (due to how PATCH REST endpoints behave). 

To circumvent this, I made changes to create [an update specific request](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3150/files#diff-d12ae64c4fb4930bcb895b1084c385c650bc43b60cb9dbc1020e9794a3975292R108-R114) that sends an empty header map to the `PATCH` endpoint which will correctly blank out any existing headers via the API. [A further change](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3150/files#diff-d12ae64c4fb4930bcb895b1084c385c650bc43b60cb9dbc1020e9794a3975292R200-R209) is needed to decode the API response and ignore empty header maps and treat is as null (this is needed to resolve the terraform state of applying a config without the header field and receiving an empty header map from the API).

NOTE: one caveat is that if a user explicitly puts an empty header map in the config and runs `terraform apply`, it will error out due to the provider treating an empty header map as null so there will be a mismatch.

cc @marcosuma @oarbusi let me know if this limitation in my `NOTE` is okay.

docs/data-sources/stream_connection.md
```diff
@@ -20,7 +20,7 @@ data "mongodbatlas_stream_connection" "example" {
 
 ## Attributes Reference
 
-* `type` - Type of connection. Can be `Cluster`, `Kafka`, `Sample`, or `AWSLambda`.
+* `type` - Type of connection. Can be `Cluster`, `Kafka`, `Sample`, `Https` or `AWSLambda`.
```
lantoli: nit: should we use alphabetical order or other criteria for the list?

docs/resources/stream_connection.md
```diff
@@ -81,12 +81,28 @@ resource "mongodbatlas_stream_connection" "test" {
...
+```terraform
+resource "mongodbatlas_stream_connection" "example-https" {
+  project_id      = var.project_id
+  instance_name   = mongodbatlas_stream_instance.example.instance_name
+  connection_name = "https_connection_tf_new"
+  type            = "Https"
+  url             = "https://example.com"
+  headers = {
+    "key1" : "value1",
+    "key2" : "value2"
```
lantoli: in a map, keys don't require quotes if content doesn't have blanks.  In case it can help, in this [doc](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/advanced_cluster%2520%2528preview%2520provider%25202.0.0%2529) we show some different ways to have keys:
```

tags = {
  env           = "dev"         # key strings without blanks can be enclosed in quotes but not required
  "tag 2"       = "val"         # enclose key strings with blanks in quotes
  (var.tag_key) = "another_val" # enclose key expressions in brackets so they can be evaluated
}
```

lantoli: also I note you use : (semicolon) and , "commas", where we use = (equal) for a similar attribute (tags & labels in advanced_cluster), I think both syntaxes are correct but maybe we should agree on one way in the examples to be more consistent

internal/service/streamconnection/model_stream_connection.go
```diff
@@ -87,9 +88,34 @@ func NewStreamConnectionReq(ctx context.Context, plan *TFStreamConnectionModel)
 		}
 	}
 
+	if !plan.Headers.IsNull() {
+		headersMap := &map[string]string{}
```
lantoli: nit: it's a bit more idiomatic to create the map without pointer:
```
headersMap := make(map[string]string)
```

and then use &headersMap in functions where pointer is needed

docs/data-sources/stream_connection.md
lantoli: also nit: i think there is small inconsistency in case, we say AWSLamba (not AwsLambda) but Https, shouldn't it be HTTPS?

internal/service/streamconnection/data_source_stream_connection_test.go
```diff
@@ -85,6 +85,35 @@ func TestAccStreamDSStreamConnection_sample(t *testing.T) {
 	})
 }
 
+func TestAccStreamDSStreamConnection_https(t *testing.T) {
```
lantoli: I see this resource has still resource and datasource tests splited. We made a team decision to join resource and data sources tests as we normally need to setup the same resources so we save some resource creation.  no need to do in this PR but I think it would be go to do it in a follow-up PR

docs/resources/stream_connection.md
```diff
@@ -102,6 +118,10 @@ If `type` is of value `Kafka` the following additional arguments are defined:
 If `type` is of value `AWSLambda` the following additional arguments are defined:
 * `aws` - The configuration for AWS Lambda connection. See [AWS](#AWS)
 
+If `type` is of value `Https` the following additional attributes are defined:
+* `url` - URL of the HTTPs endpoint that will be used for creating a connection.
+* `headers` - A map of key-value pairs for optional headers.
+
 ### Authentication
 
 * `mechanism` - Style of authentication. Can be one of `PLAIN`, `SCRAM-256`, or `SCRAM-512`.
```
lantoli, line 127: nit: similar case inconsistency, here we use all uppercase when in other attributes we use capital case, I guess it's because Atlas API

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -326,6 +381,18 @@ func sampleStreamConnectionAttributeChecks(
...
+func httpsStreamConnectionAttributeChecks(resourceName, instanceName, url string) resource.TestCheckFunc {
+	resourceChecks := []resource.TestCheckFunc{
+		checkStreamConnectionExists(),
+		resource.TestCheckResourceAttrSet(resourceName, "project_id"),
+		resource.TestCheckResourceAttr(resourceName, "instance_name", instanceName),
+		resource.TestCheckResourceAttr(resourceName, "connection_name", "ConnectionNameHttps"),
+		resource.TestCheckResourceAttr(resourceName, "type", "Https"),
+		resource.TestCheckResourceAttr(resourceName, "url", url),
+	}
+	return resource.ComposeAggregateTestCheckFunc(resourceChecks...)
```
lantoli: this works but consider using helper acc functions, e.g.:
```

func httpsStreamConnectionAttributeChecks(resourceName, instanceName, url string) resource.TestCheckFunc {
	checks := []resource.TestCheckFunc{checkStreamConnectionExists()}
	checks = acc.AddAttrSetChecks(resourceName, checks, "project_id")
	checks = acc.AddAttrChecks(resourceName, checks, map[string]string{
		"instance_name":   instanceName,
		"connection_name": "ConnectionNameHttps",
		"type":            "Https",
		"url":             url,
	})
	return resource.ComposeAggregateTestCheckFunc(checks...)
}
```

or if resource and data sources tests are joined at some point: 
```

func httpsStreamConnectionAttributeChecks(resourceName, instanceName, url string) resource.TestCheckFunc {
	setChecks := []string{"project_id"}
	mapChecks := map[string]string{
		"instance_name":   instanceName,
		"connection_name": "ConnectionNameHttps",
		"type":            "Https",
		"url":             url,
	}
	extra := []resource.TestCheckFunc{checkStreamConnectionExists()}
	return acc.CheckRSAndDS(resourceName, admin.PtrString(dataSourceName), nil, setChecks, mapChecks, extra...)
}
```

this last example can also be used only for resource passing nil instead of `admin.PtrString(dataSourceName)`

lantoli (approved):

docs/data-sources/stream_connection.md
jwongmongodb: the type for `Https` matches the API's response for `Https`. If I change the config to be `HTTPS`, there would be a mismatch when we do a READ on the API to get the response.

docs/resources/stream_connection.md
jwongmongodb: I can change it to be consistent

docs/data-sources/stream_connection.md
lantoli: that's fine, thanks

internal/service/streamconnection/resource_stream_connection_test.go
jwongmongodb: this suggestion was good, felt more clean to do the second change.

internal/service/streamconnection/data_source_stream_connection_test.go
jwongmongodb: done for this test case, I've merged them for https connection

docs/resources/stream_connection.md
```diff
@@ -81,12 +81,28 @@ resource "mongodbatlas_stream_connection" "test" {
...
+
+```terraform
+resource "mongodbatlas_stream_connection" "example-https" {
+  project_id      = var.project_id
+  instance_name   = mongodbatlas_stream_instance.example.instance_name
+  connection_name = "https_connection_tf_new"
+  type            = "Https"
+  url             = "https://example.com"
+  headers = {
+    key1 = "value1",
```
lantoli:
```suggestion
    key1 = "value1"
```

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -381,6 +443,34 @@ func clusterStreamConnectionConfig(projectID, instanceName, clusterName string)
 	`, projectID, instanceName, clusterName)
 }
 
+func httpsStreamConnectionConfig(projectID, instanceName, url, headers string) string {
```
lantoli: lowercase function are not exported so the scope is private to their package, e.g. streamconnection_test, so it's not needed to fully-qualify the name, e.g. `configHttps` , etc. can be fine, same for other unexported funcs

internal/service/streamconnection/data_source_stream_connection_test.go
lantoli: thanks for all the follow-ups

lantoli (approved):

oarbusi (approved):

PR #3180, opened 2025-03-18
feat: Adds Stream Processor update support

jwongmongodb:
## Description

This PR adds support to modify a stream processor by calling the existing ModifyStreamProcessor [endpoint](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/modifyStreamProcessor).

I also added a suite of acceptance tests to cover all the possible state transitions. The main behavior is that a stream processor must be in the `CREATED` or `STOPPED` state before it can call the ModifyStreamProcessor endpoint. If the stream processor is currently in a `STARTED` state, it must be stopped before it can be modified.

Link to any related issue(s):
[CLOUDP-301766
](https://jira.mongodb.org/browse/CLOUDP-301766)

## Type of change:

## Required Checklist:


internal/service/streamprocessor/resource_test.go
```diff
@@ -118,27 +120,145 @@ func TestAccStreamProcessor_withOptions(t *testing.T) {
 		}})
 }
 
-func TestAccStreamProcessor_createWithAutoStartAndStop(t *testing.T) {
-	var (
-		projectID     = acc.ProjectIDExecution(t)
-		processorName = "new-processor"
-		instanceName  = acc.RandomName()
-	)
+func TestAccStreamProcessor_StateTransitionsUpdates(t *testing.T) {
```
jwongmongodb, line 121: I added a whole new test suite since the current [config func](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3180/files#diff-8c4ca6bda807d9d100a1928589ba9a173c47e26c0949f41594c8bef052430540R568) would be messy to modify for testing all the updates without creating external resources (clusters or kafka). The update that I am check is the pipeline change with tumbling window since this requires no external resources to spin for each of my test case (there are many combinations)

I have three separate test suits for update tests, I remove some of the existing tests here since my new test suits should cover those flows.

The three update suites cover the following:
- Valid state to state updates 
- Valid state to empty state updates, it should use the previous state (CREATE state will have CREATE state after the update with empty state).
- Invalid state updates with expected errors.

internal/service/streamprocessor/resource_test.go
```diff
@@ -261,6 +352,188 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+			instance_name  = mongodbatlas_stream_instance.instance.instance_name
+			processor_name = %[3]q
+			pipeline       = %[4]q
+			%[5]s
+			depends_on = [mongodbatlas_stream_connection.sample]		
+		}
+		`, projectID, instanceName, processorName, pipeline, state)
+}
+
+func checkConfigAttribute(projectID, instanceName, processorName, state, expectedPipelineStr string) resource.TestCheckFunc {
```
jwongmongodb: I can rename this `checkConfigAttribute` func to have the word update in it or I can merge with the existing `composeStreamProcessorChecks` (would complicate that function if I were to do that so I am leaning with a separate function)

internal/service/streamprocessor/state_transition.go
```diff
@@ -43,6 +43,22 @@ func WaitStateTransition(ctx context.Context, requestParams *admin.GetStreamProc
 	return nil, errors.New("did not obtain valid result when waiting for stream processor state transition")
 }
 
+func ValidateUpdateStateTransition(currentState, plannedState string) (errMsg string, isValidTransition bool) {
```
jwongmongodb, line 51: I believe the validation should live in the state transition? I also added unit tests for this function that may be overkill, I can see the argument to remove or keep so I defer to the team's opinion

jwongmongodb (commented):
Initial draft review to make sure that the logic makes sense before cleaning up.

internal/service/streamprocessor/resource.go
```diff
@@ -159,63 +158,91 @@ func (r *streamProcessorRS) Update(ctx context.Context, req resource.UpdateReque
...
+	}
+
+	// update the stream processor
+	streamProcessorReq, diags := NewStreamProcessorUpdateReq(ctx, &plan)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	modifyResp, _, err := r.Client.AtlasV2.StreamsApi.ModifyStreamProcessorWithParams(ctx, &admin.ModifyStreamProcessorApiParams{
```
oarbusi: Is the update completed once we get the API response, or we should wait like we do for starting/stopping?

internal/service/streamprocessor/model.go
```diff
@@ -42,6 +42,37 @@ func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel)
 	return streamProcessor, nil
 }
 
+func NewStreamProcessorUpdateReq(ctx context.Context, plan *TFStreamProcessorRSModel) (*admin.StreamsModifyStreamProcessor, diag.Diagnostics) {
```
oarbusi: This method could return `ModifyStreamProcessorApiParams` instead of `StreamsModifyStreamProcessor`. What do you think?

docs/resources/stream_processor.md
```diff
@@ -141,7 +128,7 @@ output "stream_processors_results" {
 ### Optional
 
 - `options` (Attributes) Optional configuration for the stream processor. (see [below for nested schema](#nestedatt--options))
-- `state` (String) The state of the stream processor. Commonly occurring states are 'CREATED', 'STARTED', 'STOPPED' and 'FAILED'. Used to start or stop the Stream Processor. Valid values are `CREATED`, `STARTED` or `STOPPED`. When a Stream Processor is created without specifying the state, it will default to `CREATED` state.
+- `state` (String) The state of the stream processor. Commonly occurring states are 'CREATED', 'STARTED', 'STOPPED' and 'FAILED'. Used to start or stop the Stream Processor. Valid values are `CREATED`, `STARTED` or `STOPPED`. When a Stream Processor is created without specifying the state, it will default to `CREATED` state. When a Stream Processor is updated without specifying the state, it will default to the Previous state.
```
oarbusi, line 136: I think this detail might not be necessary. The expectation in Terraform is that a non modified attribute will not be changed. Maybe we should mention in a more general level the fact that the stream will be stopped. Maybe something similar to:
```
**NOTE**: When updating an Atlas Stream Processor, the following behavior applies:
1. If the processor is in a `STARTED` state, it will automatically be stopped before the update is applied
2. The update will be performed while the processor is in `STOPPED` state
3. If the processor was originally in `STARTED` state, it will be restarted after the update
```
docs team can confirm when they review

internal/service/streamprocessor/resource_test.go
oarbusi: really nice abstraction for the tests!

internal/service/streamprocessor/state_transition.go
oarbusi: Makes sense to put it here. I like the unit tests for this, it's a nice way to "document" all the possible state transitions

oarbusi (commented):
Really good job!

internal/service/streamprocessor/resource.go
jwongmongodb: I synced with @mongodb-matthew-normyle and as long as we get a successful response from the modify endpoint, we are good to proceed.

internal/service/streamprocessor/model.go
jwongmongodb: This is a good suggestion, it cleans up the Modify API call in the update function. I will make this change.

docs/resources/stream_processor.md
jwongmongodb: what about something like this? 

```
***NOTE***: The following behavior applies to an Atlas Stream Processor that is in a STARTED state:
1. The processor will automatically be stopped before the update is applied.
2. The update will be performed while the processor is in a `STOPPED` state.
3. The processor will be restarted after the update if the config state field is empty or set to `STARTED`.
```

davidhou17 (approved):

internal/service/streamprocessor/resource_test.go
```diff
@@ -261,6 +352,189 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+	// pipeline attribute check
+	checks = append(checks, func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[resourceName]
+		if !ok {
+			return fmt.Errorf("not found: %s", resourceName)
+		}
+
+		actualPipelineStr := rs.Primary.Attributes["pipeline"]
+
+		// JSON string comparison requires normalization because:
```
jasonrydberg: 👍

jasonrydberg (approved):
lgtm

internal/service/streamprocessor/state_transition_test.go
```diff
@@ -153,3 +153,128 @@ func responseWithState(state *string) *admin.StreamsProcessorWithStats {
...
+			wantErrMsg:     "",
+			wantValidation: true,
+		},
+
+		// Invalid transitions
+		{
+			name:           "Invalid transition - CREATED to STOPPED",
+			currentState:   CreatedState,
+			plannedState:   StoppedState,
+			wantErrMsg:     "Stream Processor must be in STARTED state to transition to STOPPED state",
```
maastha: The error message can be a constant that can be used here and in the resource

docs/resources/stream_processor.md
maastha: I don't think this behavior is specific to TF right? Is this something that can be captured in the API spec instead so we can simply add a link here?

internal/service/streamprocessor/state_transition.go
```diff
@@ -43,6 +43,22 @@ func WaitStateTransition(ctx context.Context, requestParams *admin.GetStreamProc
...
+	if currentState == plannedState {
+		return "", true
+	}
+
+	if plannedState == StoppedState && currentState != StartedState {
+		return fmt.Sprintf("Stream Processor must be in %s state to transition to %s state", StartedState, StoppedState), false
+	}
+
+	if plannedState == CreatedState && currentState != CreatedState {
```
maastha, line 60: Is there any case where the user can manually set the state to "CREATED"? Looks like it would be something that would always be returned from the API, and the user can only update the state to Start/Stop, is that correct?

jwongmongodb: The user can also put `CREATED` as a state when they first create the resource, so leaving it in the config and changing other areas will trigger an update with the `CREATED` state still in place for the update. Leaving it blanks during the creation part also defaults to the `CREATED` state.

docs/resources/stream_processor.md
jwongmongodb: The TF behavior may calling multiple endpoints at once (STOP, GET and MODIFY endpoint). Due to the nature of the stream processor, it can only be modified in the `CREATED` or `STOPPED` state. I don't think we can have one specific link to a particular endpoint like MODIFY. Perhaps we can add a requirement that the stream processor must be STOPPED before we can modify it.

.changelog/3180.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+resource/mongodbatlas_stream_processor: Support modifying a stream processor
```
lantoli:
```suggestion
resource/mongodbatlas_stream_processor: Support modifying a stream processor
```
```suggestion
resource/mongodbatlas_stream_processor: Adds update support
```

lantoli (approved):

lantoli:
nit: you can use longer branch names as long as prefixed with ticket in case it helps you to know what the branch is about, e.g. CLOUDP-301766_modify_stream_processor

docs/resources/stream_processor.md
oarbusi: Currently, the client needs to stop, modify, then start again a processor that is running. The update/modify API does not do it. So in TF we are removing this complexity to have a better user experience

internal/service/streamprocessor/state_transition.go
```diff
@@ -43,6 +43,22 @@ func WaitStateTransition(ctx context.Context, requestParams *admin.GetStreamProc
...
+		return fmt.Sprintf("Stream Processor must be in %s state to transition to %s state", CreatedState, CreatedState), false
```
EspenAlbert: [nit] Is it a transition if CreatedState = CreatedState?
Maybe a better error message is: "Cannot transition from `currentState` to `CreatedState`"

internal/service/streamprocessor/resource.go
```diff
@@ -159,63 +158,86 @@ func (r *streamProcessorRS) Update(ctx context.Context, req resource.UpdateReque
...
-			&admin.StopStreamProcessorApiParams{
+	}
+
+	// modify the stream processor
+	modifyAPIRequestParams, diags := NewStreamProcessorUpdateReq(ctx, &plan)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
```
EspenAlbert: [nit] remove linebreak. Keep update lines together

internal/service/streamprocessor/resource.go
```diff
@@ -159,63 +158,86 @@ func (r *streamProcessorRS) Update(ctx context.Context, req resource.UpdateReque
...
+	modifyResp, _, err := r.Client.AtlasV2.StreamsApi.ModifyStreamProcessorWithParams(ctx, modifyAPIRequestParams).Execute()
```
EspenAlbert:
```suggestion
	streamProcessorResp, _, err := r.Client.AtlasV2.StreamsApi.ModifyStreamProcessorWithParams(ctx, modifyAPIRequestParams).Execute()
```

EspenAlbert: Remove L219 `streamProcessorResp = modifyResp`

internal/service/streamprocessor/resource_test.go
```diff
@@ -261,6 +352,189 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+		// if the empty state is passed in the target config, this should be the same initial state of the config
+		if targetState == "" {
+			targetState = initialState
+		}
+		// For valid transitions, check the result
+		finalStep.Check = checkAttributesFromBasicUpdateFlow(projectID, instanceName, processorName, targetState, updatedPipelineWithTumblingWindow)
+	}
+	steps = append(steps, finalStep)
+
+	resource.Test(t, resource.TestCase{
```
EspenAlbert: Why not `resource.ParallelTest`?

internal/service/streamprocessor/resource_test.go
```diff
@@ -261,6 +352,189 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+	checks = append(checks, func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[resourceName]
+		if !ok {
+			return fmt.Errorf("not found: %s", resourceName)
+		}
+
+		actualPipelineStr := rs.Primary.Attributes["pipeline"]
+
+		// JSON string comparison requires normalization because:
+		// 1. Whitespace differences don't affect JSON semantics but break string comparison
```
EspenAlbert: I don't think this is necessary. In the schema we use a custom type: `fwtypes.JSONStringType`.
As it is a required field, TPF would have raised an error if it didn't semantically equal the value in the config.

See more info [here](https://docs.google.com/document/d/1iu18eHeECR2ofy51uuYOyP8dxS2Q5m8joTnEQu4Q4lQ/edit?tab=t.0#bookmark=id.zdxt0nregpvs) and in [Hashicorp docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/data-consistency-errors#acceptance-testing-all-values). Relevant snippet:

> The typical recommendation for state value checking in acceptance testing is only for Computed: true attribute… Checking state values of configured values can be removed once the resource is migrated to terraform-plugin-framework, as Terraform itself will raise any potential error.

internal/service/streamprocessor/resource_test.go
jwongmongodb: I was having issues initially running this during development since I had a more complicated update case (I was updating with connection changes) and it kept returning a dangling reference on cleanup. So this is an artifact of that development process, I put t.Parallel() for each of the tests suites above to run it in parallel, I can update this helper to use `resource.ParallelTest`.

EspenAlbert: Yes, removing `t.Parallel()` and using `resource.ParallelTest` would be my recommendation

internal/service/streamprocessor/resource_test.go
jwongmongodb: ok, I can remove this and do a straight comparison

oarbusi (approved):

internal/service/streamprocessor/resource_test.go
EspenAlbert: Thank you.

internal/service/streamprocessor/resource_test.go
EspenAlbert: TY!

internal/service/streamprocessor/state_transition.go
```diff
@@ -21,6 +21,11 @@ const (
 	FailedState     = "FAILED"
 )
 
+const (
+	ErrorUpdateStateTransition = "Stream Processor must be in %s state to transition to %s state"
+	ErrorUpdateToCreatedState  = "Cannot transition from currentState to CreatedState"
```
EspenAlbert: Meant to use the variable `currentState` not the literal

internal/service/streamprocessor/state_transition.go
jwongmongodb: oops, that was embarrassing. Made the change now

internal/service/streamprocessor/state_transition.go
jwongmongodb: oops, that was embarrassing. Made the change now

docs/resources/stream_processor.md
maastha: got it, in that case, I like @oarbusi 's suggestion. Although I'd say we should add it as a separate note to call out what happens during an update instead of adding this note to `state`.
For the `state` description I think the current one makes sense, maybe rephrase a bit like "When a Stream Processor is updated without specifying the state, it is stopped & then restored to previous state upon update completion."

internal/service/streamprocessor/state_transition.go
maastha: Thank you for the clarification!

docs/resources/stream_processor.md
jwongmongodb: this is now changed with this [commit](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3180/commits/68b9e69e387a917e11ea4667c5f8c7d78b944bd7). Can you also take another look @davidhou17 to see if these documentation changes look good?

internal/service/streamprocessor/state_transition.go
EspenAlbert: No worries. Thank you 👍

EspenAlbert (approved):

docs/data-sources/stream_processor.md
```diff
@@ -2,6 +2,11 @@
 
 `mongodbatlas_stream_processor` describes a stream processor.
 
+**NOTE**: When updating an Atlas Stream Processor, the following behavior applies:
```
AgustinBettati: I believe the intention was to add this in resource docs right? Within the context of the data source docs this is not too relevant.

AgustinBettati (commented):
LGTM, minor comment on the note in docs

docs/data-sources/stream_processor.md
jwongmongodb: yes, this was a mistake. I've moved the note to the correct template and regenerated the docs.

hshiozawa:
👍

PR #3187, opened 2025-03-19
chore: Improves handling of network errors

oarbusi:
## Description

Added a new `RetryTransport` that implements `http.RoundTripper` with automatic retry logic for transient network failures. This wraps any existing HTTP transport.

The error handling strategy is:
1. Special handling for `net.OpError` (primary network error type)
2. General `net.Error` handling (timeouts, temporary errors)
3. Pattern matching for known retriable error messages

Link to any related issue(s): CLOUDP-262912

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/config/client.go
+++ b/internal/config/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -29,16 +28,8 @@ import (
 const (
 	toolName              = "terraform-provider-mongodbatlas"
 	terraformPlatformName = "Terraform"
-
-	// Network configuration constants
-	defaultTimeout               = 30 * time.Second
-	defaultKeepAlive             = 30 * time.Second
-	defaultMaxIdleConns          = 25
-	defaultMaxIdleConnsPerHost   = 25
-	defaultIdleConnTimeout       = 90 * time.Second
-	defaultExpectContinueTimeout = 1 * time.Second
-	defaultMaxRetries            = 3
-	defaultRetryDelay            = 1 * time.Second
+	defaultMaxRetries     = 3
+	defaultRetryDelay     = 1 * time.Second
 )
 
 // MongoDBClient contains the mongodbatlas clients and configurations
@@ -88,20 +79,6 @@ type PlatformVersion struct {
 func (c *Config) NewClient(ctx context.Context) (any, error) {
 	// setup a transport to handle digest
 	transport := digest.NewTransport(cast.ToString(c.PublicKey), cast.ToString(c.PrivateKey))
-
-	baseTransport := &http.Transport{
-		DialContext: (&net.Dialer{
-			Timeout:   defaultTimeout,
-			KeepAlive: defaultKeepAlive,
-		}).DialContext,
-		MaxIdleConns:          defaultMaxIdleConns,
-		MaxIdleConnsPerHost:   defaultMaxIdleConnsPerHost,
-		Proxy:                 http.ProxyFromEnvironment,
-		IdleConnTimeout:       defaultIdleConnTimeout,
-		ExpectContinueTimeout: defaultExpectContinueTimeout,
-	}
-
-	transport.Transport = baseTransport
 
 	// Wrap with retry transport
 	retryingTransport := NewRetryTransport(transport, defaultMaxRetries, defaultRetryDelay)
--- a/internal/config/client.go
+++ b/internal/config/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -28,6 +29,16 @@ import (
 const (
 	toolName              = "terraform-provider-mongodbatlas"
 	terraformPlatformName = "Terraform"
+
+	// Network configuration constants
+	defaultTimeout               = 30 * time.Second
+	defaultKeepAlive             = 30 * time.Second
+	defaultMaxIdleConns          = 25
+	defaultMaxIdleConnsPerHost   = 25
+	defaultIdleConnTimeout       = 90 * time.Second
+	defaultExpectContinueTimeout = 1 * time.Second
+	defaultMaxRetries            = 3
+	defaultRetryDelay            = 1 * time.Second
 )
 
 // MongoDBClient contains the mongodbatlas clients and configurations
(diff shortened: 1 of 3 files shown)


internal/config/transport.go
```diff
@@ -0,0 +1,142 @@
...
+type RetryTransport struct {
+	transport  http.RoundTripper
+	maxRetries int
+	retryDelay time.Duration
+}
+
+// retriableErrors contains error messages that indicate a request should be retried.
+// These are typically transient network issues that may resolve on retry.
+var retriableErrors = []string{
+	"connection reset by peer",
```
maastha, line 21: [q] curious what was the source of these?

internal/config/transport.go
```diff
@@ -0,0 +1,142 @@
...
+// 1. First attempt is made immediately
+// 2. On failure, if the error is retriable (see isRetriableError):
+//   - Waits for retryDelay * attempt (exponential backoff)
+//   - Retries up to maxRetries times
+//
+// 3. If the error is not retriable, returns immediately
+// 4. If context is cancelled during retry, returns immediately
+// 5. If all retries are exhausted, returns the last error
+//
+// The exponential backoff helps prevent overwhelming the server during issues
```
maastha, line 50: any rate limiting that may cause issues? Although I agree 3 retries is a good number

internal/config/transport.go
maastha: Looking at the related HELP ticket, would those errors have been captured by this list?

internal/config/client.go
```diff
@@ -78,13 +80,16 @@ func (c *Config) NewClient(ctx context.Context) (any, error) {
 	// setup a transport to handle digest
 	transport := digest.NewTransport(cast.ToString(c.PublicKey), cast.ToString(c.PrivateKey))
 
+	// Wrap with retry transport
+	retryingTransport := NewRetryTransport(transport, defaultMaxRetries, defaultRetryDelay)
```
maastha, line 84: Can we also update the Realm client?

internal/config/client.go
```diff
@@ -78,13 +80,16 @@ func (c *Config) NewClient(ctx context.Context) (any, error) {
...
+
 	// initialize the client
 	client, err := transport.Client()
 	if err != nil {
 		return nil, err
 	}
 
-	client.Transport = logging.NewTransport("MongoDB Atlas", transport)
+	client.Transport = logging.NewTransport("MongoDB Atlas", retryingTransport)
```
maastha, line 92: I noticed that `logging.NewTransport` is deprecated, can we try using replacement `NewSubsystemLoggingHTTPTransport` instead?

```
logging.NewTransport is deprecated: This will log the content of every http request/response at `[DEBUG]` level, without any filtering. Any sensitive information will appear as-is in your logs. Please use NewSubsystemLoggingHTTPTransport instead.
```

internal/config/transport.go
oarbusi: I looked for some of the most common networking errors in go ([example](https://github.com/golang/go/blob/c386ed107e65f495acb6aacb0efcaade27da6a15/src/syscall/tables_js.go#L327C2-L327C12), [example 2])(https://github.com/golang/go/blob/c386ed107e65f495acb6aacb0efcaade27da6a15/src/net/net.go#L625)) Although I might have missed some that we should add.

internal/config/transport.go
```diff
@@ -0,0 +1,142 @@
...
+
+// retriableErrors contains error messages that indicate a request should be retried.
+// These are typically transient network issues that may resolve on retry.
+var retriableErrors = []string{
+	"connection reset by peer",
+	"connection refused",
+	"no such host",
+	"i/o timeout",
+	"EOF",
+	"context deadline exceeded",
```
lantoli, line 26: if I don't recall incorrectly, @AgustinBettati mentioned that "context deadline exceeded" is because a misconfiguration in our side where timeouts are not defined,  we might not want to include it here.

internal/config/transport.go
```diff
@@ -0,0 +1,142 @@
...
+	transport  http.RoundTripper
+	maxRetries int
+	retryDelay time.Duration
+}
+
+// retriableErrors contains error messages that indicate a request should be retried.
+// These are typically transient network issues that may resolve on retry.
+var retriableErrors = []string{
+	"connection reset by peer",
+	"connection refused",
```
lantoli, line 22: "connection refused" and "no such host" don't sound like retryable

internal/config/transport.go
```diff
@@ -0,0 +1,142 @@
+package config
+
+import (
```
lantoli, line 3: this sounds like a big change that will affect all networking config. I think CLI had some similar issues and solved it with things like this one: https://github.com/mongodb/mongodb-atlas-cli/blob/b5e1d76e5c92f7f9dfede46c4179259b1ba5fd9d/internal/store/store.go#L62-L72

@gssbzn might have more context

internal/config/client.go
```diff
@@ -28,6 +28,8 @@ import (
 const (
 	toolName              = "terraform-provider-mongodbatlas"
 	terraformPlatformName = "Terraform"
+	defaultMaxRetries     = 3
```
lantoli, line 31: I see there is an example of retry logic in SDK: https://github.com/mongodb/atlas-sdk-go/blob/main/examples/retry/retry.go
should we use that so we don't need to implement in ourselves?

gssbzn (commented):
https://github.com/hashicorp/go-retryablehttp ?

internal/config/transport.go
gssbzn: has TF never tried to improve on the default http client connection settings? I think before going to custom retry logic you guys should first aim to optimize your client configurations, then use something off the shelf like https://github.com/hashicorp/go-retryablehttp, then as a last resource this custom logic

internal/config/transport.go
maastha: Thinking if 502 Bad Gateway, 503 Service Unavailable, 504 Gateway Timeout are some of the status codes we should handle as part of this

internal/config/transport.go
maastha: nit: the current backoff is linear I believe not exponential

internal/config/transport.go
oarbusi: this is solving a different problem. From @gssbzn PR, I see that "Default timeouts were too long for users using ops manager with no access to Atlas and just doing mongocli --version could take a long time". 
This change is to improve on how we handle network errors and the information that we would log in cases where network issues happen

lantoli: I agree with @gssbzn to try first to adjust the settings as was done in the CLI and see if that's enough

internal/config/transport.go
AgustinBettati: That error is encountered by terraform when the default operation timeout is hit, not sure in what scenario this could be encountered in the context of an http request.

internal/config/transport.go
lantoli: @oarbusi I think they were also having network errors

oarbusi: So if we only optimize the client settings and a customer has network issues again with that new config, it will be hard for us to detect that it is in fact a network error (like it happened in the past). The goal of this was to improve handling of network errors but more importantly, give visibility on those errors, and I think only changing the client configurations won't do. I will take a look at the https://github.com/hashicorp/go-retryablehttp , if that is enough I will close this PR and open a new one using that library

gssbzn: On the CLI changes we were seeing random `"connection reset by peer"` and this were specially common on ops manager when I dug on this the problem was always on go not the server because go was trying to reuse its pool of connections and we found some recommended better defaults for http clients to use and after changing them the reports of connection errors disappeared 

Regardless of this PR I'm still surprised TF has never changed the defaults, the defaults are not production ready and is the recomendation number 1 when setting clients

gssbzn: https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

gssbzn: https://www.slingacademy.com/article/customizing-go-http-clients-for-better-performance/

oarbusi: Thank you!

oarbusi:
Closing, will follow @gssbzn [recommendations](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3187#discussion_r2007790204)

PR #3192, opened 2025-03-21
chore: Uses NewUnknownReplacements meant to replace `schemafunc.CopyUnknowns` logic and `schemafunc.NewAttributeChanges`

EspenAlbert:
## Description

Adds NewUnknownReplacements meant to replace `schemafunc.CopyUnknowns` logic and `schemafunc.NewAttributeChanges`.

The new common logic replaces the need for reflect and instead uses Terraform's `req.State.Raw.Diff(req.Config.Raw)` to find changes.
It also comes with a simplified API where each individual field can be defined instead of having to go top-down.
Note: fields can be defined at any level, for example electable_specs or instance_size are both valid.
See #3218, #3223, #3236 for how the usage is for `advanced_cluster`

See also guide on adding a new attribute [here](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3236#discussion_r2023134794)

Link to any related issue(s): CLOUDP-307851

## Type of change:

## Required Checklist:


internal/service/advancedclustertpf/resource.go
```diff
@@ -108,12 +108,12 @@ func (r *rs) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, res
 	if diags.HasError() {
 		return
 	}
-
 	useStateForUnknowns(ctx, diags, &state, &plan)
 	if diags.HasError() {
 		return
 	}
 	diags.Append(resp.Plan.Set(ctx, plan)...)
+	unknownReplacements(ctx, &req, resp)
```
EspenAlbert: This is meant to replace the `useStateForUnknowns`, but starting slowly by only handling the `mongo_db_version` attribute

internal/common/conversion/path_helpers_test.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func TestIsAttributeValueOnly(t *testing.T) {
+	assert.True(t, conversion.IsAttributeValueOnly(path.Root("replication_specs").AtListIndex(0)))
+	assert.True(t, conversion.IsAttributeValueOnly(path.Root("replication_specs").AtMapKey("myKey")))
+	assert.True(t, conversion.IsAttributeValueOnly(path.Root("replication_specs").AtSetValue(types.StringValue("myKey"))))
+}
+
+func TestAttributeNameEquals(t *testing.T) {
+	assert.True(t, conversion.AttributeNameEquals(path.Root("replication_specs").AtListIndex(0), "replication_specs"))
```
lantoli: i like the idea of the new acc test check with a map, don't know if we could do something similar here, e.g.: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/ed85e27f1a96f69c59b71cee5059f2fb1bbf5b00/internal/service/advancedclustertpf/plan_modifier_test.go#L22

internal/common/conversion/path_converter.go
```diff
@@ -0,0 +1,150 @@
...
+	GetAttribute(context.Context, path.Path, any) diag.Diagnostics
+}
+
+// AttributePathValue retrieves the value for src (state/plan/config) @ attributePath with converted path.Path, schema is needed to get the correct types.XXX (String/Object/etc.)
+func AttributePathValue(ctx context.Context, diags *diag.Diagnostics, attributePath *tftypes.AttributePath, src TPFSrc, schema TPFSchema) (attr.Value, path.Path) {
+	convertedPath, localDiags := AttributePath(ctx, attributePath, schema)
+	if localDiags.HasError() {
+		diags.Append(localDiags...)
+		return nil, convertedPath
+	}
```
lantoli:
```suggestion
	convertedPath, localDiags := AttributePath(ctx, attributePath, schema)
        diags.Append(localDiags...)
	if diags.HasError() {
		return nil, convertedPath
	}
```

I think this is better, except if you want to ignore warnings on purpose, similar for other cases

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+)
+
+func LastPart(p path.Path) string {
+	parts := strings.Split(p.String(), ".")
+	return parts[len(parts)-1]
+}
+
+func IsAttributeValueOnly(p path.Path) bool {
```
lantoli: why is it called attribute value only?

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,85 @@
...
+		return p.ParentPath().String()
+	}
+	if IsMapIndex(p) {
+		return p.ParentPath().String()
+	}
+	if IsSetIndex(p) {
+		return p.ParentPath().String()
+	}
+	return p.String()
+}
```
lantoli:
```suggestion
func StripSquareBrackets(p path.Path) string {
	if IsAttributeValueOnly(p) {
		return p.ParentPath().String()
	}
	return p.String()
}
```

internal/common/conversion/unknown.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types@v1.13.0
+func AsUnknownValue(ctx context.Context, value attr.Value) attr.Value {
+	switch v := value.(type) {
```
lantoli: could this be a map?

internal/common/conversion/path_converter.go
```diff
@@ -0,0 +1,150 @@
+package conversion
```
lantoli, line 1: as this PR is a bit big and complex, do you think it could splitted in multiple PRs, or that would be hard and PRs wouldn't make sense separated?

EspenAlbert: I wanted to show include more than just the `common/` code.
I tried to keep the chagne to a minimum in TPF, not sure what to cut out. Could skip adding the extra test for the plan_modifer_test.go.
WDYT? Anything that didn't make sense here?

internal/common/conversion/unknown.go
EspenAlbert: Dejavu 😅
I think not, since depending on the vlaue we use `v.ElementType(ctx)`

internal/common/conversion/path_helpers_test.go
EspenAlbert: I don't understand the context here. Is this related to `path_helpers_test.go`?

internal/common/conversion/path_converter.go
EspenAlbert: Thanks. Changed in a2bac7ba2ea0c449d2169a233ddb77ea5aa66ca0

internal/common/conversion/path_helpers.go
EspenAlbert: Thanks! Changed in a2bac7ba2ea0c449d2169a233ddb77ea5aa66ca0

internal/common/conversion/path_helpers.go
EspenAlbert: Renamed to `IsIndexValue` in a2bac7ba2ea0c449d2169a233ddb77ea5aa66ca0

internal/common/conversion/unknown.go
lantoli: I know, this always strikes me if it can be simplified 😬

internal/common/conversion/path_helpers_test.go
lantoli: I mean the idea of having a map of tests instead of all assert.True statements

internal/common/conversion/path_converter.go
lantoli: let me think about it

internal/common/conversion/path_helpers_test.go
EspenAlbert: Refactored in 3e8b7d08d18fdcba2fbb618e8d35700600889011 but not sure I find it more readable. WDYT?

lantoli: yeah, not sure, i think you can revert it

internal/service/advancedclustertpf/resource.go
lantoli: what about unknownReplacements getting the plan and state instead of req and resp?

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,80 @@
...
+	Changes AttributeChanges
+}
+
+func (u *UnknownReplacments[ResourceInfo]) AddReplacement(name string, call UnknownReplacementCall[ResourceInfo]) {
+	// todo: Validate the name in the schema
+	// todo: Validate the name is not already in the replacements
+	u.Replacements[name] = call
+}
+
+func (u *UnknownReplacments[ResourceInfo]) ApplyReplacments(ctx context.Context, diags *diag.Diagnostics) {
```
lantoli: typo in UnknownReplacments and ApplyReplacments

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,303 @@
...
+	if obj.IsNull() || obj.IsUnknown() {
+		return nil
+	}
+	return conversion.TFModelObject[T](ctx, obj)
+}
+
+func UpdatePlanValue[T attr.Value](ctx context.Context, diags *diag.Diagnostics, d *PlanModifyDiffer, p path.Path, value T) {
+	if localDiags := d.resp.Plan.SetAttribute(ctx, p, value); localDiags.HasError() {
+		diags.Append(localDiags...)
+	}
```
lantoli, line 189:
```suggestion
	diags.Append(d.resp.Plan.SetAttribute(ctx, p, value)...)
```

no need for generic as it's an interface, copy also warnings if any

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,303 @@
...
+}
```
lantoli:
```suggestion
func UpdatePlanValue(ctx context.Context, diags *diag.Diagnostics, d *PlanModifyDiffer, p path.Path, value attr.Value) {
	diags.Append(d.resp.Plan.SetAttribute(ctx, p, value)...)
}
```

no need for generics as it's an interface, copy also warnings if any

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,80 @@
...
+type UnknownReplacementCall[ResourceInfo any] func(ctx context.Context, stateValue ParsedAttrValue, req *UnknownReplacementRequest[ResourceInfo]) attr.Value
+
+type UnknownReplacments[ResourceInfo any] struct {
+	Differ       *PlanModifyDiffer
+	Replacements map[string]UnknownReplacementCall[ResourceInfo]
+	Info         ResourceInfo
+}
+
+// ParsedAttrValue is a wrapper around attr.Value that provides type-safe accessors to support using the same signature of functions.
+type ParsedAttrValue struct {
```
lantoli, line 33: similar to other cases, if it only has one element, consider dropping the struct and just have:
```

type ParsedAttrValue attr.Value
```

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,303 @@
...
+
+func (d *DiffTPF[T]) PlanOrStateValue() *T {
+	if d.Plan != nil {
+		return d.Plan
+	}
+	return d.State
+}
+
+func findChanges(ctx context.Context, diff []tftypes.ValueDiff, diags *diag.Diagnostics, schema conversion.TPFSchema) AttributeChanges {
+	changes := map[string]bool{}
```
lantoli:
```suggestion
	changes := make(map[string]bool)
```

nit: more idiomatic

internal/common/customplanmodifier/find_changes.go
```diff
@@ -0,0 +1,74 @@
...
+	for change := range leafChanges {
+		if strings.HasPrefix(change, addPrefix) || strings.HasPrefix(change, removePrefix) {
+			return true
+		}
+	}
+	return false
+}
+
+func (a AttributeChanges) leafChanges(removeIndex bool) map[string]bool {
+	leafChanges := map[string]bool{}
```
lantoli:
```suggestion
	leafChanges := make(map[string]bool)
```

nit: more idiomatic

internal/common/customplanmodifier/find_changes.go
```diff
@@ -0,0 +1,74 @@
...
+	removePrefix := fmt.Sprintf("%s[-", name)
+	for change := range leafChanges {
+		if strings.HasPrefix(change, addPrefix) || strings.HasPrefix(change, removePrefix) {
+			return true
+		}
+	}
+	return false
+}
+
+func (a AttributeChanges) leafChanges(removeIndex bool) map[string]bool {
```
lantoli: it looks like map[string]bool is being used like a set, it only really has values for "true". In that case it might be clearer to work with map[string]struct{}.

same for other cases

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+		if d.Value2 != nil && d.Value2.IsKnown() && !d.Value2.IsNull() {
+			if localDiags.HasError() {
+				diags.Append(localDiags...)
+				continue
+			}
+			addChangeAndParentChanges(p.String())
+		}
+	}
+	return slices.Sorted(maps.Keys(changes))
```
lantoli: are they sorted for convenience or it's really needed to work?

ok, i guess it's because this comment:
```
// The diff is sorted by the path length, for example read_only_spec is processed before read_only_spec.disk_size_gb
```

internal/common/conversion/path_helpers_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsAttributeValueOnly(t *testing.T) {
```
lantoli:
```suggestion
func TestIsIndexValue(t *testing.T) {
```

internal/common/conversion/path_helpers_test.go
```diff
@@ -0,0 +1,63 @@
...
+	assert.True(t, conversion.IsIndexValue(path.Root("replication_specs").AtListIndex(0)))
+	assert.True(t, conversion.IsIndexValue(path.Root("replication_specs").AtMapKey("myKey")))
+	assert.True(t, conversion.IsIndexValue(path.Root("replication_specs").AtSetValue(types.StringValue("myKey"))))
```
lantoli: some false assert?

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,79 @@
...
+	parentString := p.ParentPath().ParentPath().String()
+	lastPart := LastPart(p)
+	indexWithSign := strings.Replace(lastPart, "[", "[-", 1)
+	if parentString == "" {
+		return indexWithSign
+	}
+	return parentString + "." + indexWithSign
+}
+
+func StripSquareBrackets(p path.Path) string {
```
lantoli: what about rename it to TrimLastIndex?

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,303 @@
...
+		if aName, ok := step.(tftypes.AttributeName); ok {
+			if slices.Contains(keepUnknown, string(aName)) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func StateConfigDiffs[T any](ctx context.Context, diags *diag.Diagnostics, d *PlanModifyDiffer, name string, checkNestedAttributes bool) []DiffTPF[T] {
```
lantoli: StateConfigDiffs is not used, can it be deleted?

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,80 @@
...
+		req := &UnknownReplacementRequest[ResourceInfo]{
+			Info:    u.Info,
+			Path:    unknown.Path,
+			Differ:  u.Differ,
+			Changes: u.Differ.AttributeChanges,
+			Unknown: unknown.UnknownValue,
+		}
+		response := replacer(ctx, ParsedAttrValue{Value: unknown.StateValue}, req)
+		if response.IsUnknown() {
+			tflog.Info(ctx, fmt.Sprintf("Keeping unknown value in plan @ %s", strPath))
```
lantoli: not sure about having so much log info, same for others

internal/service/advancedclustertpf/resource.go
EspenAlbert: Not possible, the new differ and plan updater must have req & resp objects to do diffs and read values

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: Thanks changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Thanks changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/customplanmodifier/find_changes.go
EspenAlbert: Thanks changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/customplanmodifier/find_changes.go
EspenAlbert: Thanks changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Added a comment in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/conversion/path_helpers_test.go
EspenAlbert: Thanks changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/conversion/path_helpers_test.go
EspenAlbert: Thanks changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/conversion/path_helpers.go
EspenAlbert: Good idea! Changed in 77eb77b81d993280f144c9c4c59b936b944c2c3c

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Yes, quite a few things in this file could be deleted :D

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: Switched all log statements to debug.

internal/service/advancedclustertpf/resource.go
lantoli: ok, i'll look in more detail, my impression was that it just needed the plan, state and diags

internal/common/customplanmodifier/plan_modify_differ.go
lantoli: I think it would definitely help to delete stuff not needed (yet)

internal/common/conversion/diags.go
```diff
@@ -49,3 +49,10 @@ func AddJSONBodyErrorToDiagnostics(msgPrefix string, err error, diags *diag.Diag
 	errorJSON := string(errorBytes)
 	diags.AddError(msgPrefix, errorJSON)
 }
+
+func DiagsNonEmpty(diags *diag.Diagnostics) bool {
```
lantoli: I think DiagsNonEmpty is not being used, as in another comment I think it's better to delete from this PR everything that is not needed right now

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+	return strings.Contains(lastPart, "[Value(")
+}
+
+func HasPrefix(p, prefix path.Path) bool {
+	prefixString := prefix.String()
+	pString := p.String()
+	return strings.HasPrefix(pString, prefixString)
+}
+
+func AttributeNameEquals(p path.Path, name string) bool {
```
lantoli: is AttributeNameEquals equivalent to AttributeName(p) == name ?

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return strings.HasSuffix(lastPart, "]")
+}
+
+func IsMapIndex(p path.Path) bool {
+	lastPart := LastPart(p)
+	return strings.HasSuffix(lastPart, "\"]")
+}
+
+func IsSetIndex(p path.Path) bool {
```
lantoli, line 24: can you tell more about IsSetIndex and why is different from the rest, here we do Contains instead of HasSuffix.   Can you give an example of set index to see how it looks like and how the suffix is?

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func TrimLastIndex(p path.Path) string {
+	if IsIndexValue(p) {
+		return p.ParentPath().String()
+	}
+	return p.String()
+}
+
+func TrimLastIndexPath(p path.Path) path.Path {
```
lantoli: To reduce cognitive load I think it's normally better to export only functions that are being used in another package.  In that regard, I would keep unexported / lowercase funcs like TrimLastIndexPath, IsMapIndex or  IsSetIndex (maybe there are more)

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+	for {
```
lantoli: is the `for` on purpose?  can it happen to have multiple indexes together? (please give an example if so)

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+		if IsIndexValue(p) {
+			p = p.ParentPath()
+		} else {
+			return p
+		}
+	}
+}
+
+func ParentPathWithIndex(p path.Path, attributeName string, diags *diag.Diagnostics) path.Path {
+	for {
```
lantoli, line 87: if the for is on purpose I think a better name would have Ancestor instead of Parent,   as I think in parent as the direct ancestor. similar comment for other funcs

internal/common/conversion/type_conversion.go
```diff
@@ -104,3 +108,29 @@ func NilForUnknownOrEmptyString(primitiveAttr types.String) *string {
 	}
 	return value
 }
+
+func TFModelList[T any](ctx context.Context, diags *diag.Diagnostics, input types.List) []T {
```
lantoli, line 112: TFModelList and TFModelObject have the same implementation as in plan_modifier, so I would remove from there and update their references to this common package

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -15,9 +15,9 @@ import (
 var (
 	// Change mappings uses `attribute_name`, it doesn't care about the nested level.
 	attributeRootChangeMapping = map[string][]string{
-		"disk_size_gb":           {}, // disk_size_gb can be change at any level/spec
-		"replication_specs":      {},
-		"mongo_db_major_version": {"mongo_db_version"},
+		"disk_size_gb":      {}, // disk_size_gb can be change at any level/spec
+		"replication_specs": {},
+		// "mongo_db_major_version": {"mongo_db_version"}, // Using new plan modifier logic to test this
```
lantoli: I don't know if this is a good test after a PR I saw by Oriol that we'll keep mongo_db_version always as unknown

internal/service/advancedclustertpf/plan_modifier2.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/customplanmodifier"
+)
+
+var attributePlanModifiers = map[string]customplanmodifier.UnknownReplacementCall[PlanModifyResourceInfo]{
+	"mongo_db_version": mongoDBVersionReplaceUnknown,
```
lantoli: similar to previous comment, don't know if this is good example anymore

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,165 @@
...
+		diags.AddError("Error diffing state and config", err.Error())
+		return nil
+	}
+
+	attributeChanges := findChanges(ctx, diffStatePlan, diags, schema)
+	tflog.Debug(ctx, fmt.Sprintf("Attribute changes: %s\n", strings.Join(attributeChanges, "\n")))
+	return &PlanModifyDiffer{
+		req:              req,
+		resp:             resp,
+		stateConfigDiff:  diffStateConfig,
```
lantoli: similar to prev. comments, can we delete stateConfigDiff as we're not using it?

internal/service/advancedclustertpf/plan_modifier2.go
```diff
@@ -0,0 +1,49 @@
+package advancedclustertpf
```
lantoli: I think it's better to keep this filename as plan_modifier and rename the other one to something like plan_modifier_old and add a comment that will be eventually deleted

internal/common/conversion/path_converter.go
```diff
@@ -0,0 +1,150 @@
...
+						fmt.Sprintf("Original Error: %s", err),
+				),
+			}
+		}
+
+		fwStep, err := AttributePathStep(ctx, tfTypeStep, attrType)
+
+		if err != nil {
+			return path.Empty(), diag.Diagnostics{
+				diag.NewErrorDiagnostic(
```
lantoli, line 71: consider having a helper func for the 3 errors in the func

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,165 @@
...
+	addChangeAndParentChanges := func(change string) {
+		changes[change] = true
+		parts := strings.Split(change, ".")
+		for i := range parts[:len(parts)-1] {
+			changes[strings.Join(parts[:len(parts)-1-i], ".")] = true
+		}
+	}
+	for _, d := range diff {
+		p, localDiags := conversion.AttributePath(ctx, d.Path, schema)
+		if conversion.IsListIndex(p) {
```
lantoli: do we need to do something similar for IsMapIndex or IsSetIndex?

internal/service/advancedclustertpf/testdata/.gitignore
```diff
@@ -0,0 +1,2 @@
+ClusterTwoRepSpecsWithAutoScalingAndSpecs_*.yaml
```
lantoli: .gitignore not needed anymore

internal/service/advancedclustertpf/resource.go
lantoli: discussed offline, explore if ctx, &resp.diags, &req.State, &resp.Plan can be the arguments

internal/common/conversion/unknown.go
lantoli: I don't see references to AsUnknownValue in this PR, can it be deleted?

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,85 @@
...
+type UnknownReplacementCall[ResourceInfo any] func(ctx context.Context, stateValue ParsedAttrValue, req *UnknownReplacementRequest[ResourceInfo]) attr.Value
+
+type UnknownReplacements[ResourceInfo any] struct {
+	Differ       *PlanModifyDiffer
+	Replacements map[string]UnknownReplacementCall[ResourceInfo]
+	Info         ResourceInfo
+}
+
+// ParsedAttrValue is a wrapper around attr.Value that provides type-safe accessors to support using the same signature of functions.
+type ParsedAttrValue struct {
```
lantoli: not sure if the complexity of adding ParsedAttrValue is worth it as opposse to use attr.Value,  and maybe have some help funcs like:
```

func getAttrValueString(value attr.Value) types.String {
return value.(types.String)
}
func getAttrValueObject(value attr.Value) types.Object {
return value.(types.Object)
}

```

lantoli: or we might not even need the helper funcs as conversion is quite direct

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,85 @@
...
+	Unknown attr.Value
+	Differ  *PlanModifyDiffer
+	Diags   *diag.Diagnostics
+	Path    path.Path
+	Changes AttributeChanges
+}
+
+func (u *UnknownReplacements[ResourceInfo]) AddReplacement(name string, call UnknownReplacementCall[ResourceInfo]) {
+	// todo: Validate the name exists in the schema
+	_, existing := u.Replacements[name]
```
lantoli: nit: use always the same name, e.g. `found`

internal/common/conversion/path_helpers.go
EspenAlbert: Example:
`"advanced_configuration.custom_openssl_cipher_config_tls12[Value(\"ECDHE-RSA-AES256-GCM-SHA384\")]"`

Want to separate IsListIndex from IsMapIndex and IsSetIndex.
See new docstrings in `plan_modify_differ.go#L140`

internal/common/conversion/type_conversion.go
EspenAlbert: Changing in the next PR

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: yes, I think you are right

internal/common/conversion/path_helpers.go
EspenAlbert: Thanks. Should be better now!

internal/common/conversion/path_helpers.go
EspenAlbert: Looks like it wasn't necessary. Removed. Thank you!

internal/common/conversion/path_helpers.go
EspenAlbert: Renamed! Thank you!

internal/service/advancedclustertpf/plan_modifier.go
EspenAlbert: Decided to move resource changes to next PR: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3218

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: changed! Thank you!

internal/service/advancedclustertpf/plan_modifier2.go
EspenAlbert: Went straight for full change in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3218

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Updated doc string

internal/service/advancedclustertpf/testdata/.gitignore
EspenAlbert: Thanks! Deleted

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: refactored ✔️  Thank you!

internal/common/conversion/path_helpers.go
lantoli: thanks, so I don't still understand why we want to differentiate IsSetIndex and IsMapIndex, can we just have IsMapIndex? what is a use case where we want to treat them differently?

(also note IsMapIndex is not correct bc you have to do something like in IsListIndex impl.,  say if IsSetIndex(p) return false)

internal/common/conversion/path_helpers.go
```diff
@@ -0,0 +1,118 @@
...
+func lastPart(p path.Path) string {
+	parts := strings.Split(p.String(), ".")
+	return parts[len(parts)-1]
+}
+
+func isIndexValue(p path.Path) bool {
+	return IsMapIndex(p) || IsListIndex(p) || IsSetIndex(p)
+}
+
+func attributeNameEquals(p path.Path, name string) bool {
```
lantoli: is this func helpful now, or can be deleted?

internal/common/customplanmodifier/unknown_replacement_test.go
```diff
@@ -0,0 +1,279 @@
...
+func (r *rs) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	r.base.Metadata(ctx, req, resp)
+}
+
+func (r *rs) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	r.base.Configure(ctx, req, resp)
+}
+
+// ModifyPlan is the only method overridden in this test.
+func (r *rs) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
lantoli, line 59: as discussed offline, not sure about this test as it's very coupled to the implementation

internal/common/conversion/path_helpers.go
EspenAlbert: good point. Deleted

internal/common/customplanmodifier/unknown_replacement_test.go
EspenAlbert: Curious to hear what others think. Please react with 👍 if you find it useful or 👎  if it is just noisy/confusing

internal/common/conversion/path_helpers.go
EspenAlbert: There is a subtle difference:
- `"replication_specs[Value(\"myKey\")]"` ends with `)]`, close parenthesis+square bracket
- `"replication_specs[\"myKey\"]"` ends with `"]` quote+square bracket

See also test case `TestIndexMethods`

internal/service/advancedclustertpf/resource.go
EspenAlbert: Changed. Thank you!

internal/common/conversion/unknown.go
EspenAlbert: Thanks. Deleted

internal/common/conversion/path_converter.go
```diff
@@ -0,0 +1,150 @@
...
+	attrValue := attrType.ValueType(ctx)
+	if localDiags := src.GetAttribute(ctx, convertedPath, &attrValue); localDiags.HasError() {
+		diags.Append(localDiags...)
+		return nil, convertedPath
+	}
+	return attrValue, convertedPath
+}
+
+// AttributePath similar to the internal function in TPF, but simpler interface as argument and less logging
+func AttributePath(ctx context.Context, tfType *tftypes.AttributePath, schema TPFSchema) (path.Path, diag.Diagnostics) {
```
EspenAlbert, line 43: This is the internal function:
[func AttributePath(ctx context.Context, tfType *tftypes.AttributePath, schema fwschema.Schema) (path.Path, diag.Diagnostics) {](https://github.com/hashicorp/terraform-plugin-framework/blob/e09ec9d169c581d2606372ecdfb0113be7e3b34f/internal/fromtftypes/attribute_path.go#L17)

internal/common/conversion/path_helpers.go
lantoli: oh, I missed that different end, thanks.  and when do we want to treat them differently?
I was considering index when it's a number like [number], else set or map index

internal/common/conversion/path_converter.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	attrValue := attrType.ValueType(ctx)
+	if localDiags := src.GetAttribute(ctx, convertedPath, &attrValue); localDiags.HasError() {
+		diags.Append(localDiags...)
+		return nil, convertedPath
+	}
+	return attrValue, convertedPath
+}
+
+// AttributePath similar to the internal function in TPF, but simpler interface as argument and less logging
```
maastha: is this the only difference between the two implementations? If yes, not sure why we want to implement this on our end, would prefer to re-use the original method, or am I missing something?
If the difference is more than just "simpler interface as argument and less logging" then I think we should enrich the comment to address the "why"

internal/common/conversion/path_converter.go
```diff
@@ -0,0 +1,150 @@
...
+						fmt.Sprintf("Original Error: unknown path.PathStep type: %#v", fwStep),
+				),
+			}
+		}
+	}
+
+	return fwPath, nil
+}
+
+func AttributePathStep(ctx context.Context, tfType tftypes.AttributePathStep, attrType attr.Type) (path.PathStep, error) {
```
maastha, line 119: same as my previous comment, why do we need our own implementation for this as well? Can we reuse https://github.com/hashicorp/terraform-plugin-framework/blob/e09ec9d169c581d2606372ecdfb0113be7e3b34f/internal/fromtftypes/attribute_path_step.go#L19 ?

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// AddKeepUnknownAlways adds the attribute name to the list of attributes that should always keep unknown values. For example connection_string or state_name.
+func (u *UnknownReplacements[ResourceInfo]) AddKeepUnknownAlways(keepUnknown ...string) {
+	u.keepUnknownAttributeNames = append(u.keepUnknownAttributeNames, keepUnknown...)
+}
+
+// AddKeepUnknownOnChanges adds the attribute changed and its depending attributes to the list of attributes that should keep unknown values.
+// However, it does not infer dependencies. For example: instance_size --> disk_size_gb, and disk_gb --> disk_iops, doesn't mean instance_size --> disk_iops.
+func (u *UnknownReplacements[ResourceInfo]) AddKeepUnknownOnChanges(attributeEffectedMapping map[string][]string) {
```
maastha:
```suggestion
func (u *UnknownReplacements[ResourceInfo]) AddKeepUnknownOnChanges(attributeAffectedMapping map[string][]string) {
```

maastha: IIUC

EspenAlbert: Thank you.Changed in 7229da8a952cc379d564770aadb5d230063fcde2

internal/common/conversion/path_converter.go
EspenAlbert: Added comment in 7229da8a952cc379d564770aadb5d230063fcde2
Internal functions are not exposed in go.

internal/common/conversion/path_converter.go
EspenAlbert: Added comment in 7229da8a952cc379d564770aadb5d230063fcde2
Internal functions are not exposed in go.

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+)
+
+func NewUnknownReplacements[ResourceInfo any](ctx context.Context, state *tfsdk.State, plan *tfsdk.Plan, diags *diag.Diagnostics, schema conversion.TPFSchema, info ResourceInfo) *UnknownReplacements[ResourceInfo] {
```
AgustinBettati, line 17: q: Is there a strong reason for the type parameter here? From usages I see we always pass PlanModifyResourceInfo struct but wanted to confirm

internal/common/customplanmodifier/find_changes.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// ListIndexChanged returns true if the list at the given index has changed, false if it was added or removed
+func (a AttributeChanges) ListIndexChanged(fullPath string, index int) bool {
+	indexPath := fmt.Sprintf("%s[%d]", fullPath, index)
+	return slices.Contains(a, indexPath)
+}
+
+// ListLenChanges accepts a fullPath, e.g., "replication_specs[0].region_configs" and returns true if the length of the nested list has changed
+func (a AttributeChanges) ListLenChanges(fullPath string) bool {
```
oarbusi: nit: 
```suggestion
func (a AttributeChanges) ListLenChanged(fullPath string) bool {
```

internal/common/customplanmodifier/find_changes.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return false
+}
+
+func (a AttributeChanges) allAttributeNameChanges() map[string]struct{} {
+	nameChanges := make(map[string]struct{})
+	for _, change := range a {
+		parts := strings.Split(change, ".")
+		attributeName := parts[len(parts)-1]
+		nameChanges[attributeName] = struct{}{}
```
oarbusi, line 51: Why an empty struct instead of something simpler, bool for example? It would make the code more readable, IMO. I understand that struct{} requires 0 bytes to store, but I don't think memory will be an issue here.

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,157 @@
...
+	return readSrcStructValue[T](ctx, d.plan, p)
+}
+
+func ReadStateStructValue[T any](ctx context.Context, d *PlanModifyDiffer, p path.Path) *T {
+	return readSrcStructValue[T](ctx, d.state, p)
+}
+
+func readSrcStructValue[T any](ctx context.Context, src conversion.TPFSrc, p path.Path) *T {
+	var obj types.Object
+	if localDiags := src.GetAttribute(ctx, p, &obj); localDiags.HasError() {
```
oarbusi: localDiags not used/passed anywhere? Why is it handled differently than in `readSrcStructValues`?

internal/common/customplanmodifier/plan_modify_differ.go
```diff
@@ -0,0 +1,157 @@
...
+		changes[change.String()] = true
+		for _, p := range conversion.AncestorPaths(change) {
+			changes[p.String()] = true
+		}
+	}
+	// avoids adding change for removed region_configs inside a removed replication_specs
+	isAncestorRemoved := func(p path.Path) bool {
+		for _, a := range conversion.AncestorPaths(p) {
+			if conversion.IsListIndex(a) {
+				if _, found := changes[conversion.AsRemovedIndex(a)]; found {
```
oarbusi:
```suggestion
				if changes[conversion.AsRemovedIndex(a)]{
```
should return zero value when not found

internal/common/customplanmodifier/unknown_replacement.go
```diff
@@ -0,0 +1,124 @@
...
+	Differ       *PlanModifyDiffer
+	Replacements map[string]UnknownReplacementCall[ResourceInfo]
+	Info         ResourceInfo
+
+	keepUnknownAttributeNames []string // todo: Support validating values when adding attributes
+	keepUnknownsExtraCalls    []func(ctx context.Context, stateValue attr.Value, req *UnknownReplacementRequest[ResourceInfo]) []string
+}
+
+func (u *UnknownReplacements[ResourceInfo]) AddReplacement(name string, call UnknownReplacementCall[ResourceInfo]) {
+	// todo: Validate the name exists in the schema
```
oarbusi: CLOUDP?

oarbusi (approved):
Very long PR 😅 overall LGMT

internal/common/conversion/path_converter.go
maastha: Thank you!

internal/common/customplanmodifier/unknown_replacement.go
lantoli: I concur that we should try to limit generics, especially for the whole struct (as opposed to in a func) as it can make the code harder to follow.    but maybe it's a good use case here

internal/common/customplanmodifier/find_changes.go
lantoli: it was my suggestion, was bool before, but for me it was not clear because we only added elements with true,  to represent a `set` I think this approach is better

oarbusi: in plan_modify_differ.go the same is done with map[string]bool and usage of that map is clearer IMO.

lantoli: I hear you, and i agree if we would add true and false values, but when we only add true values is when I think it's better struct{}

internal/common/conversion/path_helpers.go
EspenAlbert: So far `set` and `map` are used with _primitive values_ `string` in this case. While the ListIndex is used with nested objects. We don't need to check for `SetLenChanges` or `MapLenChanges` only `ListLenChanges` so far

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: The ResourceInfo is meant to depend on the resource. For example, `advanced_cluster` ResourceInfo will differ from `search_deployment` or `project` ResourceInfo

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Thanks for spotting this.
I'm not 100% sure all the cases where this might happen.
I think it is used in `readSrcStructValues` to use the `conversion.TFModelList` method, but I agree we should be consistent.
I'll investigate and update

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: Thanks added in 8d87c2375af0584f9c4765e51d1765ee498c1736

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Thanks. Added in 8d87c2375af0584f9c4765e51d1765ee498c1736

internal/common/customplanmodifier/find_changes.go
EspenAlbert: Thanks changed in 8d87c2375af0584f9c4765e51d1765ee498c1736

internal/common/customplanmodifier/unknown_replacement.go
EspenAlbert: Added doc string in 0fe27800a8fca77d7a6f69bec354fa77ec5d407a
Does that help?

internal/common/conversion/path_converter.go
marcosuma: I haven't gone deep, but it smells a lot we're copying internal methods. Also, what is the impact on the policy?

internal/common/customplanmodifier/plan_modify_differ.go
EspenAlbert: Updated, choosing to remove the diags argument.
These errors should be spotted before production, but just to be sure, we log them instead of adding to "root" diagnostics. This avoids any risk of the user having a "stuck" resource due to plan modifier errors.

internal/common/conversion/path_converter.go
AgustinBettati: Did we get any additional information here?

EspenAlbert: Good point. Waiting on reply.

EspenAlbert:
Due to high effort and no demand this will be closed, possibly used if plan verbosity requests comes.

PR #3201, opened 2025-03-24
chore: Improves client configurations

oarbusi:
## Description

 Improves client configurations after the feedback received in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3187.

The goal of this PR is to optimize the client configurations. In case we notice any further networking issue, we would then start using something off the shelf like https://github.com/hashicorp/go-retryablehttp, and if that is not enough, as a last resource we will add custom logic, like in #3187 

Link to any related issue(s): CLOUDP-262912

## Type of change:

## Required Checklist:


internal/config/client.go
```diff
@@ -28,8 +29,27 @@ import (
 const (
 	toolName              = "terraform-provider-mongodbatlas"
 	terraformPlatformName = "Terraform"
+
+	timeout               = 5 * time.Second
```
lantoli, line 35: is there a way to have some tests, like waiting a connection for more than 5s and see that it times out instead of waiting as before this change?

internal/config/client.go
```diff
@@ -78,6 +98,8 @@ func (c *Config) NewClient(ctx context.Context) (any, error) {
 	// setup a transport to handle digest
 	transport := digest.NewTransport(cast.ToString(c.PublicKey), cast.ToString(c.PrivateKey))
```
lantoli: you can use NewTransportWithHTTPTransport instead and pass the transport

lantoli (approved):

internal/config/client.go
oarbusi: Done

internal/config/client.go
oarbusi: It is possible, for sure, but I don't think it adds value. This should(and is) be tested in the net/http package https://github.com/golang/go/blob/master/src/net/http/transport_test.go

lantoli: ok, i was not refering to test the core functionality, more to test the end to end connection to make sure we're using it correctly.  but it's fine

internal/config/client.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
```
lantoli, line 7: should we do the same change in CFN?

internal/config/client.go
```diff
@@ -76,7 +96,7 @@ type PlatformVersion struct {
 // NewClient func...
 func (c *Config) NewClient(ctx context.Context) (any, error) {
 	// setup a transport to handle digest
-	transport := digest.NewTransport(cast.ToString(c.PublicKey), cast.ToString(c.PrivateKey))
+	transport := digest.NewTransportWithHTTPTransport(cast.ToString(c.PublicKey), cast.ToString(c.PrivateKey), baseTransport)
```
maastha, line 102: [q] how does this PR now handle retries as mentioned in the ticket?

oarbusi: I think that after the feedback in #3187, the goal of the ticket has shifted a bit. We will optimize the client configuration as part of this ticket (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3201). In case we notice any further networking issue, we would then start using something off the shelf like https://github.com/hashicorp/go-retryablehttp, and if that is not enough, as a last resource we will add custom logic, similar to [#3187](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3187). I included this details both in this PR description and as a comment in the CLOUDP. Let me know what you think

internal/config/client.go
oarbusi: I will do it once this is approved and merged

internal/config/client.go
EspenAlbert, line 35: How were these values chosen?

internal/config/client.go
```diff
@@ -28,8 +29,27 @@ import (
...
+
+	timeout               = 5 * time.Second
+	keepAlive             = 30 * time.Second
+	maxIdleConns          = 10
+	maxIdleConnsPerHost   = 5
+	idleConnTimeout       = 30 * time.Second
+	expectContinueTimeout = 1 * time.Second
 )
 
+var baseTransport = &http.Transport{
```
marcosuma, line 43: It took me time to read the other PR and [the comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3187/files#r2007759417) before understanding the context of this change. IIUC, this change in the end is not worrying about networking changes but it's only improving the configuration "hoping" that will fix networking issues. It might even make it worse, who knows.

I am going to leave some questions, one of them blocking this PR:
1. [blocking] how were these values chosen? did we just copy Atlas CLI? took them from other providers? or what?
2. [non-blocking] what is the plan to actually address CLOUDP-262912? Context of why and how should be somewhere that it's easily referenced across the work you are doing and I am going to say again that you need a document to talk about the plan and why this change.
3. [non-blocking] have you tried to replicate networking issues? this might be hard, but how do we know these values won't make it worse for customers? we have zero metrics so we are going total blind

Going to leave here the DefaultTransport for reference that, even if not recommended, it's what we are currently using:
```
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: defaultTransportDialContext(&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
	}),
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```

lantoli: Those setting values helped CLI with networking problems, they might also help us.

We might want to ask CLI team why they chose those config values.

Biggest differences seem to be idleConnTimeout 30s instead of 90s, DialContext.timeout is 5s instead of 30s, maxIdleConns is 10 instead of 100.

I agree that we're blind here, but don't know if it's easy to reproduce the network issues.

oarbusi: Discussed the values in the one-pager. Closing this as it's no longer relevant and team is aligned on the changes in this PR

oarbusi:
Alignment has been reached on this issue through the one-pager doc. Merging this.

PR #3211, opened 2025-03-26
test: Removes check for specific MongoDB version and removes flex specific project

oarbusi:
## Description

Fixes to avoid failing tests:
- Stops checking a specific mongodb version
- Keeps mongo_db_version unknown in TPF
- Flex cluster does not need anymore a specific project

Link to any related issue(s): CLOUDP-308797

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -100,9 +100,6 @@ on:
       azure_private_endpoint_region:
         type: string
         required: true
-      mongodb_atlas_flex_project_id:
-        type: string
-        required: true
       confluent_cloud_network_id:
         type: string
         required: true
@@ -754,7 +751,6 @@ jobs:
       - name: Acceptance Tests
         env:
           MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
-          MONGODB_ATLAS_FLEX_PROJECT_ID: ${{ inputs.mongodb_atlas_flex_project_id }}
           ACCTEST_PACKAGES: ./internal/service/flexcluster
         run: make testacc
 
--- a/internal/service/advancedcluster/resource_advanced_cluster_test.go
+++ b/internal/service/advancedcluster/resource_advanced_cluster_test.go
@@ -1288,7 +1288,6 @@ func TestAccMockableAdvancedCluster_replicasetAdvConfigUpdate(t *testing.T) {
 		key   = "env"
 		value = "test"
 	}
-	mongo_db_major_version = "8.0"
 	pit_enabled = true
 	redact_client_log_data = true
 	replica_set_scaling_strategy = "NODE_TYPE"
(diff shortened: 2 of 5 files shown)


internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -1255,7 +1256,6 @@ func TestAccMockableAdvancedCluster_replicasetAdvConfigUpdate(t *testing.T) {
 			"state_name":                    "IDLE",
 			"backup_enabled":                "true",
 			"bi_connector_config.0.enabled": "true",
-			"mongo_db_major_version":        "8.0",
```
AgustinBettati, line 1258: The config for this step `configBasicReplicaset(t, projectID, clusterName, fullUpdate)` does explicitly define mongo_db_major_version to `8.0`, are we not facing a non empty plan here given the response defines the value `8.1`?

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -100,9 +100,6 @@ on:
       azure_private_endpoint_region:
         type: string
         required: true
-      mongodb_atlas_flex_project_id:
```
AgustinBettati, line 103: nice

AgustinBettati (commented):
have a doubt on the cluster test

internal/service/advancedcluster/resource_advanced_cluster_test.go
oarbusi: So the issue here is that in the first step the version is not specified, so the current default will be used. In the update we are trying to update to a previous version of MongoDB, from 8.1 to 8.0 in this case, and it does not update and 8.1 remains. 
I have now removed the mongodb version from the update config. I think it involves too much maintenance to have specific versions in the tests. When a new release is done we have to change them

internal/service/advancedclustertpf/plan_modifier.go
```diff
@@ -62,7 +62,7 @@ func useStateForUnknowns(ctx context.Context, diags *diag.Diagnostics, state, pl
 		return
 	}
 	attributeChanges := schemafunc.NewAttributeChanges(ctx, state, plan)
-	keepUnknown := []string{"connection_strings", "state_name"} // Volatile attributes, should not be copied from state
+	keepUnknown := []string{"connection_strings", "state_name", "mongo_db_version"} // Volatile attributes, should not be copied from state
```
EspenAlbert, line 64: With this change you can remove L20: 
`		"mongo_db_major_version": {"mongo_db_version"},`

Since it will always be unknown, not only when `mongo_db_major_version` changes

internal/service/advancedcluster/resource_advanced_cluster_test.go
EspenAlbert: What about setting `mongo_db_major_version = "7.0"` instead? That should always trigger an update

internal/service/advancedclustertpf/plan_modifier.go
oarbusi: done in [760e579](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3211/commits/760e579399343da7b4c6aa4a6100d5cbb506aeb8)

internal/service/advancedcluster/resource_advanced_cluster_test.go
AgustinBettati: That could be an option, with the risk it eventually becomes outdated. We do have a specific test for downgrade with pinned FCV (and the upgrade) so I think we are okay in not using `mongo_db_major_version` for this test

AgustinBettati (approved):
LGTM

PR #3230, opened 2025-03-31
chore: Define conversion function for setting API response values into TF Model (non-nested)

lantoli:
## Description

Define conversion function for setting API response values into TF Model (non-nested).
At the moment it's created in a new `autogeneration` package.

Link to any related issue(s): CLOUDP-309398

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogeneration/marshal.go
+++ b/internal/common/autogeneration/marshal.go
@@ -1,0 +1,6 @@
+package autogeneration
+
+// Unmarshal get a JSON (e.g. from an Atlas response) and unmarshals it into a Terraform model.
+func Unmarshal(raw []byte, to any) error {
+	return nil
+}
--- a/internal/common/autogeneration/marshal.go
+++ b/internal/common/autogeneration/marshal.go
@@ -44,14 +44,14 @@ func mapField(nameAttrSrc string, valueAttrSrc any, valDest reflect.Value) error
 	}
 	switch v := valueAttrSrc.(type) {
 	case string:
-		return assignField(fieldDest, types.StringValue(v))
+		return assignField(nameDest, fieldDest, types.StringValue(v))
 	case bool:
-		return assignField(fieldDest, types.BoolValue(v))
+		return assignField(nameDest, fieldDest, types.BoolValue(v))
 	case float64: // number: try int or float
-		if assignField(fieldDest, types.Float64Value(v)) == nil {
+		if assignField(nameDest, fieldDest, types.Float64Value(v)) == nil {
 			return nil
 		}
-		return assignField(fieldDest, types.Int64Value(int64(v)))
+		return assignField(nameDest, fieldDest, types.Int64Value(int64(v)))
 	case nil:
 		return nil // skip nil values, no need to set anything
 	default:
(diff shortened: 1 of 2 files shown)


internal/common/autogeneration/marshal.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+	return nil
+}
+
+func mapField(nameAttrSrc string, valueAttrSrc any, valDest reflect.Value) error {
+	nameDest := xstrings.ToPascalCase(nameAttrSrc)
+	fieldDest := valDest.FieldByName(nameDest)
+	if !fieldDest.CanSet() {
+		return nil // skip fields that cannot be set, are invalid or not found
```
oarbusi, line 44: So we don't fail in these cases? How will we know that a field that was supposed to be mapped has not been mapped?

internal/common/autogeneration/marshal_test.go
```diff
@@ -0,0 +1,147 @@
...
+// It will be updated when we add support for them.
+func TestUnmarshalUnsupportedModel(t *testing.T) {
+	testCases := map[string]struct {
+		model        any
+		responseJSON string
+	}{
+		"Int32 not supported yet as it's not being used in any model": {
+			responseJSON: `{"attr": 1}`,
+			model: &struct {
+				Attr types.Int32
```
AgustinBettati, line 106: Would there be a need for supporting Int32? I would imagine autogeneration would alway use Int64 when handling an integer type (same applies for floats). Still like the ideal of having the test.

internal/common/autogeneration/marshal.go
```diff
@@ -0,0 +1,70 @@
...
+	case nil:
+		return nil // skip nil values, no need to set anything
+	default:
+		return fmt.Errorf("not supported yet type %T for field %s", v, nameAttrSrc)
+	}
+}
+
+func assignField(nameDest string, fieldDest reflect.Value, valueDest attr.Value) error {
+	valObj := reflect.ValueOf(valueDest)
+	if !fieldDest.Type().AssignableTo(valObj.Type()) {
```
AgustinBettati, line 65: nice, this keeps it very simple for now

AgustinBettati (approved):
LGTM

internal/common/autogeneration/marshal.go
lantoli: I did it on purpose, mainly to cover the use cases where there are extra fields in API response like `links` that we don't want to have in the model.  It's true that it could happen what you say, that we miss some assignment, although i would expect some tests in the resource to catch it.  
I think it's ok to be lenient at the moment, we might want to be stricter in the future if we find this is an issue.

Note that will only skip if the JSON attribute is not in the model.  If the attribute is in the model but can't be assigned, an error will be thrown.

@oarbusi is that ok?

oarbusi: Makes sense, thanks for clarifying

oarbusi (approved):

internal/common/autogeneration/marshal_test.go
lantoli: we don't need it at the moment, so I thought it was better not to include it, but having it documented in the test and the func description.  As you say, I would go for Int64 / Float64 for numbers in the model

internal/common/autogeneration/marshal.go
```diff
@@ -0,0 +1,70 @@
...
+	if !fieldDest.CanSet() {
+		return nil // skip fields that cannot be set, are invalid or not found
+	}
+	switch v := valueAttrSrc.(type) {
+	case string:
+		return assignField(nameDest, fieldDest, types.StringValue(v))
+	case bool:
+		return assignField(nameDest, fieldDest, types.BoolValue(v))
+	case float64: // number: try int or float
+		if assignField(nameDest, fieldDest, types.Float64Value(v)) == nil {
```
EspenAlbert, line 52: @lantoli wouldn't it be better with int64 before float64?

lantoli: @EspenAlbert I think the order is irrelevant here, the success of the assignment depends only on the type of the model attribute, be it types.Float64Value or types.Int64Value.  
For instance, if the attribute type is types.Float64Value, assignment with types.Int64Value(v) will fail.

PR #3239, opened 2025-04-01
chore: Define conversion function for generating API request body from Resource Model in create (non-nested)

lantoli:
## Description

Define conversion function for generating API request body from Resource Model in create (non-nested).

`Marshal` is used to create the Atlas request from the Terraform plan model. It's called in operations that need to create an Atlas request payload, i.e. `Create` and `Update`. It's not used in `Read`, `Delete` or `Import`.

`Unmarshal` is used to update the Terraform model from the Atlas response. It's called in `Create`, `Update` and `Read`. It's not used in `Delete` or `Import`.

It supports an `autogeneration` tag in model struct fields that can have:
- `omitjson`:  Field will never be marshaled into JSON, useful for attributes not needed in Atlas requests like timeouts or computed values.
- `omitjsonupdate`: Field that is not marshaled into JSON when it's an update, this field is sent only when the resource is created.

Link to any related issue(s): CLOUDP-310079

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogeneration/marshal.go
+++ b/internal/common/autogeneration/marshal.go
@@ -4,10 +4,16 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/huandu/xstrings"
+)
+
+const (
+	tagKeyAutogeneration = "autogeneration"
+	tagValueOmitJSON     = "omitjson"
 )
 
 // Marshal gets a Terraform model and marshals it into JSON (e.g. for an Atlas request).
@@ -40,7 +46,11 @@ func Unmarshal(raw []byte, model any) error {
 func marshalAttrs(valModel reflect.Value) (map[string]any, error) {
 	objJSON := make(map[string]any)
 	for i := 0; i < valModel.NumField(); i++ {
-		attrNameModel := valModel.Type().Field(i).Name
+		attrTypeModel := valModel.Type().Field(i)
+		if strings.Contains(attrTypeModel.Tag.Get(tagKeyAutogeneration), tagValueOmitJSON) {
+			continue // skip fields with tag `omitjson`
+		}
+		attrNameModel := attrTypeModel.Name
 		attrValModel := valModel.Field(i)
 		if err := marshalAttr(attrNameModel, attrValModel, objJSON); err != nil {
 			return nil, err
--- a/internal/common/autogeneration/marshal.go
+++ b/internal/common/autogeneration/marshal.go
@@ -12,12 +12,13 @@ import (
 )
 
 const (
-	tagKeyAutogeneration = "autogeneration"
-	tagValueOmitJSON     = "omitjson"
+	tagKey           = "autogeneration"
+	tagValOmitJSON   = "omitjson"
+	tagValCreateOnly = "createonly"
 )
 
 // Marshal gets a Terraform model and marshals it into JSON (e.g. for an Atlas request).
-func Marshal(model any) ([]byte, error) {
+func Marshal(model any, isCreate bool) ([]byte, error) {
 	valModel := reflect.ValueOf(model)
 	if valModel.Kind() != reflect.Ptr {
 		panic("model must be pointer")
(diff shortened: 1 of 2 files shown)


internal/common/autogeneration/marshal.go
```diff
@@ -4,67 +4,133 @@ import (
...
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/huandu/xstrings"
 )
 
+const (
+	tagKey           = "autogeneration"
+	tagValOmitJSON   = "omitjson"
+	tagValCreateOnly = "createonly"
```
AgustinBettati: Giving some more thought to the naming, would `omitjsonupdate` over `createonly` be more consistent?

internal/common/autogeneration/marshal.go
```diff
@@ -4,67 +4,133 @@ import (
...
+)
+
+// Marshal gets a Terraform model and marshals it into JSON (e.g. for an Atlas request).
+func Marshal(model any, isCreate bool) ([]byte, error) {
```
AgustinBettati: Would leave an additional comment clarifying when attributes are sent in payload, and how use of tags changes that behaviour

internal/common/autogeneration/marshal_test.go
```diff
@@ -3,50 +3,175 @@ package autogeneration_test
...
 	"github.com/stretchr/testify/require"
 )
 
+func TestMarshalBasic(t *testing.T) {
+	model := struct {
+		AttrFloat  types.Float64 `tfsdk:"attribute_float"`
+		AttrString types.String  `tfsdk:"attribute_string"`
+		// values with tag `omitjson` are not marshaled, and they don't need to be Terraform types
+		AttrOmit            types.String `tfsdk:"attribute_omit" autogeneration:"omitjson"`
+		AttrOmitNoTerraform string       `autogeneration:"omitjson"`
```
AgustinBettati, line 19: Any particular reason why this is not types.String? Would also define it with a known string value to ensure it is never sent even if it has a value in the config

AgustinBettati (approved):
LGTM, leaving some doubts/suggestions

internal/common/autogeneration/marshal_test.go
lantoli: do you mean AttrOmitNoTerraform?  It's to prove that in this case you can use "invalid" types that are not Terraform types in the model, and that's ok because it's not sent in JSON, like explained in the comment above.
In the normal case it would panic if the model has an attribute that is not Terraform, e.g test [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/2a5577ff29d09ea81775952a2e48eda11e65585e/internal/common/autogeneration/marshal_test.go#L117)

lantoli: good idea, changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3239/commits/d6c212b4cb421ea24dbcec6ceb1db79616b4cbbf

internal/common/autogeneration/marshal.go
lantoli: good idea, changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3239/commits/d6c212b4cb421ea24dbcec6ceb1db79616b4cbbf

internal/common/autogeneration/marshal.go
lantoli: done here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3239/commits/065089b6fbe9b9dc8498c0ea0e0fcbb739a8ae8d

marcosuma (approved):

internal/common/autogeneration/marshal_test.go
AgustinBettati: Aah makes sense, I see `AttrOmit` covers the case i was mentioning, thanks!

AgustinBettati (approved):
thanks for follow ups

internal/common/autogeneration/marshal.go
```diff
@@ -10,61 +10,130 @@ import (
...
-	return mapFields(src, dest)
+	return unmarshalAttrs(objJSON, model)
+}
+
+func marshalAttrs(valModel reflect.Value, isUpdate bool) (map[string]any, error) {
+	objJSON := make(map[string]any)
+	for i := 0; i < valModel.NumField(); i++ {
+		attrTypeModel := valModel.Type().Field(i)
+		tag := attrTypeModel.Tag.Get(tagKey)
+		if tag == tagValOmitJSON {
```
EspenAlbert, line 55: @lantoli 
Would it be worth supporting multiple tags: "autogeneration:omitjson,{futureTag}"? Probably, it can happen at a later stage

lantoli: @EspenAlbert yes, it makes sense, but didn't want to anticipate until we don't need it,  this code would be very easy to adapt when neeeded

PR #3250, opened 2025-04-04
chore: Updates `mongodbatlas_cluster` to consume selected processArgs fields from the createCluster/updateCluster APIs

maastha:
## Description

Updates `mongodbatlas_cluster` to consume selected processArgs fields from the createCluster/updateCluster APIs.

Link to any related issue(s): [CLOUDP-296222](https://jira.mongodb.org/browse/CLOUDP-296222)

## Type of change:

## Required Checklist:


go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/wI2L/jsondiff v0.6.1
 	github.com/zclconf/go-cty v1.16.2
-	go.mongodb.org/atlas v0.37.0
+	go.mongodb.org/atlas v0.37.1-0.20250402194754-dd09936138bc
```
maastha, line 34: will be replaced with new release version before merge

maastha:
Successful QA test run: https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/14337702520

internal/common/conversion/pointer.go
```diff
@@ -17,3 +17,10 @@ func StringPtr(v string) *string {
 	}
 	return nil
 }
+
+func SliceFromPtr[T any](slicePtr *[]T) []T {
```
lantoli, line 21: I wonder if we have this need elsewhere and we could use there this func as well,  or we're already using another approach/func and we can use that instead of creating this func

internal/service/cluster/model_cluster.go
```diff
@@ -228,6 +240,29 @@ func expandTagSliceFromSetSchema(d *schema.ResourceData) []*matlas.Tag {
 	return res
 }
 
+func expandClusterAdvancedConfiguration(d *schema.ResourceData) *matlas.AdvancedConfiguration {
+	ac := d.Get("advanced_configuration")
+	if aclist, ok1 := ac.([]any); ok1 && len(aclist) > 0 {
```
lantoli, line 245: nit: just need check len(aclist)   as it won't have elements if there is an error, so you wouldn't need ok1

internal/service/cluster/resource_cluster.go
```diff
@@ -508,6 +508,7 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		AutoScaling:              autoScaling,
 		ProviderSettings:         providerSettings,
 		ReplicationSpecs:         replicationSpecs,
```
lantoli, line 510: there are not test changes, can we add/update tests what would fail before this change but will pass now?

lantoli (approved):

go.mod
lantoli: recommend to add a comment so we don't forget

lantoli (approved):

internal/service/cluster/resource_cluster.go
maastha: since these are backend changes (i.e. we're only fetching from a different API, no API behavior changes) I'm not sure how we can test the before/after of this, I think as long as expected attributes are set correctly that should suffice. We already have a decent coverage for advanced_configuration for this. 
Let me know if you can think of specific scenarios we can capture

internal/common/conversion/pointer.go
maastha: I couldn't find anything similar already present. I think that's because usually we use SDK Get methods that does this for us. But since cluster resource still uses manual SDK, I need to handle with this new function

PR #3256, opened 2025-04-07
chore: Support processing PATCH request properties and inferring OmitUpdateBody

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-310769

This PRs includes logic for parsing PATCH request properties. Up to now we were only relying on POST and GET request/responses + GET path params.

Having PATCH properties available, we can now mark attributes which are only present in POST request but not in PATCH (present in model as `OmitUpdateBody` instead of `AllRequestBodies`).

Refactoring was made in `mergeAttributes`, `mergeNestedAttributes`, and `addOrUpdate` unifying duplicate logic and simplifying amount of parameters passed.

**Testing:** A unit test was added to verifying correcting handling of a resource which has a property defined in POST request and then used as path param in GET/PATCH/DELETE (common pattern in our spec). As an example, running generation of Custom DB Role locally `role_name` was correctly marked as `omitjsonupdate`. For the case of Database User no changes are encountered because POST and PATCH have the same request body.

## Type of change:

## Required Checklist:


internal/service/databaseuserapi/resource_schema.go
```diff
@@ -57,16 +57,16 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
 					Attributes: map[string]schema.Attribute{
 						"href": schema.StringAttribute{
-							Optional:            true,
+							Computed:            true,
 							MarkdownDescription: "Uniform Resource Locator (URL) that points another API resource to which this response has some relationship. This URL often begins with `https://cloud.mongodb.com/api/atlas`.",
 						},
 						"rel": schema.StringAttribute{
-							Optional:            true,
+							Computed:            true,
 							MarkdownDescription: "Uniform Resource Locator (URL) that defines the semantic relationship between this resource and another API resource. This URL often begins with `https://cloud.mongodb.com/api/atlas`.",
```
AgustinBettati, line 70: result of pass by fix to ensure request properties with `readOnly: true` are not processed

tools/codegen/codespec/attribute.go
```diff
@@ -21,7 +21,12 @@ func buildResourceAttrs(s *APISpecSchema) (Attributes, error) {
 			return nil, err
 		}
 
-		attribute, err := schema.buildResourceAttr(name, s.GetComputability(name))
+		// ignores properties defined in request which are defined with readOnly (common in Atlas API Spec)
+		if schema.Schema.ReadOnly != nil && *schema.Schema.ReadOnly && isFromRequest {
+			continue
+		}
```
AgustinBettati, line 27: this was a pass by fix

tools/codegen/codespec/merge_attributes.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // mergeNestedAttributes recursively merges nested attributes
-func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, computability ComputedOptionalRequired, reqBodyUsage AttributeReqBodyUsage, isFromResponse bool) {
```
AgustinBettati, line 8: relying on parent attribute computability was causing errors once processing patch request properties, was able to remove dependency on this argument completely

tools/codegen/codespec/testdata/config.yml
```diff
@@ -0,0 +1,44 @@
+resources:
```
AgustinBettati, line 1: unified small configs into single default file in testing

tools/codegen/codespec/merge_attributes.go
```diff
@@ -17,10 +17,15 @@ func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, compu
 	for i := range newAttrs {
 		newAttr := &newAttrs[i]
 
-		if _, exists := mergedMap[newAttr.Name.SnakeCase()]; exists {
-			addOrUpdate(newAttr, computability, reqBodyUsage, mergedMap, isFromResponse)
+		if existing, exists := mergedMap[newAttr.Name.SnakeCase()]; exists {
```
lantoli: nit existing and exists are a bit confusing, what about e.g. curAttr, found := ... (cur for current, or oldAttr, etc.)

tools/codegen/codespec/merge_attributes.go
```diff
@@ -35,87 +40,106 @@ func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, compu
...
+	if first == Required || second == Required {
+		return Required
+	}
+	if first == ComputedOptional || second == ComputedOptional {
+		return ComputedOptional
+	}
+	if first == Optional || second == Optional {
+		return Optional
+	}
+	return Computed
```
lantoli, line 42: both are computed at this point, right?

tools/codegen/codespec/merge_attributes.go
```diff
@@ -35,87 +40,106 @@ func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, compu
...
+}
+
 // addOrUpdate adds or updates an attribute in the merged map, including nested attributes
-func addOrUpdate(attr *Attribute, computability ComputedOptionalRequired, reqBodyUsage AttributeReqBodyUsage, merged map[string]*Attribute, isFromResponse bool) {
-	if existingAttr, exists := merged[attr.Name.SnakeCase()]; exists {
+func addOrUpdate(newAttr *Attribute, targetComputability ComputedOptionalRequired, reqBodyUsage AttributeReqBodyUsage, merged map[string]*Attribute, isFromResponse bool) {
```
lantoli: consider refactoring this func into smaller ones

tools/codegen/codespec/merge_attributes.go
```diff
@@ -35,87 +40,106 @@ func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, compu
...
 
+type attributeDefinitionSources struct {
+	createPathParams, createRequest, updateRequest, createResponse, readResponse Attributes
+}
+
 // mergeAttributes merges attributes from different sources (path params, create/get operation bodies) and determines a single merged list of attributes.
 // Computability and reqBodyUsage values are determined as part of this process.
 // Different sources are applied in a specific order, defining the computability and reqBodyUsage value they have at each step.
-func mergeAttributes(createPathParams, createRequest, createResponse, readResponse Attributes) Attributes {
+func mergeAttributes(sources *attributeDefinitionSources) Attributes {
```
lantoli, line 104: nice to have a struct request here

lantoli (approved):

tools/codegen/codespec/merge_attributes.go
AgustinBettati: correct

tools/codegen/codespec/merge_attributes.go
AgustinBettati: Adjusted to `existingAttr, found`

tools/codegen/codespec/merge_attributes.go
AgustinBettati: Thanks, extracted a separate function `updateAttrWithNewSource` and was able to greatly simplify `mergeNestedAttributes` when revisiting duplicate logic.

tools/codegen/codespec/merge_attributes.go
```diff
@@ -35,88 +28,105 @@ func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, compu
...
+		addOrUpdate(merged, &sources.updateRequest[i], AllRequestBodies, false)
 	}
 
 	// POST/GET response body: properties not in the request body are "computed" or "computed_optional" (if a default is present), reqBodyUsage will have OmitAll not present in request body
-	for i := range createResponse {
-		if hasDefault(&createResponse[i]) {
-			addOrUpdate(&createResponse[i], ComputedOptional, OmitAll, merged, true)
-		} else {
-			addOrUpdate(&createResponse[i], Computed, OmitAll, merged, true)
-		}
```
AgustinBettati, line 111: when properties are parsed in request/responses of create/update computability is defined as Optional or ComputedOptional depending if a default value is present, this logic was redundant

internal/service/databaseuserapi/resource_schema.go
maastha: nice

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -408,6 +409,54 @@ func TestConvertToProviderSpec_nested_schemaOverrides(t *testing.T) {
...
+							ComputedOptionalRequired: codespec.Required,
+							String:                   &codespec.StringAttribute{},
+							Description:              conversion.StringPtr(testPathParamDesc),
+							ReqBodyUsage:             codespec.OmitAll,
+						},
+						{
+							Name:                     "special_param",
+							ComputedOptionalRequired: codespec.Required,
+							String:                   &codespec.StringAttribute{},
+							ReqBodyUsage:             codespec.OmitUpdateBody,
```
maastha, line 434: if an attribute cannot be updated, I think the schema should have RequiresReplace plan modifier right?
Just checking if we eventually plan to do that?

tools/codegen/codespec/attribute.go
maastha: so essentially this use-case happens if API teams for example end up using same views for request & response right?
Thinking if this is a common "pattern" we can always rely on? I'm only referring to using `isFromRequest`, agree with using ReadOnly

maastha (approved):
left some questions, overall LGTM, nice refactoring!

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
AgustinBettati: Correct, adding plan modifier would be a fairly simple change on the code generation side. There is a spike that will cover this more in detail (WRITING-29828 how we want to handle non-updatable fields), for now will leave as is and we can make a final call there.

AgustinBettati:
> this use-case happens if API teams for example end up using same views for request & response right?

@maastha correct, was seen already in both resources being generated. I suspect it can be common because API docs rendering does also filter them out when showing request details.

tools/codegen/codespec/attribute.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/pb33f/libopenapi/orderedmap"
 )
 
-func buildResourceAttrs(s *APISpecSchema) (Attributes, error) {
+func buildResourceAttrs(s *APISpecSchema, isFromRequest bool) (Attributes, error) {
```
EspenAlbert, line 11: @AgustinBettati 
Not sure about `isFromRequest`. Maybe we can use `attrSource` instead. Although it is likely only `request` or `path` are valid values (maybe query?), I think it might be more readable. Then use a comparison instead:
`&& attrSource == const.request`

tools/codegen/codespec/merge_attributes.go
```diff
@@ -35,88 +28,105 @@ func mergeNestedAttributes(existingAttrs *Attributes, newAttrs Attributes, compu
...
-		merged[attr.Name.SnakeCase()] = &newAttr
+	if existingAttr.MapNested != nil && newAttr.MapNested != nil {
+		mergeNestedAttributes(&existingAttr.MapNested.NestedObject.Attributes, newAttr.MapNested.NestedObject.Attributes, reqBodyUsage, isFromResponse)
+	} else if newAttr.MapNested != nil {
+		existingAttr.MapNested = newAttr.MapNested
 	}
 }
 
+type attributeDefinitionSources struct {
+	createPathParams, createRequest, updateRequest, createResponse, readResponse Attributes
```
EspenAlbert, line 98: Curious why `updateResponse`, `updatePathParams` are not here?

AgustinBettati: Not a strong reason, processing both `updateResponse` and `updatePathParams` will likely always be redundant given current sources being processed. We can keep it simple for now but adding them should not be hard.

tools/codegen/codespec/attribute.go
AgustinBettati: `buildResourceAttrs` is only used for building attributes coming from request or response schemas, parsing of path/query params is handled separately (reference `pathParamsToAttributes`). With this in mind the function only has 2 sources so I think a bool keeps it clear.

PR #3261, opened 2025-04-08
chore: Support nested attributes in conversion logic for generating Resource Model from API response

lantoli:
## Description

Support nested attributes in conversion logic for generating Resource Model from API response. It includes ListNest, SetNested and SingleNested. MapNested is not supported as it's not needed at the moment.

Link to any related issue(s): CLOUDP-311710

## Type of change:

## Required Checklist:


internal/common/autogeneration/marshal.go
```diff
@@ -38,7 +39,10 @@ func Marshal(model any, isUpdate bool) ([]byte, error) {
 }
 
 // Unmarshal gets a JSON (e.g. from an Atlas response) and unmarshals it into a Terraform model.
-// It supports the following Terraform model types: String, Bool, Int64, Float64.
+// It supports the following Terraform model types: String, Bool, Int64, Float64, Object, List, Set.
+// Map is not supported yet as it is not used in the Atlas API.
```
AgustinBettati: Would note that there are some definitions of maps in our API Spec. These are defined with the `type: object` + `additionalProperties`, one example is `headers` in [https streams connection](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/getStreamConnection). You can also reference an example we have in our [unit testing of api spec parsing](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/5a6037e3b884b920790d7a9ff07952255aea8cd7/tools/codegen/codespec/testdata/api-spec.yml#L454).
No concern in leaving as limitation for now but would adjust the comment to avoid confusion. We can create a follow up ticket to keep track of the work and revisit priority later as we progress.

internal/common/autogeneration/marshal.go
```diff
@@ -38,7 +39,10 @@ func Marshal(model any, isUpdate bool) ([]byte, error) {
...
+// Attributes that are in JSON but not in the model are ignored, no error is returned.
+// Object attributes that are unknown are converted to null as all values must be known in the response state.
 func Unmarshal(raw []byte, model any) error {
```
AgustinBettati: nit: Should we move `Unmarshal` function and testing to separate unmarshal files?

internal/common/autogeneration/marshal_test.go
```diff
@@ -376,25 +607,22 @@ func TestUnmarshalUnsupportedModel(t *testing.T) {
...
 		errorStr     string
 	}{
-		"JSON objects not support yet": {
-			responseJSON: `{"attr": {"key": "value"}}`,
-			errorStr:     "unmarshal not supported yet for type map[string]interface {} for field attr",
-		},
-		"JSON arrays not supported yet": {
-			responseJSON: `{"attr": [{"key": "value"}]}`,
-			errorStr:     "unmarshal not supported yet for type []interface {} for field attr",
+		"JSON maps not supported yet": {
```
AgustinBettati: nice that we capture current limitation with a test

internal/common/autogeneration/marshal.go
```diff
@@ -164,26 +173,211 @@ func unmarshalAttr(attrNameJSON string, attrObjJSON any, valModel reflect.Value)
...
+	}
+	listNew, diags := types.ListValue(elmType, elms)
+	if diags.HasError() {
+		return nil, fmt.Errorf("unmarshal failed to convert list to object: %v", diags)
+	}
+	return listNew, nil
+}
+
+func setSetAttrModel(set types.Set, arrayJSON []any) (attr.Value, error) {
+	elmType := set.ElementType(context.Background())
```
AgustinBettati: I see usage of `context.Background()` for some functions related to unmarshal, do we have any risk? Would it be better to pass the ctx coming from original terraform operation?

internal/common/autogeneration/marshal.go
```diff
@@ -164,26 +173,211 @@ func unmarshalAttr(attrNameJSON string, attrObjJSON any, valModel reflect.Value)
...
+		return types.Float64Null(), nil
+	default:
+		if objType, ok := attrType.(types.ObjectType); ok {
+			return types.ObjectNull(objType.AttributeTypes()), nil
+		}
+		return nil, fmt.Errorf("unmarshal to get null value not supported yet for type %T", attrType)
+	}
+}
+
+func convertUnknownToNull(valModel reflect.Value) {
```
AgustinBettati: should this logic be recursive?

internal/common/autogeneration/marshal.go
```diff
@@ -164,26 +173,211 @@ func unmarshalAttr(attrNameJSON string, attrObjJSON any, valModel reflect.Value)
...
+			return nil, err
+		}
+		return objNew, nil
+	case nil:
+		return nil, nil // skip nil values, no need to set anything
+	}
+	return nil, fmt.Errorf("unmarshal not supported yet for type %T", value)
+}
+
+func setObjAttrModel(obj types.Object, objJSON map[string]any) (attr.Value, error) {
```
AgustinBettati: q: Recursive logic of nested objects is done through this function right?

AgustinBettati (approved):
LGTM, leaving some questions and suggestions

internal/common/autogeneration/marshal.go
lantoli: comment changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3261/commits/c0631b02c78ca01c8fec0fe66ab279106eec55e6. ticket created: CLOUDP-312797

internal/common/autogeneration/marshal.go
```diff
@@ -164,26 +173,211 @@ func unmarshalAttr(attrNameJSON string, attrObjJSON any, valModel reflect.Value)
 	}
```
EspenAlbert: Is it worth it to log a debug message?

internal/common/autogeneration/marshal.go
lantoli: changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3261/commits/942e99bb18bace4e484bd2c0cc8f8c4bc14813a1

internal/common/autogeneration/marshal.go
lantoli: not sure, in general I prefer not to log debug messages, I don't know if this case is so important to be logged

internal/common/autogeneration/marshal.go
lantoli: I don't think there is any risk as as we don't need context specific features here such as cancel or timeouts.  We might accept in Marshal/Unmarshal the operation context and pass it to all functions, but I think this will make the code less legible and won't give any real benefit

internal/common/autogeneration/marshal.go
lantoli: correct, getTfAttr calls setObjAttrModel which calls setMapAttrModel which calls getTfAttr again,  so objects are handled recursively in that func

internal/common/autogeneration/unmarshal_test.go
```diff
@@ -0,0 +1,352 @@
...
+		AttrObjNullSent:       types.ObjectNull(objTypeTest.AttrTypes),
+		AttrObjUnknownNotSent: types.ObjectUnknown(objTypeTest.AttrTypes), // unknown values are changed to null
+		AttrObjUnknownSent:    types.ObjectUnknown(objTypeTest.AttrTypes),
+		AttrObjParent:         types.ObjectNull(objTypeParentTest.AttrTypes),
+		AttrListString:        types.ListUnknown(types.StringType),
+		AttrListObj:           types.ListUnknown(objTypeTest),
+		AttrSetString:         types.SetUnknown(types.StringType),
+		AttrSetObj:            types.SetUnknown(objTypeTest),
+	}
+	// attrUnexisting is ignored because it is in JSON but not in the model, no error is returned
```
EspenAlbert, line 87: nice with these comments

internal/common/autogeneration/unmarshal.go
```diff
@@ -0,0 +1,260 @@
...
+			return types.ObjectNull(objType.AttributeTypes()), nil
+		}
+		return nil, fmt.Errorf("unmarshal to get null value not supported yet for type %T", attrType)
+	}
+}
+
+func convertUnknownToNull(valModel reflect.Value) {
+	for i := 0; i < valModel.NumField(); i++ {
+		field := valModel.Field(i)
+		if obj, ok := field.Interface().(types.Object); ok {
```
EspenAlbert: This would only work for types.Object? What about other `types.XXX`?

internal/common/autogeneration/unmarshal.go
```diff
@@ -0,0 +1,260 @@
...
+			if obj.IsUnknown() && field.CanSet() {
+				field.Set(reflect.ValueOf(types.ObjectNull(obj.AttributeTypes(context.Background()))))
```
EspenAlbert: Maybe some debug warning?

internal/common/autogeneration/unmarshal.go
```diff
@@ -0,0 +1,260 @@
...
+	valModel = valModel.Elem()
+	if valModel.Kind() != reflect.Struct {
+		panic("model must be pointer to struct")
+	}
+	for attrNameJSON, attrObjJSON := range objJSON {
+		if err := unmarshalAttr(attrNameJSON, attrObjJSON, valModel); err != nil {
+			return err
+		}
+	}
+	convertUnknownToNull(valModel)
```
EspenAlbert: I think this would be better to do in a different step?
Then it would be easier to override and possible use values from state instead?

lantoli: @EspenAlbert  do you mean not doing it in Unmarshal, but in another public func?

internal/common/autogeneration/unmarshal.go
lantoli: you're right, creating ticket created for to null logic: CLOUDP-312809

internal/common/autogeneration/unmarshal.go
lantoli: won't do unknown to null values in this PR, will be in: CLOUDP-312809

internal/common/autogeneration/unmarshal.go
lantoli: @EspenAlbert  ticket created for to null logic: CLOUDP-312809 cc @AgustinBettati

internal/common/autogeneration/marshal.go
AgustinBettati: thanks!

internal/common/autogeneration/unmarshal.go
EspenAlbert: >  but in another public func

Yes, I believe so. Sounds good to me with a different ticket 👍

EspenAlbert (approved):
Nice. Thanks for addressing the comments

internal/common/autogeneration/marshal.go
lantoli: unknown to null logic moved to another PR

PR #3258, opened 2025-04-08
chore: Add code generation for update operation

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-311034

Update operation is identical to create with the difference of passing boolean to marshal function and handling correct PATCH API call (URL, path params, http method).

Unifying into a single function does have complexity due to different parameter types (resource.CreateRequest vs resource.UpdateRequest), and current code is fairly concise.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals I have added appropriate changelog entries.
- [ ] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

PR #3244, opened 2025-04-02
chore: Adjust model generation to include tags relevant for conversion logic of marshalling

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-310081

### code generation changes

Depending on values defined in intermediate model of each attribute (`ReqBodyUsage  AttributeReqBodyUsage`) code generation supporting adding `autogeneration:"omitjson"` or `autogeneration:"omitjsonupdate"` to TF models.

### API specs parsing/model generation changes

Adjust merge attribute logic to define either `AllRequestBodies` or `OmitAll` value for `ReqBodyUsage` of each attribute. This is currently inferred by checking if attribute was defined in the POST request body. 

Follow up ticket CLOUDP-310769 has been created for inferring `OmitUpdateBody`, this will require parsing PATCH request properties which is currently not being done.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals I have added appropriate changelog entries.
- [ ] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

PR #3265, opened 2025-04-09
chore: Support import operation in generated resources

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-311036

Import functionality is supported through a single generic function. When invoking the function each resource specifies which attributes compose the import id. These id attributes are obtained from path parameters of the GET operation.

There are some small refactors addressing comments from https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3258#discussion_r2034697421 and https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3244#discussion_r2028607015.

### Complementary manual testing
Adjusted existing database_user resource to use the generic function:
```
func (r *databaseUserRS) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
	idAttributes := []string{"project_id", "username", "auth_database_name"}
	autogeneration.GenericImportOperation(ctx, idAttributes, req, resp)
}
```

With this I was able to run a successful import:
```
terraform import mongodbatlas_database_user.user "650972848269185c55f40ca1/new-user-import/admin"
```

## Type of change:

## Required Checklist:


internal/common/autogeneration/import_test.go
```diff
@@ -0,0 +1,57 @@
+package autogeneration //nolint:testpackage // reason: do not want to expose processImportID but is useful for unit testing
```
AgustinBettati: Not a strong opinion on this, another option is to leave `processImportID` publicly exposed. 
When trying to test `GenericImportOperation` directly I did encountered more limitations and complexity due to tpf types that have to be constructed, this is why I opted for defining `processImportID` with simple function signature.

internal/service/pushbasedlogexportapi/resource.go
```diff
@@ -44,7 +48,7 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 
 	reqBody, err := autogeneration.Marshal(&plan, false)
 	if err != nil {
-		resp.Diagnostics.AddError("error during create operation", err.Error())
+		resp.Diagnostics.AddError(errorConstructingAPIRequest, err.Error())
```
lantoli: nit: build instead of construct

tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl
```diff
@@ -187,5 +191,6 @@ func (r *rs) Delete(ctx context.Context, req resource.DeleteRequest, resp *resou
 }
 
 func (r *rs) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
-	// TODO: code generation logic for import will be handled in milestone 2
+	idAttributes := []string{ {{range .IDAttributes }}"{{ . }}", {{- end }}}
```
lantoli: nit: an alternative is to use varargs instead of string array for attr names, but it has to be the last param, e.g.: autogeneration.GenericImportOperation(ctx, req, resp, "group_id", "cluster_name")

internal/service/databaseuserapi/resource.go
```diff
@@ -66,13 +70,13 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 
 	respBody, err := io.ReadAll(apiResp.Body)
```
lantoli, line 72:
```suggestion
	defer apiResp.Body.Close()
	respBody, err := io.ReadAll(apiResp.Body)
```

we need to do this before each call to `io.ReadAll`, feel free to do it in this PR or a follow-up PR, e.g. as in the SDK [here](https://github.com/mongodb/atlas-sdk-go/blob/2c1a8a0a1658ebb7bd8ef21a3b62cfd59658563d/admin/api_database_users.go#L247)

lantoli (approved):

internal/common/autogeneration/import.go
```diff
@@ -0,0 +1,40 @@
...
+//   - GenericImportOperation(ctx, []string{"project_id", "name"}, req, resp)
+//   - example import ID would be "5c9d0a239ccf643e6a35ddasdf/myCluster"
+func GenericImportOperation(ctx context.Context, idAttributes []string, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	attrValues, err := processImportID(req.ID, idAttributes)
+	if err != nil {
+		resp.Diagnostics.AddError("Error processing import ID", err.Error())
+		return
+	}
+	for attr, value := range attrValues {
+		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(attr), value)...)
```
lantoli:
```suggestion
func GenericImportOperation(ctx context.Context, attrIDs []string, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
	attrs, err := processImportID(req.ID, attrIDs)
	if err != nil {
		resp.Diagnostics.AddError("Error processing import ID", err.Error())
		return
	}
	for name, value := range attrs {
		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(name), value)...)
```

nit about naming

internal/common/autogeneration/import.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+func processImportID(importID string, idAttributes []string) (map[string]string, error) {
+	parts := strings.Split(importID, "/")
+	if len(parts) != len(idAttributes) {
+		return nil, errors.New("Expected format: " + strings.Join(idAttributes, "/"))
```
maastha: nit: make this a const so can be used in tests as is

internal/common/autogeneration/import.go
lantoli: i guess all the id attributes are strings so we never need to convert it to other types when calling State.SetAttribute

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -151,7 +151,7 @@ func TestConvertToProviderSpec_nested(t *testing.T) {
 								ElementType: codespec.String,
 							},
 							Description:  conversion.StringPtr(testFieldDesc),
-							ReqBodyUsage: codespec.OmitAll,
+							ReqBodyUsage: codespec.OmitAlways,
```
maastha, line 154: Ty!

internal/common/autogeneration/import_test.go
lantoli: I prefer to avoid linter exceptions if possible, in this case I'd have it exported and keep autogeneration_test package for tests.   You can add a comment to the func like [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/39c51132bc9810d8b301803305c44ec7aa0f0289/internal/common/conversion/schema_generation.go#L179)

maastha (approved):

internal/common/autogeneration/import_test.go
AgustinBettati: thanks, will go for this approach

tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl
AgustinBettati: Doing some searching variadic are usually used for improved readability/syntax sugar when calling the function, in this case I feel passing an array explicitly is more clear

internal/service/databaseuserapi/resource.go
AgustinBettati: thanks, following example from [go docs](https://pkg.go.dev/net/http#example-Get) calling close directly after the `ReadAll`

internal/service/databaseuserapi/resource.go
```diff
@@ -65,14 +69,15 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 	}
 
 	respBody, err := io.ReadAll(apiResp.Body)
+	apiResp.Body.Close()
 	if err != nil {
```
EspenAlbert, line 74: [nit] Why not separate the `errorProcessingAPIResponse` errors?

tools/codegen/gofilegen/codetemplate/template.go
```diff
@@ -23,6 +23,7 @@ type ResourceFileInputs struct {
 	PackageName   string
 	ResourceName  string
 	APIOperations APIOperations
+	IDAttributes  []string // e.g. ["project_id", "name"]
```
EspenAlbert: [nit] ImportIDAttributes? Or are they used in other places?

EspenAlbert (approved):
Nice! Only a few nit comments.

internal/common/autogeneration/import.go
AgustinBettati: thanks, revisited and took some suggestions

AgustinBettati: correct, have not encountered any path params in api spec that are not strings

tools/codegen/gofilegen/codetemplate/template.go
AgustinBettati: only import, agree with the suggestion 👍

internal/service/databaseuserapi/resource.go
AgustinBettati: agree, added separate `errorReadingAPIResponse` in case the read fails

PR #3269, opened 2025-04-10
refactor: Supports cleaning resources for static projects with `test-acc-tf-p-keep` prefix

EspenAlbert:
## Description

Related issue: https://jira.mongodb.org/browse/CLOUDP-314965

Supports cleaning projects with `"test-acc-tf-p-keep"` prefix, that shouldn't be deleted but have resources removed when running the cleanup (the project should be empty after tests).

We have various tests that depend on using specific projects with feature flags enabled.
However, these projects risk having extra resources that are not cleaned.

The only resources found for deletion:
```
test-acc-tf-p-keep-ear-AWS-private-endpoint (6790e57a9b41416f5c216fee) # 3 stream instances
test-acc-tf-p-keep-ear-private-endpoint (66d83bcc1fe1835125c52422) # no resources
test-acc-tf-p-keep-encryption-at-rest (67810c93a7dcf40b5f0db4c2) # no resources
```

## Sweeper Alternative considered
tl;dr: Uses the same mechanism as `go test` but with a `-sweep={regionName}` flag. I consider our current cleanup test to be easier to adapt than adding custom sweepers and a new CI workflow. 

- Docs: https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/sweepers
- AWS Example [invoke](https://github.com/hashicorp/terraform-provider-aws/blob/main/GNUmakefile#L585) and [code](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/accessanalyzer/sweep.go#L24)
- Advantage over `CheckDestroy`: If a test panics or the process terminates unexpectedly, checkDestroy never runs.
- CI Integration: Sweepers are designed to be run in CI pipelines on schedules, independent of test runs.
- The sweeper function must accept a region string parameter and return an error

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/testutil/clean/org_clean_test.go
+++ b/internal/testutil/clean/org_clean_test.go
@@ -84,7 +84,7 @@ func TestCleanProjectAndClusters(t *testing.T) {
 	projects := readAllProjects(t.Context(), t, client)
 	projectsBefore := len(projects)
 	t.Logf("found %d projects (DRY_RUN=%t)", projectsBefore, dryRun)
-	projectsToDelete := map[string]string{}
+	projectsToClean := map[string]string{}
 	projectInfos := []string{}
 	for _, p := range projects {
 		skipReason := projectSkipReason(&p, skipProjectsAfter, onlyZeroClusters)
@@ -95,14 +95,14 @@ func TestCleanProjectAndClusters(t *testing.T) {
 			continue
 		}
 		projectInfos = append(projectInfos, fmt.Sprintf("Project created at %s name %s (%s)", p.GetCreated().Format(time.RFC3339), projectName, p.GetId()))
-		projectsToDelete[projectName] = projectID
+		projectsToClean[projectName] = projectID
 	}
-	t.Logf("will try to delete %d projects:", len(projectsToDelete))
+	t.Logf("deleting project resources and optionally delete projects for %d projects", len(projectsToClean))
 	slices.Sort(projectInfos)
 	t.Log(strings.Join(projectInfos, "\n"))
 	var deleteErrors int
 	var emptyProjectCount int
-	for name, projectID := range projectsToDelete {
+	for name, projectID := range projectsToClean {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 			changes := removeProjectResources(t.Context(), t, dryRun, client, projectID)
--- a/internal/testutil/clean/org_clean_test.go
+++ b/internal/testutil/clean/org_clean_test.go
@@ -109,8 +109,8 @@ func TestCleanProjectAndClusters(t *testing.T) {
 			if changes != "" {
 				t.Logf("project %s %s", name, changes)
 			}
-			if skipReason := projectNoDeleteReason(name); skipReason != "" {
-				t.Logf("keep project empty, but no delete %s (%s), reason: %s", name, projectID, skipReason)
+			if skipProjectDelete(name) {
+				t.Logf("keep project empty, but no delete %s (%s)", name, projectID)
 				emptyProjectCount++
 				return
 			}
@@ -230,13 +230,13 @@ func projectSkipReason(p *admin.Group, skipProjectsAfter time.Time, onlyEmpty bo
 	return ""
 }
 
-func projectNoDeleteReason(name string) string {
+func skipProjectDelete(name string) bool {
 	for _, keepPrefix := range keptPrefixes {
 		if strings.HasPrefix(name, keepPrefix) {
-			return "keep project but not resources"
+			return true
 		}
 	}
-	return ""
+	return false
 }
 
 func removeClusters(ctx context.Context, t *testing.T, dryRun bool, client *admin.APIClient, projectID string) int {
--- a/internal/testutil/clean/org_clean_test.go
+++ b/internal/testutil/clean/org_clean_test.go
@@ -34,11 +34,10 @@ var (
 		"cfn-test-bot-",
 		"test-acc-tf-p-",
 	}
+	// These are projects gets their resources removed but the project itself is kept
+	// Useful when a feature flag or cloud provider is configured outside of the test
 	keptPrefixes = []string{
 		"test-acc-tf-p-keep",
-	}
-	keepProjectEmptyPrefixes = []string{
-		"test-acc-tf-p-empty",
 	}
 	projectRetryDeleteErrors = []string{
 		"CANNOT_CLOSE_GROUP_ACTIVE_ATLAS_CLUSTERS",
@@ -110,8 +109,7 @@ func TestCleanProjectAndClusters(t *testing.T) {
 			if changes != "" {
 				t.Logf("project %s %s", name, changes)
 			}
-			skipReason := projectNoDeleteReason(name)
-			if skipReason != "" {
+			if skipReason := projectNoDeleteReason(name); skipReason != "" {
 				t.Logf("keep project empty, but no delete %s (%s), reason: %s", name, projectID, skipReason)
 				emptyProjectCount++
 				return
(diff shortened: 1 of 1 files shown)


internal/testutil/clean/org_clean_test.go
```diff
@@ -37,6 +37,9 @@ var (
 	keptPrefixes = []string{
 		"test-acc-tf-p-keep",
 	}
+	keepProjectEmptyPrefixes = []string{
+		"test-acc-tf-p-empty",
```
lantoli: is there any case where we want to keep the project without deleting anything in it?  are they with prefix test-acc-tf-p-keep?     also I can imagine that in keep ones we want to keep some type of resource (e.g. private endpoint) but delete others (like clusters).     so not sure if it will be more general to use only one prefix project, e.g. test-acc-tf-p-keep, and then use "keep" in resources we want to keep inside, e.g. cluster "test-acc-tf-c-keep-mycluster would me not to delete that cluster.

internal/testutil/clean/org_clean_test.go
```diff
@@ -87,25 +90,32 @@ func TestCleanProjectAndClusters(t *testing.T) {
...
+	var emptyProjectCount int
 	for name, projectID := range projectsToDelete {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 			changes := removeProjectResources(t.Context(), t, dryRun, client, projectID)
 			if changes != "" {
 				t.Logf("project %s %s", name, changes)
 			}
+			skipReason := projectNoDeleteReason(name)
+			if skipReason != "" {
```
lantoli:
```suggestion
			if skipReason := projectNoDeleteReason(name); skipReason != "" {
```

nit

internal/testutil/clean/org_clean_test.go
AgustinBettati: Agree we can preserve `test-acc-tf-p-keep` prefix for all cases. 
Not sure how complex it would be to start adding different `keep` prefixes for each resource type, as for this PR maybe we can stick to `test-acc-tf-p-keep` (preserves any resources) and `test-acc-tf-p-keep-empty` (removes all resources)?

internal/testutil/clean/org_clean_test.go
```diff
@@ -37,6 +37,9 @@ var (
 	keptPrefixes = []string{
 		"test-acc-tf-p-keep",
 	}
+	keepProjectEmptyPrefixes = []string{
```
AgustinBettati: changes LGTM, would only double check that for the following projects we don't have any static resources we are relying on that could be deleted. (I dont believe its the case)
EAR-example (6792ac6ceaa5601d0f45dc95)
test-acc-tf-p-keep-ear-AWS-private-endpoint (6790e57a9b41416f5c216fee)
test-acc-tf-p-keep-ear-private-endpoint (66d83bcc1fe1835125c52422)
test-acc-tf-p-keep-encryption-at-rest (67810c93a7dcf40b5f0db4c2)

internal/testutil/clean/org_clean_test.go
lantoli: @AgustinBettati FYI, see also comments in the doc

internal/testutil/clean/org_clean_test.go
EspenAlbert: Thanks. Changed in 819a31dcd2ed0c25dd689b3fecec86cf43ca27d5

internal/testutil/clean/org_clean_test.go
EspenAlbert: Checked. Only `test-acc-tf-p-keep-ear-AWS-private-endpoint` has a few steam_instances, but those are ok to delete

internal/testutil/clean/org_clean_test.go
EspenAlbert: Changed `test-acc-tf-p-keep` instead

internal/testutil/clean/org_clean_test.go
```diff
@@ -34,6 +34,8 @@ var (
 		"cfn-test-bot-",
 		"test-acc-tf-p-",
 	}
+	// These are projects gets their resources removed but the project itself is kept
```
lantoli:
```suggestion
	// keptPrefixes has the prefix of the projects that we want to delete their resources but keep the projects themselves.
```

nit

internal/testutil/clean/org_clean_test.go
```diff
@@ -87,25 +89,31 @@ func TestCleanProjectAndClusters(t *testing.T) {
...
+	var emptyProjectCount int
 	for name, projectID := range projectsToDelete {
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 			changes := removeProjectResources(t.Context(), t, dryRun, client, projectID)
 			if changes != "" {
 				t.Logf("project %s %s", name, changes)
 			}
+			if skipReason := projectNoDeleteReason(name); skipReason != "" {
+				t.Logf("keep project empty, but no delete %s (%s), reason: %s", name, projectID, skipReason)
```
lantoli: log text and skiReason (keep project but not resources) seem a bit repetitive.  maybe instead of projectNoDeleteReason we can just have a keepProject or similar func that returns just a bool

internal/testutil/clean/org_clean_test.go
```diff
@@ -204,11 +211,6 @@ func removeProjectResources(ctx context.Context, t *testing.T, dryRun bool, clie
 }
 
 func projectSkipReason(p *admin.Group, skipProjectsAfter time.Time, onlyEmpty bool) string {
```
lantoli, line 213: what's the difference between projectSkipReason and projectNoDeleteReason?

internal/testutil/clean/org_clean_test.go
EspenAlbert: Agree. Changed in 47ad1b7e09e308fab9404b4872f1df889c79bef3

internal/testutil/clean/org_clean_test.go
EspenAlbert: projectSkipReason avoids cleaning the project alltogether. (All non bot projects are not cleaned by default)
skipProjectDelete cleans resources but doesn't delete the project.

internal/testutil/clean/org_clean_test.go
```diff
@@ -87,25 +89,31 @@ func TestCleanProjectAndClusters(t *testing.T) {
...
 		projectInfos = append(projectInfos, fmt.Sprintf("Project created at %s name %s (%s)", p.GetCreated().Format(time.RFC3339), projectName, p.GetId()))
-		projectID := p.GetId()
 		projectsToDelete[projectName] = projectID
 	}
 	t.Logf("will try to delete %d projects:", len(projectsToDelete))
 	slices.Sort(projectInfos)
 	t.Log(strings.Join(projectInfos, "\n"))
 	var deleteErrors int
+	var emptyProjectCount int
 	for name, projectID := range projectsToDelete {
```
AgustinBettati: should we adjust the log to mention the current behaviour? "deleting project resources and optionally delete project"
same for variable `projectsToDelete` maybe `projectsToCleanAndDelete` makes it more clear?

AgustinBettati (approved):
LGTM, just a suggestion for improving clarity

internal/testutil/clean/org_clean_test.go
lantoli: thanks for the clarification

lantoli (approved):

internal/testutil/clean/org_clean_test.go
EspenAlbert: Good idea. Changed in 975741bbb77058e5bd348f94274290131cc81edd

PR #3280, opened 2025-04-14
feat: Adds description field to `mongodbatlas_database_user`

EspenAlbert:
## Description

Adds description field to `mongodbatlas_database_user`

Link to any related issue(s): CLOUDP-288713

## Considerations
- Choosing `optional` only attribute
  - Implies extra logic to align config value with API value since `null` != `""` in TPF. (Similar to recent resource_policy implementation)
  - Also, there is no way of setting the description to `null` in the SDK. Once it is set, it will always be `""` since we cannot dump `null` (to be discussed based on this [example](https://gist.github.com/EspenAlbert/e9cd45f80f033336d6482cd9c334d3d2).)
- Will handle test refactoring in a follow-up issue: https://jira.mongodb.org/browse/CLOUDP-312885
- Choosing to keep model_database_user_test.go simple, all paths are covered in acceptance tests.
  - state null plan null
  - state some value plan null
  - state null plan not-null
  - state set and plan set

## Type of change:

## Required Checklist:


internal/service/databaseuser/data_source_database_user_test.go
```diff
@@ -30,6 +30,7 @@ func TestAccConfigDSDatabaseUser_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(dataSourceName, "roles.0.role_name", roleName),
 					resource.TestCheckResourceAttr(dataSourceName, "roles.0.database_name", "admin"),
 					resource.TestCheckResourceAttr(dataSourceName, "labels.#", "2"),
+					resource.TestCheckNoResourceAttr(dataSourceName, "description"),
```
oarbusi, line 33: Why `TestCheckNoResourceAttr`? Shouldn't we check that we set it? (and set it in the resource)

internal/service/databaseuser/model_database_user.go
```diff
@@ -11,42 +11,48 @@ import (
...
 		Scopes:       NewMongoDBAtlasScopes(scopesModel),
 	}
 
-	if statePasswordValue != dbUserModel.Password {
+	if statePasswordValue != plan.Password {
 		// Password value has been modified or no previous state was present. Password is only updated if changed in the terraform configuration CLOUDP-235738
-		result.Password = dbUserModel.Password.ValueStringPointer()
+		result.Password = plan.Password.ValueStringPointer()
+	}
+	if plan.Description.IsNull() && !stateDescriptionValue.Equal(plan.Description) {
```
lantoli, line 52: just to confirm, what happens if we don't have this condition?

wouldn't result.Description be "" in this case because of previous: `Description:  plan.Description.ValueStringPointer(),` ?

internal/service/databaseuser/resource_database_user.go
```diff
@@ -204,28 +208,28 @@ func (r *databaseUserRS) Schema(ctx context.Context, req resource.SchemaRequest,
 }
 
 func (r *databaseUserRS) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
lantoli, line 210: nit: maybe in a follow-up PR we want to unify RS and DS tests in the same test file and tests

internal/service/databaseuser/resource_database_user.go
```diff
@@ -204,28 +208,28 @@ func (r *databaseUserRS) Schema(ctx context.Context, req resource.SchemaRequest,
...
 
-	diags := req.Plan.Get(ctx, &databaseUserPlan)
+	diags := req.Plan.Get(ctx, &plan)
 	resp.Diagnostics.Append(diags...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
-	dbUserReq, d := NewMongoDBDatabaseUser(ctx, types.StringNull(), databaseUserPlan)
+	dbUserReq, d := NewMongoDBDatabaseUser(ctx, types.StringNull(), types.StringNull(), plan)
```
lantoli: nit: use diags or localDiags as in other places

internal/service/databaseuser/resource_database_user.go
```diff
@@ -240,20 +244,20 @@ func (r *databaseUserRS) Create(ctx context.Context, req resource.CreateRequest,
...
-	projectID := databaseUserState.ProjectID.ValueString()
-	authDatabaseName := databaseUserState.AuthDatabaseName.ValueString()
+	username := state.Username.ValueString()
+	projectID := state.ProjectID.ValueString()
+	authDatabaseName := state.AuthDatabaseName.ValueString()
 
 	// Use the ID only with the IMPORT operation
-	if databaseUserState.ID.ValueString() != "" && (username == "" || projectID == "" || authDatabaseName == "") {
-		projectID, username, authDatabaseName, err = SplitDatabaseUserImportID(databaseUserState.ID.ValueString())
+	if state.ID.ValueString() != "" && (username == "" || projectID == "" || authDatabaseName == "") {
```
lantoli: I understand this change is not related to this PR, but wondering why we have a different Read logic when importing, instead of Import setting the attributes we need in Read so we don't need to do a special case in Read for import

lantoli (approved):

internal/service/databaseuser/model_database_user_test.go
```diff
@@ -65,6 +65,7 @@ var (
 		GroupId:      projectID,
 		DatabaseName: authDatabaseName,
 		Username:     username,
+		Description:  conversion.Pointer(""),
```
oarbusi, line 68: I would add a var `emptyDescription = ""` and then here use `&emptyDescription`

internal/service/databaseuser/resource_database_user_test.go
```diff
@@ -203,13 +207,14 @@ func TestAccConfigRSDatabaseUser_withLabels(t *testing.T) {
...
 					resource.TestCheckResourceAttr(resourceName, "username", username),
+					resource.TestCheckResourceAttr(resourceName, "description", description2),
 					resource.TestCheckResourceAttr(resourceName, "password", "test-acc-password"),
 					resource.TestCheckResourceAttr(resourceName, "auth_database_name", "admin"),
 					resource.TestCheckResourceAttr(resourceName, "labels.#", "2"),
 				),
 			},
 			{
-				Config: acc.ConfigDatabaseUserWithLabels(projectID, username, "read",
+				Config: acc.ConfigDatabaseUserWithLabels(projectID, username, "read", "",
```
oarbusi, line 217: Nice test!

oarbusi (approved):
LGTM

marcosuma (approved):

internal/service/databaseuser/data_source_database_user_test.go
EspenAlbert: It is an optional field and should be `null` when it is not set in the config

internal/service/databaseuser/model_database_user.go
EspenAlbert: No, ValueStringPointer would be nil and therefore not dumped.
Cannot use `ValueString()` since we don't want to set it to `""` when it is not defined.

internal/service/databaseuser/model_database_user_test.go
EspenAlbert: Not sure, I feel this is more readable. With an intermediate var you would need to check that var. Here it is all in one line

internal/service/databaseuser/resource_database_user.go
EspenAlbert: See PR description

internal/service/databaseuser/model_database_user.go
```diff
@@ -91,6 +98,10 @@ func NewTfDatabaseUserModel(ctx context.Context, model *TfDatabaseUserModel, dbU
 		// The Password is not retuned from the endpoint so we use the one provided in the model
 		databaseUserModel.Password = model.Password
 	}
+	if dbUser.Description != nil && model != nil && model.Description.IsNull() {
+		// null != "" in TPF:  Error: Provider produced inconsistent result after apply. .description: was null, but now cty.StringVal("")
+		databaseUserModel.Description = model.Description
+	}
```
AgustinBettati: In this case are we removing possibility of drift detection?
Lets say description was modified in Atlas UI, with this read logic description value would preserve state value ignoring value from API.
I believe this can be fixed by checking that dbUser.Description is an empty string as well.

internal/service/databaseuser/resource_database_user.go
EspenAlbert: Wow, yeah you are right. Moved to Import function in e860fa93fcc8c47d3a66a40b3d54f395fa7cb1ba

internal/service/databaseuser/resource_database_user.go
EspenAlbert: Thanks. Changed in 27168d99a9772bc0b3f07fe95b8a612a479250de

AgustinBettati (commented):
one doubt otherwise LGTM

internal/service/databaseuser/model_database_user_test.go
oarbusi: IMO it's more readable with the var but I don't have a strong opinion

internal/service/databaseuser/model_database_user.go
lantoli: thanks

internal/service/databaseuser/resource_database_user.go
lantoli: thanks

internal/service/databaseuser/model_database_user.go
EspenAlbert: Good catch.
This turned out to be harder than I expected.
Changed in 6a6b44f4b9c4afd792ee4643145a0dced0003cdc

Tested manually

internal/service/databaseuser/model_database_user.go
```diff
@@ -94,16 +94,15 @@ func NewTfDatabaseUserModel(ctx context.Context, model *TfDatabaseUserModel, dbU
...
 	}
-	if dbUser.Description != nil && model != nil && model.Description.IsNull() {
+	if inModel != nil && outModel.Description.Equal(types.StringValue("")) && inModel.Description.IsNull() {
 		// null != "" in TPF:  Error: Provider produced inconsistent result after apply. .description: was null, but now cty.StringVal("")
-		databaseUserModel.Description = model.Description
+		outModel.Description = types.StringNull()
 	}
-
-	return databaseUserModel, nil
+	return outModel, nil
```
AgustinBettati, line 105: nice way of keeping it readable 👍

AgustinBettati (approved):
LGTM

kanchana-mongodb (approved):
LGTM

PR #3281, opened 2025-04-15
chore: Convert unknown values to null values

lantoli:
## Description

Convert unknown values to null values

Link to any related issue(s): CLOUDP-312809

## Type of change:

## Required Checklist:


internal/common/autogen/response_model.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// PrepareResponseModel is called before the Terraform operations set the model response.
+// Unknown attributes are converted to null.
+// Empty lists and sets are converted to null to avoid error "inconsistent result after apply".
```
EspenAlbert: Is this safe? What if the user explicitly sets a list or set to `[]`? Wouldn't this create the same error?

lantoli: @EspenAlbert  you're right, what about passing the config, so we can check if it's in the config or not to set null or empty?

lantoli: or only update attributes that are computed and not in the config?  this will avoid inconsistency errors although we might be hiding some issues

internal/common/autogen/response_model.go
```diff
@@ -0,0 +1,126 @@
...
+func PrepareResponseModel(model any) error {
```
AgustinBettati: nit: If we converge to this function having the sole purpose of setting nil to unkown values I would rename the function to make this explicit

internal/common/autogen/response_model.go
EspenAlbert: I like the idea of passing config, as it would be very useful to avoid the inconsistent errors, but maybe some logic so it doesn't override "real" values. For example API returning a non empty list should never be overriden, only null / empty values

internal/common/autogen/response_model.go
lantoli: it was called that before but I changed it in case it's used for more purposes, I'm inclined to keep as it is in case we detect other preparations we need to do, but not strong opinion to change it back if you prefer that

internal/common/autogen/response_model.go
lantoli: @EspenAlbert let me try to change Unmarshal instead of this func, so it doesn't change model if it has no elements and an empty list comes from JSON

lantoli: @EspenAlbert cc @AgustinBettati move to Unmarshal, I think it's better now, please take a look: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3281/commits/5d37660934112b00556233b9fff1d79b5b29aee4

internal/common/autogen/unmarshal_test.go
```diff
@@ -303,6 +303,45 @@ func TestUnmarshalNestedAllTypes(t *testing.T) {
 	assert.Equal(t, modelExpected, model)
 }
 
+func TestUnmarshalZeroLenCollections(t *testing.T) {
```
EspenAlbert, line 329: nice!

EspenAlbert (approved):
LGTM.

internal/common/autogen/response_model.go
lantoli: @AgustinBettati renamed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3281/commits/b4ef52970d3f8bf01164460d31aa93a64d67f3e1

internal/common/autogen/unmarshal.go
```diff
@@ -185,10 +185,13 @@ func setObjAttrModel(obj types.Object, objJSON map[string]any) (attr.Value, erro
 
 func setListAttrModel(list types.List, arrayJSON []any) (attr.Value, error) {
 	elmType := list.ElementType(context.Background())
-	elms, err := getCollectionElements(arrayJSON, elmType)
+	elms, err := getCollectionElements(arrayJSON, elmType, list.Elements())
 	if err != nil {
 		return nil, err
 	}
+	if len(elms) == 0 && len(list.Elements()) == 0 {
+		return list, nil // keep current list if both model and JSON lists are zero-len (empty or null) so config is preserved
```
AgustinBettati: could be good to also capture the why: avoid inconsistent result after apply when user explicilty sets an empty list in config.

internal/common/autogen/unknown.go
```diff
@@ -0,0 +1,118 @@
...
+		valNew, err := prepareAttr(value)
+		if err != nil {
+			return err
+		}
+		field.Set(reflect.ValueOf(valNew))
+	}
+	return nil
+}
+
+func prepareAttr(value attr.Value) (attr.Value, error) {
```
AgustinBettati, line 37: q: Just for thought, is there common logic related to iterating over tf model here compared to unmarshal? Just thinking trade offs of having this logic separated vs together with unmarshall.

AgustinBettati (approved):
LGTM, really like the thorough testing

internal/common/autogen/unknown.go
lantoli: The "navigation" logic is similar but the work done is different so I'm not sure we can extract some common code.  And I don't think there is a big benefit of having them together, maybe a slightly better performance, but I think it's cleaner to have both logics separated

internal/common/autogen/unmarshal.go
lantoli: doc update here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3281/commits/07b2a99c17c233d415567a06493df9e5b74150b2

PR #3303, opened 2025-04-25
chore: Support nested attributes with same name avoiding generation of duplicate model names

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-315376

**Problem**: In the case of Stream Instance and Cluster resource we encountered compilation errors when generating resources. This is due to the schemas having nested attributes with same name present in different nested objects.
Example: Cluster resource generated `TFComputeModel` twice as `compute` nested attribute is defined both in `auto_scaling` and `analytics_auto_scaling`.

**Solution**: All nested models have the parent model name as part of their name, ensuring no duplicate models. For case of Compute model we now have `TFReplicationSpecsRegionConfigsAnalyticsAutoScalingComputeModel` and `TFReplicationSpecsRegionConfigsAutoScalingComputeModel`.

## Type of change:

## Required Checklist:


EspenAlbert (commented):
Are these the same objects in the API Spec?
Why not re-use?

internal/service/clusterapi/resource_schema.go
```diff
@@ -0,0 +1,673 @@
...
+	InstanceSize  types.String  `tfsdk:"instance_size"`
+	NodeCount     types.Int64   `tfsdk:"node_count"`
+}
+type TFReplicationSpecsRegionConfigsAutoScalingModel struct {
+	Compute types.Object `tfsdk:"compute"`
+	DiskGb  types.Object `tfsdk:"disk_gb"`
+}
+type TFReplicationSpecsRegionConfigsAutoScalingComputeModel struct {
+	Enabled           types.Bool   `tfsdk:"enabled"`
+	MaxInstanceSize   types.String `tfsdk:"max_instance_size" autogen:"omitjson"`
```
EspenAlbert, line 647: Why is this `omitjson`?

internal/service/streaminstanceapi/resource_schema.go
```diff
@@ -0,0 +1,508 @@
...
+			"_id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Unique 24-hexadecimal character string that identifies the project.",
+			},
+			"cloud_provider": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "Label that identifies the cloud service provider where MongoDB Cloud performs stream processing. Currently, this parameter only supports AWS and AZURE.",
+			},
+			"connections": schema.ListNestedAttribute{
+				Computed:            true,
```
EspenAlbert, line 24: Computed only?

internal/service/clusterapi/resource_schema.go
```diff
@@ -0,0 +1,673 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			"mongo_dbmajor_version": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "MongoDB major version of the cluster. Set to the binary major version. \n\nOn creation: Choose from the available versions of MongoDB, or leave unspecified for the current recommended default in the MongoDB Cloud platform. The recommended version is a recent Long Term Support version. The default is not guaranteed to be the most recently released version throughout the entire release cycle. For versions available in a specific project, see the linked documentation or use the API endpoint for [project LTS versions endpoint](#tag/Projects/operation/getProjectLtsVersions).\n\n On update: Increase version only by 1 major version at a time. If the cluster is pinned to a MongoDB feature compatibility version exactly one major version below the current MongoDB version, the MongoDB version can be downgraded to the previous major version.",
```
lantoli, line 243: nit: we might want to limit description size

tools/codegen/gofilegen/schema/typed_model.go
```diff
@@ -45,7 +45,7 @@ func generateModelObjType(attrs codespec.Attributes, name string) CodeStatement
 	}
 }
 
-func getNestedModel(attribute *codespec.Attribute, withObjTypes bool) *CodeStatement {
+func getNestedModel(attribute *codespec.Attribute, parentName string, withObjTypes bool) *CodeStatement {
```
lantoli: nit: should we call it ancestorsName instead of parentName as it can potentially have not only the direct parent but all ancestors?

lantoli (approved):
LGTM, nice what a small change is needed

internal/service/clusterapi/resource_schema.go
AgustinBettati: Yes this is a bug for sure, either inconsistency in the API Spec or bug in our API Spec parsing logic. Will try to dive deeper next week to see the root cause.

internal/service/streaminstanceapi/resource_schema.go
AgustinBettati: Yes, this one is actually correct in terms of how the API Spec is defined. `connections` cannot be provided in requests of Stream Instance CRUD however it is present in the responses as read only.

As for our manual resources we decided to not include `connections` as there are separate resource and data sources for handling connections.

AgustinBettati:
@EspenAlbert responding here

> Are these the same objects in the API Spec?

Yes

> Why not re-use?

Conserving the original references to common models, both in the intermediate model and in code generation logic, adds complexity. As of now I cant see significant drawbacks of having multiple models with same content, mainly as all code is fully generated. We can always revisit if we find additional drawbacks.

tools/codegen/gofilegen/schema/typed_model.go
AgustinBettati: good point, adjusted

internal/service/clusterapi/resource_schema.go
AgustinBettati: Yes this might be a hard criteria to determine. Will copy generated docs into personal provider so they can be viewed directly in terraform registry, hopefully we can get good feedback from product in case current schema docs require significant changes.

lantoli:
@AgustinBettati @EspenAlbert I think it's ok as it is now, we might consider the improvement of keeping a set of models already generated so we only generate links, etc. once, if they really have the same content always. then we can go back to no need the ancestor prefix.

edit: or maybe reconsider this PR and do it here directly keeping the name without ancestors but generating it only once keeping the set of already generated models

AgustinBettati:
> we might consider the improvement of keeping a set of models already generated so we only generate links once

Yes this could be an alternative. We have to keep in mind that additional logic in generation using common models also can imply additional logic in conversion logic to find the correct nested model (currently we don't need this, but the use case may come in the future).

My take is that at the moment the drawback is small (potential models with same content), as we continue covering more cases we can assess if there are more drawbacks to this approach which keeps model and code generation simple (KISS principle).

PR #3307, opened 2025-04-29
chore: Adding CI tests for working project and resource policy generated resources

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-315050

In addition to adding acceptance tests a fix was made for supporting successful updates in project and resource policy resources. This is due to the PATCH request using path parameters that are computed-only and not present in the Plan.

Stream instance CI test have not been added as the resource currently faces 2 limitations:

- `_id` property causes errors in our conversion logic
- Generated schema is not correct as POST and PATCH do not have the same request structure

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/clusterapi/resource.go
+++ b/internal/service/clusterapi/resource.go
@@ -74,13 +74,16 @@ func (r *rs) Read(ctx context.Context, req resource.ReadRequest, resp *resource.
 
 func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
 	var plan TFModel
+	var state TFModel
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	// Path params are grabbed from state as they may be computed-only and not present in the plan
 	pathParams := map[string]string{
-		"groupId": plan.GroupId.ValueString(),
-		"name":    plan.Name.ValueString(),
+		"groupId": state.GroupId.ValueString(),
+		"name":    state.Name.ValueString(),
 	}
 	callParams := config.APICallParams{
 		VersionHeader: apiVersionHeader,
--- a/internal/service/projectapi/resource_test.go
+++ b/internal/service/projectapi/resource_test.go
@@ -32,10 +32,11 @@ func TestAccProjectAPI_basic(t *testing.T) {
 				Config: configBasic(orgID, projectName, true),
 				Check:  checkBasic(),
 			},
-			{
-				Config: configBasic(orgID, projectName, false),
-				Check:  checkBasic(),
-			},
+			// complete removal does not send property in PATCH
+			// {
+			// 	Config: configBasic(orgID, projectName, false),
+			// 	Check:  checkBasic(),
+			// },
 			{
 				ResourceName:      resourceName,
 				ImportStateIdFunc: importStateIDFunc(resourceName),
--- a/internal/service/projectapi/resource_test.go
+++ b/internal/service/projectapi/resource_test.go
@@ -13,7 +13,7 @@ import (
 
 const resourceName = "mongodbatlas_project_api.test"
 
-func TestAccDatabaseUserAPI_basic(t *testing.T) {
+func TestAccProjectAPI_basic(t *testing.T) {
 	var (
 		orgID       = os.Getenv("MONGODB_ATLAS_ORG_ID")
 		projectName = acc.RandomProjectName()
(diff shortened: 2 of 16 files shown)


tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl
```diff
@@ -86,12 +86,15 @@ func (r *rs) Read(ctx context.Context, req resource.ReadRequest, resp *resource.
...
+	var state TFModel
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	// Path params are grabbed from state as they may be computed-only and not present in the plan
 	pathParams := map[string]string{ {{range .APIOperations.Update.PathParams }}
-		"{{ .CamelCaseName }}": plan.{{ .PascalCaseName }}.ValueString(),
+		"{{ .CamelCaseName }}": state.{{ .PascalCaseName }}.ValueString(),
```
AgustinBettati, line 97: Includes a fix detected in project update. `Id` attribute which is computed only is not present in Plan and must be grabbed from State. Applied this for all path params. We have to double check how this impacts the get requests of long last operations.

lantoli: I think it would be ok if you change:
```
CallParams: readAPICallParams(&plan),
```
to: 
```
CallParams: readAPICallParams(&state),
```

almost at the end of Update op

lantoli: I think it's not needed but you might want to also do this change in Create in order to keep consistency when calling readAPICallParams

internal/service/clusterapi/resource.go
```diff
@@ -74,13 +74,16 @@ func (r *rs) Read(ctx context.Context, req resource.ReadRequest, resp *resource.
...
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	// Path params are grabbed from state as they may be computed-only and not present in the plan
 	pathParams := map[string]string{
-		"groupId": plan.GroupId.ValueString(),
-		"name":    plan.Name.ValueString(),
+		"groupId": state.GroupId.ValueString(),
```
lantoli, line 85: just to confirm that we don't need to do this in Create?

internal/service/projectapi/resource_test.go
```diff
@@ -0,0 +1,129 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: configBasic(orgID, projectName, false),
+				Check:  checkBasic(),
+			},
+			{
+				Config: configBasic(orgID, projectName, true),
+				Check:  checkBasic(),
+			},
+			// complete removal does not send property in PATCH
```
lantoli: add TODO to know it's something to be addressed?

lantoli (approved):

EspenAlbert (approved):
Nice

tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl
AgustinBettati: Thanks, adjusted the update operation with suggestions, makes sense.

With create operation it is slightly more complex. Creation only has a Plan, and no State. Let say we have a resource like Project or Resource Policy, if we defined the readAPICallParams using the Plan (only info available at that point) computed-only attributes will not be available. Instead of passing params directly to the function, we might have to instead pass a function that received a model and extracts necessary info for obtaining path params. As of now we dont have an example resource were this would fail.

internal/service/clusterapi/resource.go
AgustinBettati: clarified above, we can keep the conversation there

internal/service/projectapi/resource_test.go
AgustinBettati: added

lantoli (approved):
LGTM, thanks for the follow-ups

tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl
lantoli: I don't think that will happen for any resource in Create.  We have an initial plan (from config) that is augmented with API response so all computed values are added to the "plan" model that will be sent as the response state.   So calling read with this model will have all required attributes, e.g. id in project resource

AgustinBettati: I think the problem comes if in the create operation we have a wait configuration. In this case the readAPICallParams that are passed for calling the get will be empty (the computed-only) when invoking `HandleCreate`. FYI merging this PR, if we see this is an actual problem I will create a follow up ticket

lantoli: let's continue offline

PR #3353, opened 2025-05-27
feat: Adds support for streams accountDetails endpoint

kpatel71716:
## Description

Adding support for the MongoDB Streams accountDetails endpoint to unblock current customers

[JIRA: STREAMS-1727](https://jira.mongodb.org/browse/STREAMS-1727)

Link to any related issue(s):

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ x ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [ x  ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [  x ] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [ x  ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [ x  ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [  x ] I have added any necessary documentation (if appropriate)
- [  x ] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals I have added appropriate changelog entries.
- [ ] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

PR #3364, opened 2025-05-29
feat: Adds support for streams accountDetails endpoint

kpatel71716:
## Description

Adding support for the MongoDB Streams accountDetails endpoint to unblock current customers

PR is copied from https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3353 as checks won't pass on PRs from forked repo

[JIRA: STREAMS-1727](https://jira.mongodb.org/browse/STREAMS-1727)

Link to any related issue(s):

## Type of change:

## Required Checklist:

## Further comments
See previous conversations on https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3353


docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
+# Data Source: mongodbatlas_stream_account_details
+
+`mongodbatlas_stream_account_details` returns the Account ID, and the VPC ID for the group and region specified.
```
jwilliams-mongo: we should clarify that these are AWS IDs. also tweak slightly to remove passive voice.
```suggestion
`mongodbatlas_stream_account_details` returns the AWS Account ID and the AWS VPC ID for the group and region that you specify.
```

docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
...
+  value = data.mongodbatlas_stream_account_details.account_details.vpc_id
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `cloud_provider` (String) One of "aws", "azure" or "gcp".
```
jwilliams-mongo: use monospace for field values instead of quotes. 
```suggestion
- `cloud_provider` (String) One of `aws`, `azure`, or `gcp`.
```

docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
...
+- `project_id` (String) Unique 24-hexadecimal digit string that identifies your project. Use the [/groups](#tag/Projects/operation/listProjects) endpoint to retrieve all projects to which the authenticated user has access.
```
jwilliams-mongo: I don't think this relative link will work from the terraform provider docs. use a link like the one on L49

docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
...
+
+**NOTE**: Groups and projects are synonymous terms. Your group id is the same as your project id. For existing groups, your group/project id remains the same. The resource and corresponding endpoints use the term groups.
+- `region_name` (String) The cloud provider specific region name, i.e. "US_EAST_1" for cloud provider "aws".
```
jwilliams-mongo:
```suggestion
- `region_name` (String) The cloud provider specific region name, i.e. `US_EAST_1` for cloud provider `aws`.
```

docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Read-Only
+
```
jwilliams-mongo, line 42: question: some of these are intuitive enough (like azure_subscription_id), but do these other fields apply only for specific cloud providers? If so, it might help to provide that context in the descriptions.

docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
...
+- `aws_account_id` (String) The AWS Account ID.
+- `azure_subscription_id` (String) The Azure Subscription ID.
+- `cidr_block` (String) The VPC CIDR Block.
+- `virtual_network_name` (String) The name of the virtual network.
+- `vpc_id` (String) The VPC ID.
+
+For more information see: [MongoDB Atlas API - Returns the Account ID, and the VPC ID for the group and region specified](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/getAccountDetails) Documentation.
```
jwilliams-mongo:
```suggestion
For more information see [MongoDB Atlas API - Returns the Account ID, and the VPC ID for the group and region specified](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/getAccountDetails).
```

jwilliams-mongo: The API name is a bit verbose. My recommendation would be to change that upstream to something that aligns with the style of the other endpoint names.

examples/mongodbatlas_stream_account_details/README.md
```diff
@@ -0,0 +1,9 @@
+# MongoDB Atlas Provider - AWS Confluent Privatelink for Atlas Streams 
+
+This example shows how to get Atlas Streams account details for a project
```
jwilliams-mongo:
```suggestion
This example shows how to get Atlas Streams account details for a project.
```

internal/service/streamaccountdetails/data_source_schema.go
```diff
@@ -0,0 +1,73 @@
...
+				MarkdownDescription: "The VPC CIDR Block.",
+			},
+			"cloud_provider": schema.StringAttribute{
+				Required:            true,
+				Description:         "One of \"aws\", \"azure\" or \"gcp\".",
+				MarkdownDescription: "One of \"aws\", \"azure\" or \"gcp\".",
+			},
+			"project_id": schema.StringAttribute{
+				Required:            true,
+				Description:         "Unique 24-hexadecimal digit string that identifies your project. Use the [/groups](#tag/Projects/operation/listProjects) endpoint to retrieve all projects to which the authenticated user has access.\n\n**NOTE**: Groups and projects are synonymous terms. Your group id is the same as your project id. For existing groups, your group/project id remains the same. The resource and corresponding endpoints use the term groups.",
```
jwilliams-mongo: replace the relative link here with an absolute one

internal/service/streamaccountdetails/data_source_schema.go
```diff
@@ -0,0 +1,73 @@
...
+				MarkdownDescription: "Unique 24-hexadecimal digit string that identifies your project. Use the [/groups](#tag/Projects/operation/listProjects) endpoint to retrieve all projects to which the authenticated user has access.\n\n**NOTE**: Groups and projects are synonymous terms. Your group id is the same as your project id. For existing groups, your group/project id remains the same. The resource and corresponding endpoints use the term groups.",
```
jwilliams-mongo: replace the relative link here with an absolute one

templates/data-sources/stream_account_details.md.tmpl
```diff
@@ -0,0 +1,10 @@
+# {{.Type}}: {{.Name}}
+
+`{{.Name}}` returns the Account ID, and the VPC ID for the group and region specified.
```
jwilliams-mongo: see comments on the matching string above

templates/data-sources/stream_account_details.md.tmpl
```diff
@@ -0,0 +1,10 @@
...
+
+## Example Usages
+{{ tffile (printf "examples/mongodbatlas_stream_account_details/main.tf" )}}
+
+{{ .SchemaMarkdown | trimspace }}
+
+For more information see: [MongoDB Atlas API - Returns the Account ID, and the VPC ID for the group and region specified](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/getAccountDetails) Documentation.
```
jwilliams-mongo:
```suggestion
For more information see [MongoDB Atlas API - Returns the Account ID, and the VPC ID for the group and region specified](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/getAccountDetails).
```

jwilliams-mongo (commented):
few comments. many of them likely apply also to the upstream API documentation. Ideally they should be fixed in both places. Up to the terraform folks to decide whether it's appropriate to fix them upstream first before pulling in changes here.

internal/service/streamaccountdetails/data_source_schema.go
```diff
@@ -0,0 +1,73 @@
...
+				Validators: []validator.String{
+					stringvalidator.RegexMatches(regexp.MustCompile("^([a-f0-9]{24})$"), ""),
+				},
```
AgustinBettati: nit: We tend to avoid  regex or enum validations within the provider, leaving this up to the API directly. This avoids duplicate logic on both sides.

internal/service/streamaccountdetails/data_source_schema.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+)
+
+func DataSourceSchema(ctx context.Context) schema.Schema {
+	return schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"aws_account_id": schema.StringAttribute{
+				Computed:            true,
+				Description:         "The AWS Account ID.",
+				MarkdownDescription: "The AWS Account ID.",
```
AgustinBettati: to reduce verbosity you can define only `MarkdownDescription` and make use of `UpdateSchemaDescription` to set the value of `Description` ([example](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/streamprocessor/resource.go#L43))

internal/service/streamaccountdetails/model.go
```diff
@@ -0,0 +1,29 @@
...
+	accountDetails *admin.AccountDetails,
+) (*TFStreamAccountDetailsModel, diag.Diagnostics) {
+	return &TFStreamAccountDetailsModel{
+		ProjectId:           types.StringValue(projectID),
+		AwsAccountId:        types.StringPointerValue(accountDetails.AwsAccountId),
+		AzureSubscriptionId: types.StringPointerValue(accountDetails.AzureSubscriptionId),
+		CidrBlock:           types.StringPointerValue(accountDetails.CidrBlock),
+		CloudProvider:       types.StringValue(cloudProvider),
+		RegionName:          types.StringValue(region),
+		VirtualNetworkName:  types.StringPointerValue(accountDetails.VirtualNetworkName),
```
AgustinBettati, line 26: q (not fully related to data source implementation): Do we know why `virtualNetworkName` is not showing in API docs? https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Streams/operation/getAccountDetails

AgustinBettati (commented):
overall LGTM, leaving some suggestions and doubts

internal/service/streamaccountdetails/model.go
kpatel71716: Screenshot 2025-05-30 at 7 35 32 AM

It looks like a quirk of the docs where it won't show up unless the cloudProvider is selected in the response schema. It's a bit confusing though I imagine the intention was to show that the response fields depend on the cloud provider

internal/service/streamaccountdetails/data_source_schema.go
kpatel71716: Good to know! Removed!

internal/service/streamaccountdetails/data_source_schema.go
kpatel71716: Misses on my end, fixed!

docs/data-sources/stream_account_details.md
kpatel71716: I agree and we have added this to our list of potential changes in the next version of our API suite, though since this endpoint is used in prod, I don't believe we can update that at this time

docs/data-sources/stream_account_details.md
kpatel71716: Good question, I made some changes to provide a bit more context. The fields returned depend on the cloud provider so `virtual_network_name` and `virtual_network_name` are only present in the response for Azure while `aws_account_id` and `vpc_id` will be present for AWS

docs/data-sources/stream_account_details.md
kpatel71716: Yep, miss on my end for not replacing these, fixed!

kpatel71716:
> few comments. many of them likely apply also to the upstream API documentation. Ideally they should be fixed in both places. Up to the terraform folks to decide whether it's appropriate to fix them upstream first before pulling in changes here.

I agree that the docs need clarification.  I will also fix those upstream but can do that in a follow-up ticket if that is fine with the terraform and docs folks (that way we can unblock the customer needing this feature first)

lantoli:
note that acceptances tests are failing in PR checks, e.g. [execution](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/15347258727/job/43186854239?pr=3364)

internal/service/streamaccountdetails/data_source_test.go
```diff
@@ -0,0 +1,78 @@
...
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyStreamInstance,
+		Steps: []resource.TestStep{
+			{
+				Config: StreamAccountDetailsConfig(projectID, "aws", "US_EAST_1", clusterName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttrSet(dataSourceName, "project_id"),
```
lantoli: nit: check for existance of required attributes is not needed in tests, Terraform ensures this, useful for Computed values though, ideally using `TestCheckResourceAttr` if you know the expected value, as `TestCheckResourceAttrSet` just checks it has some value

internal/service/streamaccountdetails/model.go
AgustinBettati: Oh I see, that makes sense, thanks for clarifying 👍

kpatel71716:
> note that acceptances tests are failing in PR checks, e.g. [execution](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/15347258727/job/43186854239?pr=3364)

Yep, it looks like the cluster being created isn't in us-east-1, causing a downstream issue that makes the test fail. Looking into a fix for this before re-requesting reviews

kpatel71716:
I added a [commit](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3364/commits/8999f69b9f29ff22a32cf63f84ca95f1cf0e3723) to allow for cluster creation in us-east-1 for acceptance tests. This is required for the `getAccountDetails` endpoint since it requires a cluster and us-west-2 is not a supported region for this endpoint.

The commit has cluster creation default to us-west-2 if us-east-1 is not selected but please let me know if this logic should be adjusted and/or if this change is acceptable

The currently failing acceptance tests are unrelated and a known issue

internal/service/streamaccountdetails/data_source_test.go
```diff
@@ -0,0 +1,44 @@
...
+				),
+			},
+		},
+	})
+}
+
+func StreamAccountDetailsConfig(projectID, cloudProvider, regionName string) string {
+	return fmt.Sprintf(`
+
+		data "mongodbatlas_stream_account_details" "test_details" {
```
AgustinBettati:
```suggestion
	return fmt.Sprintf(`
		data "mongodbatlas_stream_account_details" "test_details" {
```

internal/testutil/acc/atlas.go
```diff
@@ -62,7 +62,11 @@ func deleteCluster(projectID, name string) {
 	}
 }
 
-func clusterReq(name, projectID string) admin.ClusterDescription20240805 {
+func clusterReq(name, projectID, regionName string) admin.ClusterDescription20240805 {
+	if regionName != constant.UsEast1 {
+		regionName = constant.UsWest2
+	}
```
AgustinBettati: since we are adding flexibility for what region to use, I would allow any value so we dont have to adjust these hardcoded values in the future

internal/testutil/acc/shared_resource.go
```diff
@@ -89,10 +90,23 @@ func ProjectIDExecutionWithCluster(tb testing.TB, totalNodeCount int) (projectID
 // This function relies on using an execution project and returns its id.
 // When `MONGODB_ATLAS_CLUSTER_NAME` and `MONGODB_ATLAS_PROJECT_ID` are defined it will be used instead of creating resources. This is useful for local execution but not intended for CI executions.
 func ClusterNameExecution(tb testing.TB, populateSampleData bool) (projectID, clusterName string) {
+	tb.Helper()
+	return ClusterNameExecutionWithRegion(tb, "", populateSampleData)
```
AgustinBettati:
```suggestion
	return ClusterNameExecutionWithRegion(tb, constant.UsWest2, populateSampleData)
```
Will help with readability, makes it clear what region ClusterNameExecution uses without going into implementation.

internal/testutil/acc/shared_resource.go
```diff
@@ -89,10 +90,23 @@ func ProjectIDExecutionWithCluster(tb testing.TB, totalNodeCount int) (projectID
...
+// When `MONGODB_ATLAS_CLUSTER_NAME` and `MONGODB_ATLAS_PROJECT_ID` are defined it will be used instead of creating resources. This is useful for local execution but not intended for CI executions.
+// Valid regions are US_EAST_1, US_WEST_2, or an empty string
+func ClusterNameExecutionWithRegion(tb testing.TB, regionName string, populateSampleData bool) (projectID, clusterName string) {
 	tb.Helper()
 	SkipInUnitTest(tb)
 	require.True(tb, sharedInfo.init, "SetupSharedResources must called from TestMain test package")
 
+	if regionName != "" && regionName != constant.UsEast1 && regionName != constant.UsWest2 {
+		require.NoError(tb, fmt.Errorf("the region supplied for cluster creation is invalid: %s", regionName))
+	}
```
AgustinBettati: Similar to comment above, would allow any value for regionName so we don't need to adjust this in the future if we want to use another region

internal/service/streamaccountdetails/data_source_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/constant"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/acc"
+)
+
+func TestAccStreamAccountDetailsDS_basic(t *testing.T) {
+	var (
+		projectID, _   = acc.ClusterNameExecutionWithRegion(t, constant.UsEast1, false)
```
AgustinBettati: Is this endpoint limited to a set of regions? What is the experience of the end user? 
If its the case want to make sure we have a clear error message on the API side if an unsupported region is used. Potentially we could also clarify this in our docs (I see we currently don't mention anything in API docs/tf docs.

AgustinBettati (approved):
LGTM, have a doubt related to supported regions and small suggestion on latest change for defining custom region.

internal/testutil/acc/shared_resource.go
```diff
@@ -89,10 +90,23 @@ func ProjectIDExecutionWithCluster(tb testing.TB, totalNodeCount int) (projectID
...
 func ClusterNameExecution(tb testing.TB, populateSampleData bool) (projectID, clusterName string) {
+	tb.Helper()
+	return ClusterNameExecutionWithRegion(tb, "", populateSampleData)
+}
+
+// ClusterNameExecutionWithRegion returns the name of a created cluster for the execution of the tests in the resource package.
+// This function relies on using an execution project and returns its id.
+// When `MONGODB_ATLAS_CLUSTER_NAME` and `MONGODB_ATLAS_PROJECT_ID` are defined it will be used instead of creating resources. This is useful for local execution but not intended for CI executions.
+// Valid regions are US_EAST_1, US_WEST_2, or an empty string
+func ClusterNameExecutionWithRegion(tb testing.TB, regionName string, populateSampleData bool) (projectID, clusterName string) {
```
EspenAlbert: I think the [`GetClusterInfo`](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/STREAMS-1727/internal/testutil/acc/cluster.go#L64) would be more appropriate instead of adding more logic here.
Then in the test you can edit to depend on the resource directly and create it as part of the config instead of outside.

internal/service/streamaccountdetails/data_source_test.go
kpatel71716: I agree, there are a lot of usability issues with this endpoint that we (the Atlas Streams) team need to document and fix (I've had many issues working with this endpoint myself for this ticket). I have a ticket created [here](https://jira.mongodb.org/browse/CLOUDP-321656) as a follow-up

Right now the endpoint does not work unless the user has a created cluster in the region specified in the query params. And Atlas Streams only supports creating a streams tenant  in a handful of regions so a user needs to have used one of those regions as well, which is confusing

marcosuma:
@kpatel71716 thanks for your PR! Referring to @jwilliams-mongo's comment, can you share with us the CLOUDP you'll use to update the API docs? That way we can unblock this but make sure we follow-up on the API changes.

docs/data-sources/stream_account_details.md
```diff
@@ -0,0 +1,49 @@
+# Data Source: mongodbatlas_stream_account_details
```
marcosuma, line 1: I am not sure calling it _details is right. I think it should be just called `mongodbatlas_stream_account` - @AgustinBettati @lantoli @EspenAlbert @oarbusi  thoughts?

internal/testutil/acc/shared_resource.go
kpatel71716: This worked perfectly, switched! Was unaware this could be done, thank you

kpatel71716:
> @kpatel71716 thanks for your PR! Referring to @jwilliams-mongo's comment, can you share with us the CLOUDP you'll use to update the API docs? That way we can unblock this but make sure we follow-up on the API changes.

Sure thing, I have [CLOUDP-321656](https://jira.mongodb.org/browse/CLOUDP-321656) created for our team to handle  some exceptions better but mostly to improve the documentation around this endpoint and standardize across this repo , mms, and our documentation pages

We are also considering feedback for a new proposed major version of this api so I will gather the thoughts already in this PR as well as any other feedback for our proposal

docs/data-sources/stream_account_details.md
AgustinBettati: No strong opinion, the path `https://cloud.mongodb.com/api/atlas/v2/groups/{groupId}/streams/accountDetails` and operation id `getAccountDetails` use this term but I can see how it is redundant

marcosuma: fair point, we should stick to the API name (even though I don't really like it)

internal/testutil/acc/shared_resource.go
EspenAlbert: Awesome!
Maybe something we can add to the `contributing/testing-best-practices.md` file 🤔

EspenAlbert (approved):
Great job! Thank you for addressing the comments 👏

docs/data-sources/stream_account_details.md
lantoli: i agree to stick to API although i don't like either, also it can be confused with plural data sources

internal/service/streamaccountdetails/data_source_test.go
```diff
@@ -0,0 +1,52 @@
...
+					resource.TestCheckResourceAttrSet(dataSourceName, "vpc_id"),
+					resource.TestCheckNoResourceAttr(dataSourceName, "azure_subscription_id"),
+					resource.TestCheckNoResourceAttr(dataSourceName, "virtual_network_name"),
+				),
+			},
+		},
+	})
+}
+
+func StreamAccountDetailsConfig(clusterInfoStr, projectID, cloudProvider, regionName, clusterInfoNameRef string) string {
```
lantoli: StreamAccountDetailsConfig can be unexported (lowercase). and then, name can be simplified to something like `configBasic` as we do in other tests, as the scope is restricted to package  streamaccountdetails_test

lantoli (approved):

lantoli:
nit: in the PR description, for checks instead of using [ x  ] use [x] without blanks so they will be real checks

internal/testutil/acc/shared_resource.go
kpatel71716: More than happy to contribute to the doc. I'll put up a separate change to this!

internal/service/streamaccountdetails/data_source_test.go
kpatel71716: good idea, done!

docs/data-sources/stream_account_details.md
kpatel71716: agreed on the confusing naming, will stick with the API but noting this as we audit the state of our existing apis

kpatel71716:
@jwilliams-mongo does this lgtm from a docs perspective?

jwilliams-mongo (approved):

PR #3404, opened 2025-06-12
chore: Make action to run script and commit changes reusable outside of repo

oarbusi:
## Description

Make action to run script and commit changes reusable outside of repo
Follows structure of other reusable action clean-atlas-org.

Changes have been tested with changelog generation action. See [commit](https://github.com/mongodb/terraform-provider-mongodbatlas/commit/d6b3b5ad73731a65b6a243d89ac3e124b0b43e84) and [action run](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/15631168829/job/44035646275)

Also, has been tested outside of this repo. See [commit](https://github.com/mongodb/mongodbatlas-cloudformation-resources/commit/275e283d230f167876292b099d27a402cec54680) and [action run](https://github.com/mongodb/mongodbatlas-cloudformation-resources/actions/runs/15631487324/job/44036692869) and the [PR](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/1365) (still in draft)

Link to any related issue(s): CLOUDP-324019

## Type of change:

## Required Checklist:


.github/templates/run-script-and-commit/action.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      with:
+        fetch-depth: 0
+        token: ${{ inputs.apix_bot_pat }}
+
+    - name: Set up Go
+      uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5
+      with:
+        go-version-file: ${{ format('{0}go.mod', inputs.repo-path) }}
```
EspenAlbert, line 44: This might need to be flexible, for example, in CFN where there is no `go.mod` file in the root path

oarbusi: this is already flexible, input.repo-path is used to provide path where the go.mod is. See where I am using [this](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/1365/files#diff-e81e13daadd1745de51d8b8d85fce1fcd9be355732b64d60a6b56e18c28388caR128)

.github/workflows/release.yml
```diff
@@ -54,16 +54,19 @@ jobs:
...
-      file_to_commit: 'docs/index.md'
-      commit_message: 'chore: Updates examples link in index.md for ${{ github.event.inputs.version_number }} release'
-    secrets:
-      apix_bot_pat: ${{ secrets.APIX_BOT_PAT }}
-      remote: https://svc-apix-bot:${{ secrets.APIX_BOT_PAT }}@github.com/${{ github.repository }}  
-      gpg_private_key: ${{ secrets.APIX_BOT_GPG_PRIVATE_KEY }}
-      passphrase: ${{ secrets.APIX_BOT_PASSPHRASE }}      
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
```
EspenAlbert, line 59: How does the checkout work?
Looks like we are doing checkout before calling the template and in the template.
We use fetch-depth=0 in the template but not outside.
Also, we don't specify the checkout path, maybe that is needed?

.github/templates/run-script-and-commit/action.yaml
EspenAlbert: Nice :D

.github/workflows/release.yml
oarbusi: To have available the action template, we need to checkout. In this case it wouldn't be 100% necessary, but it's to be consistent. See in [CFN usage of this](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/1365), we checkout the TF repo (to get the action template defined in this PR) and then inside the action we check out the CFN code to be able to run the scripts defined there. Hope that makes sense

EspenAlbert: Sounds good, maybe we can add a comment here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3404/files#diff-2b54d47c44009e167c79ba7eb9d595a4dd48988568f3d15b09e9ada86d225eb8R35

EspenAlbert (approved):
Nice!

.github/templates/run-script-and-commit/action.yaml
```diff
@@ -0,0 +1,70 @@
+name: "Run Script and Commit Changes"
```
marcosuma, line 1: is there any security risk in making this reusable outside? To understand the scope: any other repository in GitHub could potentially use this?

marcosuma: I wonder if we could limit to only mongodb organization

lantoli: does this mean that anybody can call this GHA and get their changes signed with our private key? (can be a security risk even if the commit is in their repo)

oarbusi: keys are an input that has to be provided wherever this is used, so signing with our keys is only possible in our repos where we set the keys as secrets

oarbusi: tldr: no risk in making this reusable. We are just avoiding having to copy and paste this into different repos

marcosuma: yup, I just think that if someone outside our team decides to start using it, it will raise a question of having to maintain it. I know we want to avoid copy/paste, but I think we should do it only if access to the GHA can be limited to ourselves only, not full public.

oarbusi: As far as I know, it can't be limited because code is public. I don't think having this public (it already is in this PR) should be an issue in terms of maintenance. This repo is licensed under [MLP-2.0](https://github.com/mongodb/terraform-provider-mongodbatlas?tab=MPL-2.0-1-ov-file#readme), and it provides the software ["as is" ](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/1922fedc0a6aaba84532117f529d55aabf51df30/LICENSE#L266). Apart from that, we already have one template here: https://github.com/mongodb/terraform-provider-mongodbatlas/tree/master/.github/templates/clean-atlas-org,

AgustinBettati: > I just think that if someone outside our team decides to start using it, it will raise a question of having to maintain it

I would say while external users can leverage the action template, it would be at their own risk. [Publishing actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/publishing-actions-in-github-marketplace) seems like the more intentional approach for exposing workflow job steps to the broader community (including versioning, documentation, and more).

.github/workflows/release.yml
```diff
@@ -190,21 +196,24 @@ jobs:
       && needs.release.result == 'success'
       && needs.release-config.outputs.is_official_release == 'true'
       && needs.compliance.result == 'success'
-    uses: ./.github/workflows/run-script-and-commit.yml
```
AgustinBettati, line 193: [q] Why did we have to move from a reusable workflow template over to an action? Cant the CFN repo use the reusable workflow template?

oarbusi: The switch from a reusable workflow template to a composite action is mainly for flexibility and ease of reuse. While the CFN repo is in the same org and technically could use the reusable workflow template we had, composite actions are much easier to consume as a step in any workflow, not just as a top-level workflow.

PR #3458, opened 2025-07-02
chore: Singletons in auto-generated resources

lantoli:
## Description

Singletons in auto-generated resources. This is part of the spike: WRITING-29828.

- Generate `auditing_api` and `maintenance_window_api` resources.
- `StaticRequestBody` to support Delete operation with PATCH or PUT.

Link to any related issue(s): CLOUDP-328945

## Type of change:

## Required Checklist:


internal/config/client.go
```diff
@@ -290,7 +290,7 @@ func (c *MongoDBClient) UntypedAPICall(ctx context.Context, params *APICallParam
 
 	apiResp, err := untypedClient.CallAPI(apiReq)
 
-	if apiResp.StatusCode >= 300 {
+	if apiResp != nil && apiResp.StatusCode >= 300 {
```
lantoli: avoid panic in some circumstances such as bad internet connection

internal/service/maintenancewindow/resource_maintenance_window_test.go
```diff
@@ -42,6 +42,7 @@ func TestAccConfigRSMaintenanceWindow_basic(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { acc.PreCheckBasic(t) },
 		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroy,
```
lantoli, line 45: improve curated resource tests checking destroy

internal/serviceapi/auditingapi/resource.go
```diff
@@ -0,0 +1,142 @@
...
+		RelativePath:  "/api/atlas/v2/groups/{groupId}/auditLog",
+		PathParams:    pathParams,
+		Method:        "PATCH",
+	}
+	reqHandle := autogen.HandleDeleteReq{
+		Resp:             resp,
+		Client:           r.Client,
+		State:            &state,
+		CallParams:       &callParams,
+		FixedRequestBody: `{"enabled": "false"}`,
```
lantoli: example to do Delete op with PATCH

tools/codegen/config.yml
```diff
@@ -1,4 +1,22 @@
...
+    create:
+      path: /api/atlas/v2/groups/{groupId}/auditLog
+      method: PATCH
+    update:
+      path: /api/atlas/v2/groups/{groupId}/auditLog
+      method: PATCH
+    delete:
+      path: /api/atlas/v2/groups/{groupId}/auditLog
+      method: PATCH
+      fixed_request_body: '{"enabled": "false"}'
```
lantoli: config for Delete with PATCH sending a request body

tools/codegen/config.yml
```diff
@@ -41,6 +59,7 @@ resources:
             "DELETING",
           ]
         target_states: ["DELETED"] # DELETED is a special state value when API returns 404 or empty object
+    version_header: application/vnd.atlas.2024-08-05+json
```
lantoli, line 62: add explicit version_header to all auto-generated resources instead of using the latest one

tools/codegen/config/config_model.go
```diff
@@ -14,9 +14,10 @@ type Resource struct {
 }
 
 type APIOperation struct {
-	Wait   *Wait  `yaml:"wait"`
-	Path   string `yaml:"path"`
-	Method string `yaml:"method"`
+	Wait             *Wait  `yaml:"wait"`
+	Path             string `yaml:"path"`
+	Method           string `yaml:"method"`
+	FixedRequestBody string `yaml:"fixed_request_body"` // use at the moment for Delete when it's done with a PATCH or PUT and needs to send a fixed request body
```
maastha: I'm not clear why we want to call this "fixed" RequestBody & not simply "RequestBody" or something?

internal/service/maintenancewindow/resource_maintenance_window_test.go
maastha: nice catch that this wasn't already there

internal/config/client.go
maastha: can we use http_status_checks.go?

tools/codegen/config/config_model.go
lantoli: sure, I'm open to other names, @AgustinBettati any suggestion?  Fixed was to highlight that we'll always send this request body in this operation

internal/config/client.go
lantoli: Those checks are for matching a specific value, here we want >= , not sure if it's worth it to add a new func in http_status_checks like StatusNotSuccesful if it's only going to be used from here.

internal/common/autogen/unmarshal.go
```diff
@@ -17,6 +17,9 @@ import (
 // Map is not supported yet, will be done in CLOUDP-312797.
 // Attributes that are in JSON but not in the model are ignored, no error is returned.
 func Unmarshal(raw []byte, model any) error {
+	if isEmptyJSON(raw) {
```
EspenAlbert, line 20: Why is this needed?

internal/config/client.go
EspenAlbert: Can we add some why comment here?
The error returned from `untypedClient.CallAPI(apiReq)`, why is it not enough?
Why not have this logic in the SDK?

internal/common/autogen/unmarshal.go
lantoli: because some operations return an empty payload (e.g. PATCH maintenance window), we don't want to fail there, we just don't update anything in the model

tools/codegen/config/config_model.go
EspenAlbert: I like it. It allows us to extend later, for example:
```go
type APIOperation struct {
    TemplateRequestBody TemplateBody
}

type TemplateBody struct {
    TemplateStr string
    AttributeReplacements map[string]string
}
```

EspenAlbert (approved):

internal/config/client.go
lantoli: at the moment I think it's better to keep the funcs PrepareRequest, CallAPI, MakeApiError separate as they have different responsibilities (and params) and they're following how SDK works for regular typed calls.  

Added a comment:   https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3458/commits/71b1bef4a03ec6c1d6b108f59d3d247132298f6c

internal/common/autogen/unmarshal.go
EspenAlbert: Comment maybe?

lantoli: comment added here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3458/commits/f94e2d70b3216bb07d83f5475eec25cf7ad951fa

tools/codegen/config.yml
```diff
@@ -1,4 +1,22 @@
...
+    read:
+      path: /api/atlas/v2/groups/{groupId}/auditLog
+      method: GET
+    create:
+      path: /api/atlas/v2/groups/{groupId}/auditLog
+      method: PATCH
+    update:
+      path: /api/atlas/v2/groups/{groupId}/auditLog
+      method: PATCH
+    delete:
```
AgustinBettati, line 14: q: Do we or would we want to support delete being a no-op? Thinking case of project settings.

internal/config/client.go
AgustinBettati: While very similar, we could follow the same conditionals SDK does when using these functions ([reference](https://github.com/mongodb/atlas-sdk-go/blob/main/admin/api_streams.go#L852-L857))

tools/codegen/config/config_model.go
AgustinBettati: Potentially `StaticRequestBody`? Both convey the intention imo so no strong opinion

AgustinBettati (approved):
LGTM, liked the static request body for improving delete UX when possible 👍

tools/codegen/config.yml
lantoli: i thought about it but didn't what to add the functionality until we don't need it.  In the case of project settings, don't we want to reset some values in delete?

internal/config/client.go
lantoli: thanks, changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3458/commits/bca42d9a3b42c33e28c9066026bac910e89105a1

tools/codegen/config/config_model.go
lantoli: I like that and we can use Dynamic for @EspenAlbert idea, changed here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3458/commits/f6607baac21b18fd9a4e63eab82c02dfd9957215

tools/codegen/config.yml
AgustinBettati: I wouldn't delegate the reseting logic to our side as we start to be opinionated on what those values are. If resetting is needed I think the API should expose a DELETE to control that behaviour (case of maintenance window).
Okay in not implementing this now, don't see unknowns in supporting this later on when needed.

lantoli: yes, i agree DELETE would be the best option here

PR #3479, opened 2025-07-08
test: Fixes streams connection kafka update tests

jwongmongodb:
## Description

The test for the kafka connection with VPC has a step where it is changing from a public networking type to a VPC type which is invalid and was fixed in our API with the following [PR](https://github.com/10gen/mms/pull/133418). This change updates the test to directly test the POST kafka connection with VPC. I also added a test to show the invalid updates on the networking type for the kafka connection.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/streamconnection/resource_stream_connection_test.go
+++ b/internal/service/streamconnection/resource_stream_connection_test.go
@@ -5,6 +5,7 @@ import (
 	_ "embed"
 	"fmt"
 	"os"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
@@ -89,12 +90,6 @@ func TestAccStreamRSStreamConnection_kafkaNetworkingVPC(t *testing.T) {
 		CheckDestroy:             CheckDestroyStreamConnection,
 		Steps: []resource.TestStep{
 			{
-				Config: networkPeeringConfig + configureKafka(projectID, instanceName, "user", "rawpassword", "localhost:9092", "earliest", kafkaNetworkingPublic, true),
-				Check: resource.ComposeAggregateTestCheckFunc(
-					checkKafkaAttributes(resourceName, instanceName, "user", "rawpassword", "localhost:9092", "earliest", networkingTypePublic, true, true),
-				),
-			},
-			{
 				Config: networkPeeringConfig + configureKafka(projectID, instanceName, "user", "rawpassword", "localhost:9092", "earliest", kafkaNetworkingVPC, true),
 				Check:  checkKafkaAttributes(resourceName, instanceName, "user", "rawpassword", "localhost:9092", "earliest", networkingTypeVPC, true, true),
 			},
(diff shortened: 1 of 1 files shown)


internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -88,12 +89,6 @@ func TestAccStreamRSStreamConnection_kafkaNetworkingVPC(t *testing.T) {
 		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
 		CheckDestroy:             CheckDestroyStreamConnection,
 		Steps: []resource.TestStep{
```
jwongmongodb, line 91: this test is for creating a VPC kafka connection, so there is no need to update from a public kafka to a private one.

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -109,6 +104,35 @@ func TestAccStreamRSStreamConnection_kafkaNetworkingVPC(t *testing.T) {
 	})
 }
 
+func TestAccStreamRSStreamConnection_invalidKafkaNetworkingUpdates(t *testing.T) {
```
jwongmongodb: This may be testing too much, I can remove if others feel like the endpoint test should handle this

EspenAlbert:
I'm not sure if it's worth mentioning. But an alternative is to use the PlanModifier RequiresReplace:
This will make Terraform delete the resource and then create it again on changes.
```go
			"networking": schema.SingleNestedAttribute{
				Optional: true,
				Computed: true,
				PlanModifiers: []planmodifier.Object{
					objectplanmodifier.UseStateForUnknown(),
				},
					"access": schema.SingleNestedAttribute{
						Required: true,
							"type": schema.StringAttribute{
								Required: true,
									stringplanmodifier.RequiresReplace(), // will force a delete+create
								},
							},
							"connection_id": schema.StringAttribute{
								Optional: true,
							},
						},
					},
				},
			},
```

jwongmongodb:
> I'm not sure if it's worth mentioning. But an alternative is to use the PlanModifier RequiresReplace: This will make Terraform delete the resource and then create it again on changes.
> 
> ```go
> 			"networking": schema.SingleNestedAttribute{
> 				Optional: true,
> 				Computed: true,
> 				PlanModifiers: []planmodifier.Object{
> 					objectplanmodifier.UseStateForUnknown(),
> 				},
> 				Attributes: map[string]schema.Attribute{
> 					"access": schema.SingleNestedAttribute{
> 						Required: true,
> 						Attributes: map[string]schema.Attribute{
> 							"type": schema.StringAttribute{
> 								Required: true,
> 								PlanModifiers: []planmodifier.String{
> 									stringplanmodifier.RequiresReplace(), // will force a delete+create
> 								},
> 							},
> 							"connection_id": schema.StringAttribute{
> 								Optional: true,
> 							},
> 						},
> 					},
> 				},
> 			},
> ```

~~This makes sense to me, if the user actually wanted to switch the type, they would have to manually delete it anyways. At least this way the plan will suggest this change to them if they changed the `access.type` and they can proceed destroying and creating it again (I believe the plan will suggest this in a requires replace).~~

EDIT: I think this change may need to be postponed for later since right now our connection status are not surfaced so deleting a VPC kafka node is not synchronous (there is a [proposal](https://docs.google.com/document/d/1Gxe_ysPMfcCC2Hw1i0OIPcEdAIeRFPn8j6HSVGQqsR4/edit?tab=t.0#heading=h.my8l8kwcujpo) to fix this in Q3)

internal/service/streamconnection/resource_stream_connection_test.go
EspenAlbert: Suggestion. Change the 2nd step to use `networkingTypePublic` and move it directly to the test above instead and delete this extra test
Normally, a good practice is only testing one thing at a time.
But with Terraform acceptance tests we can do one thing per test step instead 😅

This way, step 1 tests the networkingVPC and 2nd step tests the error condition

jwongmongodb: ok, I've moved this test into the existing public SSL test -> private VPC invalid update

EspenAlbert (approved):

PR #3515, opened 2025-07-22
feat: Network_peering long-running operation improvements

lantoli:
## Description

Network_peering long-running operation improvements

Link to any related issue(s): CLOUDP-332657

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fmt and formatted my code
- [x] If changes include deprecations or removals I have added appropriate changelog entries.
- [x] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

## Further comments

**Time-out with `delete_on_create_timeout`:**

error1

**Time-out with `delete_on_create_timeout` and cleanup failing:**

error2

**Time-out without `delete_on_create_timeout`:**

error3

PR #3525, opened 2025-07-24
feat: Long-running operation improvements for `mongodbatlas_flex_cluster` resource

lantoli:
## Description

Implements long-running operation improvements for flex cluster resource, adding proper timeout support and delete_on_create_timeout functionality.

### Changes
- Added timeout support: Flex cluster Create, Update, and Delete operations now respect configured timeouts block
- Added delete_on_create_timeout: New optional field (defaults to true) that automatically deletes cluster if creation times out
- Fixed data source: Resolved issue where data source incorrectly included resource-only timeout fields
- Added timeout testing: New acceptance test validates timeout and cleanup functionality
- Updated function signatures: CreateFlexCluster, UpdateFlexCluster, and DeleteFlexCluster now accept timeout parameters for use by advanced cluster services

Link to any related issue(s): CLOUDP-333401

## Type of change:

## Required Checklist:


internal/service/flexcluster/data_source.go
```diff
@@ -25,10 +26,17 @@ type ds struct {
...
 	resp.Schema = conversion.DataSourceSchemaFromResource(ResourceSchema(ctx), &conversion.DataSourceSchemaRequest{
-		RequiredFields: []string{"project_id", "name"},
+		RequiredFields:  []string{"project_id", "name"},
+		OverridenFields: dataSourceOverridenFields(),
 	})
 }
 
+func dataSourceOverridenFields() map[string]dsschema.Attribute {
+	return map[string]dsschema.Attribute{
+		"delete_on_create_timeout": nil,
```
oarbusi, line 36: do as timeout, remove it by default in `DataSourceSchemaFromResource`

internal/service/flexcluster/resource.go
```diff
@@ -221,6 +225,11 @@ func CreateFlexCluster(ctx context.Context, projectID, clusterName string, flexC
 	return flexClusterResp, nil
 }
 
+// TODO: keeping CreateFlexCluster to avoid changes in adv_cluster and running their acc tests while doing this PR, remove before merging.
```
oarbusi, line 228: create CLOUDP to handle this before finishing epic

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -462,7 +462,7 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 	if isFlex {
 		flexClusterReq := advancedclustertpf.NewFlexCreateReq(clusterName, d.Get("termination_protection_enabled").(bool), conversion.ExpandTagsFromSetSchema(d), replicationSpecs)
-		flexClusterResp, err := flexcluster.CreateFlexCluster(ctx, projectID, clusterName, flexClusterReq, connV2.FlexClustersApi)
+		flexClusterResp, err := flexcluster.CreateFlexCluster(ctx, projectID, clusterName, flexClusterReq, connV2.FlexClustersApi, &timeout)
```
marcosuma, line 465: why are we passing timeout only here and not for delete and update?

Also this method is called from advanced_cluster. Are we changing something also on that resource?

internal/service/flexcluster/resource_test.go
```diff
@@ -26,6 +26,26 @@ func TestAccFlexClusterRS_failedUpdate(t *testing.T) {
...
+	var (
+		projectID   = acc.ProjectIDExecution(t)
+		clusterName = acc.RandomName()
+		provider    = "AWS"
+		region      = "US_EAST_1"
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyFlexCluster, // resource is deleted when creation times out
```
marcosuma: bear with me, but I think the destroy always happens at the end of a test, right? (I am assuming this because of my memories and also reading the tests below. 
If so, how can we make sure we actually really deleted because of the business logic?

internal/service/flexcluster/resource_test.go
```diff
@@ -122,6 +142,25 @@ func configBasic(projectID, clusterName, provider, region string, terminationPro
...
+		resource "mongodbatlas_flex_cluster" "test" {
+			project_id = %[1]q
+			name       = %[2]q
+			provider_settings = {
+				backing_provider_name = %[3]q
+				region_name           = %[4]q
+			}
+			delete_on_create_timeout = %[5]t
+			timeouts = {
+				create = "1s"
```
marcosuma: nit: not sure if I am missing something, but because of `MinTimeout: 3 * time.Second` I guess this is "too fast", right?

internal/common/customplanmodifier/create_only.go
```diff
@@ -0,0 +1,51 @@
...
+func (d *createOnlyAttributePlanModifier) PlanModifyBool(ctx context.Context, req planmodifier.BoolRequest, resp *planmodifier.BoolResponse) {
+	validateCreateOnly(req.PlanValue, req.StateValue, req.Path, &resp.Diagnostics)
+}
+
+// validateCreateOnly checks if an attribute value has changed and adds an error if it has
+func validateCreateOnly(planValue, stateValue attr.Value, attrPath path.Path, diagnostics *diag.Diagnostics,
+) {
+	if !stateValue.IsNull() && !stateValue.Equal(planValue) {
+		diagnostics.AddError(
+			fmt.Sprintf("%s cannot be updated", attrPath),
```
EspenAlbert, line 47: [nit] Would improve the details by showing the plan & stateValue.

internal/service/advancedcluster/resource_advanced_cluster.go
```diff
@@ -1540,7 +1540,7 @@ func resourceUpdateFlexCluster(ctx context.Context, flexUpdateRequest *admin.Fle
 	projectID := ids["project_id"]
 	clusterName := ids["cluster_name"]
 
-	_, err := flexcluster.UpdateFlexCluster(ctx, projectID, clusterName, flexUpdateRequest, connV2.FlexClustersApi)
+	_, err := flexcluster.UpdateFlexCluster(ctx, projectID, clusterName, flexUpdateRequest, connV2.FlexClustersApi, nil)
```
EspenAlbert: Why are we not passing the timeout? (Passing `nil` it looks like)

internal/service/flexcluster/resource.go
```diff
@@ -203,7 +250,43 @@ func splitFlexClusterImportID(id string) (projectID, clusterName *string, err er
 	return
 }
 
-func CreateFlexCluster(ctx context.Context, projectID, clusterName string, flexClusterReq *admin.FlexClusterDescriptionCreate20241113, client admin.FlexClustersApi) (*admin.FlexClusterDescription20241113, error) {
+// resolveTimeout extracts the appropriate timeout duration from the model for the given operation
+func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string, diags *diag.Diagnostics) time.Duration {
```
EspenAlbert: This looks almost identical to:
[func resolveTimeout(ctx context.Context, t *timeouts.Value, operationName string, diags *diag.Diagnostics) time.Duration {](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/342d64a23d3ceb20292cbf69c296fa4c0e3b13ea/internal/service/advancedclustertpf/resource.go#L509)

Can we try to refactor?

internal/service/flexcluster/resource_test.go
```diff
@@ -122,6 +142,25 @@ func configBasic(projectID, clusterName, provider, region string, terminationPro
...
+				update = "1s"
```
EspenAlbert: why update and delete?

marcosuma (approved):

internal/service/flexcluster/resource_test.go
oarbusi: We want to always timeout in this test, so if we set the timeout to less than minTimeout(Smallest time to wait before refreshes), the first refresh it will have already timed out

internal/service/flexcluster/resource.go
oarbusi: good point, done in [3cfc917](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/3cfc917f11e898a99a15d0dd564863368f9e3567)

internal/service/flexcluster/resource_test.go
oarbusi: right, they are not necessary in this test. Removed in [bb7c257](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/bb7c2570438c8711eb7dc8f97868849d2d113709)

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: good catch, fixed in [56b6dc8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/56b6dc8da2d15cb659e646082e0d4fc8c3ea3a1d)

internal/service/advancedcluster/resource_advanced_cluster.go
oarbusi: good catch, fixed in [56b6dc8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/56b6dc8da2d15cb659e646082e0d4fc8c3ea3a1d).
We are doing changes in advanced cluster to support both timeout and delete_on_create_timeout for flex clusters in advanced cluster resource

internal/service/flexcluster/resource_test.go
oarbusi: good point, I have added this explicit check in [b0d39a8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/b0d39a851e34732d8c6b42e7485bb3224994dbe5)

internal/service/flexcluster/resource_schema.go
```diff
@@ -145,20 +146,52 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				},
 				MarkdownDescription: "Method by which the cluster maintains the MongoDB versions.",
 			},
+			"delete_on_create_timeout": schema.BoolAttribute{
+				Optional: true,
+				PlanModifiers: []planmodifier.Bool{
+					customplanmodifier.CreateOnlyBoolPlanModifier(),
+				},
+
+				MarkdownDescription: "Indicates whether to delete the resource if creation times out. Default is `true`. When Terraform apply fails, it returns immediately without waiting for cleanup to complete. If you suspect a transient error, wait before retrying to allow resource deletion to finish.",
```
AgustinBettati, line 155:
```suggestion
				MarkdownDescription: "Indicates whether to delete the created resource if a timeout is reached when waiting for completion. Default is `true`. If you suspect a transient error, wait before retrying to allow resource deletion to finish.",
```
The 3rd sentence was confusing to me, would just leave as one description.

internal/service/flexcluster/state_transition.go
```diff
@@ -40,12 +40,16 @@ func WaitStateTransition(ctx context.Context, requestParams *admin.GetFlexCluste
 	return nil, errors.New("did not obtain valid result when waiting for flex cluster state transition")
 }
 
-func WaitStateTransitionDelete(ctx context.Context, requestParams *admin.GetFlexClusterApiParams, client admin.FlexClustersApi) error {
+func WaitStateTransitionDelete(ctx context.Context, requestParams *admin.GetFlexClusterApiParams, client admin.FlexClustersApi, timeout *time.Duration) error {
+	deleteTimeout := constant.DefaultTimeout
+	if timeout != nil {
+		deleteTimeout = *timeout
+	}
```
AgustinBettati: for both WaitStateTransition and WaitStateTransitionDelete would avoid receiving timeout as a pointer and instead received a know value directly. In `resolveTimeout` function we are already ensuring timeout will be defined and have logic for defining the default value.

internal/service/flexcluster/resource_test.go
```diff
@@ -150,3 +193,19 @@ func checksFlexCluster(projectID, clusterName string, terminationProtectionEnabl
 	checks = acc.AddAttrChecks(dataSourcePluralName, checks, pluralMap)
 	return acc.CheckRSAndDS(resourceName, &dataSourceName, &dataSourcePluralName, attrSet, attrMap, checks...)
 }
+
+// checkClusterNotExists verifies that the flex cluster does not exist,
+// confirming that delete_on_create_timeout logic successfully cleaned up the resource
+func checkClusterNotExists(projectID, clusterName string) resource.TestCheckFunc {
```
AgustinBettati: not certain how long the delete operation takes, but would have expected the GET still responds but with a status of deleting. Can you confirm that when running this test the POST is actually sent?

AgustinBettati (approved):

internal/service/flexcluster/resource_schema.go
oarbusi: This is following the description in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3515/files#diff-e9d48cdf6bde5c047a1a852d6c3275175fe0a3d47f7ab543facdf8663b57fc57R334 which was reviewed by docs team.
I am open to changing it, but I would loop in docs team to review this and then change it in all resources.

internal/service/flexcluster/resource_test.go
oarbusi: Not sure if I understand what you mean by "POST is actually sent". The test triggers the creation(and POST request is sent), then it times out (due to timeout being 1s) and then DELETE is sent because delete_on_create_timeout is true

AgustinBettati: Gave this a quick debug run and can see the the check `checkClusterNotExists` is never run, this is likely because the step received the error and only asserts `ExpectError`. My doubt originally came because I expect the delete operation to take some seconds/minutes, with the flex cluster in a `DELETING` status.

internal/service/flexcluster/state_transition.go
oarbusi: makes sense, changed in [55f67aa](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/55f67aa9146420be990cf68d92cac92fdb8132d9)

maastha (approved):

internal/service/advancedclustertpf/resource.go
```diff
@@ -620,7 +597,7 @@ func handleFlexUpdate(ctx context.Context, diags *diag.Diagnostics, client *conf
 	clusterName := plan.Name.ValueString()
 	flexCluster, err := flexcluster.UpdateFlexCluster(ctx, plan.ProjectID.ValueString(), clusterName,
 		GetFlexClusterUpdateRequest(configReq.Tags, configReq.TerminationProtectionEnabled),
-		client.AtlasV2.FlexClustersApi)
+		client.AtlasV2.FlexClustersApi, nil)
```
marcosuma: timeouts here?

internal/service/flexcluster/resource_test.go
marcosuma: @oarbusi for these tests, it's very easy to assume they do things - it's a good practice to "make the test fail" first (E.g. you could check something you know it's wrong and expect the test to fail), or debug like Agustin did.

oarbusi: Good point, I missed the check not being run. It should be enough to have the error check to make sure that the resource has been deleted because delete_on_create_timeout is true. Removed it in [8d2d1f0](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/8d2d1f01bae27c11bd0e1275a90c92ba6c4bd80a)

marcosuma: @oarbusi I think now you went back to the previous code where I left the question: how do we know the delete happened because of our business logic or because the cleanup phase of the test? am I missing something? you're now only checking the text of the error

oarbusi: validating that the error contains "will run cleanup because delete_on_create_timeout is true" is a valid way of making sure that the resource has been deleted because delete_on_create_timeout is true. This is also done for the other resource https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3515

internal/service/advancedclustertpf/resource.go
oarbusi: fixed in [37b192a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/37b192a9123768784055a8d9120701dfab6723d7)

internal/service/flexcluster/resource_test.go
oarbusi: Can't explicitly check this in a step check because the apply actually fails. Validating the error message should be enough to make sure the clean up is being done because delete_on_create_timeout is true

internal/service/advancedclustertpf/resource.go
```diff
@@ -612,15 +589,15 @@ func handleFlexUpgrade(ctx context.Context, diags *diag.Diagnostics, client *con
...
+func handleFlexUpdate(ctx context.Context, diags *diag.Diagnostics, client *config.MongoDBClient, waitParams *ClusterWaitParams, plan *TFModel) *TFModel {
 	configReq := normalizeFromTFModel(ctx, plan, diags, false)
 	if diags.HasError() {
 		return nil
 	}
 	clusterName := plan.Name.ValueString()
 	flexCluster, err := flexcluster.UpdateFlexCluster(ctx, plan.ProjectID.ValueString(), clusterName,
 		GetFlexClusterUpdateRequest(configReq.Tags, configReq.TerminationProtectionEnabled),
-		client.AtlasV2.FlexClustersApi)
+		client.AtlasV2.FlexClustersApi, waitParams.Timeout)
```
marcosuma, line 600: non-blocking, but you've got a couple of comments telling you were missing to pass the timeouts object. This hints at the fact that this code is not tested, otherwise some test should have failed.

Without blocking this PR, can you add in the epic a task to add tests that prove input timeout fields are used in the resource when calling update/delete?

internal/service/flexcluster/resource_test.go
marcosuma: ok I think it's a fair compromise because I can see we're unit testing the link between the error text and the function call in `TestCleanupOnErrorCalledForATimeout` .

internal/service/advancedclustertpf/resource.go
oarbusi: I see what you mean. Initially I wanted to cover only `mongodbatlas_flex_cluster` resource in this PR, that is why I left those calls without timeout(the missing timeouts where on the advanced_cluster resource calls), but eventually they will need to be done for consistency, so I added them here. Will add a tests in advanced_cluster for this

marcosuma: I am not sure it's only advanced cluster. What I mean is that there should be a test somewhere, for both resources, that proves that it's not just the default timeout that gets applied, but we're taking in consideration the inputs from the customer for create,delete and update.

it's not a blocker, but given two people made you notice there was some `nil` passed instead of `&timeout` and no test failed, I am just being mindful of potential coverage gaps.

oarbusi: sure, will add it in this PR

oarbusi: added the tests in [13b41ce](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/13b41ce9afd820a16c1203e709a202d941367787)

internal/service/advancedcluster/resource_advanced_cluster_test.go
```diff
@@ -3262,39 +3233,93 @@ func configFlexCluster(t *testing.T, projectID, clusterName, providerName, regio
...
+		deleteOnCreateTimeout = true
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyFlexCluster,
+		Steps: []resource.TestStep{
+			{
+				Config:      configFlexCluster(t, projectID, clusterName, "AWS", "US_EAST_1", "", acc.TimeoutConfig(&createTimeout, nil, nil, false), false, &deleteOnCreateTimeout),
+				ExpectError: regexp.MustCompile("context deadline exceeded"), // with the current implementation, this is the error that is returned
```
oarbusi, line 3285: Haven't changed how `delete_on_create_timeout` is handled in advanced_cluster, this is why this is the error we get.

internal/service/flexcluster/resource_test.go
```diff
@@ -26,29 +26,83 @@ func TestAccFlexClusterRS_failedUpdate(t *testing.T) {
...
+		clusterName           = acc.RandomName()
+		provider              = "AWS"
+		region                = "US_EAST_1"
+		createTimeout         = "1s"
+		deleteOnCreateTimeout = true
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             acc.CheckDestroyFlexCluster, // verify cleanup was effective
```
AgustinBettati: Noticed that CheckDestroy is not being run, likely because the last step ends with an error. Can you confirm? If that the case I am struggling to see how we can assert that the delete call is effectively being made.

oarbusi: Actually, in this test it is expected for the checkDestroy to not be run and is redundant. Because the resource is not successfully created due to the create timeout being reached, we trigger the delete on create.
 Checking that the error is "will run cleanup because delete_on_create_timeout is true" is enough for us to verify that the delete is triggered

oarbusi: I have removed the redundant checkDestroy in [d1d777c](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/d1d777cba675942ceaeeb8c97aed450e77fb4193) and improved the test name in [66c72ef](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3525/commits/66c72efffea3ae6fcd55571f65b4cb53cbc16e1e)

AgustinBettati: Thanks, yes I agree the specific error message does give us some form of verification of the cleanup.

AgustinBettati (approved):
LGTM

PR #3536, opened 2025-07-30
feat: Long-running operation improvements for `mongodbatlas_cloud_backup_snapshot` resource

oarbusi:
## Description

Long-running operation improvements for `mongodbatlas_cloud_backup_snapshot` resource
- Adds `delete_on_create_timeout` attribute to indicate whether to delete the resource if its creation times out

Link to any related issue(s): CLOUDP-334355

## Type of change:

## Required Checklist:


internal/service/cloudbackupsnapshot/resource.go
```diff
@@ -121,10 +122,20 @@ func Resource() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"delete_on_create_timeout": { // Don't use Default: true to avoid unplanned changes when upgrading from previous versions.
+				Type:        schema.TypeBool,
+				Optional:    true,
+				ForceNew:    true,
```
oarbusi, line 128: necessary because if not, we encounter the following error:
```
 Internal validation of the provider failed! This is always a bug
  | with the provider itself, and not a user issue. Please report
  | this bug:
  | 
  | resource mongodbatlas_cloud_backup_snapshot: No Update defined, must set ForceNew on: []string{"delete_on_create_timeout"}
```

marcosuma: > to avoid unplanned changes when upgrading from previous versions

considering we're releasing this in 2.0.0, isn't it ok?

marcosuma: > necessary because if not, we encounter the following error:

and why are we encountering this error? it says "no update defined" but I am not clear what that means. (and will this happen to all SDKv2?)

internal/service/cloudbackupsnapshot/resource.go
```diff
@@ -155,12 +166,20 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		Target:     []string{"completed", "failed"},
 		Refresh:    resourceRefreshFunc(ctx, requestParams, connV2),
 		Timeout:    d.Timeout(schema.TimeoutCreate) - time.Minute,
-		MinTimeout: 60 * time.Second,
-		Delay:      1 * time.Minute,
+		MinTimeout: minTimeout,
+		Delay:      minTimeout,
```
marcosuma: even though I see we're not consistent across resources, why are we setting the same value? it looks like it's usually different (but again, not sure based on what we choose it). It also seems weird from the name of the variable

marcosuma (approved):

internal/service/cloudbackupsnapshot/resource.go
oarbusi: we discussed this in the previous resource https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3515#discussion_r2224884939 
We want to simplify minTimeout and Delay as previous implementation had mixed values for create, update, delete for no apparent reason. Having the same value makes sense because Delay is the time to wait before the first check, MinTimeOut is the time between checks. I think this is okay in order to have more simplicity. Let me know if you think otherwise

internal/service/cloudbackupsnapshot/resource.go
oarbusi: this resource can't be updated and there is no update operation defined, that is why we have this error

oarbusi: > considering we're releasing this in 2.0.0, isn't it ok?

yes it would be ok. I discussed this with @lantoli last week after his PR(https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3515) doing this. He preferred and went with the current approach
There are 2 options:
- No default in the schema, but we do the default in the code(current approach)
  - no plan changes
  - default behavior does not impact the plan or state, avoiding friction on customers
- Default in the schema
  - this would cause plan changes when upgrading
    - it would be "fine" because it's a major version
  -  for already created resources, users would have this unplanned changed in their plan, related to the create operation, which is no longer relevant in that case and could cause confusion or friction
  - more verbosity in the plan even though for the majority of users it wouldn't be relevant

I was in favor of having the default on the schema but in the end I changed my mind, I think we can provide a better UX with almost no impact on the code and future autogeneration efforts: only difference in the code if we don't have the default in the schema is [L173 ](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3536/files#diff-77f40b859156979f1f239e0ab190e7fac344bb38dee4fedf8358e3246478b4eaR173) of this same file

Let me know what you think

internal/service/cloudbackupsnapshot/resource.go
```diff
@@ -121,10 +122,20 @@ func Resource() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"delete_on_create_timeout": { // Don't use Default: true to avoid unplanned changes when upgrading from previous versions.
```
EspenAlbert, line 125: Why not use the CreateOnly plan modifier?

internal/service/cloudbackupsnapshot/resource.go
marcosuma: > for already created resources, users would have this unplanned changed in their plan, related to the create operation, which is no longer relevant in that case and could cause confusion or friction

this is a good point. sgtm

internal/service/cloudbackupsnapshot/resource.go
```diff
@@ -155,12 +166,20 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
-	_, err = stateConf.WaitForStateContext(ctx)
-	if err != nil {
-		return diag.FromErr(err)
+	_, errWait := stateConf.WaitForStateContext(ctx)
+	deleteOnCreateTimeout := true // default value when not set
+	if v, ok := d.GetOkExists("delete_on_create_timeout"); ok {
+		deleteOnCreateTimeout = v.(bool)
+	}
+	errWait = cleanup.HandleCreateTimeout(deleteOnCreateTimeout, errWait, func(ctxCleanup context.Context) error {
+		_, errCleanup := connV2.CloudBackupsApi.DeleteReplicaSetBackup(ctxCleanup, groupID, clusterName, snapshot.GetId()).Execute()
```
EspenAlbert, line 178: Quite interesting how the name use `ReplicaSet` (DeleteReplicaSetBackup).
Looks like there is equvialent methods for Sharded (`DeleteShardedClusterBackup`)... Are we only supporting Replicaset?

internal/service/cloudbackupsnapshot/resource.go
marcosuma: I just found it weird that `minTimeout` is also for the `Delay` property, perhaps we can call the const variable with a name of how much time it represents? e.g oneMinute (not 100% confident about Go best practices, but that could be an option)

internal/service/cloudbackupsnapshot/resource.go
oarbusi: because this is an SDKv2 resource

internal/service/cloudbackupsnapshot/resource.go
oarbusi: It seems that although this specific method is used in the delete, both are supported and it works for both SHARDED and REPLICASET. There are tests for both cases: `TestAccBackupRSCloudBackupSnapshot_sharded` and `TestAccBackupRSCloudBackupSnapshot_basic` (replicaset)

internal/service/cloudbackupsnapshot/resource.go
oarbusi: Good idea, changed to `oneMinute` to improve clarity in [b1fa80c](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3536/commits/b1fa80c231075ae2bfb537e1f9998d61d7f5b9b3)

PR #3543, opened 2025-07-31
feat: Long-running operation improvements for `mongodbatlas_privatelink_endpoint` resource 

oarbusi:
## Description

Long-running operation improvements for `mongodbatlas_privatelink_endpoint` resource

- Adds `delete_on_create_timeout` attribute to indicate whether to delete the resource if its creation times out

Link to any related issue(s): CLOUDP-334365

## Type of change:

## Required Checklist:


internal/service/privatelinkendpoint/resource.go
```diff
@@ -134,15 +143,22 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
-	_, err = stateConf.WaitForStateContext(ctx)
-	if err != nil {
-		return diag.FromErr(fmt.Errorf(errorPrivateLinkEndpointsCreate, err))
+	_, errWait := stateConf.WaitForStateContext(ctx)
+	deleteOnCreateTimeout := true // default value when not set
+	if v, ok := d.GetOkExists("delete_on_create_timeout"); ok {
+		deleteOnCreateTimeout = v.(bool)
+	}
+	errWait = cleanup.HandleCreateTimeout(deleteOnCreateTimeout, errWait, func(ctxCleanup context.Context) error {
+		return deletePrivateEndpointAndWait(ctxCleanup, connV2, projectID, providerName, privateEndpoint.GetId(), d.Timeout(schema.TimeoutDelete))
```
oarbusi: if we don't wait for the deletion to complete, tests fail
```
Error: error when destroying resource
        
        error deleting project (688b8c1fe0bf0a19dae81081):
        https://cloud-dev.mongodb.com/api/atlas/v2/groups/688b8c1fe0bf0a19dae81081
        DELETE: HTTP 409 Conflict (Error code:
        "CANNOT_CLOSE_GROUP_ACTIVE_ATLAS_PRIVATE_ENDPOINT_SERVICES") Detail: Cannot
        close group while it has active private endpoint services; please terminate
        all private endpoint services. Reason: Conflict. Params: [],
        BadRequestDetail: 
```
this could happen in production as well if users want to remove projects that failed to create a privatelink_endpoint but have not yet been deleted

marcosuma: if we're waiting, we then need to change the docs for this field?

marcosuma: what you found here though looks like a general issue. Isn't it the same for cluster for example? a project cannot be deleted if a cluster is still there.

oarbusi: Included in CLOUDP-335200 the fact that we will need to document this behaviour in the delete_on_create_timeout field. 
I have found 2 different issues when implementing this:
- Case 1: Deletion fails if resource has not reached a certain state
- Case 2: Resource A cannot be deleted if it has a Resource B
  - If create of Resource B fails, it won't be stored in the state and not managed by TF, making it difficult for users to delete that resource(go into UI, import resource in TF and then delete)

For both cases, I think we should try to delete the resource and not leave it dangling. Because this resources take some time to be deleted, this waiting logic is needed in order to make sure that it's deleted

oarbusi: As per the last discussion, I changed it so that the resourec doesn't wait for the delete on create to complete

marcosuma: not urgent, but it'd be good to leave a comment here with a why we're not waiting

oarbusi: we don't wait for the deletion on create timeout because:
- Avoid over extending from initial timeout user defines for an operation.
- If CREATE is timing out, it is possible it happens due to capacity issues or something similar so waiting for DELETE to complete can be unreliable in that scenario.

internal/service/privatelinkendpoint/resource.go
```diff
@@ -134,15 +143,23 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		Pending:    []string{"INITIATING", "DELETING"},
 		Target:     []string{"WAITING_FOR_USER", "FAILED", "DELETED", "AVAILABLE"},
 		Refresh:    refreshFunc(ctx, connV2, projectID, providerName, privateEndpoint.GetId()),
-		Timeout:    d.Timeout(schema.TimeoutCreate) - time.Minute, // If using a CRUD function with a timeout, any StateChangeConf timeouts should be configured below that duration to avoid returning the SDK context: deadline exceeded error instead of the retry logic error.
```
AgustinBettati, line 137: nit: I believe we can remove the - 1 now. The `- time.Minute` is relevant when using `CreateContext` which handles the timeout in the main context and can lead to the error `context: deadline exceeded` ([reference](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#statechangeconf)). Since we are using `CreateWithoutTimeout` we wont have the risk of getting a  `deadline exceeded error`.

AgustinBettati (approved):
LGTM

carriecwk (approved):
lgtm!

PR #3568, opened 2025-08-07
feat: Add new resource mongodbatlas_cloud_user_project_assignment

csanx:
## Description

Add new resource: `mongodbatlas_cloud_user_project_assignment`

Link to any related issue(s): [CLOUDP-333176](https://jira.mongodb.org/browse/CLOUDP-333176)

## Type of change:

## Required Checklist:


internal/service/clouduserprojectassignment/resource.go
```diff
@@ -0,0 +1,242 @@
...
+
+	newCloudUserProjectAssignmentModel, diags := NewTFModel(ctx, projectID, apiResp)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, newCloudUserProjectAssignmentModel)...)
+}
+
+func fetchProjectUser(ctx context.Context, connV2 *admin.APIClient, projectID, userID, username string) (*admin.GroupUserResponse, error) {
```
csanx, line 76: `fetchProjectUser` function is introduced here and reused in #3569 to extract common functionality between the resource and the data source.

internal/service/clouduserprojectassignment/model.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, role := range rolesToRemove {
+		removeRequests = append(removeRequests, &admin.AddOrRemoveGroupRole{
+			GroupRole: role,
+		})
+	}
+
+	return addRequests, removeRequests, nil
+}
+
+func diffRoles(oldRoles, newRoles []string) (toAdd, toRemove []string) {
```
marcosuma, line 78: not urgent, can be done in a follow-up ticket - but this method should have its unit test, or you can have a test for `NewAtlasUpdateReq` that checks it always generates the right request based on diff

internal/service/clouduserprojectassignment/resource.go
```diff
@@ -0,0 +1,242 @@
...
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Error adding role %s to user(%s) in ProjectID(%s):", addReq.GroupRole, username, projectID),
+				err.Error(),
+			)
+			return
+		}
+	}
+
+	for _, removeReq := range removeRequests {
+		_, _, err := connV2.MongoDBCloudUsersApi.RemoveProjectRole(ctx, projectID, userID, removeReq).Execute()
```
marcosuma, line 182: not blocking - if this API fails, the update will be partially applied (only the addprojectRole). This should be documented at minimum, but also we should check if we're consistent in similar situations

cc @maastha

internal/service/clouduserprojectassignment/resource.go
```diff
@@ -0,0 +1,242 @@
...
+	ok, parts := conversion.ImportSplit(req.ID, 2)
+	if !ok {
+		resp.Diagnostics.AddError("invalid import ID format", "expected 'project_id/user_id' or 'project_id/username', got: "+importID)
+		return
+	}
+	projectID, user := parts[0], parts[1]
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("project_id"), projectID)...)
+
+	emailRegex := regexp.MustCompile(`^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$`)
```
marcosuma, line 245: not blocking - we don't have any util method that does that already?

internal/service/clouduserprojectassignment/model_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+func TestNewProjectUserRequest(t *testing.T) {
+	ctx := t.Context()
+	expectedRoles, _ := types.SetValueFrom(ctx, types.StringType, testProjectRoles)
+
+	testCases := map[string]struct {
+		plan     *clouduserprojectassignment.TFModel
+		expected *admin.GroupUserRequest
+	}{
+		"Complete model": {
```
maastha, line 148: also add test cases for empty/nil values (for all unit tests)

internal/service/clouduserprojectassignment/resource.go
maastha: by documented, do you mean add a comment here or mention something in the public documentation?

internal/service/clouduserprojectassignment/model.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	addProjectUserReq := admin.GroupUserRequest{
+		Username: plan.Username.ValueString(),
+		Roles:    roleNames,
+	}
+	return &addProjectUserReq, nil
+}
+
+func NewAtlasUpdateReq(ctx context.Context, plan, state *TFModel) (addRequests, removeRequests []*admin.AddOrRemoveGroupRole, diags diag.Diagnostics) {
```
maastha: I'm thinking if instead of getting roles from the state here, would it be better to compare against what is returned by the API instead? That ways any partially applied calls can be handled correctly WDYT?

internal/service/clouduserprojectassignment/resource.go
marcosuma: I think we should publicly document it? because I can see a question coming, like: "the update failed but then I logged in and saw some users were added ..."

internal/service/clouduserprojectassignment/resource_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/testutil/acc"
+)
+
+var resourceName = "mongodbatlas_cloud_user_project_assignment.test"
+
+func TestAccCloudUserProjectAssignmentRS_basic(t *testing.T) {
+	resource.ParallelTest(t, *basicTestCase(t))
+}
+
+func basicTestCase(t *testing.T) *resource.TestCase {
```
maastha, line 29: should we also have a test case to import an existing (ACTIVE) project user? is that possible?

internal/service/clouduserprojectassignment/model.go
csanx: Do you mean making another API call before calling this method in the Update?

internal/service/clouduserprojectassignment/model.go
csanx: Added a test for `NewAtlasUpdateReq`

jwilliams-mongo (commented):
@csanx are there any docs updates for this feature? If not, LGTM

jwilliams-mongo:
> @csanx are there any docs updates for this feature? If not, LGTM

I see https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3578/files so I'll review that and approve this

internal/service/clouduserprojectassignment/resource_schema.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			"first_name": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "First or given name that belongs to the MongoDB Cloud user.",
+			},
+			"project_id": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "Unique 24-hexadecimal digit string that identifies your project. Use the [/groups](#tag/Projects/operation/listProjects) endpoint to retrieve all projects to which the authenticated user has access.\n\n**NOTE**: Groups and projects are synonymous terms. Your group id is the same as your project id. For existing groups, your group/project id remains the same. The resource and corresponding endpoints use the term groups.",
+				Validators: []validator.String{
+					stringvalidator.RegexMatches(regexp.MustCompile("^([a-f0-9]{24})$"), ""),
```
AgustinBettati: Would remove the regex validation just for consistency with other resources

internal/service/clouduserprojectassignment/resource.go
```diff
@@ -0,0 +1,242 @@
...
+		resp.Diagnostics.AddError(fmt.Sprintf("error deleting user(%s) from ProjectID(%s):", username, projectID), err.Error())
+		return
+	}
+}
+
+func (r *rs) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	importID := req.ID
+	ok, parts := conversion.ImportSplit(req.ID, 2)
+	if !ok {
+		resp.Diagnostics.AddError("invalid import ID format", "expected 'project_id/user_id' or 'project_id/username', got: "+importID)
```
AgustinBettati: q: Why do we support import with `user_id` as well? Does it help with a particular migration journey? Asking as this does add some complexity to our read operation.

internal/service/clouduserprojectassignment/resource_migration_test.go
```diff
@@ -0,0 +1,138 @@
...
+					},
+				},
+				Config: removeProjectInvitationConfigThird(username, projectName, orgID, roles),
+			},
+			mig.TestStepCheckEmptyPlan(removeProjectInvitationConfigThird(username, projectName, orgID, roles)),
+		},
+	})
+}
+
+func originalConfigFirst(username, projectName, orgID string, roles []string) string {
```
AgustinBettati:
```suggestion
func legacyProjectInvitationConfig(username, projectName, orgID string, roles []string) string {
```
Would make more explicit this is using the deprecated resource

internal/service/clouduserprojectassignment/resource_migration_test.go
```diff
@@ -0,0 +1,138 @@
...
+			project_id  = mongodbatlas_project.mig_test.id
+			username    = local.username
+			roles       = local.roles
+		}
+
+		resource "mongodbatlas_cloud_user_project_assignment" "user_mig_test" {
+			project_id = mongodbatlas_project.mig_test.id
+			username   = local.username
+			roles      = local.roles
+		}
```
AgustinBettati, line 106: Can `user_project_assignment` be created with `project_invitation` already existing? I suppose this is how the API behaves but wanted to confirm.

internal/service/clouduserprojectassignment/resource_migration_test.go
```diff
@@ -0,0 +1,138 @@
...
+		resource "mongodbatlas_project" "mig_test" {
+			name 	= %[3]q
+			org_id  = %[4]q
+		}
+
+		resource "mongodbatlas_cloud_user_project_assignment" "user_mig_test" {
+			project_id = mongodbatlas_project.mig_test.id
+			username   = local.username
+			roles      = local.roles
+		}
```
AgustinBettati, line 127: When `project_invitation` is deleted it has not impact on `cloud_user_project_assignment`?

AgustinBettati (commented):
LGTM, leaving some doubts which seem more API related

internal/service/clouduserprojectassignment/resource.go
csanx: There's a method for this but in the acceptance tests package. Since this same regex is duplicated across 2 of the other new resources, we could consider extracting it into a shared utility function in a follow-up ticket, WDYT?

csanx:
@jwilliams-mongo Yes, my upcoming PRs will add documentation for all new resources and data sources. The one you mention is related to another resource actually

internal/service/clouduserprojectassignment/model_test.go
csanx: Done [59950ba](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3568/commits/59950ba14111760da2db4b341de818fcbc15433e)

internal/service/clouduserprojectassignment/resource_migration_test.go
csanx: Yes, after running this, the state shows:
```
{
      "mode": "managed",
      "type": "mongodbatlas_project_invitation",
      "name": "mig_test",
      ...
    },
{
      "mode": "managed",
      "type": "mongodbatlas_cloud_user_project_assignment",
      "name": "user_mig_test",
      ...
    },
```
We can also see in Cloud Dev that the same user is invited twice.

internal/service/clouduserprojectassignment/resource_migration_test.go
csanx: No, they act as two independent invitations. Deleting one just removes that specific invitation.

internal/service/clouduserprojectassignment/resource.go
marcosuma: yup, no problem

internal/service/clouduserprojectassignment/resource_test.go
csanx: Done

internal/service/clouduserprojectassignment/model.go
maastha: yes

internal/service/clouduserprojectassignment/resource.go
maastha: yes but after failure when they run terraform plan again they will see the exact status of what roles were added vs which were not so I think that should suffice

marcosuma (approved):

internal/service/clouduserprojectassignment/resource.go
marcosuma: ok, you have a fair point with this one. I'd say an extra clarification in the doc doesn't hurt, especially in case of customers reaching out. But it's a nice to have.

internal/service/clouduserprojectassignment/resource_migration_test.go
AgustinBettati: I see, thanks. This sentence for the Tech design helped me understand this migration journey, potentially we can leave it as a comment:

> we only need to consider the migration path for mongodbatlas_project_invitation resources with a PENDING invite (similar to org_invites) - which are not discoverable by the new APIs - so the only way to migrate here is to re-create the pending users.

internal/service/clouduserprojectassignment/resource.go
csanx: It’s mainly for consistency, since we provide both keys in all Cloud User resources

internal/service/clouduserprojectassignment/resource_migration_test.go
csanx: Got it, I’ll take that into account

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -601,8 +602,10 @@ jobs:
           MONGODB_ATLAS_PROJECT_OWNER_ID: ${{ inputs.mongodb_atlas_project_owner_id }}
           MONGODB_ATLAS_ORG_ID: ${{ inputs.mongodb_atlas_org_id }}
           MONGODB_ATLAS_USERNAME: ${{ vars.MONGODB_ATLAS_USERNAME }}
+          MONGODB_ATLAS_USERNAME_2: ${{ vars.MONGODB_ATLAS_USERNAME_2 }}
```
maastha, line 605: is there any difference between MONGODB_ATLAS_USERNAME & MONGODB_ATLAS_USERNAME_2? or we're just adding for more test cases? thinking if we should add a comment here on what MONGODB_ATLAS_USERNAME_2 signifies

jwilliams-mongo (approved):

maastha (approved):

.github/workflows/acceptance-tests-runner.yml
csanx: I believe `MONGODB_ATLAS_USERNAME` has the Org Owner role, so if no Project Owner is specified, that user is assigned automatically. Then, when I try to use the `user_project_assignment` resource, I get this error:
`POST: HTTP 400 Bad Request (Error code: "USER_ALREADY_IN_GROUP")`
That’s why I used `MONGODB_ATLAS_USERNAME_2` instead. I’ll add a comment in the code to clarify this.

PR #3552, opened 2025-08-05
doc: Updates `delete_on_create_timeout` description as suggested and use it in the examples

oarbusi:
## Description

- Updates `delete_on_create_timeout` description and use it in the examples
- Includes `delete_on_create_timeout` addition in the 2.0 upgrade guide

Link to any related issue(s): CLOUDP-335200

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [x] Documentation fix/enhancement

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [ ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals I have added appropriate changelog entries.
- [ ] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

PR #3595, opened 2025-08-19
feat: Changes default value of `delete_on_create_timeout` to true in `mongodbatlas_advanced_cluster` and `mongodbatlas_search_deployment` resources

oarbusi:
## Description

Changes default value of `delete_on_create_timeout` to true in `mongodbatlas_advanced_cluster` and `mongodbatlas_search_deployment` resources.
- Documentation changes were done in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3552
- For `mongodbatlas_search_deployment` resource, implementation of the `delete_on_create_timeout` logic has been changed to use the one we are using in the other resources. This is to unify the implementation and be consistent across the resources where we can. In `mongodbatlas_advanced_cluster` is not possible because the creation logic is more complex and there are multiple API calls, making it necessary to use the `ReplaceContextDeadlineExceededDiags` and `deferCall` implementation

Link to any related issue(s): CLOUDP-335991

## Type of change:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.

## Required Checklist:

### DIFF
--- a/.changelog/3595.txt
+++ b/.changelog/3595.txt
@@ -1,0 +1,7 @@
+```release-note:breaking-change
+resource/mongodbatlas_advanced_cluster: Changes default value of `delete_on_create_timeout` to `true`
+```
+
+```release-note:breaking-change
+resource/mongodbatlas_search_deployment: Changes default value of `delete_on_create_timeout` to `true`
+```
--- a/internal/service/advancedclustertpf/resource.go
+++ b/internal/service/advancedclustertpf/resource.go
@@ -137,7 +137,7 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 	isFlex := IsFlex(latestReq.ReplicationSpecs)
 	projectID, clusterName := waitParams.ProjectID, waitParams.ClusterName
 	clusterDetailStr := fmt.Sprintf("Cluster name %s (project_id=%s).", clusterName, projectID)
-	if plan.DeleteOnCreateTimeout.ValueBool() {
+	if cleanup.ResolveDeleteOnCreateTimeout(plan.DeleteOnCreateTimeout) {
 		var deferCall func()
 		ctx, deferCall = cleanup.OnTimeout(
 			ctx, waitParams.Timeout, diags.AddWarning, clusterDetailStr, DeleteClusterNoWait(r.Client, projectID, clusterName, isFlex),
--- a/internal/service/searchdeployment/resource.go
+++ b/internal/service/searchdeployment/resource.go
@@ -3,7 +3,6 @@ package searchdeployment
 import (
 	"context"
 	"errors"
-	"fmt"
 	"regexp"
 	"time"
 
@@ -68,18 +67,6 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 	if diags.HasError() {
 		return
 	}
-	if plan.DeleteOnCreateTimeout.ValueBool() {
-		var deferCall func()
-		deleteOnTimeout := func(newCtx context.Context) error {
-			cleanup.ReplaceContextDeadlineExceededDiags(diags, createTimeout)
-			_, err := connV2.AtlasSearchApi.DeleteAtlasSearchDeployment(newCtx, projectID, clusterName).Execute()
-			return err
-		}
-		ctx, deferCall = cleanup.OnTimeout(
-			ctx, createTimeout, diags.AddWarning, fmt.Sprintf("Search Deployment %s, (%s)", clusterName, projectID), deleteOnTimeout,
-		)
-		defer deferCall()
-	}
 	if _, _, err := connV2.AtlasSearchApi.CreateAtlasSearchDeployment(ctx, projectID, clusterName, &createReq).Execute(); err != nil {
 		diags.AddError("error during search deployment creation", err.Error())
 		return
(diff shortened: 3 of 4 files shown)


.changelog/3595.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:breaking-change
```
lantoli, line 1: I wouldn't say this is a breaking change (although it's definitely a behavioral change) because this param is not used for existing clusters as it's only read in Create, only when you create a new cluster.

internal/service/searchdeployment/resource.go
```diff
@@ -68,25 +67,20 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 	if diags.HasError() {
 		return
 	}
-	if plan.DeleteOnCreateTimeout.ValueBool() {
```
marcosuma, line 71: curious: why have you changed this code? wouldn't have it continued to work after switching the default to true?

lantoli, line 71: nice that you took the opportunity to simplify

internal/service/searchdeployment/resource.go
```diff
@@ -68,25 +67,20 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
...
 	}
 
 	deploymentResp, err := WaitSearchNodeStateTransition(ctx, projectID, clusterName, connV2.AtlasSearchApi,
 		RetryTimeConfig(createTimeout, minTimeoutCreateUpdate))
+	err = cleanup.HandleCreateTimeout(cleanup.ResolveDeleteOnCreateTimeout(plan.DeleteOnCreateTimeout), err, func(ctxCleanup context.Context) error {
+		_, err := connV2.AtlasSearchApi.DeleteAtlasSearchDeployment(ctxCleanup, projectID, clusterName).Execute()
+		if err != nil {
+			return err
+		}
+		return nil
```
lantoli:
```suggestion
		return err
```

you can just return err

.changelog/3595.txt
marcosuma, line 1: sorry, I might be missing something obvious - but how are we changing the default value to true? I would have expected the schema to change but maybe we're not doing it to avoid unnecessary breaking changes?

Related to this: I would have expected some test to fail since we're changing the default to true. Wasn't that the case?

internal/service/searchdeployment/resource.go
oarbusi: fixed in [961cf66](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3595/commits/961cf6633b2ae3c5cbc07d578cc308c66d60a1dc)

internal/service/advancedclustertpf/resource.go
```diff
@@ -137,7 +137,7 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 	isFlex := IsFlex(latestReq.ReplicationSpecs)
 	projectID, clusterName := waitParams.ProjectID, waitParams.ClusterName
 	clusterDetailStr := fmt.Sprintf("Cluster name %s (project_id=%s).", clusterName, projectID)
-	if plan.DeleteOnCreateTimeout.ValueBool() {
+	if cleanup.ResolveDeleteOnCreateTimeout(plan.DeleteOnCreateTimeout) {
```
lantoli, line 140: should we have some acc tests to check that default value true is working?

should we have some mig test to see that there are no plan changes or issues when updating provider to this version with a different default value?

lantoli (approved):

.changelog/3595.txt
marcosuma: I think it's still a breaking change because the same customer will get a different behaviour when creating their new future cluster, so to get the "same" previous experience, they need to take an action

internal/service/searchdeployment/resource.go
oarbusi: in the PR description I put more details about this:

> For mongodbatlas_search_deployment resource, implementation of the delete_on_create_timeout logic has been changed to use the one we are using in the other resources. This is to unify the implementation and be consistent across the resources where we can. In mongodbatlas_advanced_cluster is not possible because the creation logic is more complex and there are multiple API calls, making it necessary to use the ReplaceContextDeadlineExceededDiags and deferCall implementation


basically, only advanced_cluster needs the more complex implementation with deferCall and the ReplaceContextDeadlineExceededDiags due to the fact that multiple API calls are done in advanced_cluster resource. Because search_deployment is only a single POST, it can be simplified to use the implementation we are using in all other resources. This was also stated in the long-running operations spike in the "implementation considerations" section

.changelog/3595.txt
oarbusi: Like the other resources, we are not defining the default value in the schema to avoid causing plan changes to already existing resources when users upgrade. This will avoid confusing customers and avoid them from thinking: why delete_on_create_timeout is now true if the cluster/search deployment already exists?. 
So in this case, the default value is defined in `ResolveDeleteOnCreateTimeout` method that we now use in both resources to get the value of the attribute(if not defined it returns true).

On the tests aspect, tests related to delete_on_create_timeout attribute were always explicitly setting `delete_on_create_timeout = true` so that's why changing the default didn't cause a failure(although because now the error message is different in search_deployment, a tests in fact failed due to the error message being different, see [97205bb](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3595/commits/97205bb56d4ea25871bddff9bee6b36376b508d1) commit with the change to fix it). On the advanced_cluster, the error message did not change and didn't add a explicit test for this because I think it's a bit of overkill to test both the attribute set to true and attribute not set, more so considering that we sometime run out of resources when testing the advanced_cluster resource.
I have added a step in the search_deployment test to test that the behavior is the same if delete_on_create_timeout is set to true or not set at all in [4c8953a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3595/commits/4c8953a48ca929434fc02df9f9f6652055ae6a38). Hope that helps

internal/service/advancedclustertpf/resource.go
oarbusi: I have added a step in the search_deployment test to test that the behavior is the same if delete_on_create_timeout is set to true or not set at all in [4c8953a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3595/commits/4c8953a48ca929434fc02df9f9f6652055ae6a38). Hope that helps

marcosuma (approved):

EspenAlbert (approved):

AgustinBettati (approved):
LGTM, nice to see the reduced complexity in our final implementations

erabil-mdb (approved):
LGTM!

PR #3610, opened 2025-08-21
feat: Add workspace_name field in stream_connection resource and datasource

kpatel71716:
## Description

JIRA: [CLOUDP-339895](https://jira.mongodb.org/browse/CLOUDP-339895), following the plan discussed with APIx team in [CLOUDP-336156](https://jira.mongodb.org/browse/CLOUDP-336156)

As part of an ongoing effort to rename stream instances to stream workspaces, we will support a `workspace_name` field for the `stream_connection` resource and datasource that will function identically to 1instance_name`. To avoid confusion, only one of instance_name or workspace name will be allowed to be set

In a future major version of terraform, we will remove the instance_name field

Link to any related issue(s):

## Type of change:

## Required Checklist:


.changelog/3610.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:enhancement
```
marcosuma: should we say here that `instance_name` is deprecated now?

docs/data-sources/stream_connection.md
```diff
@@ -11,17 +11,30 @@ subcategory: "Streams"
...
 
 ## Argument Reference
 
 * `project_id` - (Required) Unique 24-hexadecimal digit string that identifies your project.
-* `instance_name` - (Required) Human-readable label that identifies the stream instance.
+* `instance_name` - (Deprecated) Human-readable label that identifies the stream instance. Attribute is deprecated and will be removed in version 3.0.0 in favor of `workspace_name`.
+* `workspace_name` - (Optional) Human-readable label that identifies the stream instance. Conflicts with `workspace_name`.
 * `connection_name` - (Required) Human-readable label that identifies the stream connection. In the case of the Sample type, this is the name of the sample source.
 
+~> **NOTE:** Either `workspace_name` or `workspace_name` must be provided, but not both. These fields are functionally identical and `workspace_name` is an alias for `workspace_name`. `workspace_name` should be used instead of `workspace_name`.
```
marcosuma:
```suggestion
~> **NOTE:** Either `workspace_name` or `instance_name` must be provided, but not both. These fields are functionally identical and `workspace_name` is an alias for `workspace_name`. `workspace_name` should be used instead of `workspace_name`.
```

looks like a typo? is there any api public docs link here to be attached?

docs/resources/stream_connection.md
```diff
@@ -152,10 +152,13 @@ resource "mongodbatlas_stream_connection" "example-https" {
 ## Argument Reference
 
 * `project_id` - (Required) Unique 24-hexadecimal digit string that identifies your project.
-* `instance_name` - (Required) Human-readable label that identifies the stream instance.
+* `instance_name` - (Deprecated) Human-readable label that identifies the stream instance. Attribute is deprecated and will be removed in version 3.0.0 in favor of `workspace_name`.
```
marcosuma:
```suggestion
* `instance_name` - (Deprecated) Human-readable label that identifies the stream instance. Attribute is deprecated and will be removed in following major versions in favor of `workspace_name`.
```

suggesting to be a bit less specific here

marcosuma: applies to others too

examples/mongodbatlas_stream_connection/main.tf
```diff
@@ -58,7 +58,7 @@ resource "mongodbatlas_stream_connection" "example-kafka-plaintext" {
 
 resource "mongodbatlas_stream_connection" "example-kafka-oauthbearer" {
   project_id      = var.project_id
-  instance_name   = mongodbatlas_stream_instance.example.instance_name
+  workspace_name   = mongodbatlas_stream_instance.example.instance_name
```
marcosuma: can you remind me if we're also deprecating the field in mongodbatlas_stream_instance ?

internal/service/streamconnection/data_source_stream_connection.go
```diff
@@ -38,15 +62,21 @@ func (d *streamConnectionDS) Read(ctx context.Context, req datasource.ReadReques
 
 	connV2 := d.Client.AtlasV2
 	projectID := streamConnectionConfig.ProjectID.ValueString()
-	instanceName := streamConnectionConfig.InstanceName.ValueString()
+	effectiveWorkspaceName := getEffectiveWorkspaceName(&streamConnectionConfig)
```
marcosuma: curious why we call it effective?

internal/service/streamconnection/data_source_stream_connection.go
```diff
@@ -25,7 +29,27 @@ type streamConnectionDS struct {
 
 func (d *streamConnectionDS) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
 	resp.Schema = conversion.DataSourceSchemaFromResource(ResourceSchema(ctx), &conversion.DataSourceSchemaRequest{
-		RequiredFields: []string{"project_id", "instance_name", "connection_name"},
+		RequiredFields: []string{"project_id", "connection_name"},
+		OverridenFields: map[string]dsschema.Attribute{
```
marcosuma, line 35: what do we mean here with overridden?

internal/service/streamconnection/data_source_stream_connection.go
marcosuma: will this now fail if they pass instanceSize only?

kpatel71716: The check below will error if workspace_name and instance_name are both undefined. If both are defined, then there will also be an error in NewTFStreamConnectionWithInstanceName

So one of workspace_name and instance_name is essentially required

kpatel71716: I just picked it but I don't mind switching to something like getWorkspaceOrInstanceName. My thought was that getWorkspaceName might be unclear, but I am fine with that too if it sounds simpler to you

examples/mongodbatlas_stream_connection/main.tf
kpatel71716: Yep, we will deprecate that as well in favor of a new `mongodbatlas_stream_instance` resource/datasource

docs/data-sources/stream_connection.md
kpatel71716: Thanks! Silly miss on my end. We don't have docs ready at this time but I created a [ticket](https://jira.mongodb.org/browse/CLOUDP-347566) for us to do so when they are ready

.changelog/3610.txt
kpatel71716: Added it in a separate note section. Does it make sense to just keep that note in the `enhancement` blocks?

marcosuma: can't remember off the top of my head, team might be able to provide better insights here (or look at previous changelogs to see how we declared a deprecated field

examples/mongodbatlas_stream_connection/main.tf
marcosuma: unrelated to this PR, but if it's a whole new resource, we'll need to implement the `moved` block to make the migration process easy for customers. We'll also need a migration guide and examples how to do that

internal/service/streamconnection/data_source_stream_connection.go
marcosuma: I would opt for a more explicit name, in this case `getWorkspaceOrInstanceName`

marcosuma: ah ok, so for whoever continues to use `instance_name` there will be no impact, right?

.changelog/3610.txt
```diff
@@ -0,0 +1,11 @@
...
+resource/mongodbatlas_stream_connection: Adds workspace_name field as alias for instance_name.
+```
+
+```release-note:enhancement
+data-source/mongodbatlas_stream_connection: Adds support for GCP as a Cloud Provider.
```
lantoli: remove ending period from all changelog sentences

.changelog/3610.txt
lantoli: info about deprecation notes: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/changelog-process.md#deprecations

internal/service/streamconnection/resource_schema.go
```diff
@@ -28,10 +29,26 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				},
 			},
 			"instance_name": schema.StringAttribute{
-				Required: true,
+				Optional: true,
```
lantoli, line 34: add `DeprecationMessage` in instance_name in all schemas

internal/service/streamconnection/resource_schema.go
```diff
@@ -28,10 +29,26 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(path.Expressions{
+						path.MatchRelative().AtParent().AtName("workspace_name"),
+					}...),
+				},
+			},
+			"workspace_name": schema.StringAttribute{
+				Optional: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
```
lantoli, line 46: this means that if user changes workspace_name, Delete and Create will be called instead of Update.

qq: is this the behavior we want? there are 3 main alternatives:
- Destroy and create a new resource (current approach)
- Handle workspace_name changes in Update
- Raise an error that workspace_name can't be updated

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -400,6 +400,57 @@ func TestAccStreamRSStreamConnection_AWSLambda(t *testing.T) {
 	})
 }
 
+func TestAccStreamRSStreamConnection_workspaceName(t *testing.T) {
+	var (
+		projectID, instanceName = acc.ProjectIDExecutionWithStreamInstance(t)
+		connectionName          = "workspace-name-test"
```
lantoli: consider using a random name with acc.RandomName()

.changelog/3610.txt
manupedrozo: Curious, do we ever raise warnings diagnostics for upcoming attribute deprecations to notify customers that way as well?

lantoli (approved):

.changelog/3610.txt
lantoli: it's done automatically if following my [comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3610#discussion_r2381547461), message in `DeprecationMessage` will be shown.

Just to clarify, the deprecation happens right now, it's the message, the deletion will happen in the future.

manupedrozo: That's what I was looking for, nice!

lantoli:
`tf-validate` PR check fail will probably get fixed if you bring latest from master to this branch, there was a fix here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3713

carriecwk (approved):
lgtm

examples/mongodbatlas_stream_connection/main.tf
kpatel71716: Good call, I added a note for myself in [CLOUDP-335390](https://jira.mongodb.org/browse/CLOUDP-335390)

internal/service/streamconnection/data_source_stream_connection.go
kpatel71716: agreed, updated!

internal/service/streamconnection/resource_schema.go
kpatel71716: Done!

.changelog/3610.txt
kpatel71716: Thanks, updated the notes

internal/service/streamconnection/data_source_stream_connection.go
kpatel71716: This was unnecessary. I removed it

internal/service/streamconnection/data_source_stream_connection.go
kpatel71716: yep, that's correct!

internal/service/streamconnection/resource_schema.go
kpatel71716: Double checked and discussed with our team. This is intentional behavior. The same happens with the equivalent instance_name attribute today

internal/service/streamconnection/data_source_stream_connection.go
kpatel71716: Actually looks like we still need this so that terraform can use the workspace_name field in place of the instance name field. Without this we get errors due to terraform thinking that the datasource is being modified due to the mapping of the instance_name and workspace_name fields

.changelog/3610.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:note
+resource/mongodbatlas_stream_connection: Deprecates the `instance_name` attribute `workspace_name`. All configurations using `instance_name` should be updated to use the new `workspace_name` attribute instead
```
LijieZhang1998:
```suggestion
resource/mongodbatlas_stream_connection: Deprecates the `instance_name` attribute. All configurations using `instance_name` should be updated to use the new `workspace_name` attribute instead
```

internal/service/streamconnection/data_source_stream_connections.go
```diff
@@ -42,13 +75,17 @@ func (d *streamConnectionsDS) Read(ctx context.Context, req datasource.ReadReque
 
 	connV2 := d.Client.AtlasV2
 	projectID := streamConnectionsConfig.ProjectID.ValueString()
-	instanceName := streamConnectionsConfig.InstanceName.ValueString()
+	workspaceName := getWorkspaceOrInstanceNameForDS(&streamConnectionsConfig)
```
LijieZhang1998: I'm not sure if we want it consistent.
```suggestion
	workspaceOrInstanceName := getWorkspaceOrInstanceNameForDS(&streamConnectionsConfig)
```

internal/service/streamconnection/model_stream_connection.go
```diff
@@ -117,12 +117,28 @@ func NewStreamConnectionUpdateReq(ctx context.Context, plan *TFStreamConnectionM
 	return streamConnection, nil
 }
 
-func NewTFStreamConnection(ctx context.Context, projID, instanceName string, currAuthConfig *types.Object, apiResp *admin.StreamsConnection) (*TFStreamConnectionModel, diag.Diagnostics) {
-	rID := fmt.Sprintf("%s-%s-%s", instanceName, projID, conversion.SafeString(apiResp.Name))
+func NewTFStreamConnection(ctx context.Context, projID, workspaceName string, currAuthConfig *types.Object, apiResp *admin.StreamsConnection) (*TFStreamConnectionModel, diag.Diagnostics) {
+	return NewTFStreamConnectionWithInstanceName(ctx, projID, "", workspaceName, currAuthConfig, apiResp)
```
LijieZhang1998: [q] Is `NewTFStreamConnectionWithInstanceName` to replace `NewTFStreamConnection`? `NewTFStreamConnection ` becomes a one-line method. So I'm good with just refactoring it without creating a new method.

LijieZhang1998 (approved):
LGTM 👍 Just have a few small things.

internal/service/streamconnection/model_stream_connection.go
kpatel71716: good point, I refactored to just having one method

kpatel71716:
Hi @lantoli @marcosuma . Apologies for the noise + delay on this PR, we found a bug after initial review that caused cascading issues trying to fix, but the PR is passing all checks now.

Would y'all be able to re-review this PR? Especially [this](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3610/files#diff-d12ae64c4fb4930bcb895b1084c385c650bc43b60cb9dbc1020e9794a3975292R121) change to preserver backwards compatibility with `instance_name`

internal/service/streamconnection/model_stream_connection.go
```diff
@@ -117,12 +117,24 @@ func NewStreamConnectionUpdateReq(ctx context.Context, plan *TFStreamConnectionM
...
-	rID := fmt.Sprintf("%s-%s-%s", instanceName, projID, conversion.SafeString(apiResp.Name))
+// determines if the original model was created with instance_name or workspace_name and sets the appropriate field
+func NewTFStreamConnection(ctx context.Context, projID, instanceName, workspaceName string, currAuthConfig *types.Object, apiResp *admin.StreamsConnection) (*TFStreamConnectionModel, diag.Diagnostics) {
+	if workspaceName != "" && instanceName != "" {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic("Attribute \"workspace_name\" cannot be specified when \"instance_name\" is specified", "")}
+	}
+
+	// Generate ID with workspace prefix only for workspace_name to distinguish from instance_name
+	// Keep original format for instance_name to maintain backward compatibility
+	var rID string
```
lantoli: why do we have ID computed attribute? It was required in SDKv2 but it's not needed in TPF, for instance you can see the package paths:
```
"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
"github.com/hashicorp/terraform-plugin-sdk/v2/helper/id"
```

ID makes sense if it's calculated and returned by the server - Atlas, but not so useful if generated in client, Terraform, and we're in TPF as this resource.

internal/service/streamconnection/model_stream_connection.go
```diff
@@ -117,12 +117,24 @@ func NewStreamConnectionUpdateReq(ctx context.Context, plan *TFStreamConnectionM
 	return streamConnection, nil
 }
 
-func NewTFStreamConnection(ctx context.Context, projID, instanceName string, currAuthConfig *types.Object, apiResp *admin.StreamsConnection) (*TFStreamConnectionModel, diag.Diagnostics) {
-	rID := fmt.Sprintf("%s-%s-%s", instanceName, projID, conversion.SafeString(apiResp.Name))
+// determines if the original model was created with instance_name or workspace_name and sets the appropriate field
```
lantoli:
```suggestion
// NewTFStreamConnection determines if the original model was created with instance_name or workspace_name and sets the appropriate field.
```

nit: doc standard for funcs is to start with func name, and end all comments with period.

internal/service/streamconnection/data_source_stream_connections.go
```diff
@@ -67,11 +104,12 @@ func (d *streamConnectionsDS) Read(ctx context.Context, req datasource.ReadReque
...
-	PageNum      types.Int64               `tfsdk:"page_num"`
-	ItemsPerPage types.Int64               `tfsdk:"items_per_page"`
-	TotalCount   types.Int64               `tfsdk:"total_count"`
+	ID            types.String              `tfsdk:"id"`
+	ProjectID     types.String              `tfsdk:"project_id"`
+	InstanceName  types.String              `tfsdk:"instance_name"`
+	WorkspaceName types.String              `tfsdk:"workspace_name"`
+	Results       []TFStreamConnectionModel `tfsdk:"results"`
+	PageNum       types.Int64               `tfsdk:"page_num"`
+	ItemsPerPage  types.Int64               `tfsdk:"items_per_page"`
```
lantoli, line 113: not probably for this PR, but we're preferring not to expose pagination parameter and instead just return the whole list of items

internal/service/streamconnection/model_stream_connection_test.go
```diff
@@ -452,6 +473,43 @@ func TestStreamConnectionsSDKToTFModel(t *testing.T) {
...
+	SDKResp             *admin.PaginatedApiStreamsConnection
+	providedConfig      *streamconnection.TFStreamConnectionsDSModel
+	expectedErrorString string
+	name                string
+}
+
+func TestStreamConnectionsSDKToTFModelError(t *testing.T) {
+	testCases := []connectionsSDKToTFModelErrorTestCase{
+		{
+			name: "With workspace name and instance name",
```
lantoli: we're prefering to use a map from test name to params instead of having the test name in the params struct, you can see other examples.

internal/service/streamconnection/model_stream_connection_test.go
```diff
@@ -452,6 +473,43 @@ func TestStreamConnectionsSDKToTFModel(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, diags := streamconnection.NewTFStreamConnections(t.Context(), tc.providedConfig, tc.SDKResp)
+			if !diags.HasError() {
+				t.Fatalf("expected error but got none")
+			}
+			if !strings.Contains(diags.Errors()[0].Summary(), tc.expectedErrorString) {
+				t.Fatalf("expected error %s but got %s", tc.expectedErrorString, diags.Errors()[0].Summary())
+			}
```
lantoli:
```suggestion
			assert.True(t, diags.HasError())
			assert.Contains(t, diags.Errors()[0].Summary(), tc.expectedErrorString)
```

can use assert

lantoli (approved):

internal/service/streamconnection/model_stream_connection.go
kpatel71716: Thanks for catching this. It seems I added unnecessary validation that led to errors that made me think I had to make changes here. I'll revert those to keep things much more simple

lantoli: can you remove ID attribute entirely?

kpatel71716: I am a bit confused on how that would work. I am probably missing something but from my understanding, we need to be able to set the instance_name/workspace_name fields or else we will always fail reading [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/437b816522d84ed9f604f4025dfe8c76b7764e1a/internal/service/streamconnection/resource_stream_connection.go#L180)

Is there a way around this using the basetypes and id packages above if we need the workspace/instance for the api call in the linked code?

lantoli: yes, my point is why we have `id` in the schema

kpatel71716: We need it avoid errors converting the schema to `streamconnection.TFStreamConnectionModel`. While the model needs the ID field so that the `ImportStateRequest` can have the information it needs in the ID field.

If we had access to the connection_name and instance_name in the ImportState() function then we wouldn't need to construct and parse an ID with that information. I am not sure if there is an example of this anywhere in the codebase though. fwiw the original implementation also contained ID so it's existed before these changes. But happy to create a follow-up ticket for any refactor once I understand if there is any way we can avoid using ID

internal/service/streamconnection/model_stream_connection_test.go
kpatel71716: Thanks! Ended up removing this test but noted for future reference

internal/service/streamconnection/data_source_stream_connections.go
kpatel71716: that sgtm, I just took the existing pattern but will make a JIRA ticket for us to track cleaning this up

Edit: https://jira.mongodb.org/browse/CLOUDP-352309

marcosuma (approved):

internal/service/streamconnection/model_stream_connection.go
kpatel71716: @lantoli let me know if you have any thoughts on the above. Happy to explore any ideas in a follow-up. Or we can circle back to it later

lantoli: I know this ID is not from this PR so feel free not to do anything in this PR.

I think there is some confusion, ID in the Import is not the schema attribute but the value passed to the import command or block, e.g:
```

terraform import mongodbatlas_stream_connection.test "DefaultInstance-12251446ae5f3f6ec7968b13-NewConnection"      # ID will be "DefaultInstance-12251446ae5f3f6ec7968b13-NewConnection" 

terraform import mongodbatlas_stream_connection.test "hello"      # ID will be "hello" 


import {
  to = mongodbatlas_stream_connection.test
  id = "my-id" # ID will be "my-id"
}
```

So this ID is passed to ImportState, current code is getting this value, splitting by dashes and populating the corresponding values, but there is no need for ID schema attribute in the resource:
```
func (r *streamConnectionRS) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
	workspaceName, projectID, connectionName, err := splitStreamConnectionImportID(req.ID)
	if err != nil {
		resp.Diagnostics.AddError("error splitting stream connection import ID", err.Error())
		return
	}
	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("instance_name"), workspaceName)...)
	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("workspace_name"), workspaceName)...)
	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("project_id"), projectID)...)
	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("connection_name"), connectionName)...)
}
```

this implementation is correct, we get the "id" from import and fill it the main attributes, no need to fill it evertything as Read is called just after Import, so you only need to fill in the fields required by Read, as done in the current code.

but there is no need for ID in the resource schema or datasource schemas

kpatel71716: Ah ok, thank for you bearing with me and explaining! That makes sense to me, I'll create a follow-up ticket for our team to tackle this since to clean up our logic and link

kpatel71716:
@mongodb/docs-cloud-team can we get a review on this PR? I believe the original reviewer is no longer at the company so it needs another review

jwilliams-mongo (approved):

PR #3658, opened 2025-09-08
refactor: Uses the new create-only plan modifier in flex cluster

EspenAlbert:
## Description

**Update: After changes in #3747 this is a bit simplified.

Uses the new create-only plan modifier with default for delete_on_create_timeout.

Additional changes:
- A follow-up PR will be done to use the `CreateOnlyBoolWithDefault` for the other resources

Link to any related issue(s): CLOUDP-343190

## Type of change:

## Required Checklist:


internal/service/flexcluster/resource.go
```diff
@@ -80,8 +80,7 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 	flexClusterResp, err := CreateFlexCluster(ctx, projectID, clusterName, flexClusterReq, connV2.FlexClustersApi, &createTimeout)
 
 	// Handle timeout with cleanup logic
-	deleteOnCreateTimeout := cleanup.ResolveDeleteOnCreateTimeout(tfModel.DeleteOnCreateTimeout)
-	err = cleanup.HandleCreateTimeout(deleteOnCreateTimeout, err, func(ctxCleanup context.Context) error {
+	err = cleanup.HandleCreateTimeout(tfModel.DeleteOnCreateTimeout.ValueBool(), err, func(ctxCleanup context.Context) error {
```
EspenAlbert, line 83: Removed logic

internal/service/flexcluster/resource_schema.go
```diff
@@ -148,8 +148,9 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			},
 			"delete_on_create_timeout": schema.BoolAttribute{
 				Optional: true,
+				Computed: true,
 				PlanModifiers: []planmodifier.Bool{
-					customplanmodifier.CreateOnlyBoolPlanModifier(),
+					customplanmodifier.CreateOnlyBoolWithDefault(true),
```
EspenAlbert, line 153: This change will be applied to the other resources in a follow up PR:
- encryptionatrestprivateendpoint
- pushbasedlogexport
- streamprocessor

internal/common/customplanmodifier/create_only_default_bool.go
```diff
@@ -0,0 +1,95 @@
...
+// This is useful for attributes only supported in create and not in update.
+// It shows a helpful error message helping the user to update their config to match the state.
+// Never use a schema.Default for create only attributes, instead use `WithDefault`, the default will lead to plan changes that are not expected after import.
+// Implement CopyFromPlan if the attribute is not in the API Response.
+func CreateOnlyBool() planmodifier.Bool {
+	return &createOnlyBoolPlanModifier{}
+}
+
+// CreateOnlyBoolWithDefault sets a default value on create operation that will show in the plan.
+// This avoids any custom logic in the resource "Create" handler.
```
AgustinBettati, line 16: The read will also require no changes in terms of setting the default value right? If so we can add in the comment as well.

internal/common/customplanmodifier/create_only_string.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	planmodifier "github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+// CreateOnlyString creates a plan modifier that prevents updates to string attributes.
+// This is useful for attributes only supported in create and not in update.
+// It shows a helpful error message helping the user to update their config to match the state.
+// Never use a schema.Default for create only attributes, instead use `WithDefault`, the default will lead to plan changes that are not expected after import.
+// No default value implemented for string until we have a use case.
+// Implement CopyFromPlan if the attribute is not in the API Response.
```
AgustinBettati: What is this exactly referring to? Not sure how relevant this guidance is, it applies for any attribute which is not present in API response.

AgustinBettati (commented):
LGTM, lets just make sure to clean up the boolean plan modifier implementing PlanModifyString. 
Like the approach of splitting by type and clear functions for supporting default value.

internal/common/customplanmodifier/create_only_bool.go
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+	if isUpdated(req.StateValue, req.PlanValue) {
+		d.addDiags(&resp.Diagnostics, req.Path, req.StateValue)
+	}
+	if !IsKnown(req.PlanValue) {
+		resp.PlanValue = req.StateValue
+	}
+}
+
```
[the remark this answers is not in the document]
AgustinBettati: This is a good catch 👍

internal/common/customplanmodifier/create_only_bool.go
```diff
@@ -0,0 +1,95 @@
...
+type createOnlyBoolPlanModifier struct {
+	defaultBool *bool
+}
+
+func (d *createOnlyBoolPlanModifier) Description(ctx context.Context) string {
+	return d.MarkdownDescription(ctx)
+}
+
+func (d *createOnlyBoolPlanModifier) MarkdownDescription(ctx context.Context) string {
+	return "Ensures the update operation fails when updating an attribute. If the read after import don't equal the configuration value it will also raise an error."
```
marcosuma:
```suggestion
	return "Ensures the update operation fails when updating an attribute. If the read after import doesn't equal the configuration value it will also raise an error."
```

internal/common/customplanmodifier/create_only_default_bool.go
```diff
@@ -0,0 +1,95 @@
...
+	return d.MarkdownDescription(ctx)
+}
+
+func (d *createOnlyBoolPlanModifier) MarkdownDescription(ctx context.Context) string {
+	return "Ensures the update operation fails when updating an attribute. If the read after import don't equal the configuration value it will also raise an error."
+}
+
+// isCreate uses the full state to check if this is a create operation
+func isCreate(t *tfsdk.State) bool {
+	return t.Raw.IsNull()
```
marcosuma, line 38: nit, but I am surprised there isn't a native method that doesn't tell already if it's a create or not.

Related to this: is it correct to assume that the null state is only when it's creation? what about import? or can this be called at some point after a deletion?

internal/common/customplanmodifier/create_only_bool.go
```diff
@@ -0,0 +1,95 @@
+package customplanmodifier
```
marcosuma, line 1: shouldn't we unit test these plan modifiers?

internal/common/customplanmodifier/create_only_bool.go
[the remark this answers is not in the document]
marcosuma: I guess this method goes to the other class `create_only_string`?

internal/service/flexcluster/resource_schema.go
```diff
@@ -22,14 +22,14 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"project_id": schema.StringAttribute{
 				Required: true,
 				PlanModifiers: []planmodifier.String{
-					customplanmodifier.CreateOnlyStringPlanModifier(),
+					customplanmodifier.CreateOnlyString(),
```
marcosuma: this won't have any breaking change, will it?

internal/service/pushbasedlogexport/resource_schema.go
```diff
@@ -59,7 +59,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"delete_on_create_timeout": schema.BoolAttribute{
 				Optional: true,
 				PlanModifiers: []planmodifier.Bool{
-					customplanmodifier.CreateOnlyBoolPlanModifier(),
+					customplanmodifier.CreateOnlyBool(),
```
marcosuma: Double checking here: will this also be usable for SDKv2? just remembered [of this comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3536/files/b0313a417fcd34b4b205651534c5943e7f6a3784#r2245207121).

Feel free to correct me if I am saying something incorrect - I guess it won't, which means that I am wondering if the "import" use case needs to be handled also for the SDKv2 resources? cc @oarbusi

internal/service/pushbasedlogexport/resource_schema.go
```diff
@@ -59,7 +59,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"delete_on_create_timeout": schema.BoolAttribute{
```
marcosuma, line 59: @EspenAlbert @AgustinBettati chatting more with @oarbusi , I reflected on one detail I missed during the explanation. The plan change during the import would only occur if `delete_on_create_timeout` is explicitly set in the configuration used for the import. If not, there is no issue with the existing logic. 

That means: if in the import this value is not specified by the practitioner (like it really should be, as there's no reason why it should be set), the current code doesn't have any limitation. Considered that, I can still figure there's a value in improving the overall experience, but I see less urgency in making this change.

Is there any other consideration I missed?

internal/common/customplanmodifier/create_only_default_bool.go
EspenAlbert: This depends on the API Behavior.
Adding the line from above docstring: `// If the attribute is not in the API Response implement CopyFromPlan behavior when converting API Model to TF Model.`

internal/common/customplanmodifier/create_only_default_bool.go
EspenAlbert: Correct. Import doesn't call the plan modifier. 
There will be no plan after deletion.

internal/common/customplanmodifier/create_only_bool.go
EspenAlbert: These are tested by the acceptance tests for project & flex resources:
From project:
- TestAccProject_withFalseDefaultSettings
- TestMigProject_withFalseDefaultAlertSettings
- TestMigProject_withTrueDefaultAlertSettings

It is not straightforward to unit test them as they are called in a specific context.

internal/common/customplanmodifier/create_only_bool.go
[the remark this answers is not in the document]
EspenAlbert: Yes, forgot to delete it from the `createOnlyBoolPlanModifier`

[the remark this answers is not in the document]
EspenAlbert: Removed in 75e97a70562f1aecfe7f5abf4ee046f18de1c830

internal/common/customplanmodifier/create_only_string.go
EspenAlbert: Updated, WDYT?

internal/common/customplanmodifier/create_only_bool.go
EspenAlbert: Applied in 75e97a70562f1aecfe7f5abf4ee046f18de1c830

internal/service/flexcluster/resource_schema.go
EspenAlbert: No.

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: This is a good point.
https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/plan-modification#migration-notes 
We could investigate using DiffSuppress function for this, will have to do another investigation.

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: Correct. Nothing else apart from the ones documented here; https://docs.google.com/document/d/1iu18eHeECR2ofy51uuYOyP8dxS2Q5m8joTnEQu4Q4lQ/edit?tab=t.0#bookmark=id.qliuhcp0nzu1

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: The DiffSuppress is on the resource level:
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.37.0/helper/schema#CustomizeDiffFunc 
So, I don't think it is worth changing the logic for the SDKv2 resources.

The behavior of a non-empty plan after import for these resources when an explicit config value is set, I think, is ok to keep for now; otherwise, we would need to add the check of `Null` to `defined-value` in all `Resources.Update` methods.

internal/common/customplanmodifier/create_only_default_bool.go
marcosuma: nit, but if one day for some reason we decide to use the plan modifier to the import (not sure if that is even possible?) this method would become incorrect.

internal/service/pushbasedlogexport/resource_schema.go
oarbusi: Problem I see is not being consistent between SDKv2 and TPF resources. If we do this only in the TPF resources and not in SDKv2, we are not consistent and we might cause more confusion than clarity

internal/common/customplanmodifier/create_only_default_bool.go
EspenAlbert: image
src: https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md 

As seen by the diagram, the Plan Resource Change is always called after ReadResource when importing.
I don't think this lifecycle will change.

internal/service/pushbasedlogexport/resource_schema.go
marcosuma: ok, I was of the impression that it was something that was "always" going to happen when an import was made. My take then is that this is not worth it a refactoring now that we are very close to the release. 

As it comes down to best practices, I still see the value but I would postpone this change to perhaps a minor release. This will help us focusing in getting 2.0.0 out with less moving pieces.

@AgustinBettati @EspenAlbert @lantoli  @maastha @marcabreracast  @oarbusi looking for feedback here, I'd like to make a call together.

oarbusi: Thinking on this more, I think that if we can't provide the same experience in SDKv2 and TPF, I would maybe hold this off until we can, because not having the same UX across resources can be confusing for customers

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: I'll try to do it in the follow-up PR to see how much work it is:
https://jira.mongodb.org/browse/CLOUDP-343190?focusedCommentId=7658580&focusedId=7658580&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-7658580

internal/service/pushbasedlogexport/resource_schema.go
oarbusi: Agree on postponing this and including it in a minor release instead of 2.0.0

AgustinBettati: > would only occur if delete_on_create_timeout is explicitly set in the configuration used for the import

Yeah this is very much an edge case, defining this attribute when importing will not make sense and can be mitigated with a docs clarification. Im ok with keeping 2.0 release as lean as possible and then we revisit.

EspenAlbert: I am leaning towards merging. The plan behavior is more explicit, and we are following hashicorp best practices with this refactor (using computed for an attribute with a default value)

Since this plan modifier is already released now in 1.41 we would need to keep both implementations which adds tech debt.

lantoli: I'm ok to merge this PR if it's ready

lantoli (approved):

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: It is ready, but I think the team is not convinced. @marcosuma @AgustinBettati @oarbusi

marcosuma: @lantoli @EspenAlbert maybe I was misunderstood - I am not saying we shouldn't merge. I am saying we shouldn't link this to 2.0.0 as there's really no strong reason to do so.

if we merge just this PR, we'll have:
- this resource (TPF) using this version of plan modifier
- other resources (TPF) using the other approach but they can be updated
- other resources (SDKv2) which would stay with a different behavior.

What is the plan to align all of those? Why do you want to link this to 2.0.0?

EspenAlbert: Ok. Allthough no blocker for 2.0.0 I think we should merge it. 
Some clarifications If we merge:
- flex cluster resource (TPF) using this version of plan modifier for delete_on_create_timeout with default=true
- other resources (TPF) using the plan modifier for delete_on_create_timeout but without a default (set in create handler instead) --> I expect to follow up with using default instead after merging.
- other resources (SDKv2) with delete_on_create_timeout which would stay with an allowed plan update after import (when the attribute is changed). --> I can look into updating this today.

lantoli (commented):
after offline discussion, let's wait until 2.0.0 is released

lantoli:
we're good to go

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: Was decided to postpone

internal/common/customplanmodifier/create_only_default_bool.go
```diff
@@ -12,35 +12,46 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-// CreateOnlyAttributePlanModifierWithBoolDefault sets a default value on create operation that will show in the plan.
+// CreateOnlyBool creates a plan modifier that prevents updates to boolean attributes.
+// This is useful for attributes only supported in create and not in update.
+// It shows a helpful error message helping the user to update their config to match the state.
+// Never use a schema.Default for create only attributes, instead use `WithDefault`, the default will lead to plan changes that are not expected after import.
+// If the attribute is not in the API Response implement CopyFromPlan behavior when converting API Model to TF Model.
+func CreateOnlyBool() planmodifier.Bool {
```
manupedrozo: Is there an advantage in using this one instead of [create_only.go](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/65d7a6c8d6d8b56b6cb705eb55792bb19249bcf5/internal/common/customplanmodifier/create_only.go#L18-L17) when no default is provided? 
Should we only allow usage of this modifier with a default?

internal/service/flexcluster/resource_test.go
```diff
@@ -104,10 +104,11 @@ func basicTestCase(t *testing.T) *resource.TestCase {
...
 			{
-				ResourceName:      resourceName,
-				ImportStateIdFunc: acc.ImportStateIDFuncProjectIDClusterName(resourceName, "project_id", "name"),
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            resourceName,
+				ImportStateIdFunc:       acc.ImportStateIDFuncProjectIDClusterName(resourceName, "project_id", "name"),
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"delete_on_create_timeout"},
```
AgustinBettati, line 111: q: Why is this ignore now needed? Cant fully track how current changes impact resulting state after an import

internal/service/pushbasedlogexport/resource_schema.go
AgustinBettati: As we revisit the PR can be good to clarifying if we will leave a difference between TPF and SDK v2 resources for `delete_on_create_timeout`, and if so what is that UX difference.

internal/service/flexcluster/resource_test.go
EspenAlbert: Because it is set to `true` by default now.
The old implementation used `true` as the default in step 1.
But changed the value to `null` in step 2 (since it was optional-only), so this wasn't needed.

internal/common/customplanmodifier/create_only_default_bool.go
EspenAlbert: Good idea, can always add it back later if needed.
Changed in 21a2274f38f1e8a45470b3bc790d0d89de5b3460

manupedrozo: Nice, can also move to a `bool` instead of a `*bool` and remove nil checks.

manupedrozo (approved):

internal/common/customplanmodifier/create_only_default_bool.go
EspenAlbert: true, changed in 7bc11f1022aa6312031bd75787c43ffb467f0093

internal/service/flexcluster/resource_test.go
AgustinBettati: makes sense, from UX perspective since it is O+C during an import user wont be forced to define it so we are good 👍

AgustinBettati (approved):
changes LGTM, as per [thread](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3658#discussion_r2329723797) think it can be good clarify how the end state will look for TPF and SDK v2 resources in terms of `delete_on_create_timeout` behaviour.

internal/service/pushbasedlogexport/resource_schema.go
EspenAlbert: Updated the Jira ticket:
tl;dr: Don't see a concern in adding the extra error condition in the two resources. Will do it in follow-up PR

Impacted SDKv2 resources: only networkpeering and onlinearchive (other resources have no update, can reuse logic for looking for changes)

cloudbackupsnapshot <-- no update method (unaffected)
cloudprovideraccessetup <-- no update
clusteroutagesimulation <-- no update
networkpeering <-- affected, can raise an error here
onlinearchive <-- affected, can raise an error here
privatelinkendpoint <-- no update
privatelinkendpointservice <-- no update

PR #3700, opened 2025-09-18
feat: Adds support for Service Account credentials as provider inputs, environment variables and AWS Secrets Manager

oarbusi:
## Description

Adds support for Service Account credentials as provider inputs, environment variables and AWS Secrets Manager
Includes a test that tests the authentication using Service Account credentials through environment variables that runs in CI. Provider inputs and AWS Secrets have been tested manually.

**Important:** followup tickets will implement refresh policy, thread-safe access to the token, and final hierarchy of credentials

Changes in `provider_mock.go`  and `http_mocker.go` are due to the following problem:
- The Service Account authentication implementation introduced OAuth2 background operations that use the same HTTP client as the test mocking infrastructure. This caused data races when tests tried to modify the HTTP client's Transport while OAuth2 goroutines were simultaneously reading from it.

To solve this:
- Modified the test provider mocking to create copies of HTTP clients before applying test modifications, rather than modifying the shared client directly. This isolates test modifications from production OAuth2 operations.

Changes in `internal/service/advancedclustertpf/main_test.go` are due to the following problem:

- After commit cec3112d “error instead of warning when no credentials are set”, provider `Configure` fails without credentials. Several unit tests previously relied on running with no creds, causing Terraform to hang and hit the 2m timeout.
- Providing Service Account creds isn’t viable for these tests because the provider eagerly fetches an OAuth2 token using a fresh HTTP client (not intercepted by our HTTP mocker), which would attempt real network calls.
- **Fix:** In internal/service/advancedclustertpf/main_test.go TestMain, set dummy Digest credentials and clear Service Account vars to force the Digest path, which our mocker fully intercepts:
  - Set: MONGODB_ATLAS_PUBLIC_API_KEY=dummy, MONGODB_ATLAS_PRIVATE_API_KEY=dummy
  - Unset: MONGODB_ATLAS_CLIENT_ID, MONGODB_ATLAS_CLIENT_SECRET, TF_VAR_CLIENT_ID, TF_VAR_CLIENT_SECRET
- **Impact:** Restores test stability post-credential validation change; no production behavior change.

Link to any related issue(s): CLOUDP-345759

## Type of change:

## Required Checklist:

### DIFF
--- a/.changelog/3700.txt
+++ b/.changelog/3700.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-provider: Support Service Account as credentials to authenticate the provider
+provider: Supports Service Account as credentials to authenticate the provider
 ```
--- a/.changelog/3700.txt
+++ b/.changelog/3700.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+provider: Support Service Account as credentials to authenticate the provider
+```
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -481,7 +481,6 @@ jobs:
           terraform_version: ${{ inputs.terraform_version }}
           terraform_wrapper: false    
       - id: sts-assume-role
-        if: matrix.auth_method == 'sts_assume_role'
         name: Generate STS Temporary credential for acceptance testing
         shell: bash
         env:
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -171,6 +171,10 @@ on:
         required: true
       mongodb_atlas_rp_public_key:
         required: true
+      mongodb_atlas_client_id:
+        required: true
+      mongodb_atlas_client_secret:
+        required: true
       azure_directory_id:
         required: true
       azure_resource_group_name:
@@ -245,7 +249,7 @@ jobs:
       mustTrigger: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.test_group == '' )  }}
     outputs: # ensure resources are sorted alphabetically
       advanced_cluster: ${{ steps.filter.outputs.advanced_cluster == 'true' || env.mustTrigger == 'true' }}
-      assume_role: ${{ steps.filter.outputs.assume_role == 'true' || env.mustTrigger == 'true' }}
+      authentication: ${{ steps.filter.outputs.authentication == 'true' || env.mustTrigger == 'true' }}
       autogen: ${{ steps.filter.outputs.autogen == 'true' || env.mustTrigger == 'true' }}
       backup: ${{ steps.filter.outputs.backup == 'true' || env.mustTrigger == 'true' }}
       control_plane_ip_addresses: ${{ steps.filter.outputs.control_plane_ip_addresses == 'true' || env.mustTrigger == 'true' }}
@@ -278,7 +282,7 @@ jobs:
         filters: |
           advanced_cluster:
             - 'internal/service/advancedclustertpf/*.go'
-          assume_role:
+          authentication:
             - 'internal/provider/*.go'
           autogen:
             - 'internal/common/autogen/*.go'
@@ -460,11 +464,14 @@ jobs:
             ./internal/service/advancedclustertpf
         run: make testacc
 
-  assume_role:
+  authentication:
     needs: [ change-detection, get-provider-version ]
-    if: ${{ needs.change-detection.outputs.assume_role == 'true' || inputs.test_group == 'assume_role' }}
+    if: ${{ needs.change-detection.outputs.authentication == 'true' || inputs.test_group == 'authentication' }}
     runs-on: ubuntu-latest
     permissions: {}
+    strategy:
+      matrix:
+        auth_method: [sts_assume_role, service_account]
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
         with:
@@ -477,6 +484,7 @@ jobs:
           terraform_version: ${{ inputs.terraform_version }}
           terraform_wrapper: false    
       - id: sts-assume-role
+        if: matrix.auth_method == 'sts_assume_role'
         name: Generate STS Temporary credential for acceptance testing
         shell: bash
         env:
@@ -485,7 +493,8 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.aws_access_key_id }}
           ASSUME_ROLE_ARN: ${{ vars.ASSUME_ROLE_ARN }}
         run: bash ./scripts/generate-credentials-with-sts-assume-role.sh
-      - name: Acceptance Tests
+      - name: Acceptance Tests (STS Assume Role)
+        if: matrix.auth_method == 'sts_assume_role'
         env:
           MONGODB_ATLAS_PUBLIC_KEY: ""
           MONGODB_ATLAS_PRIVATE_KEY: ""
@@ -496,6 +505,16 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ steps.sts-assume-role.outputs.aws_access_key_id }}
           AWS_SECRET_ACCESS_KEY: ${{ steps.sts-assume-role.outputs.aws_secret_access_key }}
           AWS_SESSION_TOKEN: ${{ steps.sts-assume-role.outputs.AWS_SESSION_TOKEN }}
+          MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
+          ACCTEST_PACKAGES: ./internal/provider
+        run: make testacc
+      - name: Acceptance Tests (Service Account)
+        if: matrix.auth_method == 'service_account'
+        env:
+          MONGODB_ATLAS_PUBLIC_KEY: ""
+          MONGODB_ATLAS_PRIVATE_KEY: ""
+          MONGODB_ATLAS_CLIENT_ID: ${{ secrets.mongodb_atlas_client_id }}
+          MONGODB_ATLAS_CLIENT_SECRET: ${{ secrets.mongodb_atlas_client_secret }}
           MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
           ACCTEST_PACKAGES: ./internal/provider
         run: make testacc
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -469,9 +469,6 @@ jobs:
     if: ${{ needs.change-detection.outputs.authentication == 'true' || inputs.test_group == 'authentication' }}
     runs-on: ubuntu-latest
     permissions: {}
-    strategy:
-      matrix:
-        auth_method: [sts_assume_role, service_account]
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
         with:
@@ -484,7 +481,6 @@ jobs:
           terraform_version: ${{ inputs.terraform_version }}
           terraform_wrapper: false    
       - id: sts-assume-role
-        if: matrix.auth_method == 'sts_assume_role'
         name: Generate STS Temporary credential for acceptance testing
         shell: bash
         env:
@@ -494,7 +490,6 @@ jobs:
           ASSUME_ROLE_ARN: ${{ vars.ASSUME_ROLE_ARN }}
         run: bash ./scripts/generate-credentials-with-sts-assume-role.sh
       - name: Acceptance Tests (STS Assume Role)
-        if: matrix.auth_method == 'sts_assume_role'
         env:
           MONGODB_ATLAS_PUBLIC_KEY: ""
           MONGODB_ATLAS_PRIVATE_KEY: ""
@@ -510,7 +505,6 @@ jobs:
           ACCTEST_PACKAGES: ./internal/provider
         run: make testacc
       - name: Acceptance Tests (Service Account)
-        if: matrix.auth_method == 'service_account'
         env:
           MONGODB_ATLAS_PUBLIC_KEY: ""
           MONGODB_ATLAS_PRIVATE_KEY: ""
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -469,9 +469,6 @@ jobs:
     if: ${{ needs.change-detection.outputs.authentication == 'true' || inputs.test_group == 'authentication' }}
     runs-on: ubuntu-latest
     permissions: {}
-    strategy:
-      matrix:
-        auth_method: [sts_assume_role, service_account]
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
         with:
@@ -493,8 +490,7 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.aws_access_key_id }}
           ASSUME_ROLE_ARN: ${{ vars.ASSUME_ROLE_ARN }}
         run: bash ./scripts/generate-credentials-with-sts-assume-role.sh
-      - name: Acceptance Tests (STS Assume Role)
-        if: matrix.auth_method == 'sts_assume_role'
+      - name: Acceptance Tests
         env:
           MONGODB_ATLAS_PUBLIC_KEY: ""
           MONGODB_ATLAS_PRIVATE_KEY: ""
@@ -505,14 +501,6 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ steps.sts-assume-role.outputs.aws_access_key_id }}
           AWS_SECRET_ACCESS_KEY: ${{ steps.sts-assume-role.outputs.aws_secret_access_key }}
           AWS_SESSION_TOKEN: ${{ steps.sts-assume-role.outputs.AWS_SESSION_TOKEN }}
-          MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
-          ACCTEST_PACKAGES: ./internal/provider
-        run: make testacc
-      - name: Acceptance Tests (Service Account)
-        if: matrix.auth_method == 'service_account'
-        env:
-          MONGODB_ATLAS_PUBLIC_KEY: ""
-          MONGODB_ATLAS_PRIVATE_KEY: ""
           MONGODB_ATLAS_CLIENT_ID: ${{ secrets.mongodb_atlas_client_id }}
           MONGODB_ATLAS_CLIENT_SECRET: ${{ secrets.mongodb_atlas_client_secret }}
           MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -506,6 +506,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ steps.sts-assume-role.outputs.aws_secret_access_key }}
           AWS_SESSION_TOKEN: ${{ steps.sts-assume-role.outputs.AWS_SESSION_TOKEN }}
           MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
+          ACCTEST_REGEX_RUN: '^TestAccSTSAssumeRole'
           ACCTEST_PACKAGES: ./internal/provider
         run: make testacc
       - name: Acceptance Tests (Service Account)
@@ -516,6 +517,7 @@ jobs:
           MONGODB_ATLAS_CLIENT_ID: ${{ secrets.mongodb_atlas_client_id }}
           MONGODB_ATLAS_CLIENT_SECRET: ${{ secrets.mongodb_atlas_client_secret }}
           MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
+          ACCTEST_REGEX_RUN: '^TestAccServiceAccount'
           ACCTEST_PACKAGES: ./internal/provider
         run: make testacc
 
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -481,6 +481,7 @@ jobs:
           terraform_version: ${{ inputs.terraform_version }}
           terraform_wrapper: false    
       - id: sts-assume-role
+        if: matrix.auth_method == 'sts_assume_role'
         name: Generate STS Temporary credential for acceptance testing
         shell: bash
         env:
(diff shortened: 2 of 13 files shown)


kanchana-mongodb (approved):

.changelog/3700.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+provider: Support Service Account as credentials to authenticate the provider
```
lantoli:
```suggestion
provider: Supports Service Account as credentials to authenticate the provider
```

internal/config/client.go
```diff
@@ -97,15 +105,55 @@ type UAMetadata struct {
...
-	// Don't change logging.NewTransport to NewSubsystemLoggingHTTPTransport until all resources are in TPF.
-	tfLoggingTransport := logging.NewTransport("Atlas", digestTransport)
-	client := &http.Client{Transport: tfLoggingTransport}
 
-	optsAtlas := []matlasClient.ClientOpt{matlasClient.SetUserAgent(userAgent(c))}
+	var client *http.Client
+	var optsAtlas []matlasClient.ClientOpt
+
+	// Determine authentication method based on available credentials
+	switch authMethod := resolveAuthMethod(c); authMethod {
```
lantoli:
```suggestion
	switch resolveAuthMethod(c) {
```

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -497,6 +506,18 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ steps.sts-assume-role.outputs.aws_secret_access_key }}
           AWS_SESSION_TOKEN: ${{ steps.sts-assume-role.outputs.AWS_SESSION_TOKEN }}
           MONGODB_ATLAS_LAST_VERSION: ${{ needs.get-provider-version.outputs.provider_version }}
+          ACCTEST_REGEX_RUN: '^TestAccSTSAssumeRole'
+          ACCTEST_PACKAGES: ./internal/provider
+        run: make testacc
+      - name: Acceptance Tests (Service Account)
```
lantoli, line 507: I'm not sure about the benefits of using matrix here with if statements, what about having the authentication job with just consecutive steps Acceptance Tests (STS Assume Role) and Acceptance Tests (Service Account)?

if it's because parallel execution, then I think it's clearer to have two groups, although if tests are fast it should be ok to be in serial

internal/config/client.go
```diff
@@ -319,3 +367,13 @@ func userAgent(c *Config) string {
 
 	return strings.Join(parts, " ")
 }
+
+func resolveAuthMethod(c *Config) string {
+	if c.ClientID != "" && c.ClientSecret != "" {
+		return "service_account"
+	}
+	if c.PublicKey != "" && c.PrivateKey != "" {
+		return "digest"
```
lantoli: you can use the consts you created.

nit: consider using a `enum` approach that has type check enforcment, e.g. [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/74019a3908a31210004ae4d36dd8030d7734dfed/tools/codegen/codespec/model.go#L145)

internal/provider/credentials.go
```diff
@@ -74,16 +74,18 @@ func configureCredentialsSTS(cfg *config.Config, secret, region, awsAccessKeyID,
...
 		return *cfg, err
 	}
-	if secretData.PrivateKey == "" {
-		return *cfg, fmt.Errorf("secret missing value for credential PrivateKey")
-	}
 
-	if secretData.PublicKey == "" {
-		return *cfg, fmt.Errorf("secret missing value for credential PublicKey")
+	switch {
+	case secretData.PrivateKey != "" && secretData.PublicKey != "":
```
lantoli: nit: these conditions seem duplicated with the ones resolveAuthMethod, don't know if that logic can be shared

internal/provider/provider.go
```diff
@@ -188,6 +190,14 @@ func (p *MongodbtlasProvider) Schema(ctx context.Context, req provider.SchemaReq
 				Optional:    true,
 				Description: "AWS Security Token Service provided session token.",
 			},
+			"client_id": schema.StringAttribute{
+				Optional:    true,
+				Description: "MongoDB Atlas Client ID.",
```
lantoli: should we add more info? e.g.:
```
MongoDB Atlas Client ID for SA (Service Account)
```

.github/workflows/acceptance-tests-runner.yml
oarbusi: IMO it's a nice way to split the different credentials test considering that the tests are in the same package/file

internal/provider/provider.go
```diff
@@ -450,6 +456,28 @@ func setDefaultValuesWithValidations(ctx context.Context, data *tfMongodbAtlasPr
...
+
+	if data.ClientSecret.ValueString() == "" {
+		data.ClientSecret = types.StringValue(MultiEnvDefaultFunc([]string{
+			"MONGODB_ATLAS_CLIENT_SECRET",
+			"TF_VAR_CLIENT_SECRET",
+		}, "").(string))
+	}
+
+	// Check if any valid authentication method is provided
+	hasDigestAuth := data.PublicKey.ValueString() != "" && data.PrivateKey.ValueString() != ""
```
lantoli: nit: these conditions seem duplicated in other parts of the code, don't know if we could unify that logic

.github/workflows/acceptance-tests-runner.yml
lantoli: it looks like a bad smell to have a matrix and then an `if` statement for each value (matrix is normally used when you want to execute exactly the same except because the variable that is changing), i think it's clearer just having two steps in that job.  
anyway, it's ok, feel free to keep it and disregard this comment.

internal/provider/provider.go
```diff
@@ -61,7 +61,7 @@ const (
 	MongodbGovCloudQAURL           = "https://cloud-qa.mongodbgov.com"
 	MongodbGovCloudDevURL          = "https://cloud-dev.mongodbgov.com"
 	ProviderConfigError            = "error in configuring the provider."
-	MissingAuthAttrError           = "either Atlas Programmatic API Keys or AWS Secrets Manager attributes must be set"
+	MissingAuthAttrError           = "either Atlas Programmatic API Keys, Service Accounts, or AWS Secrets Manager attributes must be set"
```
lantoli:
```suggestion
	MissingAuthAttrError           = "either AWS Secrets Manager, Service Accounts, or Atlas Programmatic API Keys attributes must be set"
ProviderMetaUserAgentExtra     = "user_agent_extra"
```

nit: text in hierarchical order

internal/provider/provider_sdk2.go
```diff
@@ -417,6 +424,28 @@ func setDefaultsAndValidations(d *schema.ResourceData) diag.Diagnostics {
...
+	}
+
+	// Check if any valid authentication method is provided
+	hasDigestAuth := d.Get("public_key").(string) != "" && d.Get("private_key").(string) != ""
+	hasServiceAccountAuth := d.Get("client_id").(string) != "" && d.Get("client_secret").(string) != ""
+
+	if !hasDigestAuth && !hasServiceAccountAuth && !awsRoleDefined {
+		diagnostics = append(diagnostics, diag.Diagnostic{Severity: diag.Warning, Summary: MissingAuthAttrError})
+	}
+
```
lantoli: this logic seems very similar to provider.go, consider extracting some common logic if applicable

internal/config/client.go
oarbusi: changed to enum here [57a4873](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/57a487304de2ca52dcdfed6819ee0775de70532d)

internal/provider/provider.go
oarbusi: changed in [01513e7](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/01513e7499f704481a0858f4022741838f0b1560)

internal/provider/provider.go
oarbusi: done in [01513e7](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/01513e7499f704481a0858f4022741838f0b1560)

internal/config/client.go
```diff
@@ -97,15 +140,56 @@ type UAMetadata struct {
...
-	optsAtlas := []matlasClient.ClientOpt{matlasClient.SetUserAgent(userAgent(c))}
+	var client *http.Client
+	var optsAtlas []matlasClient.ClientOpt
+
+	// Determine authentication method based on available credentials
+	switch ResolveAuthMethod(c) {
+	case ServiceAccount:
+		conf := clientcredentials.NewConfig(c.ClientID, c.ClientSecret)
+		// Override TokenURL and RevokeURL if custom BaseURL is provided
+		if c.BaseURL != "" {
```
lantoli, line 154: I think we're not supporting SA for the old SDK that is used in deprecated cluster resources, can you confirm?

similarly for Realm API, GetRealmClient will fail if API keys are not found.

Adding a question in the doc to verify if it's in the scope

internal/provider/provider_sdk2.go
oarbusi: refactored all this in [16d36b8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/16d36b8d7918532f32665f70cac8af86ebba1215)

internal/provider/provider.go
oarbusi: refactored all this in [16d36b8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/16d36b8d7918532f32665f70cac8af86ebba1215)

internal/provider/credentials.go
oarbusi: refactored all this in [16d36b8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/16d36b8d7918532f32665f70cac8af86ebba1215)

.github/workflows/acceptance-tests-runner.yml
oarbusi: no strong opinion on this from me, removed in [09f516d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/09f516d55eb4213127867a57705f8e8aeb4f7ed0)

internal/config/client.go
oarbusi: To my understanding, the old deprecated SDK does not support service accounts([link](https://github.com/mongodb/go-client-mongodb-atlas?tab=readme-ov-file#authentication)), so we won't support it in terraform. For realm, I haven't considered it, I can check

internal/provider/credentials.go
lantoli: nice

internal/config/client.go
```diff
@@ -40,6 +44,37 @@ const (
...
+
+// IsServiceAccountAuth checks if client ID/secret credentials are present
+func IsServiceAccountAuth(clientID, clientSecret string) bool {
+	return clientID != "" && clientSecret != ""
+}
+
+// HasValidAuthCredentials checks if any valid authentication method is provided
+func HasValidAuthCredentials(publicKey, privateKey, clientID, clientSecret string) bool {
+	return IsDigestAuth(publicKey, privateKey) || IsServiceAccountAuth(clientID, clientSecret)
+}
```
lantoli:
```suggestion
func IsDigestAuth(cp CredentialProvider) bool {
	return cp.GetPublicKey != "" && cp.GetPrivateKey != ""
}

// IsServiceAccountAuth checks if client ID/secret credentials are present
func IsServiceAccountAuth(cp CredentialProvider) bool {
	return cp.GetClientID != "" && cp.GetClientSecret != ""
}

// HasValidAuthCredentials checks if any valid authentication method is provided
func HasValidAuthCredentials(cp CredentialProvider) bool {
	return IsDigestAuth(cp) || IsServiceAccountAuth(cp)
}
```

interface can be used so need to pass explicitly the variables, similar to ResolveAuthMethod param

oarbusi: done in [106af6b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/106af6bd4e9416b9fa787c48f30b21cebc25da76)

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -460,9 +464,9 @@ jobs:
             ./internal/service/advancedclustertpf
         run: make testacc
 
-  assume_role:
+  authentication:
```
marcosuma, line 467: @oarbusi probably unrelated to this PR, but we'll need ways to test the business logic of the ordering between the various auth mechanisms

internal/config/client.go
marcosuma: how is this conversation landing?

internal/config/client.go
```diff
@@ -97,15 +140,56 @@ type UAMetadata struct {
...
-	tfLoggingTransport := logging.NewTransport("Atlas", digestTransport)
-	client := &http.Client{Transport: tfLoggingTransport}
 
-	optsAtlas := []matlasClient.ClientOpt{matlasClient.SetUserAgent(userAgent(c))}
+	var client *http.Client
+	var optsAtlas []matlasClient.ClientOpt
+
+	// Determine authentication method based on available credentials
+	switch ResolveAuthMethod(c) {
+	case ServiceAccount:
```
marcosuma, line 151: I am guessing there will be another `case` to support when the token is passed directly?

internal/provider/credentials.go
```diff
@@ -74,16 +74,18 @@ func configureCredentialsSTS(cfg *config.Config, secret, region, awsAccessKeyID,
...
 		return *cfg, err
 	}
-	if secretData.PrivateKey == "" {
-		return *cfg, fmt.Errorf("secret missing value for credential PrivateKey")
-	}
 
-	if secretData.PublicKey == "" {
-		return *cfg, fmt.Errorf("secret missing value for credential PublicKey")
+	switch config.ResolveAuthMethod(&secretData) {
+	case config.Digest:
```
marcosuma, line 79: what if they set both? doesn't the order matter? (in general I would keep order consistency for the `case`s you're adding

internal/provider/provider.go
```diff
@@ -61,7 +61,7 @@ const (
 	MongodbGovCloudQAURL           = "https://cloud-qa.mongodbgov.com"
 	MongodbGovCloudDevURL          = "https://cloud-dev.mongodbgov.com"
 	ProviderConfigError            = "error in configuring the provider."
-	MissingAuthAttrError           = "either Atlas Programmatic API Keys or AWS Secrets Manager attributes must be set"
+	MissingAuthAttrError           = "either AWS Secrets Manager, Service Accounts, or Atlas Programmatic API Keys attributes must be set"
```
marcosuma: so we have token, SAs and PAK
Then we have AWS Secrets Manager that is a meta auth, meaning it can contain Token, SA or PAK. Am I right? If so, this error is probalby mixing up things

internal/provider/provider.go
```diff
@@ -450,6 +456,30 @@ func setDefaultValuesWithValidations(ctx context.Context, data *tfMongodbAtlasPr
...
+
+	if data.ClientSecret.ValueString() == "" {
+		data.ClientSecret = types.StringValue(MultiEnvDefaultFunc([]string{
+			"MONGODB_ATLAS_CLIENT_SECRET",
+			"TF_VAR_CLIENT_SECRET",
+		}, "").(string))
+	}
+
+	// Check if any valid authentication method is provided
+	if !config.HasValidAuthCredentials(&config.Config{
```
marcosuma, line 474: this is not erroring if more than one auth mechanism is set, is it?

marcosuma (commented):
> Important: followup tickets will implement refresh policy, thread-safe access to the token, etc


Based on your description, I not sure I understand how we're handling the different priorities between:
- environment variables, IDs and Secrets Manager
- token, SAs and PAK

internal/config/client.go
oarbusi: yes exactly

internal/provider/credentials.go
oarbusi: we'll work on hierarchy in following tickets. For now, service account has priority over digest auth if both are set

internal/provider/provider.go
oarbusi: no, just a warning(like it was before)

internal/provider/provider.go
oarbusi: before this change, it was checking if PAK and AWS secret manager were set, if not, warning was raised. Now that we are adding service accounts we are changing the error message to include service accounts and we have reordered the credentials in the warnings to follow hierarchy

.github/workflows/acceptance-tests-runner.yml
oarbusi: we'll work on hierarchy in following tickets, and add the tests you are referring to

oarbusi:
> Based on your description, I not sure I understand how we're handling the different priorities between:
> 
> * environment variables, IDs and Secrets Manager
> * token, SAs and PAK

@marcosuma for now we have not changed anything in terms of priority/hierarchy of credentials. AWS Secret manager has the priority, then SA, and finally PAK. WE first look for provider inputs and then environment variables. Direct usage of Token is not yet implemented and also we have not implemented refresh policy of the token we generate with the SA. This will all come on following PRs

.github/workflows/acceptance-tests-runner.yml
marcosuma: ok I just want to make sure we're not missing an opportunity to avoid scattered logic in the code base. We should aim at a single place where priority, warnings and auth logic is managed. If that's the plan, all good.

internal/provider/provider.go
marcosuma: I just don't see the connection with https://docs.google.com/document/d/1PuAwTTNbVLUsqMH9wmmHKUCbEojhkQ_oHL2ONxwXZhs/edit?tab=t.0#bookmark=id.9ywd4yj7k619

How I see it is that there are two level of priorities, and this error message hints that we're mixing them

lantoli (approved):
LGTM, thanks for following-up the comments.  Consider doing old SDK and Realm changes in another PR

internal/config/client.go
```diff
@@ -97,15 +140,56 @@ type UAMetadata struct {
...
+		// 1. OAuth2 token caching: The oauth2 library only caches tokens after successful acquisition
+		// 2. Early credential validation: Fail fast during provider init rather than first resource operation
+		// 3. Performance: Subsequent requests use cached tokens instead of blocking for token acquisition
+		_, err := tokenSource.Token()
+		if err != nil {
+			return nil, fmt.Errorf("failed to acquire OAuth2 token: %w", err)
+		}
+
+		oauthClient := auth.NewClient(ctx, tokenSource)
+		tfLoggingTransport := logging.NewTransport("Atlas", oauthClient.Transport)
```
lantoli, line 180: nice that we're also logging calls to `api/oauth/token` so we can better diagnose possible network issues
```
Network Request Start: POST https://cloud-dev.mongodb.com/api/oauth/token (started at 2025-09-23T09:35:57.396658+02:00)
Network Request Complete: POST https://cloud-dev.mongodb.com/api/oauth/token - Status: 200 (Success) - Duration: 779.585333ms```

File: .github/workflows/acceptance-tests-runner.yml (hunk 12 above)
Comment (oarbusi) [reply .github/workflows/acceptance-tests-runner.yml:467]: yes that's the idea

File: internal/provider/provider.go (hunk 15 above)
Comment (oarbusi) [reply internal/provider/provider.go]: I think that's the wrong link, but the thing with AWS Secret manager is that it needs some inputs in the provider block to define secret name, aws role arn, etc. This error message is returned when there are no credentials set up at all

File: internal/config/client.go (hunk 10 above)
Comment (oarbusi) [reply internal/config/client.go:154]: So after testing both old SDK and the realm SDK:
- old deprecated SDK works with service accounts: we use the oauth client we create for the new SDK into the old and it just works(this is already working in this PR :) ) 
- realm SDK does not work with service accounts, because realm API does not support service accounts and only supports digest auth

File: internal/provider/credentials.go (hunk 14 above)
Comment (marcosuma) [reply internal/provider/credentials.go:79]: ok thanks - nit, but please make sure in the description we call out that hierarchy is not fully handled - at least for me I couldn't fully understand what to expect from the PR

File: internal/provider/provider.go (hunk 15 above)
Comment (marcosuma) [inline internal/provider/provider.go]:
```suggestion
	MissingAuthAttrError           = "either AWS Secrets Manager, Service Accounts or Atlas Programmatic API Keys attributes must be set"
```

marcosuma: sorry for the link, now it should be good 😄 

```
resp.Diagnostics.AddWarning(ProviderConfigError, MissingAuthAttrError)
```

correct me if I am wrong, but you are using this const to leave a warning, not an error. But your comment says `	// Check if any valid authentication method is provided` So should that really be an `AddError`?
I can see this is similar to the existing code, but as per [my comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/files#r2371439178) I'd like if we improve it.

IMO this is the part that it's confusing:
1. **error** if there is **no authentication** (neither through AWS SM, input creds nor env var)
2. **warning** if there is **more than one authentication** is provided (across all the combination between first- and second-level of priority)

Pseudo-code-wise, I'd expect a code that follows a PD structure:

```
if !hasAWSSecretManager and !hasInputCreds and !hasEnvVar then
   raiseError

if !hasMoreThanOneAuth then
   raiseWarning
   authMechanism = getHighestPriorityAuth(...)
```

internal/provider/credentials.go
oarbusi: added that the hierarchy will be implemented in following PR in the "important" note in the PR description

internal/provider/provider.go
oarbusi: Fair enough, we can change this so that we return an error instead of a warning. The behavior does not change because now we will fail early instead of doing unauthenticated requests which fail. Done in [cec3112](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/cec3112d438e0535837408bdab6c59be5e8e008b)
error looks like the following:
Screenshot 2025-09-23 at 14 21 57

oarbusi: changed here [a0e4e0a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3700/commits/a0e4e0a96a63f2007131f0b11f750d8c270ba679)

internal/provider/provider.go
oarbusi: on the warning when multiple credentials are set up, the idea is to do it when we implement credential hierarchy in CLOUDP-345764, CLOUDP-345765 and CLOUDP-345766

internal/provider/provider.go
marcosuma:
```
	if !config.HasValidAuthCredentials(&config.Config{
                PublicKey:    data.PublicKey.ValueString(),
		PrivateKey:   data.PrivateKey.ValueString(),
		ClientID:     data.ClientID.ValueString(),
		ClientSecret: data.ClientSecret.ValueString(),
	}) && !awsRoleDefined
```

is that equivalent to `if !hasAWSSecretManager and !hasInputCreds and !hasEnvVar then`?

from a quick look I am not sure about it because it seems `HasValidAuthCredentials` is only checking input creds (but I could be wrong) and nothing about env var.

oarbusi: yes it's equivalent. At this point in the code `data` has either the provider input or env var values, so checking data is like checking `hasInputCreds` and `hasEnvVar` of your pseudocode

marcosuma: ok maybe add a comment at the top of the if?

internal/provider/provider.go
marcosuma: to finalize this comment, we're not returning error (as it should be) if no auth is provided

PR #3711, opened 2025-09-24
chore: Implement Service Account token caching & thread-safe concurrent access

lantoli:
Implement Service Account token caching & thread-safe concurrent access.

Refresh policy will be implemented in a follow-up PR.

Also:
- Don't send AdvancedClusterPreview in User Agent anymore as the environment variable is not longer used.

Link to any related issue(s): CLOUDP-345762

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/config/client.go
+++ b/internal/config/client.go
@@ -175,7 +175,6 @@ func (c *Config) NewClient(ctx context.Context) (any, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to acquire OAuth2 token: %w", err)
 		}
-
 		oauthClient := auth.NewClient(ctx, tokenSource)
 		tfLoggingTransport := logging.NewTransport("Atlas", oauthClient.Transport)
 		oauthClient.Transport = tfLoggingTransport
--- a/internal/config/client.go
+++ b/internal/config/client.go
@@ -25,8 +25,6 @@ import (
 
 	"github.com/mongodb/terraform-provider-mongodbatlas/version"
 
-	"go.mongodb.org/atlas-sdk/v20250312007/auth/clientcredentials"
-
 	"go.mongodb.org/atlas-sdk/v20250312007/auth"
 )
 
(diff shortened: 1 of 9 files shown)


internal/testutil/unit/provider_mock.go
```diff
@@ -46,32 +46,16 @@ func (p *ProviderMocked) Configure(ctx context.Context, req fwProvider.Configure
 	if !ok {
 		p.t.Fatal("Failed to cast ResourceData to MongoDBClient")
 	}
-
```
lantoli, line 49: revert changes as not needed anymore

internal/service/advancedcluster/main_test.go
```diff
@@ -8,18 +8,6 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	// Only modify credentials for unit tests. Preserve GH Actions env in acceptance (TF_ACC=1).
```
lantoli, line 11: revert changes and move logic to init() in internal/testutil/acc/factory.go

internal/config/service_account.go
```diff
@@ -0,0 +1,51 @@
...
+	mu           sync.Mutex
+}{}
+
+func tokenSource(ctx context.Context, c *Config, base http.RoundTripper) (auth.TokenSource, error) {
+	saInfo.mu.Lock()
+	defer saInfo.mu.Unlock()
+
+	if saInfo.tokenSource != nil {
+		if saInfo.clientID != c.ClientID || saInfo.clientSecret != c.ClientSecret || saInfo.baseURL != c.BaseURL {
+			return nil, fmt.Errorf("service account credentials changed")
```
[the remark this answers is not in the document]
lantoli: at the moment it's ok as we don't expect this error to happen

internal/config/service_account.go
```diff
@@ -0,0 +1,51 @@
...
+	"golang.org/x/oauth2"
+)
+
+var saInfo = struct {
+	tokenSource  auth.TokenSource
+	clientID     string
+	clientSecret string
+	baseURL      string
+	mu           sync.Mutex
+}{}
```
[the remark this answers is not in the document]
lantoli: these values are also in different parts of this process and in environment variables

internal/config/service_account.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx = context.WithValue(ctx, auth.HTTPClient, &http.Client{Transport: base})
+	token, err := conf.TokenSource(ctx).Token()
+	if err != nil {
+		return nil, err
+	}
+	saInfo.clientID = c.ClientID
+	saInfo.clientSecret = c.ClientSecret
+	saInfo.baseURL = c.BaseURL
+	// TODO: token will be refreshed in a follow-up PR
+	saInfo.tokenSource = oauth2.StaticTokenSource(token)
```
oarbusi, line 49: why this instead of `conf.TokenSource(ctx)`?

internal/config/client.go
```diff
@@ -100,15 +96,14 @@ type MongoDBClient struct {
...
 type Config struct {
-	AssumeRole                      *AssumeRole
-	PublicKey                       string
-	PrivateKey                      string
-	BaseURL                         string
-	RealmBaseURL                    string
-	TerraformVersion                string
-	ClientID                        string
-	ClientSecret                    string
-	PreviewV2AdvancedClusterEnabled bool
```
marcosuma, line 111: did we forget about these?

internal/config/service_account.go
```diff
@@ -0,0 +1,51 @@
...
+
+var saInfo = struct {
+	tokenSource  auth.TokenSource
+	clientID     string
+	clientSecret string
+	baseURL      string
+	mu           sync.Mutex
+}{}
+
+func tokenSource(ctx context.Context, c *Config, base http.RoundTripper) (auth.TokenSource, error) {
```
marcosuma, line 23: curious: where is the concurrent access coming from?

context: if we're saying in the same terraform execution (e.g. terraform apply/plan - this is the process context, right) we're saying there are multiple threads that arrive here to get a token. I am wondering is it always the same token we're returning?

oarbusi: multiple operations can be done in parallel(https://developer.hashicorp.com/terraform/cli/commands/apply#parallelism-n) in the same terraform apply/plan. So we need to be sure the token used by all the parallel process is the same and if one of the threads needs to refresh it, it is refreshed only once and the other threads will use the newly generated token instead of trying to generate a new one. Hope that makes the motivation clearer

internal/config/service_account.go
oarbusi: seems like we are going backwards using static token source here instead of just TokenSource as we were doing

lantoli: I don't think we can use TokenSource here as it's not thread-safe, and for instance we're sharing it with old & new SDK, apart from the resource parallelism you mention in another comment. 
We would need a TokenSource wrapper with serialized access.

Anyway the follow-up PR will change this.

internal/config/client.go
lantoli: yes, leftover, added in the PR description

internal/config/service_account.go
marcosuma: ok great , just want to make sure here's where **we are not** creating a token for each thread, but I think you all got this

marcosuma (approved):

oarbusi (approved):

PR #3723, opened 2025-09-30
chore: Allow SA for `mongodbatlas_project_api_key resource`

lantoli:
## Description

Allow SA for `mongodbatlas_project_api_key` resource by changing `getKeyDetails` implementation so it works with PAK or SA credentials.

Note: it's ok to do it in master as this change is valid for PAK and SA.

Link to any related issue(s): CLOUDP-348041

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/projectapikey/model_project_api_key.go
+++ b/internal/service/projectapikey/model_project_api_key.go
@@ -67,23 +67,24 @@ func sameRoles(roles1, roles2 []string) bool {
 
 // getKeyDetails returns nil error and nil details if not found as it's not considered an error
 func getKeyDetails(ctx context.Context, connV2 *admin.APIClient, apiKeyID string) (*admin.ApiKeyUserDetails, string, error) {
-	root, _, err := connV2.RootApi.GetSystemStatus(ctx).Execute()
+	resp, _, err := connV2.OrganizationsApi.ListOrgs(ctx).Execute()
 	if err != nil {
 		return nil, "", err
 	}
-	for _, role := range root.ApiKey.GetRoles() {
-		if orgID := role.GetOrgId(); orgID != "" {
-			key, _, err := connV2.ProgrammaticAPIKeysApi.GetOrgApiKey(ctx, orgID, apiKeyID).Execute()
-			if err != nil {
-				if admin.IsErrorCode(err, "API_KEY_NOT_FOUND") {
-					return nil, orgID, nil
-				}
-				return nil, orgID, fmt.Errorf("error getting api key information: %s", err)
-			}
-			return key, orgID, nil
-		}
+	orgIDs := resp.GetResults()
+	if len(orgIDs) == 0 {
+		return nil, "", fmt.Errorf("no organizations found")
 	}
-	return nil, "", nil
+	// At present a PAK or SA belongs to exactly one organization. If this changes in the future, this logic will need to be updated.
+	orgID := orgIDs[0].GetId()
+	key, _, err := connV2.ProgrammaticAPIKeysApi.GetOrgApiKey(ctx, orgID, apiKeyID).Execute()
+	if err != nil {
+		if admin.IsErrorCode(err, "API_KEY_NOT_FOUND") {
+			return nil, orgID, nil
+		}
+		return nil, orgID, fmt.Errorf("error getting api key information: %s", err)
+	}
+	return key, orgID, nil
 }
 
 func validateUniqueProjectIDs(d *schema.ResourceData) error {


internal/service/projectapikey/model_project_api_key.go
```diff
@@ -67,23 +67,24 @@ func sameRoles(roles1, roles2 []string) bool {
...
-				}
-				return nil, orgID, fmt.Errorf("error getting api key information: %s", err)
-			}
-			return key, orgID, nil
+	orgIDs := resp.GetResults()
+	if len(orgIDs) == 0 {
+		return nil, "", fmt.Errorf("no organizations found")
+	}
+	// At present a PAK or SA belongs to exactly one organization. If this changes in the future, this logic will need to be updated.
+	orgID := orgIDs[0].GetId()
```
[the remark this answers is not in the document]
lantoli: at the moment there is only 1 org

internal/service/projectapikey/model_project_api_key.go
```diff
@@ -67,23 +67,24 @@ func sameRoles(roles1, roles2 []string) bool {
...
-					return nil, orgID, nil
-				}
-				return nil, orgID, fmt.Errorf("error getting api key information: %s", err)
-			}
-			return key, orgID, nil
+	orgIDs := resp.GetResults()
+	if len(orgIDs) == 0 {
+		return nil, "", fmt.Errorf("no organizations found")
+	}
+	// At present a PAK or SA belongs to exactly one organization. If this changes in the future, this logic will need to be updated.
```
manupedrozo, line 78: Nice, should we create a ticket on our side to track this?

manupedrozo (approved):

internal/service/projectapikey/model_project_api_key.go
```diff
@@ -67,23 +67,24 @@ func sameRoles(roles1, roles2 []string) bool {
 
 // getKeyDetails returns nil error and nil details if not found as it's not considered an error
 func getKeyDetails(ctx context.Context, connV2 *admin.APIClient, apiKeyID string) (*admin.ApiKeyUserDetails, string, error) {
-	root, _, err := connV2.RootApi.GetSystemStatus(ctx).Execute()
+	resp, _, err := connV2.OrganizationsApi.ListOrgs(ctx).Execute()
```
marcosuma, line 70: This was a good catch. any chance this getSystemStatus has different permission level and/or returns different amount of information? 

You're here picking the first orgID, could that be different than the one we currently fetch?

AgustinBettati: Alternatively we could leverage any project from `project_assignments` and obtain the orgID from the response of  [getProject](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-getgroup), holding same assumption of `At present a PAK or SA belongs to exactly one organization`

internal/service/projectapikey/model_project_api_key.go
lantoli: not sure as there is not any work to do, just wait in case this assumption changes in the future, I'll add more info in the next comment

internal/service/projectapikey/model_project_api_key.go
lantoli: According to the [doc](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-listorgs) this new endpoint only needs Organization Member role which is the standard role so I think we're ok with the permissions.

Note that the assumption of PAK or SA belonging to only 1 org was already being made (although not explicitly documented). 

> You're here picking the first orgID, could that be different than the one we currently fetch?

@marcosuma  We got confirmation that this can't happen now, APK and SA only belong to 1 org. Before this PR we would iterate through all the roles from the logged APK until finding an org role and returning the associated OrgID, so the OrgID must be the same as the new way, that directly returns the OrgID the PAK or SA belongs to.

In the future PAK or SA could belong to multiple orgs. But in that case more changes would be needed. For example project resource would need to get the OrgID (and the 
[Create project endpoint](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-creategroup) to be updated to receive it). 

Atlas API and some Terraform resources are currently working with the (implicit) idea of a current  org.

@AgustinBettati  yes, that would work although seems similar, do you see some benefit to get it from some project id?     Also the user could potentially mix assignments from projects in multiple orgs (so we would depend to pick the OrgID on the order), although that won't work at the moment.

oarbusi (approved):

PR #3736, opened 2025-10-01
chore: Support generation of singleton resources with no delete operation

manupedrozo:
## Description

Adding support for generating singleton resources that have no delete operation defined in the API.

An example, is the newly generated: `project_settings_api`.
- Tests to be added in a separate ticket

Link to any related issue(s): CLOUDP-347905

Removed from PR but keeping for context:
```markdown
Re-generated existing resources to use the updated case formatting (which fixes API mapping for the updated attributes):
- `maintenance_window_api`
- `database_user_api`
- `cluster_api`

#### On attribute case formatting:
Fixed some instances of case formatting conversion causing mismatches between the TF model and the input expected by the API.

These issues come from the conversion from `camelCase` to `snake_case` and viceversa, which is "irreversible". 

For example: An API attribute called `testAPIAttribute` would get converted to `test_apiattribute` and then converted back to `testApiattribute`, which does not match the original `testAPIAttribute` that the API expects as input to POST/PATCH operations, nor what we expect as a response to GET operations.

The problem is not fully solved, since we do not type child objects in the generated TFModels, child object attributes are forced to go through a "irreversible" conversion when marshalling/unmarshalling (we cannot use reflection).

Further work is needed for a robust solution. Typing child objects should help, I am probably missing context here though.
```

## Type of change:

## Required Checklist:


internal/common/autogen/marshal.go
```diff
@@ -62,7 +63,7 @@ func marshalAttrs(valModel reflect.Value, isUpdate bool) (map[string]any, error)
 }
 
 func marshalAttr(attrNameModel string, attrValModel reflect.Value, objJSON map[string]any, isUpdate, includeNullOnUpdate bool) error {
-	attrNameJSON := toJSONName(attrNameModel)
+	attrNameJSON := stringcase.Uncapitalize(attrNameModel)
```
marcosuma: reading your description, this is not about "Support generation of singleton resources" but fixing other issues, right? This is probably where it is worth it having a separate PR, because by quickly looking at the PR it seems like there are many changes that are related to the fix and not to "Support generation of singleton resources"

manupedrozo: Yes, this targets the "attribute case formatting" side of things. I had to fix this issue as part of the generation of the new `project_settings_api` resource, so changes ended up together. It was not quick to go back and split it into two PRs but I can do that if this is a blocker.

marcosuma: thanks for explaining, it might make the PR harder to review but ok to keep it as it is if it's hard to split (I haven't checked myself)

tools/codegen/codespec/model.go
```diff
@@ -25,10 +27,10 @@ type Resource struct {
 }
 
 type APIOperations struct {
-	Create        APIOperation
-	Read          APIOperation
-	Update        APIOperation
-	Delete        APIOperation
+	Create        *APIOperation
+	Read          *APIOperation
+	Update        *APIOperation
```
AgustinBettati: Would revert to leaving value type (no pointer) so it is explicit which operations are optional and which not. As of these changes we now correctly handle the optional delete.

tools/codegen/config.yml
```diff
@@ -274,6 +278,54 @@ resources:
...
+      method: PATCH
+    update:
+      path: /api/atlas/v2/groups/{groupId}/settings
+      method: PATCH
+    version_header: application/vnd.atlas.2023-01-01+json
+    schema:
+      aliases:
+        # "AI" is all uppercase and does not play nice with camelCase to snake_case conversion.
+        is_data_explorer_gen_aifeatures_enabled: is_data_explorer_gen_ai_features_enabled
+        is_data_explorer_gen_aisample_document_passing_enabled: is_data_explorer_gen_ai_sample_document_passing_enabled
```
AgustinBettati, line 295: lets sync offline, curious about this usage as currently alias config is very limited (used for path param renaming) and not considered in conversion logic.

tools/codegen/gofilegen/codetemplate/template.go
```diff
@@ -27,11 +27,11 @@ type ResourceFileInputs struct {
 }
 
 type APIOperations struct {
+	Create        *Operation
+	Read          *Operation
+	Update        *Operation
```
AgustinBettati: Similar to change in the codespec/model.go, dont believe we currently have a case where these operations are not defined

tools/codegen/gofilegen/resource/resource_file_test.go
```diff
@@ -3,9 +3,9 @@ package resource_test
 import (
 	"testing"
 
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/autogen/stringcase"
 	"github.com/mongodb/terraform-provider-mongodbatlas/tools/codegen/codespec"
 	"github.com/mongodb/terraform-provider-mongodbatlas/tools/codegen/gofilegen/resource"
-	"github.com/mongodb/terraform-provider-mongodbatlas/tools/codegen/stringcase"
```
AgustinBettati, line 8: Can we add a unit test here defining a model that does not have the delete operation?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -517,6 +559,56 @@ func TestConvertToProviderSpec_pathParamPresentInPostRequest(t *testing.T) {
 	runTestCase(t, tc)
 }
 
+func TestConvertToProviderSpec_singletonResourceNoDeleteOperation(t *testing.T) {
```
AgustinBettati, line 520: nice

AgustinBettati (commented):
Singleton resource changes LGTM with small doubt on model/config adjustment.

Related to attribute case formatting lets sync offline. If possible highly agree we split into separate PR, even if we currently reach a non-functional project settings resource. This way we can look into the problem separately, understand different cases, and see what areas require adjustments.

tools/codegen/codespec/model.go
manupedrozo: That was my initial approach, ended up going with pointer for all to make the code for handling these more consistent (instead of dealing with a ref in some cases and a value in others). 

Note that in both cases we get a panic if trying to generate a resource with these undefined, so using value type does not prevent that.

My suggestion: Use the same type here and add a check to raise an explicit error when required operations are not defined.

What do you think?

tools/codegen/gofilegen/codetemplate/template.go
manupedrozo: See other comment, whatever we decide we can apply here as well.

tools/codegen/gofilegen/resource/resource_file_test.go
manupedrozo: Yes, good catch! [3fe8675](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3736/commits/3fe8675028bdbbae516305677f615e7a3cc5955a)

tools/codegen/config.yml
manupedrozo: Yes, this case does work (with the case formatting changes) since we are just adding a "_". 
If it was changing any letter it would indeed break the runtime conversion.

manupedrozo (commented):
Thanks for the comments @marcosuma @AgustinBettati. 
I'll split out the case formatting changes since they affect all generated resources (probably worth its own ticket).
@AgustinBettati yes let's sync tomorrow on both parts.

internal/common/autogen/marshal.go
manupedrozo: Done. Not hard, just not quick :)

internal/serviceapi/projectsettingsapi/resource_schema.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+	}
+}
+
+type TFModel struct {
+	GroupId                                         types.String `tfsdk:"group_id" autogen:"omitjson"`
+	IsCollectDatabaseSpecificsStatisticsEnabled     types.Bool   `tfsdk:"is_collect_database_specifics_statistics_enabled"`
+	IsDataExplorerEnabled                           types.Bool   `tfsdk:"is_data_explorer_enabled"`
+	IsDataExplorerGenAiFeaturesEnabled              types.Bool   `tfsdk:"is_data_explorer_gen_ai_features_enabled"`
+	IsDataExplorerGenAiSampleDocumentPassingEnabled types.Bool   `tfsdk:"is_data_explorer_gen_ai_sample_document_passing_enabled"`
```
manupedrozo, line 67: For example, these two attributes don't work without the case formatting changes since the API expects `xxxGenAIxxx` and we are passing/reading `xxxGenAixxx`.

internal/serviceapi/projectsettingsapi/resource.go
```diff
@@ -0,0 +1,120 @@
+// Code generated by terraform-provider-mongodbatlas using `make generate-resource`. DO NOT EDIT.
+
+package projectsettingsapi
```
lantoli, line 3: nit: i know it's in the PR description that will be done later, but i really think it's a good idea to have a basic acc test in the same PR to show that the resource works

tools/codegen/codespec/testdata/api-spec.yml
```diff
@@ -565,3 +608,8 @@ components:
       properties:
         levelField1:
           type: string
+    TestSingletonResource:
+      type: object
+      properties:
+        flag:
+          type: boolean
```
lantoli: end last line with newline, also for other files

tools/codegen/codespec/model.go
lantoli: don't have all the context but i'm inclined to have pointer only for optional operations

tools/codegen/codespec/testdata/api-spec.yml
AgustinBettati: there is a vs code setting you can adjust so it does it automatically for you

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -49,6 +57,25 @@ func ToCodeSpecModel(atlasAdminAPISpecFilePath, configPath string, resourceName
 	return &Model{Resources: results}, nil
 }
 
+func validateRequiredOperations(resourceConfigs map[string]config.Resource) error {
```
AgustinBettati, line 60: thanks for the extra care here

AgustinBettati (approved):
LGTM 🚀  thanks for addressing comments

internal/serviceapi/projectsettingsapi/resource.go
manupedrozo: Agree, will try to do that moving forward. We need to fix the case formatting issues in order to tests all attributes for this one.

PR #3740, opened 2025-10-02
chore: Fix case formatting of generated TF model fields not matching API fields

manupedrozo:
## Description

Fixed case formatting conversion causing mismatches between the generated TF model and the API-defined fields.

These issues come from the conversion from `camelCase` to `snake_case` and viceversa, which is "irreversible". 

For example: An API attribute called `testAPIAttribute` would get converted to `test_apiattribute` and then converted back to `testApiattribute`, which does not match the original `testAPIAttribute` that the API expects as input to POST/PATCH operations, nor what we expect as a response to GET operations.

This PR fixes the mapping for the TF **root** model.
Since we do not type child objects in the generated TF models, child object attributes are forced to go through an "irreversible" conversion when marshalling/unmarshalling in runtime (we cannot use reflection). 
> Typing of child objects will be covered in another ticket.

Re-generated existing resources to use the updated case formatting:
- `maintenance_window_api`
- `database_user_api`
- `cluster_api`
- `project_settings_api`

Link to any related issue(s): CLOUDP-349085

## Type of change:

## Required Checklist:


internal/serviceapi/clusterapi/resource_test.go
```diff
@@ -43,7 +43,7 @@ func TestAccClusterAPI_basic(t *testing.T) {
 				ImportStateVerifyIdentifierAttribute: "name",
 				ImportStateVerifyIgnore: []string{
 					"retain_backups_enabled",   // This field is TF specific and not returned by Atlas, so Import can't fill it in.
-					"mongo_db_major_version",   // Risks plan change of 8 --> 8.0 (always normalized to `major.minor`)
+					"mongo_dbmajor_version",    // Risks plan change of 8 --> 8.0 (always normalized to `major.minor`)
```
marcosuma, line 46: q: why did this change? I was expecting to stay as `mongo_db_major_version` ?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -38,55 +38,64 @@ func TestConvertToProviderSpec(t *testing.T) {
 					Attributes: codespec.Attributes{
 						{
 							Name:                     "bool_default_attr",
+							PascalCaseName:           "BoolDefaultAttr",
```
marcosuma: maybe missing some context, but shouldn't we have a test case like `ThisIsAnAPIField` that becomes `this_is_an_api_field`?

marcosuma (approved):

internal/serviceapi/clusterapi/resource_test.go
manupedrozo: Since this attribute was previously incorrectly formatted (`mongoDbMajorVersion` instead of `mongoDBMajorVersion`), it did not show up in the state as it was wrongly mapped. 

`mongo_db_major_version` was not the correct value here (see old schema [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/serviceapi/clusterapi/resource_schema.go#L218)), it just wasn't doing anything because the mapping was broken.

 In order to get `mongo_db_major_version`, we'd need to add an alias or do some updates to our API field -> TF schema attribute name conversion (which we are discussing).

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
manupedrozo: The conversion from camelCase (API field) to snake_case (TF schema) is not modified. 
Something like `ThisIsAnAPIField` is still being mapped to `this_is_an_apifield`.

We are discussing updating this conversion logic so that it handles the multiple-uppercase letters as expected. On top of that, we can provide manual aliases for specific cases.

This PR is focused on the TF model case formatting, so that the mapping from TF model to the API fields works correctly.
Following the same example, before we would have had:
```go
type TFModel struct {
	ThisIsAnApifield  types.List   `tfsdk:"this_is_an_apifield"` // <- wrong casing, won't map to API correctly
	// ThisIsAnApifield is sent to the API as `thisIsAnApifield`, which is not the correct field name
}
```
And now we have:
```go
type TFModel struct {
	ThisIsAnAPIField  types.List   `tfsdk:"this_is_an_apifield"` // <- correct casing, maps to API correctly
	// ThisIsAnAPIField is sent to to the API as `thisIsAnAPIField`, which is the correct field name
}
```
In both cases, the TF schema attribute name stays the same.

AgustinBettati: Thanks for the very clear example. 
Can we create a followup ticket to improve the aspect of `API field -> TF schema attribute name conversion`? Should not be complex to reach the expected `this_is_an_api_field` or `mongo_db_major_version`, we can evaluate when to priotize this work.

internal/common/autogen/name.go
```diff
@@ -1,17 +0,0 @@
...
-	return xstrings.ToCamelCase(name)
-}
-
-func toTfSchemaName(name string) string {
-	return xstrings.ToSnakeCase(name)
-}
-
-func toTfModelName(name string) string {
-	return xstrings.ToPascalCase(name)
-}
```
AgustinBettati, line 17: nit: I find it easier to follow code when functions use terms `JSONName`, `TFSchemaName`, `TFModelName` vs direct `camelCase`, `snakeCase`, `PascalCase`. Not sure if after these changes some of these function became redundant.

internal/common/autogen/stringcase/string_case.go
```diff
@@ -46,3 +48,39 @@ func FromCamelCase(input string) SnakeCaseString {
...
+	return xstrings.ToSnakeCase(str)
+}
+
+func Capitalize(str string) string {
+	return capitalization(str, true)
+}
+
+func Uncapitalize(str string) string {
+	return capitalization(str, false)
+}
```
AgustinBettati, line 66: Should we use xstring.FirstRuneToUpper/xstring.FirstRuneToLower ([reference](https://github.com/huandu/xstrings?tab=readme-ov-file#package-xstrings-functions))?

tools/codegen/codespec/model.go
```diff
@@ -78,6 +80,7 @@ type Attribute struct {
 	CustomType               *CustomType
 	ComputedOptionalRequired ComputedOptionalRequired
 	Name                     stringcase.SnakeCaseString
+	PascalCaseName           string
```
AgustinBettati: just a thought, should we rename these variables so that the intention is more clear? e.g.`TFSchemaName`, `APIPropertyName`.

AgustinBettati (approved):
LGTM

internal/common/autogen/name.go
manupedrozo: I agree, wanted to make it explicit for the time being since we should actually remove the usage of these.

internal/common/autogen/stringcase/string_case.go
manupedrozo: Nice catch. Given that we need to change camelCase -> snake_case conversion for a custom one, I don't think we'll be using xstring at all. I'll come back to this once its clear.

tools/codegen/codespec/model.go
manupedrozo: Yes, let me do that. Second one should be `TFModelName`.

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
manupedrozo: Yes, we need a custom conversion instead of using xstring for this. Should be straightforward.

tools/codegen/codespec/model.go
manupedrozo: Done at [5ecbd3](https://github.com/mongodb/terraform-provider-mongodbatlas/commit/5ecbd3273235a899c6019c39b4e93301b558774f)

lantoli (approved):

PR #3756, opened 2025-10-07
chore: Add codegen intermediate model serialization

manupedrozo:
## Description

Add intermediate model serialization to enable manual changes of the generated resource models when necessary.

This explicitly separates the two codegen steps:
  1) Process the API spec and generate intermediate resource models
  2) Use intermediate resource models to generate Go code

1\. Intermediate models are generated based on the codegen config (config.yaml) and serialized into yaml files under `tools/codegen/models/<resource_name>.yaml`

2\. The `models/` dir is traversed and resources are generated based on the files present. 
\- A resource may not be present in the config.yaml but have a (likely manually modified) serialized model.

When a `resourceName` argument is provided to generate a single resource, the intermediate model is first generated (or skipped if the resource is not in the config). Then the serialized model file is read and the resource code generated.

Link to any related issue(s): CLOUDP-347912

## Type of change:

## Required Checklist:


tools/codegen/codespec/model.go
```diff
@@ -21,69 +21,69 @@ type Model struct {
...
+	Timeouts                 *TimeoutsAttribute         `yaml:"timeouts,omitempty"`
+	SingleNested             *SingleNestedAttribute     `yaml:"single_nested,omitempty"`
+	SetNested                *SetNestedAttribute        `yaml:"set_nested,omitempty"`
+	Description              *string                    `yaml:"description,omitempty"`
+	DeprecationMessage       *string                    `yaml:"deprecation_message,omitempty"`
+	CustomType               *CustomType                `yaml:"custom_type,omitempty"`
+	ComputedOptionalRequired ComputedOptionalRequired   `yaml:"computed_optional_required"`
+	TFSchemaName             stringcase.SnakeCaseString `yaml:"tf_schema_name"`
+	TFModelName              string                     `yaml:"tf_model_name"`
+	ReqBodyUsage             AttributeReqBodyUsage      `yaml:"req_body_usage"`
```
AgustinBettati, line 84: We could consider changing the type of ReqBodyUsage to string similar to how we handle ComputedOptionalRequired, makes it readable in the model.

tools/codegen/main.go
```diff
@@ -27,21 +30,61 @@ func main() {
 		log.Fatalf("an error occurred when downloading Atlas Admin API spec: %v", err)
 	}
 
-	model, err := codespec.ToCodeSpecModel(specFilePath, configPath, resourceName)
-	if err != nil {
-		log.Fatalf("an error occurred while generating codespec.Model: %v", err)
+	{
```
AgustinBettati, line 33: curious about the redundant curly brackets, added intentional to separate related statements? We can also consider extracting all the model serialization to separate function to keep main more concise

AgustinBettati (approved):
nice!

tools/codegen/models/maintenance_window_api.yaml
```diff
@@ -0,0 +1,116 @@
...
+        path: /api/atlas/v2/groups/{groupId}/maintenanceWindow
+        static_request_body: ""
+    read:
+        http_method: GET
+        path: /api/atlas/v2/groups/{groupId}/maintenanceWindow
+        static_request_body: ""
+    update:
+        http_method: PATCH
+        path: /api/atlas/v2/groups/{groupId}/maintenanceWindow
+        static_request_body: ""
```
AgustinBettati: nit: can we put an omit empty on this one as well?

tools/codegen/main.go
manupedrozo: Yes, just adding the explicit scope. I moved to a function initially and discarded it since it just made things harder to follow, especially the file writing/reading.

tools/codegen/codespec/model.go
manupedrozo: We can do that. [69d125f](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3756/commits/69d125fb0916226899af847f999ca514d7ea7c25)

tools/codegen/models/maintenance_window_api.yaml
manupedrozo: Yes, [69d125f](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3756/commits/69d125fb0916226899af847f999ca514d7ea7c25)

tools/codegen/models/auditing_api.yaml
```diff
@@ -0,0 +1,64 @@
+schema:
+    description: Updates the auditing configuration for the specified project. The auditing configuration defines the events that MongoDB Cloud records in the audit log. To use this resource, the requesting Service Account or API Key must have the Project Owner role. This feature isn't available for `M0`, `M2`, `M5`, or serverless clusters.
+    attributes:
+        - bool:
```
lantoli, line 4: Seems strange that there is not a "type" attribute and bool, string, etc. are directly as attribute root elements.  
For instance, is it valid an attribute with both bool and string?    

i suppose it's to mimic TF schema structure? That's limited by Go so I don't know if we can do it differently in our model, or better to keep it in this way

manupedrozo: Its 1 to 1 with the Attribute struct yes. Having multiple would be valid serialization wise, its up to the codegen logic how that is handled (see order [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/65d7a6c8d6d8b56b6cb705eb55792bb19249bcf5/tools/codegen/gofilegen/schema/schema_attribute.go#L29-L28)).

I wouldn't make any change to the serialization. 
We can evaluate representing the type differently in the model (like a type enum + any).

lantoli: ok

PR #3801, opened 2025-10-21
chore: Add support for typed nested objects in autogen

manupedrozo:
## Description

Adding a custom Object terraform type for autogen usage which keeps a reference to the underlying type, enabling traversal of the entire struct hierarchy during marshal & unmarshal operations.

Previously we could only access struct tags of the root generated TF Model:
- Nested object attributes were converted on a best-effort basis using field names, causing mismatches between the API property names and TF schema attribute names in some cases. 
- Logic that depends on tags (e.g. `autogen:"omitjsonupdate"`) could not be applied to attributes in nested objects.

With this change, we now access tags for nested objects during conversion and apply the same logic as for the root TF Model.

Support for other nested attribute types (list, set & map) will be added other PRs.

Added a temporary `use_custom_nested_types` flag to the autogen config so we can control the resources for which we use the new custom types until they are all implemented. Mixing custom nested types and non-custom nested types does not work as it breaks the hierarchy traversal (disabled custom types for `cluster_api`). 

Custom TF types docs: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom

Link to any related issue(s): CLOUDP-349095

## Type of change:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.

## Required Checklist:


internal/testutil/unit/provider_mock.go
```diff
@@ -51,6 +51,7 @@ func (p *ProviderMocked) Configure(ctx context.Context, req fwProvider.Configure
 	originalClient := client.AtlasV2.GetConfig().HTTPClient
 	if originalClient == nil {
 		p.t.Fatal("HTTPClient is nil, mocking will fail")
+		return // Unnecessary return. Avoids staticcheck issue.
```
manupedrozo, line 54: Unrelated. Fix for failing git pre-commit hook.

tools/codegen/config.yml
```diff
@@ -61,6 +61,7 @@ resources:
         target_states: ["DELETED"] # DELETED is a special state value when API returns 404 or empty object
     version_header: application/vnd.atlas.2024-08-05+json
     schema:
+      use_custom_nested_types: false
```
manupedrozo, line 64: Temporarily disabling usage of custom nested types for cluster_api until we implement typing support in other nested attributes.

internal/common/autogen/marshal.go
```diff
@@ -100,28 +102,36 @@ func getModelAttr(val attr.Value) (any, error) {
...
 	case types.Set:
-		return getListAttr(v.Elements())
+		return getListAttr(v.Elements(), isUpdate)
 	case jsontypes.Normalized:
 		var valueJSON any
 		if err := json.Unmarshal([]byte(v.ValueString()), &valueJSON); err != nil {
 			return nil, fmt.Errorf("marshal failed for JSON custom type: %v", err)
 		}
 		return valueJSON, nil
+	case customtype.ObjectValueInterface:
```
lantoli, line 118: update/create unit tests for any change in marshal/unmarshal

internal/common/autogen/unknown.go
```diff
@@ -40,9 +41,30 @@ func prepareAttr(value attr.Value) (attr.Value, error) {
 		return value, nil
 	}
 	ctx := context.Background()
+
+	if v, ok := value.(customtype.ObjectValueInterface); ok {
```
lantoli, line 45: can't it go in the value.(type) switch?

(also as in previous comment please add unit tests)

internal/common/autogen/customtype/object.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+type ObjectValueInterface interface {
+	basetypes.ObjectValuable
+	ValuePtrAsAny(ctx context.Context) (any, diag.Diagnostics)
+	NewObjectValue(ctx context.Context, value any) ObjectValueInterface
+	NewObjectValueNull(ctx context.Context) ObjectValueInterface
+}
+
+func (v ObjectValue[T]) NewObjectValue(ctx context.Context, value any) ObjectValueInterface {
```
lantoli, line 132: nit: as this implementation seems like it has a lot of boilerplate, don't know if we can take advantage of some existing structs/funcs to reduce code

internal/common/customtype/object.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if valueType.Kind() != reflect.Struct {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic(
+			"Error getting value attribute types",
+			fmt.Sprintf(`%T has usupported type: %s`, value.Interface(), valueType),
+		)}
+	}
+
+	attributeTypes := make(map[string]attr.Type)
+	for i := 0; i < valueType.NumField(); i++ {
```
lantoli:
```suggestion
	for i := range valueType.NumField() {
```

a little bit more Go idiomatic

internal/common/autogen/unknown.go
manupedrozo: We need to handle custom types before the value.IsUnknown() check below since the logic is different for these types. Will move to a switch as we add other custom types and then revisit the general structure once all custom types are implemented.

internal/common/autogen/customtype/object.go
manupedrozo: Besides the usage of the underlying `basetypes` functions when needed, everything else needs to be specific to this type. 
Did you have anything in mind?

internal/common/customtype/object.go
manupedrozo: Nice, updated older usages as well. [a18325b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3801/commits/a18325bd6a230b104de8f04f8a119312ee641b0d)

internal/common/autogen/marshal.go
manupedrozo: Good call, added [a18325b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3801/commits/a18325bd6a230b104de8f04f8a119312ee641b0d)

internal/common/autogen/customtype/object.go
```diff
@@ -0,0 +1,209 @@
+package customtype
```
AgustinBettati, line 1: should we place this package within autogen directory? Don't believe we would ever want regular resources leveraging these custom types

internal/common/customtype/object.go
```diff
@@ -0,0 +1,209 @@
...
+	"fmt"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// Custom types: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom
```
AgustinBettati: leaving a why and intended usage can definitely help for future reference/new joiners

internal/common/autogen/customtype/object.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return valuePtr, diags
+}
+
+func getAttributeTypes[T any](ctx context.Context) (map[string]attr.Type, diag.Diagnostics) {
+	var t T
+	return valueToAttributeTypes(ctx, reflect.ValueOf(t))
+}
+
+func valueToAttributeTypes(ctx context.Context, value reflect.Value) (map[string]attr.Type, diag.Diagnostics) {
```
AgustinBettati, line 202: q: Will there be potential for code reuse when implementing custom type of other nested types?

internal/serviceapi/resourcepolicyapi/resource_schema.go
```diff
@@ -17,6 +18,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"created_by_user": schema.SingleNestedAttribute{
 				Computed:            true,
 				MarkdownDescription: "The user that last updated the atlas resource policy.",
+				CustomType:          customtype.NewObjectType[TFCreatedByUserModel](ctx),
```
AgustinBettati, line 21: question for my understanding: Currently we leverage the type defined in the TFModel (e.g. customtype.ObjectValue[TFCreatedByUserModel]) during conversion logic. Why is the custom type required in the schema definition? Is there some form of error returned by TPF due to mismatch of TFModel with TF schema?

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -82,11 +82,14 @@ func apiSpecResourceToCodeSpecModel(oasResource APISpecResource, resourceConfig
 	updateOp := oasResource.UpdateOp
 	readOp := oasResource.ReadOp
 
-	createPathParams := pathParamsToAttributes(createOp)
-	createRequestAttributes := opRequestToAttributes(createOp)
-	updateRequestAttributes := opRequestToAttributes(updateOp)
-	createResponseAttributes := opResponseToAttributes(createOp)
-	readResponseAttributes := opResponseToAttributes(readOp)
+	// TODO remove useCustomNestedTypes usage once typing is supported for all nested attributes: object, list, set & map - CLOUDP-349095
```
AgustinBettati: As part of this clean up task can we make sure to also remove redundant conversion logic? Maybe a separate CLOUDP capturing these cleanup aspects can be good.

AgustinBettati (commented):
Good to add test as Leo mentioned. 
Added some questions but overall LGTM, great job on the incremental approach

internal/common/customtype/object.go
manupedrozo: Extending this comment and adding usage [a18325b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3801/commits/a18325bd6a230b104de8f04f8a119312ee641b0d)

internal/common/autogen/customtype/object.go
manupedrozo: This conversion will be reused yes!

tools/codegen/codespec/api_to_provider_spec_mapper.go
manupedrozo: Yes, we'll completely replace non-custom types across the autogen code. 
I will create tickets for other types as well. Updated the ref here [4890593](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3801/commits/489059354562cb3969e297e793ecc0844cae0d52)

internal/common/autogen/customtype/object.go
manupedrozo: Agree, moving it [a18325b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3801/commits/a18325bd6a230b104de8f04f8a119312ee641b0d)

internal/serviceapi/resourcepolicyapi/resource_schema.go
manupedrozo: This type is the one that TF will use for this attribute as opposed to the default `basetypes.ObjectType`. 
For example, when marshaling we traverse the TF schema and check each attribute's type. 
We need the type for this attribute in the schema to be `customtype.ObjectType[T]` and not `basetypes.ObjectType` so that we can access the underlying type `T`.

internal/common/autogen/marshal_test.go
```diff
@@ -282,6 +284,91 @@ func TestMarshalUpdateNull(t *testing.T) {
...
+			}),
+		}),
+	}
+
+	const expectedCreateJSON = `
+		{
+			"attrObjectBasic": {
+				"attrInt": 1,
+				"attrObjectOmitUpdate": {}
+			},
```
AgustinBettati, line 354: nice test, this got me thinking if we should also add a test case of multiple uppercase letters (e.g. awsIAMType) at both root and now nested level for object type.

AgustinBettati (approved):

internal/common/autogen/marshal_test.go
manupedrozo: Sure, added here + rebase [5ef1ee5](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3801/commits/5ef1ee5d8a90fb30655062bc0bf770c4ff35d3c4)

internal/common/autogen/customtype/object.go
lantoli: no, it was just in case HC provided some helpers

internal/common/autogen/marshal.go
lantoli: nit: i normally prefer merge to rebase to keep history when it's already published, e.g. this commit link is not working anymore. 

at the end there is not really big difference because a squashed commit is done to merge the PR,  merge/rebase only affects the PR branch

internal/testutil/unit/provider_mock.go
lantoli: nit: curious why this lint issue is not flagged in PR checks, maybe you need to run "make tools" to have the same Go linter version

internal/common/autogen/customtype/object.go
```diff
@@ -0,0 +1,237 @@
...
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+/*
+	Custom Object type used in auto-generated code to enable the generic marshal/unmarshal operations to access nested attribute struct tags during conversion.
```
lantoli, line 15: consider moving this to doc.go as it's more Go idiomatic for package documentation, e.g. [here](https://nirdoshgautam.medium.com/a-guide-to-effective-go-documentation-952f346d073f).

although it's true we're not using it in TF yet, so feel free to discard this comment.

cc @AgustinBettati

lantoli (approved):

internal/testutil/unit/provider_mock.go
manupedrozo: Yeah still getting the error after a make tools.
Its the same version as the [workflow](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/77dd9a83d5788440f2a209a4445eaa536f9631e5/.github/workflows/code-health.yml#L59) so probably something else is not matching.
golangci-lint --version: `golangci-lint has version 2.4.0 built with go1.25.0`

internal/common/autogen/customtype/object.go
manupedrozo: Worth considering - keeping in mind that these types will be exclusively used in autogen code, so I don't think it will provide much value.

PR #3806, opened 2025-10-22
test: Avoids flaky test by simplifying the check, ensuring count > 0 instead of reading API response and trying exact match

EspenAlbert:
## Description

Avoids flaky test by simplifying the check, ensuring count > 0 instead of reading API response and trying exact match.
Notice the use of the new `statecheck` and `knownvalue` terraform-plugin-testing packages.

The problem with this test is that it relies on all users in the **organization** to be stable during its test run. But our test suite is running multiple jobs in parallel, so that guarantee is not fulfilled.

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/atlasuser/data_source_atlas_users_test.go
+++ b/internal/service/atlasuser/data_source_atlas_users_test.go
@@ -21,7 +21,10 @@ func TestAccConfigDSAtlasUsers_ByOrgID(t *testing.T) {
 		dataSourceName = "data.mongodbatlas_atlas_users.test"
 		orgID          = os.Getenv("MONGODB_ATLAS_ORG_ID")
 	)
-	resource.Test(t, resource.TestCase{ // does not run in parallel to avoid changes in fetched users during execution
+	// We can only ensure count > 0 because this test relies on all users in the organization to be stable during its test run.
+	// Our test suite is running multiple jobs in parallel, so that guarantee is not fulfilled.
+	// We should isolate these runs by using separate organizations.
+	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { acc.PreCheckBasic(t) },
 		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
 		Steps: []resource.TestStep{
@@ -37,10 +40,10 @@ func TestAccConfigDSAtlasUsers_ByOrgID(t *testing.T) {
 						dataSourceName,
 						tfjsonpath.New("total_count"),
 						knownvalue.Int64Func(func(v int64) error {
-							if v < 1 {
-								return fmt.Errorf("expected total_count to be > 0, got %d", v)
+							if v > 0 {
+								return nil
 							}
-							return nil
+							return fmt.Errorf("expected total_count to be > 0, got %d", v)
 						}),
 					),
 				},
--- a/internal/service/atlasuser/data_source_atlas_users_test.go
+++ b/internal/service/atlasuser/data_source_atlas_users_test.go
@@ -1,15 +1,15 @@
 package atlasuser_test
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-	"github.com/hashicorp/terraform-plugin-testing/terraform"
-	admin20241113 "go.mongodb.org/atlas-sdk/v20241113005/admin"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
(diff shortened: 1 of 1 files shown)


internal/service/atlasuser/data_source_atlas_users_test.go
```diff
@@ -20,18 +20,30 @@ func TestAccConfigDSAtlasUsers_ByOrgID(t *testing.T) {
 	var (
 		dataSourceName = "data.mongodbatlas_atlas_users.test"
 		orgID          = os.Getenv("MONGODB_ATLAS_ORG_ID")
-		users          = fetchOrgUsers(t, orgID)
 	)
-	checks := []resource.TestCheckFunc{testAccCheckMongoDBAtlasOrgWithUsersExists(dataSourceName)} // check that org has at least one user
```
marcosuma, line 25: certainly great that we're making a test less flaky, but at the same time we're reducing its coverage. Do we know why current checks are being flaky?

EspenAlbert: The problem with this test is that it relies on all users in the **organization** to be stable during its test run. But our test suite is running multiple jobs in parallel, so that guarantee is not fulfilled

marcosuma: Consider adding this context as a comment on the new checks you are doing. Something like: "we can only ensure count > 0 because this test relies on all users in the organization to be stable during its test run. But our test suite is running multiple jobs in parallel, so that guarantee is not fulfilled. We should isolate these runs by using separate organizations."

In fact, have we thought about a separate organization then? Maybe in a follow-up refactoring?

internal/service/atlasuser/data_source_atlas_users_test.go
```diff
@@ -20,18 +20,30 @@ func TestAccConfigDSAtlasUsers_ByOrgID(t *testing.T) {
...
+						knownvalue.StringExact(orgID),
+					),
+					statecheck.ExpectKnownValue(
+						dataSourceName,
+						tfjsonpath.New("total_count"),
+						knownvalue.Int64Func(func(v int64) error {
+							if v < 1 {
+								return fmt.Errorf("expected total_count to be > 0, got %d", v)
+							}
+							return nil
```
lantoli:
```suggestion
							if v >  0 {
								return nil
							}
							return fmt.Errorf("expected total_count to be > 0, got %d", v)
```

nit: to have code follow the error text

lantoli (approved):

lantoli:
TestAccProjectAPIKey_recreateWhenDeletedExternally is [failing](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/18727933085/job/53417934848?pr=3806), does this PR have already the code with the time sleep to fix it?

internal/service/atlasuser/data_source_atlas_users_test.go
EspenAlbert: Good point added to the PR.
I think the other tests are capturing the specific `results[i]` field coverage so I am not concerned about less coverage.

For the separate organizations, I would suggest instead supporting this with a test_util function. Then we can programmatically create new orgs during test runs and use them in tests like this.
Our current approach of creating a new org and updating Github variables (dev+qa) doesn't scale well.
WDYT?

internal/service/atlasuser/data_source_atlas_users_test.go
EspenAlbert: Thank you! Addressed in 7fcc6fea84f6f008f65a22476f2b97c63dc19078

internal/service/atlasuser/data_source_atlas_users_test.go
manupedrozo: +1 on programmatic creation when possible to avoid all the manual work/maintenance. Worth a CLOUDP?

EspenAlbert:
> TestAccProjectAPIKey_recreateWhenDeletedExternally

It does 😢 Added to the pinned thread

manupedrozo (approved):

internal/service/atlasuser/data_source_atlas_users_test.go
EspenAlbert: Created [\[CLOUDP-353438\] [TF] Support Creating Org Programmatically For Tests](https://jira.mongodb.org/browse/CLOUDP-353438)

lantoli: not sure about creating an org for this, sounds like an expensive resource to create in each Test Suite or PR commits changing it

EspenAlbert: Merging. We can continue the discussion in the triage or technical sync

PR #3808, opened 2025-10-23
chore: Fix flaky tests in stream test group

lantoli:
## Description

Fix flaky tests in stream test group. Create a dependency to the resource from the datasources

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/streamconnection/resource_stream_connection_test.go
+++ b/internal/service/streamconnection/resource_stream_connection_test.go
@@ -21,6 +21,7 @@ data "mongodbatlas_stream_connection" "test" {
 		project_id = mongodbatlas_stream_connection.test.project_id
 		workspace_name = mongodbatlas_stream_connection.test.workspace_name
 		connection_name = mongodbatlas_stream_connection.test.connection_name
+		depends_on = [mongodbatlas_stream_connection.test]
 }
 `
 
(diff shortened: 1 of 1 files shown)


internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -21,13 +21,15 @@ data "mongodbatlas_stream_connection" "test" {
 		project_id = mongodbatlas_stream_connection.test.project_id
 		workspace_name = mongodbatlas_stream_connection.test.workspace_name
 		connection_name = mongodbatlas_stream_connection.test.connection_name
+		depends_on = [mongodbatlas_stream_connection.test]
```
manupedrozo, line 24: Doesn't `project_id = mongodbatlas_stream_connection.test.project_id` already create the dependency?

EspenAlbert (approved):
Let me check my understanding.
This dependency already existed since the sources depends on `project_id` and `workspace_name` of the resource, but by being explicit it will apply in more cases?

I think it is because both those attributes are known at plan time, so the data source doesn't actually wait. Am I correct?

lantoli:
@EspenAlbert cc @manupedrozo correct, the dependency must be with Computed values,  TF doesn't need to wait to execute resource to know values for project_id and workspace_name so can parallelize resource and datasource calls.

internal/service/streamconnection/resource_stream_connection_test.go
lantoli: see [this comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3808#issuecomment-3435471836)

manupedrozo:
So for resources the existing dependency would have been enough but datasources behave differently and require the explicit depends_on: https://discuss.hashicorp.com/t/dependency-model-depends-on-vs-data-reference/71578/2

Nice catch!

PR #3817, opened 2025-10-25
chore: Use timeouts attributes in autogen resources

lantoli:
## Description

Use timeouts attributes in autogen resources.

Link to any related issue(s): CLOUDP-352321

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogen/handle_operations.go
+++ b/internal/common/autogen/handle_operations.go
@@ -49,9 +49,6 @@ type HandleCreateReq struct {
 
 func HandleCreate(ctx context.Context, req HandleCreateReq) {
 	d := &req.Resp.Diagnostics
-	if d.HasError() {
-		return
-	}
 	bodyReq, err := Marshal(req.Plan, false)
 	if err != nil {
 		addError(d, opCreate, errBuildingAPIRequest, err)
@@ -88,9 +85,6 @@ type HandleReadReq struct {
 
 func HandleRead(ctx context.Context, req HandleReadReq) {
 	d := &req.Resp.Diagnostics
-	if d.HasError() {
-		return
-	}
 	bodyResp, apiResp, err := callAPIWithoutBody(ctx, req.Client, req.CallParams)
 	if notFound(bodyResp, apiResp) {
 		req.Resp.State.RemoveResource(ctx)
@@ -123,9 +117,6 @@ type HandleUpdateReq struct {
 
 func HandleUpdate(ctx context.Context, req HandleUpdateReq) {
 	d := &req.Resp.Diagnostics
-	if d.HasError() {
-		return
-	}
 	bodyReq, err := Marshal(req.Plan, true)
 	if err != nil {
 		addError(d, opUpdate, errBuildingAPIRequest, err)
@@ -164,9 +155,6 @@ type HandleDeleteReq struct {
 
 func HandleDelete(ctx context.Context, req HandleDeleteReq) {
 	d := &req.Resp.Diagnostics
-	if d.HasError() {
-		return
-	}
 	var err error
 	if req.StaticRequestBody == "" {
 		_, _, err = callAPIWithoutBody(ctx, req.Client, req.CallParams)
--- a/internal/common/autogen/handle_operations.go
+++ b/internal/common/autogen/handle_operations.go
@@ -49,6 +49,9 @@ type HandleCreateReq struct {
 
 func HandleCreate(ctx context.Context, req HandleCreateReq) {
 	d := &req.Resp.Diagnostics
+	if d.HasError() {
+		return
+	}
 	bodyReq, err := Marshal(req.Plan, false)
 	if err != nil {
 		addError(d, opCreate, errBuildingAPIRequest, err)
@@ -117,6 +120,9 @@ type HandleUpdateReq struct {
 
 func HandleUpdate(ctx context.Context, req HandleUpdateReq) {
 	d := &req.Resp.Diagnostics
+	if d.HasError() {
+		return
+	}
 	bodyReq, err := Marshal(req.Plan, true)
 	if err != nil {
 		addError(d, opUpdate, errBuildingAPIRequest, err)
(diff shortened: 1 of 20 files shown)


internal/common/autogen/handle_operations.go
```diff
@@ -49,6 +49,9 @@ type HandleCreateReq struct {
 
 func HandleCreate(ctx context.Context, req HandleCreateReq) {
 	d := &req.Resp.Diagnostics
```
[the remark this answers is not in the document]
lantoli: that's fine

tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl
```diff
@@ -49,6 +49,10 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 		PathParams: pathParams,
 		Method: "{{ .APIOperations.Create.HTTPMethod }}",
 	}
+	{{with .APIOperations.Create.Wait -}}
+	timeout, localDiags := plan.Timeouts.Create(ctx, {{ .TimeoutSeconds }}*time.Second)
+	resp.Diagnostics.Append(localDiags...)
```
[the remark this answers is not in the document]
lantoli: that's fine

internal/serviceapi/streamprocessorapi/resource.go
```diff
@@ -36,9 +37,6 @@ func (r *rs) Schema(ctx context.Context, req resource.SchemaRequest, resp *resou
 func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	var plan TFModel
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
-	if resp.Diagnostics.HasError() {
```
EspenAlbert, line 39: Why are we removing this?

lantoli: as in the PR description, we are moving it to the beginning of the handle methods like HandleCreate, etc.
No need to check it in each single resource when it can be done in the common code

EspenAlbert: I see, so the line `plan.Timeouts.Create(ctx, 300*time.Second)` will not have some weird error if this fails earlier?

tools/codegen/gofilegen/resource/testdata/wait-configuration.golden.go
```diff
@@ -48,6 +46,8 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 		PathParams:    pathParams,
 		Method:        "POST",
 	}
+	timeout, localDiags := plan.Timeouts.Create(ctx, 300*time.Second)
```
EspenAlbert, line 52: Why 300s default?

EspenAlbert (approved):
Just a few questions. I think it makes the code a bit less readable, but no strong opinion.

internal/serviceapi/streamprocessorapi/resource.go
lantoli: uhm, let me try that

tools/codegen/gofilegen/resource/testdata/wait-configuration.golden.go
lantoli: define in config file, same we had in line 60:
```
TimeoutSeconds:    300,
```

internal/serviceapi/auditingapi/resource.go
```diff
@@ -36,9 +36,6 @@ func (r *rs) Schema(ctx context.Context, req resource.SchemaRequest, resp *resou
 func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	var plan TFModel
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
```
manupedrozo, line 41: I appreciate the effort to reduce template size. In this case I think its clearer to leave this check here right after the Plan.Get() call instead of waiting until the create/update/read operation to check the diags.

I understand that following this approach would also mean adding a check for the timeout creation. 
Given that checking errors after a function call is a typical go pattern, I think its ok to generate some extra code for it.

No strong feelings either way, we can always go back to generating checks if necessary.

manupedrozo (approved):

internal/serviceapi/auditingapi/resource.go
lantoli: I think it's clearer just to set all the input params for handle operations, and the doing the "real" logic including error checks in handler functions.
Good thing about diagnostics is that it keeps a list of errors, warning, etc. so we don't lose information even if we accumulate some of them and we don't do HasErrors every time.

as in [this comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3817#discussion_r2464799561),  there is no need to do additional checks as timeouts won't fail (it will just get the default value), but if both @manupedrozo @EspenAlbert prefer keeping the check, I'll do that, reverting those changes.

lantoli: timeouts never fails (it only logs info sometimes in current implementation) but will also add hasErrors to follow the same pattern

internal/serviceapi/streamprocessorapi/resource.go
lantoli: simulated that setting the plan fails, it works fine because timeouts just get the default one, and then HandleCreate checks HasErrors in the beginning and displays the error

tools/codegen/gofilegen/resource/testdata/wait-configuration.golden.go
EspenAlbert: Ty! I just wasn't sure, couldn't see it in the changes 👍

internal/serviceapi/auditingapi/resource.go
EspenAlbert: Feel free to keep it; no strong opinion, and in general, the handler functions will have better test coverage.
Anything added to the resources before calling the handler function must ensure that they can handle default values.

lantoli: I think current approach has better tradeoffs (keeping resources just with input params and move away logic that is common) but also think it's important to get consensus, reverted here: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3817/commits/e9036353f9a611cfb1faeccfd5589415e0ae1b40

manupedrozo: Thank you!

PR #3818, opened 2025-10-27
feat: Adds GCP Confluent privatelink support to streamprivatelink endpoint resource + datasource

kpatel71716:
## Description

The Atlas Stream Processing team has added Admin API functionality to support GCP Privatelink with Confluent. We would like to add terraform functionality by adding a new `service_attachment_uris` attribute that is ONLY valid with GCP and must be provided for the GCP provider type and ONLY with confluent. This attribute is NOT valid for AWS/Azure and both `service_attachment_uris` and `service_endpoint_id` cannot be provided

Link to any related issue(s):

[CLOUDP-350439](https://jira.mongodb.org/browse/CLOUDP-350439) 

## Type of change:

## Required Checklist:


.changelog/3818.txt
```diff
@@ -0,0 +1,10 @@
+```release-note:enhancement
+resource/mongodbatlas_stream_privatelink_endpoint: Adds support for GCP Confluent
```
marcosuma, line 2:
```suggestion
resource/mongodbatlas_stream_privatelink_endpoint: Adds support for GCP Confluent
```
```

Is it missing a ``` ?

**Review by xargom — APPROVED** [review]
lgtm

**Review by marcosuma — APPROVED** [review]

**Review by malexandert — APPROVED** [review]

File: internal/service/streamprivatelinkendpoint/model.go [hunk 2]
```diff
@@ -39,15 +40,37 @@ func NewTFModel(ctx context.Context, projectID string, apiResp *admin.StreamsPri
 	}
 	result.DnsSubDomain = subdomain
 
+	serviceAttachmentUris, diagsServiceAttachment := types.ListValueFrom(ctx, types.StringType, apiResp.GetGcpServiceAttachmentUris())
```
EspenAlbert: Q: Is this always returned from the API?
Possibly, we want to use the `apiResp.GcpServiceAttachmentUris` instead of the `apiResp.GetGcpServiceAttachmentUris()`.

I think with the current implementation this might fail since the attribute is `optional` only and this will set the terraform state to `[]` which is not the same as `null` in the TPF

internal/service/streamprivatelinkendpoint/model.go
```diff
@@ -39,15 +40,37 @@ func NewTFModel(ctx context.Context, projectID string, apiResp *admin.StreamsPri
...
 
 func NewAtlasReq(ctx context.Context, plan *TFModel) (*admin.StreamsPrivateLinkConnection, diag.Diagnostics) {
 	diags := diag.Diagnostics{}
 
 	if plan.Vendor.ValueString() == VendorConfluent {
-		if plan.ServiceEndpointId.IsNull() {
-			diags.AddError(fmt.Sprintf("service_endpoint_id is required for vendor %s", VendorConfluent), "")
+		provider := plan.Provider.ValueString()
+
+		// Validate that exactly one of service_endpoint_id or service_attachment_uris is provided
```
EspenAlbert, line 56: Q: How does the API handle switching between these two? Let's say the user needs to go from `service_attachment_uris` to `service_endpoint_id`. Do we need to send an empty `service_attachment_uris = []` or it is enough to only specify `service_endpoint_id`?

kpatel71716: In that scenario, just passing in service_endpoint_id will suffice. However, we don't support modifications of private link endpoints at this time so from a terraform perspective, it would be deleted and recreated

internal/service/streamprivatelinkendpoint/model.go
kpatel71716: Good q, This field can be null, it's not always returned. I checked this code in the debugger and compared it to other examples and the default TFModel structure value is the same in both scenarios i.e. `types.ListNull(types.StringType)`.

If there is concern, I can make the field in the TFModel a pointer but I am not sure if that is preferred

.changelog/3818.txt
kpatel71716: Good catch, fixed

internal/service/streamprivatelinkendpoint/model.go
```diff
@@ -39,15 +40,37 @@ func NewTFModel(ctx context.Context, projectID string, apiResp *admin.StreamsPri
...
+		hasServiceEndpointID := !plan.ServiceEndpointId.IsNull() && plan.ServiceEndpointId.ValueString() != ""
+		hasServiceAttachmentUris := !plan.ServiceAttachmentUris.IsNull() && len(plan.ServiceAttachmentUris.Elements()) > 0
+
+		if !hasServiceEndpointID && !hasServiceAttachmentUris {
+			diags.AddError("Either service_endpoint_id or service_attachment_uris must be provided for CONFLUENT vendor", "")
+		}
+		if hasServiceEndpointID && hasServiceAttachmentUris {
```
lantoli, line 63: do we need to have all these checks in the client side instead of just letting Atlas fail?  (main benefit that message are more clear, main con that we're duplicating logic between client and server, if Atlas messages are good enough we shouldn't need it)

lantoli (approved):

internal/service/streamprivatelinkendpoint/model.go
kpatel71716: that's fair. I will verify what the Atlas errors are are defer to Atlas if the errors are descriptive enough. I will reply here

internal/service/streamprivatelinkendpoint/model.go
EspenAlbert: Ahh, I see. 
In that case I would opt for a RequiresReplace plan modifier, which makes it easier for the user instead of manually having to run terraform destroy and apply.
See example here:

["ip_address": schema.StringAttribute{](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/f04234c7fc9a665de26bba981ed56f26156d746e/internal/service/projectipaccesslist/resource_schema.go#L43-L48)
```go
    "ip_address": schema.StringAttribute{
        Optional: true,
        Computed: true,
            stringplanmodifier.RequiresReplace(),
        },
```

I understand this would be a big change to do now, but worth keeping in mind for the future.

internal/service/streamprivatelinkendpoint/model.go
EspenAlbert: Yes, I would be concerned that the value from the config `null` will not match the `[]` here.

Consider the following example: (Feel free to put it in a test file and experiment.)
```go
func TestShowTypesListNull(t *testing.T) {
	// Demonstrate the problem with apiResp.GetGcpServiceAttachmentUris() always returns []string{}, if it is not set, we should use the pointer so we get ListValueNull() 
	ctx := context.Background()
	valEmpty, diags := types.ListValueFrom(ctx, types.StringType, []string{})
	assert.False(t, diags.HasError())
	log.Printf("empty: %v", valEmpty) // empty: []
	var myPointerList *[]string
	valNull, diags := types.ListValueFrom(ctx, types.StringType, myPointerList)
	require.False(t, diags.HasError())
	log.Printf("null: %v", valNull) // null: <null>
	isEqual := valEmpty.Equal(valNull)
	log.Printf("isEqual: %t", isEqual) // isEqual: false
}
```

internal/service/streamprivatelinkendpoint/model.go
kpatel71716: Made a ticket [here](https://jira.mongodb.org/browse/CLOUDP-356651) to improve our errors in mms. We have validation for this endpoint but should enhance the logic with more descriptive errors. That way we can simplify the TF logic. Removed some cloud-provider specific logic to keep the terraform code simpler

internal/service/streamprivatelinkendpoint/model.go
kpatel71716: Good point! I went ahead and added RequiresReplace since it makes sense to do it as part of this PR for now

internal/service/streamprivatelinkendpoint/model.go
kpatel71716: Got it thank you, I made the field a pointer, and added a check on the length of the api response to determine if we need to populate the field in the TFModel. Please let me know if that looks good

kpatel71716: Actually my previous change was too ambitious and caused errors. Went with a simple length check and used types.ListNull since that works with the tests provided as well

EspenAlbert: Sorry for the confusion, but my comment meant to say to use the pointer on the apiResp directly.
- ✅ `apiResp.GcpServiceAttachmentUris` instead of
- ❌ `apiResp.GetGcpServiceAttachmentUris()`

This handles the case of `[]` / `null` correctly (seen in the test example so you avoid the conditional)

kpatel71716: Thanks for explaining, I overthought this, fixed!

EspenAlbert: TY!

EspenAlbert (approved):
Thank you for addressing all the feedback!

kpatel71716:
Failing test is unrelated, I am investigating. Merging this PR for now

PR #3819, opened 2025-10-27
chore: Add support for typed nested sets in autogen

manupedrozo:
## Description

Adding a custom `NestedSet` terraform type for autogen usage.

Note that NestedSet supports nested objects but **not** Terraform attr types, e.g. `types.string`. These will be covered in the next PR (same as how it was done for lists).

Custom TF types docs: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom

Link to any related issue(s): CLOUDP-353170

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogen/customtypes/nested_set.go
+++ b/internal/common/autogen/customtypes/nested_set.go
@@ -1,0 +1,206 @@
+package customtypes
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+/*
+	Custom Nested Set type used in auto-generated code to enable the generic marshal/unmarshal operations to access nested attribute struct tags during conversion.
+	Custom types docs: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom
+
+	Usage:
+		- Schema definition:
+			"sample_nested_object_set": schema.SetNestedAttribute{
+				...
+				CustomType: customtypes.NewNestedSetType[TFSampleNestedObjectModel](ctx),
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"string_attribute": schema.StringAttribute{...},
+					},
+				},
+			}
+
+		- TF Models:
+			type TFModel struct {
+				SampleNestedObjectSet customtypes.NestedSetValue[TFSampleNestedObjectModel] `tfsdk:"sample_nested_object_set"`
+				...
+			}
+
+			type TFSampleNestedObjectModel struct {
+				StringAttribute types.String `tfsdk:"string_attribute"`
+				...
+			}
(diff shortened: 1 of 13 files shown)


internal/common/autogen/customtypes/nested_set.go
```diff
@@ -0,0 +1,206 @@
+package customtypes
+
+import (
```
lantoli, line 3: shouldn't we have unit tests for all custom types?

internal/common/autogen/customtypes/nested_set.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return NestedSetValue[T]{SetValue: basetypes.NewSetUnknown(elemType)}
+}
+
+func (v NestedSetValue[T]) Equal(o attr.Value) bool {
+	other, ok := o.(NestedSetValue[T])
+	if !ok {
+		return false
+	}
+	return v.SetValue.Equal(other.SetValue)
```
lantoli, line 178: is SetValue the one with the logic to compare two sets independently of order and repeated elements?

internal/common/autogen/customtypes/nested_set.go
```diff
@@ -0,0 +1,206 @@
...
+	_ NestedSetValueInterface = NestedSetValue[struct{}]{}
+)
+
+type NestedSetType[T any] struct {
+	basetypes.SetType
+}
+
+func NewNestedSetType[T any](ctx context.Context) NestedSetType[T] {
+	elemType, diags := getElementType[T](ctx)
+	if diags.HasError() {
```
lantoli, line 53: when can this raise errors?

lantoli (approved):

internal/common/autogen/customtypes/nested_set.go
manupedrozo: If the traversal of the type via reflection fails due to an unexpected attribute type, missing tfsdk tag, etc.

internal/common/autogen/customtypes/nested_set.go
manupedrozo: Maybe something to consider once all custom types are implemented if we want to dedicate extra effort for tests. Existing tests should be covering these already so I don't think the ROI is worth it as of now.

internal/common/autogen/customtypes/nested_set.go
manupedrozo: Yes, it does the following:
```go
func (s SetValue) Equal(o attr.Value) bool {
    ...
    if len(s.elements) != len(other.elements) {
		return false
	}

	for _, elem := range s.elements {
		if !other.contains(elem) {
			return false
		}
	}

    return true
}
```

For extra context, see the note at: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom#equal-attr-value-bool

lantoli: @manupedrozo thanks, where is the duplication not considered, when creating it?

manupedrozo: Deduplication is not done at the type level, you can create a basetype.SetValue like:
```go
sv := types.StringValue("set1")
setValue, diags := basetypes.NewSetValue(types.StringType, []attr.Value{sv, sv})
```

PR #3822, opened 2025-10-28
chore: Support merging of response custom nested list elements into plan/state

manupedrozo:
## Description

Merging API response custom nested lists into the plan/state nested list based on element order. 
This is similar to how the handwritten code works and how the non-custom nested list behaves in autogen.

Note that for custom sets, we still do not perform the merge as we cannot make the assumption that the order of elements in the plan/state and the API response is the same.

Enabling usage of custom types in `custom_db_role_api` which was previously disabled due to this missing behavior. Specifically, when `actions[#].resources[#].cluster` is set to `false` in the plan, the [API](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-creategroupcustomdbrolerole) does not return the `false` value back. Without the merging behavior, it creates an inconsistency. 

Link to any related issue(s): CLOUDP-353170

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogen/unmarshal.go
+++ b/internal/common/autogen/unmarshal.go
@@ -327,16 +327,37 @@ func getListValueTFAttr(ctx context.Context, arrayJSON []any, list customtypes.L
 }
 
 func getNestedListValueTFAttr(ctx context.Context, arrayJSON []any, list customtypes.NestedListValueInterface) (attr.Value, error) {
-	if len(arrayJSON) == 0 && list.Len() == 0 {
+	oldSlicePtr, diags := list.SlicePtrAsAny(ctx)
+	if diags.HasError() {
+		return nil, fmt.Errorf("unmarshal failed to convert list: %v", diags)
+	}
+	oldSliceVal := reflect.ValueOf(oldSlicePtr).Elem()
+	oldSliceLen := oldSliceVal.Len()
+
+	if len(arrayJSON) == 0 && oldSliceLen == 0 {
 		// Keep current list if both model and JSON lists are zero-len (empty or null) so config is preserved.
 		// It avoids inconsistent result after apply when user explicitly sets an empty list in config.
 		return list, nil
 	}
 
 	slicePtr := list.NewEmptySlicePtr()
-	err := unmarshalNestedSlice(arrayJSON, slicePtr)
-	if err != nil {
-		return nil, err
+	sliceVal := reflect.ValueOf(slicePtr).Elem()
+	sliceVal.Set(reflect.MakeSlice(sliceVal.Type(), len(arrayJSON), len(arrayJSON)))
+
+	for i, item := range arrayJSON {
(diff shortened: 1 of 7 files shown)


internal/common/autogen/unmarshal.go
```diff
@@ -327,16 +327,37 @@ func getListValueTFAttr(ctx context.Context, arrayJSON []any, list customtypes.L
...
+
+	for i, item := range arrayJSON {
+		elementVal := sliceVal.Index(i)
+		if i < oldSliceLen {
+			elementVal.Set(oldSliceVal.Index(i))
+		}
+		elementPtr := elementVal.Addr().Interface()
+		objJSON, ok := item.(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("unmarshal of array item failed to convert to object: %v", item)
```
lantoli, line 355: in general I would avoid to show the item value in messages, e.g. it could be a sensitive value,  show item names is fine, not sure what is shown here

lantoli (approved):

internal/common/autogen/unmarshal.go
manupedrozo: Following the non-custom types error messages which are also printing the value.
I do agree with you on this not being a good approach (even if sensitive data is not returned by the API). 
It's one of the things we should look at before putting an autogen resource in production. 
cc @AgustinBettati

internal/common/autogen/unmarshal_test.go
```diff
@@ -140,10 +141,10 @@ func TestUnmarshalNestedAllTypes(t *testing.T) {
 		AttrCustomListString:        customtypes.NewListValueUnknown[types.String](ctx),
 		AttrListObj:                 types.ListUnknown(objTypeTest),
 		AttrCustomNestedList: customtypes.NewNestedListValue[modelCustomTypeTest](ctx, []modelCustomTypeTest{
+			modelCustomTypeBasic,
 			{
-				// these attribute values are irrelevant, they will be overwritten with JSON values
-				AttrString:    types.StringValue("different_string"),
-				AttrInt:       types.Int64Value(999),
+				AttrString:    types.StringValue("existing not overwritten"),
```
AgustinBettati, line 146: Nice coverage. Something to consider is TestUnmarshalNestedAllTypes is becoming very big. Not high priority but we could consider refactoring to having separate tests for these particular scenarios, with common code so each test definition doesn't imply too much boilerplate

AgustinBettati (approved):

internal/common/autogen/unmarshal.go
AgustinBettati: Yes very good point, good to double check. Specifically thinking during marshalling if we return values present in the TF Model which can include the raw sensitive values from config. I will create a ticket to not forget this aspect.

manupedrozo: Ty

PR #3840, opened 2025-11-04
chore: Improve TF schema attribute name generation

manupedrozo:
## Description

Improving the generation of TF schema attribute names to provide better default UX.

- Implemented a custom snake case conversion that treats multiple consecutive uppercase letters as part of the same word, except for the last one.

    **Examples:**
    Before: `MongoDBMajorVersion` -> `mongo_dbmajor_version`
    After: `MongoDBMajorVersion` -> `mongo_db_major_version`

    Before: `AwsIAMType` -> `aws_iamtype`
    After: `AwsIAMType` -> `aws_iam_type`

- Removed `xstrings` usage from `stringcase` package.
- Removed `snake_case` to `camelCase` conversions across the autogen code. 
    Converting from camelCase to snake_case is fine but its not a reversible operation (see [#3740](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3740)).
- Config aliases are no longer case converted (preventing potential inconsistencies).

    **Example:**
    To add an alias for the cluster name in path: /api/atlas/v2/groups/{groupId}/clusters/{clusterName}
    Before we would provide `cluster_name: name`. `cluster_name` would be converted to `clusterName` and replaced in the path.
    Now we directly provide `clusterName: name`.

Link to any related issue(s): CLOUDP-349435

## Type of change:

## Required Checklist:

### DIFF
--- a/tools/codegen/codespec/testdata/config-nested-schema-overrides.yml
+++ b/tools/codegen/codespec/testdata/config-nested-schema-overrides.yml
@@ -15,6 +15,7 @@ resources:
     version_header: application/vnd.atlas.2035-01-01+json
     schema:
       aliases:
+        groupId: projectId
         group_id: project_id
         nested_list_array_attr.inner_num_attr: inner_num_attr_alias
         outer_object.nested_level1.level_field1: level_field1_alias
--- a/tools/codegen/config.yml
+++ b/tools/codegen/config.yml
@@ -61,9 +61,9 @@ resources:
         target_states: ["DELETED"] # DELETED is a special state value when API returns 404 or empty object
     version_header: application/vnd.atlas.2024-08-05+json
     schema:
-      ignores: ["links", "mongo_dbemployee_access_grant.links"]
+      ignores: ["links", "mongo_db_employee_access_grant.links"]
       aliases:
-        cluster_name: name # path param name does not match the API request property
+        clusterName: name # path param name does not match the API request property
       # These attributes are optional but when not configured, Atlas returns a default value.
       # If a nested attribute is optional/computed, all its optional child attributes must also be defined as optional/computed to avoid non-empty plans.
       overrides:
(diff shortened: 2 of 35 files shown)


tools/codegen/config.yml
```diff
@@ -296,10 +296,6 @@ resources:
       method: PATCH
     version_header: application/vnd.atlas.2023-01-01+json
     schema:
-      aliases:
```
EspenAlbert, line 299: nice with simplified config!

tools/codegen/models/auditing_api.yaml
```diff
@@ -62,3 +62,4 @@ operations:
         path: /api/atlas/v2/groups/{groupId}/auditLog
     version_header: application/vnd.atlas.2023-01-01+json
 name: auditing_api
+packageName: auditingapi
```
EspenAlbert, line 65: [nit] Aren't these always following the same rule. Do we need to be explicit?

manupedrozo: No need to be explicit but would rather calculate once and avoid maintaining consistency in multiple places.
We now set it once during spec generation and reuse it during code generation, which also makes it configurable in the intermediate model.

EspenAlbert (approved):
Nice just a few nit comments

AgustinBettati (approved):
nice!

tools/codegen/codespec/testdata/config-nested-schema-overrides.yml
```diff
@@ -15,6 +15,7 @@ resources:
     version_header: application/vnd.atlas.2035-01-01+json
     schema:
       aliases:
+        groupId: projectId
         group_id: project_id
         nested_list_array_attr.inner_num_attr: inner_num_attr_alias
         outer_object.nested_level1.level_field1: level_field1_alias
```
AgustinBettati, line 21: nit: should we adjust these overrides? Not certain if the snake case is still supported

manupedrozo: These are still supported!
We can provide aliases for TF schema attributes. I expect that we'll do this for group_id -> project_id before releasing an autogen resource.

PR #3871, opened 2025-11-11
feat: Adds max tier size attribute to stream workspace resource and datasource

kpatel71716:
## Description

 adds the `max_tier_size` attribute to the stream_workspace rs and ds as well as the stream_workspaces ds. Not implemented for stream_instance to encourage customers to move to a non-deprecated feature. Due to our conversion logic, there are currently maxTierSize references in the stream instance models but that is not populated for the deprecated stream_instance rs and ds. We have a follow-up to explore refactoring the mapping logic

Link to any related issue(s): CLOUDP-358363

## Type of change:

## Required Checklist:

### DIFF
--- a/docs/resources/stream_workspace.md
+++ b/docs/resources/stream_workspace.md
@@ -59,7 +59,8 @@ resource "mongodbatlas_stream_workspace" "example" {
 
 ### Stream Config
 
-* `tier` - (Required) Selected tier for the Stream Instance. Configures Memory / VCPU allowances. The [MongoDB Atlas API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Streams/operation/creategroupstreamworkspace) describes the valid values.
+* `max_tier_size` - (Optional) Max tier size for the Stream Workspace. Configures Memory / VCPU allowances.
+* `tier` - (Optional) Selected tier for the Stream Workspace. Configures Memory / VCPU allowances. The [MongoDB Atlas API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Streams/operation/creategroupstreamworkspace) describes the valid values.
 
 
 ## Attributes Reference
--- a/internal/service/streaminstance/model_stream_instance.go
+++ b/internal/service/streaminstance/model_stream_instance.go
@@ -61,7 +61,7 @@ func NewTFStreamInstance(ctx context.Context, apiResp *admin.StreamsTenant) (*TF
 	}
 	var streamConfig = types.ObjectNull(StreamConfigObjectType.AttrTypes)
 	apiStreamConfig := apiResp.StreamConfig
-	if apiStreamConfig != nil && (apiStreamConfig.Tier != nil || apiStreamConfig.MaxTierSize != nil) {
+	if apiStreamConfig != nil {
 		returnedStreamConfig, diagsStreamConfig := types.ObjectValueFrom(ctx, StreamConfigObjectType.AttrTypes, TFInstanceStreamConfigSpecModel{
 			MaxTierSize: types.StringPointerValue(apiStreamConfig.MaxTierSize),
 			Tier:        types.StringPointerValue(apiStreamConfig.Tier),
--- a/internal/service/streaminstance/model_stream_instance.go
+++ b/internal/service/streaminstance/model_stream_instance.go
@@ -61,9 +61,10 @@ func NewTFStreamInstance(ctx context.Context, apiResp *admin.StreamsTenant) (*TF
 	}
 	var streamConfig = types.ObjectNull(StreamConfigObjectType.AttrTypes)
 	apiStreamConfig := apiResp.StreamConfig
-	if apiStreamConfig != nil && apiStreamConfig.Tier != nil {
+	if apiStreamConfig != nil && (apiStreamConfig.Tier != nil || apiStreamConfig.MaxTierSize != nil) {
 		returnedStreamConfig, diagsStreamConfig := types.ObjectValueFrom(ctx, StreamConfigObjectType.AttrTypes, TFInstanceStreamConfigSpecModel{
-			Tier: types.StringPointerValue(apiStreamConfig.Tier),
+			MaxTierSize: types.StringPointerValue(apiStreamConfig.MaxTierSize),
+			Tier:        types.StringPointerValue(apiStreamConfig.Tier),
 		})
 		streamConfig = returnedStreamConfig
 		diags.Append(diagsStreamConfig...)
--- a/internal/service/streaminstance/model_stream_instance_test.go
+++ b/internal/service/streaminstance/model_stream_instance_test.go
@@ -278,8 +278,7 @@ func tfRegionObject(t *testing.T, cloudProvider, region string) types.Object {
 func tfStreamConfigObject(t *testing.T, tier string) types.Object {
 	t.Helper()
 	streamConfig, diags := types.ObjectValueFrom(t.Context(), streaminstance.StreamConfigObjectType.AttrTypes, streaminstance.TFInstanceStreamConfigSpecModel{
-		MaxTierSize: types.StringNull(), // Set to null for test cases that don't specify it
-		Tier:        types.StringValue(tier),
+		Tier: types.StringValue(tier),
 	})
 	if diags.HasError() {
 		t.Errorf("failed to create terraform data process region model: %s", diags.Errors()[0].Summary())
--- a/internal/service/streaminstance/model_stream_instance_test.go
+++ b/internal/service/streaminstance/model_stream_instance_test.go
@@ -278,7 +278,8 @@ func tfRegionObject(t *testing.T, cloudProvider, region string) types.Object {
 func tfStreamConfigObject(t *testing.T, tier string) types.Object {
 	t.Helper()
 	streamConfig, diags := types.ObjectValueFrom(t.Context(), streaminstance.StreamConfigObjectType.AttrTypes, streaminstance.TFInstanceStreamConfigSpecModel{
-		Tier: types.StringValue(tier),
+		MaxTierSize: types.StringNull(), // Set to null for test cases that don't specify it
+		Tier:        types.StringValue(tier),
 	})
 	if diags.HasError() {
 		t.Errorf("failed to create terraform data process region model: %s", diags.Errors()[0].Summary())
--- a/internal/service/streamworkspace/model.go
+++ b/internal/service/streamworkspace/model.go
@@ -3,7 +3,9 @@ package streamworkspace
 import (
 	"context"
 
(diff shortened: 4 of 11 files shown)


marcosuma (approved):

jvincent-mongodb (approved):

internal/service/streaminstance/model_stream_instance.go
```diff
@@ -61,9 +61,10 @@ func NewTFStreamInstance(ctx context.Context, apiResp *admin.StreamsTenant) (*TF
 	}
 	var streamConfig = types.ObjectNull(StreamConfigObjectType.AttrTypes)
 	apiStreamConfig := apiResp.StreamConfig
-	if apiStreamConfig != nil && apiStreamConfig.Tier != nil {
+	if apiStreamConfig != nil && (apiStreamConfig.Tier != nil || apiStreamConfig.MaxTierSize != nil) {
 		returnedStreamConfig, diagsStreamConfig := types.ObjectValueFrom(ctx, StreamConfigObjectType.AttrTypes, TFInstanceStreamConfigSpecModel{
-			Tier: types.StringPointerValue(apiStreamConfig.Tier),
+			MaxTierSize: types.StringPointerValue(apiStreamConfig.MaxTierSize),
```
lantoli, line 66: why they're treated together, can't tier or max tier be provided independtly?

internal/service/streaminstance/model_stream_instance_test.go
```diff
@@ -278,7 +278,8 @@ func tfRegionObject(t *testing.T, cloudProvider, region string) types.Object {
 func tfStreamConfigObject(t *testing.T, tier string) types.Object {
 	t.Helper()
 	streamConfig, diags := types.ObjectValueFrom(t.Context(), streaminstance.StreamConfigObjectType.AttrTypes, streaminstance.TFInstanceStreamConfigSpecModel{
-		Tier: types.StringValue(tier),
+		MaxTierSize: types.StringNull(), // Set to null for test cases that don't specify it
```
lantoli: is this needed to be set explictly? i would think it's the default value if not set

lantoli (approved):

docs/resources/stream_workspace.md
```diff
@@ -59,7 +59,8 @@ resource "mongodbatlas_stream_workspace" "example" {
 
 ### Stream Config
 
-* `tier` - (Required) Selected tier for the Stream Instance. Configures Memory / VCPU allowances. The [MongoDB Atlas API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Streams/operation/creategroupstreamworkspace) describes the valid values.
+* `max_tier_size` - (Optional) Max tier size for the Stream Workspace. Configures Memory / VCPU allowances.
```
lantoli, line 62: what does max_tier_size exactly do, does it change tier if appropiate?  In that case therere will a drift in the config, say tier was SP10 and max_tier is SP30 and tier want to be changed to SP30, as it's SP10 in the Terraform config, it will try to go back to SP10, so we recommend to use ignore_changes, similarly to auto-scaling in clusters.  is this behavior what happens?

marcosuma: It's a very good questions and good catch. I was assuming that max would throw validation error when the tier is changed over that value, but Leo makes me think that was not a good assumption

kpatel71716: Apologies, I should have clarified in the description. `max_tier_size` as well as `tier` serve as static values rather than values that will change as a user tries to allocate more/less resources by changing a tier

The `stream_workspace` resource and data_source stores essentially a `default` tier and a `max` tier so that when a stream processor in a given workspace is started, it must fall within those limitations

We have downstream validation in both mms and on the stream processing platform side to enforce that a user is within the guardrails they set in the stream_workspace when they try to start a stream processor along with other edge cases

internal/service/streaminstance/model_stream_instance_test.go
kpatel71716: That's correct, removed

internal/service/streaminstance/model_stream_instance.go
kpatel71716: Are you thinking that we can change line 64 to simply `if apiStreamConfig != nil  {` ? That should be fine

if streamConfig is specified, downstream we check that tier is provided. But maxTierSize is optional regardless. Ultimately that means that if maxTier is specified, then tier is provided as well

docs/resources/stream_workspace.md
kpatel71716: To add, there will be a downstream validation error on max_tier_size being smaller than tier when creating/updating the stream_workspace resource

internal/service/streamworkspace/model.go
```diff
@@ -28,8 +30,23 @@ func newStreamWorkspaceCreateReq(ctx context.Context, plan *TFModel) (*admin.Str
 		if diags := plan.StreamConfig.As(ctx, streamConfig, basetypes.ObjectAsOptions{}); diags.HasError() {
 			return nil, diags
 		}
+		var maxTierSize *string
+		if !streamConfig.MaxTierSize.IsNull() && !streamConfig.MaxTierSize.IsUnknown() {
```
EspenAlbert: This can be greatly simplified by using `maxTierSize := streamConfig.MaxTierSize.ValueString()`

Ran this to sanity check:
```go
func TestStrinUnknownIsEmpty(t *testing.T) {
	unknownString := types.StringUnknown()
	require.True(t, unknownString.ValueString() == "")
	nullString := types.StringNull()
	require.True(t, nullString.ValueString() == "")
}
```

internal/service/streamworkspace/move_state.go
```diff
@@ -108,9 +108,11 @@ func stateMover(ctx context.Context, req resource.MoveStateRequest, resp *resour
 			tier := schemafunc.GetAttrFromStateObj[string](configObj, "tier")
 
 			objValue, diags := types.ObjectValue(map[string]attr.Type{
-				"tier": types.StringType,
+				"max_tier_size": types.StringType,
```
EspenAlbert, line 111: @lantoli I think we can stick to only setting primary identifier and the rest will be read in the read operation?

internal/service/streamworkspace/resource_test.go
```diff
@@ -25,7 +25,8 @@ func TestAccStreamWorkspaceRS_basic(t *testing.T) {
 				Config: streamsWorkspaceConfig(projectID, workspaceName, region, cloudProvider),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					streamsWorkspaceAttributeChecks(resourceName, workspaceName, region, cloudProvider),
-					resource.TestCheckResourceAttr(resourceName, "stream_config.tier", "SP30"),
+					resource.TestCheckResourceAttr(resourceName, "stream_config.max_tier_size", "SP30"),
```
EspenAlbert, line 28: Do we want to test error condition if the max_tier_size is invalid?

internal/service/streamworkspace/resource_schema.go
```diff
@@ -46,6 +46,10 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Optional: true,
 				Computed: true,
 				Attributes: map[string]schema.Attribute{
+					"max_tier_size": schema.StringAttribute{
```
EspenAlbert, line 49: Is these fields always returned by the backend?
Seems like we can UseStateForUnknown here?

docs/resources/stream_workspace.md
marcosuma: ok cool, as long as everything is sync and driven by validation error I am fine. We just have to avoid side effects, meaning that a user starts "something" and the server readjust some values (side effects) to make them compatible

internal/service/streamworkspace/model.go
kpatel71716: @EspenAlbert if I do something like:

```
maxTierSize := streamConfig.MaxTierSize.ValueString()
		tier := streamConfig.Tier.ValueString()
		streamTenant.StreamConfig = &admin.StreamConfig{
			MaxTierSize: &maxTierSize,
			Tier:        &tier,
		}
```

Then it seems like we end up trying to pass empty strings  into the downstream api request which are invalid enum values. Same thing with streamConfig.MaxTierSize.ValueStringPointer(). I did get the following to work but lmk if I missed anything:
```
var maxTierSize *string
		if !streamConfig.MaxTierSize.IsNull() && !streamConfig.MaxTierSize.IsUnknown() {
			maxTierSize = streamConfig.MaxTierSize.ValueStringPointer()
		}
```

internal/service/streamworkspace/resource_test.go
kpatel71716: I am fine passing this through and relying on our downstream error handling and error messages but can discuss if it's a blocker

internal/service/streamworkspace/resource_schema.go
kpatel71716: The field is returned if it exists in the backend but otherwise will not be in the response. Added UseStateForUnknown

internal/service/streamworkspace/resource_test.go
EspenAlbert: No blocker!

internal/service/streamworkspace/model.go
EspenAlbert: I was thinking you can replace:
`if !streamConfig.MaxTierSize.IsNull() && !streamConfig.MaxTierSize.IsUnknown()`
with:
`if streamConfig.MaxTierSize.ValueString() != ""`

EspenAlbert (approved):
Thanks for addressing comments. Non-blocking comments left

internal/service/streamworkspace/model.go
kpatel71716: Ah yeah, fixed. Good idea!

internal/service/streamworkspace/move_state.go
kpatel71716: I agree with the idea. I tried some different approaches earlier since max_tier_size will not exist for a stream_instance resource/ds terraform config that is being migrated to stream_workspace. However I kept running into areas. Happy to explore in a follow-up if I missed anything

lantoli: @kpatel71716 yes, @EspenAlbert is right, as Read is called just after MoveState, we try to have the minimum logic in MoveState, filling only the fields that Read will need.

Take into account that anything set in MoveState will be overriden by what's done in Read, so it's not useful (and adds unnecessary complexity) to set other attributes in MoveState.

kpatel71716: Thanks @lantoli @EspenAlbert, still learning the behavior on my end so this is great to know! I'll follow up with a PR to clean this up once I am back from OOO!

PR #3880, opened 2025-11-12
chore: Generate alert configuration resource and add acc tests

manupedrozo:
## Description

Generate `alert_configuration_api` and add acc tests based on the handwritten `alert_configuration`.

Link to any related issue(s): CLOUDP-356406

## Type of change:

## Required Checklist:


Makefile
```diff
@@ -111,7 +111,7 @@ tools:  ## Install the dev tools (dependencies)
 	go install github.com/icholy/gomajor@latest
 	go install github.com/terraform-linters/tflint@v0.52.0
 	go install github.com/rhysd/actionlint/cmd/actionlint@latest
-	go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@latest
+	go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@v0.38.0 # Pinning version since v0.39.0 fails to fix files with large structs. See versions at https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment?tab=versions
```
AgustinBettati, line 114: don't see a concern but curious was this a bug on their latest version (0.39.0)?

internal/serviceapi/alertconfigurationapi/resource_test.go
```diff
@@ -0,0 +1,748 @@
...
+					resource.TestCheckResourceAttr(resourceName, "matchers.#", "1"),
+					resource.TestCheckResourceAttrSet(resourceName, "metric_threshold.metric_name"),
+				),
+			},
+			{
+				ResourceName:            resourceName,
+				ImportStateIdFunc:       importStateProjectIDFunc(resourceName),
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"group_id", "updated"},
```
lantoli: why do we need to ignore group_id? shouldn't change

internal/serviceapi/alertconfigurationapi/resource_test.go
```diff
@@ -0,0 +1,748 @@
...
+		}
+		id := rs.Primary.Attributes["id"]
+		projectID := rs.Primary.Attributes["group_id"]
+		if id == "" || projectID == "" {
+			return fmt.Errorf("checkExists, attributes not found for: %s", resourceName)
+		}
+		_, _, err := acc.ConnV2().AlertConfigurationsApi.GetAlertConfig(context.Background(), projectID, id).Execute()
+		if err != nil {
+			return fmt.Errorf("the Alert Configuration(%s) does not exist", id)
+		}
```
lantoli:
```suggestion
		if _, _, err := acc.ConnV2().AlertConfigurationsApi.GetAlertConfig(context.Background(), projectID, id).Execute(); err != nil {
			return fmt.Errorf("the Alert Configuration(%s) does not exist", id)
		}
```

nit, err scope is restricted to this block

lantoli (approved):

internal/serviceapi/alertconfigurationapi/resource_test.go
manupedrozo: Good catch! Removing the project_id ignores from the handwritten resource tests as well. Not sure why they were added. [006b01d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3880/commits/006b01d7f05204f8793c8f67e92fe71b393349e4)

internal/serviceapi/alertconfigurationapi/resource_test.go
manupedrozo: 👌 applied [006b01d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3880/commits/006b01d7f05204f8793c8f67e92fe71b393349e4)

internal/serviceapi/alertconfigurationapi/resource_test.go
```diff
@@ -0,0 +1,747 @@
+package alertconfigurationapi_test
```
AgustinBettati, line 1: nice, great to see coverage over the full API and handling redacted sensitive values in response

tools/codegen/config.yml
```diff
@@ -543,3 +543,53 @@ resources:
...
+            optional: true
+            computed: true
+        notifications.delay_min:
+          computability:
+            optional: true
+            computed: true
+        notifications.interval_min:
+          computability:
+            optional: true
+            computed: true
```
AgustinBettati: so these are both getting default values returned from the API right?

AgustinBettati (approved):
LGTM

tools/codegen/config.yml
manupedrozo: Followed handwritten resource where these are optional + computed.
However, as per my testing only delay_min is computed, interval_min is not, so good call out!
Removing this override in [6da4587](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3880/commits/6da4587b3295d342f33c4fe9f3620776aefb8125)
We should also remove the computed flag from the handwritten resource as a followup: [ptr](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/alertconfiguration/resource.go#L290-L293) - thoughts?

Makefile
manupedrozo: It only started happening today, checked versions and 0.39 was released yesterday.
Can change back to latest when fixed, haven't looked if there is an associated bug.

manupedrozo:
Tests failing due to S3 issue. Unrelated to PR changes. Oncall looking into it with upstream team.

tools/codegen/config.yml
AgustinBettati: Thanks! 

Adjusting the existing `alert_configuration` resource schema in this case could risk a breaking change. Optional+Computed implies users can potentially not have the attribute in the config but have a value in the state. As soon as the attribute changes from Optional+Computed to Optional this would trigger a plan change in which terraform tries to unset the attribute. We could create a ticket and flag as candidate for including in a 3.0 release.

manupedrozo: Got it, thanks! To clarify, a user having the value in the state but not config is not expected, correct? Or is there a common flow where this may happen?

AgustinBettati: Given API behaviour should not be common scenario: Potentially a user that defined the attribute and then removes it, leaving the config with no attribute but the state will still have it present.

Issue #3869, opened 2025-11-10
[Bug]: Can't directly change alerts notification type, need to delete it first


YusDyr:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Provider Version

1.14.0

### Terraform Version

1.9.0

### Terraform Edition

Terraform Open Source (OSS)

### Current Behavior

When I created alerts with one notification type, such as, for example, `PAGER_DUTY` or `SLACK` and now want to change it to another notification type, such as `TEAMS`, I constantly got errors related to notification-specific parameters. It seems like the terraform checks first compatibility between **existing** notification parameters and **new** notification type even despite these conflicting parameters should be deleted on update.

### Terraform configuration to reproduce the issue

```hcl
variable "environment" {
  description = "Name of the environment where table is being deployed."
  type        = string
}

variable "alerts_no_primary" {
  description = "[no_primary] Alert settings"
  type = object({
    create        = optional(bool, true)
    enable        = optional(bool, true)
    interval_min  = optional(number, 30)
    delay_min     = optional(number, 0)
    notifications = optional(set(string), ["PD", "Teams"])
  })
  default = {}
}

variable "alerts_notifications_config" {
  description = "Alerts notification configs, currently for PD and SLACK"
  type = map(object({
    notification_type = string
    delay_min         = optional(number) # Number of minutes to wait after an alert condition is detected before sending out the first notification.
    interval_min      = optional(number) # Number of minutes to wait between successive notifications for unacknowledged alerts that are not resolved. The minimum value is 5. **NOTE** `PAGER_DUTY`, `VICTOR_OPS`, and `OPS_GENIE` notifications do not return this value. The notification interval must be configured and managed within each external service.
    team_name         = optional(string) # Label for the team that receives this notification
    channel_name      = optional(string) # Attributes for SLACK
    teams_webhook_url = optional(string) # Webhook URL for Teams in case of multiple Teams channel are used
  }))
  default = {
    "PD" = {
      notification_type = "PAGER_DUTY"
    }
    "Teams" = {
      notification_type = "MICROSOFT_TEAMS"
    }
  }
  validation {
    condition     = alltrue([for o in var.alerts_notifications_config : contains(["SLACK", "PAGER_DUTY", "MICROSOFT_TEAMS", "DATADOG"], o.notification_type)])
[...]

PR #3882, opened 2025-11-13
fix: Removes interval_min validation blocking notification type updates in `alert_configuration`

EspenAlbert:
## Description

Remove validation that prevented updating alert notifications from types
supporting interval_min to PAGER_DUTY/OPS_GENIE/VICTOR_OPS. Since
interval_min is optional and not preserved via UseStateForUnknown, it will
only be sent to the API if explicitly set in config, allowing updates to
succeed.

Link to any related issue(s): #3869 CLOUDP-358194

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/alertconfiguration/model.go
+++ b/internal/service/alertconfiguration/model.go
@@ -2,7 +2,6 @@ package alertconfiguration
 
 import (
 	"fmt"
-	"strings"
 
 	"go.mongodb.org/atlas-sdk/v20250312009/admin"
 
@@ -13,15 +12,6 @@ import (
 
 func NewNotificationList(list []TfNotificationModel) (*[]admin.AlertsNotificationRootForGroup, error) {
 	notifications := make([]admin.AlertsNotificationRootForGroup, len(list))
-
-	for i := range list {
-		if !list[i].IntervalMin.IsNull() && list[i].IntervalMin.ValueInt64() > 0 {
-			typeName := list[i].TypeName.ValueString()
-			if strings.EqualFold(typeName, pagerDuty) || strings.EqualFold(typeName, opsGenie) || strings.EqualFold(typeName, victorOps) {
-				return nil, fmt.Errorf(`'interval_min' must not be set if type_name is 'PAGER_DUTY', 'OPS_GENIE' or 'VICTOR_OPS'`)
-			}
-		}
-	}
 
 	for i := range list {
 		n := &list[i]
--- a/internal/service/alertconfiguration/resource.go
+++ b/internal/service/alertconfiguration/resource.go
@@ -290,6 +290,9 @@ func (r *alertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						"interval_min": schema.Int64Attribute{
 							Optional: true,
 							Computed: true,
+							Validators: []validator.Int64{
+								ValidIntervalMin(),
+							},
 						},
 						"mobile_number": schema.StringAttribute{
 							Optional: true,
--- a/internal/service/alertconfiguration/validator_interval_min.go
+++ b/internal/service/alertconfiguration/validator_interval_min.go
@@ -1,0 +1,86 @@
+package alertconfiguration
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/helpers/validatordiag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+// IntervalMinValidator validates that interval_min is not set for notification types
+// that don't support it (PAGER_DUTY, OPS_GENIE, VICTOR_OPS).
+//
+// API-level validation cannot catch this error because when the notification type
+// doesn't support interval_min, the nested_type in the API payload omits the field,
+// so it's never sent to the API. Client-side validation ensures users get immediate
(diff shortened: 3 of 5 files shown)


EspenAlbert:
Failing test is data federation with s3 access error (investigation going on with upstream team)

internal/service/alertconfiguration/validator_interval_min.go
```diff
@@ -0,0 +1,86 @@
...
+	return "'interval_min' must not be set if type_name is 'PAGER_DUTY', 'OPS_GENIE' or 'VICTOR_OPS'"
+}
+
+func (v IntervalMinValidator) MarkdownDescription(ctx context.Context) string {
+	return v.Description(ctx)
+}
+
+func (v IntervalMinValidator) ValidateInt64(ctx context.Context, req validator.Int64Request, resp *validator.Int64Response) {
+	// If the value is unknown/null/0, there is nothing to validate.
+	if req.ConfigValue.ValueInt64() <= 0 {
```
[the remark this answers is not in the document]
EspenAlbert: Check the documentation of ValueInt64(), you can see that both those cases returns 0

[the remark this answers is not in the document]
EspenAlbert: Check the documentation of ValueInt64(), you can see that both those cases returns 0

jvincent-mongodb (approved):

internal/service/alertconfiguration/resource.go
```diff
@@ -290,6 +290,9 @@ func (r *alertConfigurationRS) Schema(ctx context.Context, req resource.SchemaRe
 						"interval_min": schema.Int64Attribute{
 							Optional: true,
 							Computed: true,
+							Validators: []validator.Int64{
+								ValidIntervalMin(),
+							},
```
AgustinBettati, line 295: q: If intervalMin is send in a request using incorrect type does the API return any error or simply ignores? Just thinking if we can avoid this validation from our side.

internal/service/alertconfiguration/validator_interval_min.go
```diff
@@ -0,0 +1,86 @@
...
+		return
+	}
+
+	// Parse the path to find which notification index we're validating
+	// Path format: notification[0].interval_min
+	pathStr := req.Path.String()
```
AgustinBettati: relying on path string format looks a little risky but I guess we would catch this quickly if it changes in a future version

internal/service/alertconfiguration/model.go
```diff
@@ -14,15 +13,6 @@ import (
...
 	notifications := make([]admin.AlertsNotificationRootForGroup, len(list))
 
-	for i := range list {
-		if !list[i].IntervalMin.IsNull() && list[i].IntervalMin.ValueInt64() > 0 {
-			typeName := list[i].TypeName.ValueString()
-			if strings.EqualFold(typeName, pagerDuty) || strings.EqualFold(typeName, opsGenie) || strings.EqualFold(typeName, victorOps) {
-				return nil, fmt.Errorf(`'interval_min' must not be set if type_name is 'PAGER_DUTY', 'OPS_GENIE' or 'VICTOR_OPS'`)
-			}
-		}
-	}
```
AgustinBettati, line 24: q: When value is unset in the config, doesnt the `IntervalMin` value appear as unknown in the plan? Would expect validation is not ecountered, probably missing something.

EspenAlbert: I am thinking the same, especially since we are not using UseStateForUnknown.
However, no reason to have it here as we are doing the validation in the schema.

internal/service/alertconfiguration/resource.go
EspenAlbert: Good question! According to the validator's documentation comment (lines 17-20 in `validator_interval_min.go`), the API doesn't actually receive `interval_min` when the notification type doesn't support it - the nested_type in the API payload omits the field entirely, so it's never sent to the API. This means API-level validation cannot catch this error.

I think it would be misleading if we don't raise an error if the user sets it for a type that doesn't support it

internal/service/alertconfiguration/validator_interval_min.go
EspenAlbert: Agreed - parsing the path string format (e.g., `notification[0].interval_min`) is a bit fragile, but it's a necessary workaround given the validator's constraints. The validator only receives the attribute value and path, not the full notification context, so we need to extract the notification index from the path to check the corresponding `type_name`.

If the path format changes in a future Terraform version, the validator will gracefully skip validation (lines 52-55) when it can't parse the index, so it won't break functionality - it just won't validate in that case. We'd likely catch this during testing since the validator is covered by tests.

manupedrozo: Should be able to simplify the parsing and de-risk with:
```go
	notificationPath := req.Path.ParentPath() // <- This would give notification[0]

	var notification TfNotificationModel
	diags := req.Config.GetAttribute(ctx, notificationPath, &notification) // <- This should give you the notification at index 0 directly, removing the need to parse the index.
	if diags.HasError() {
		// If we can't read the notification, skip validation (might be unknown during plan)
		return
	}
```
Haven't tested it so may be missing something.

EspenAlbert: Great suggestion! 9bbec776c24fe764142d60d43de4b40348f0efcd

manupedrozo (approved):

AgustinBettati (approved):

PR #3860, opened 2025-11-07
chore: Support dynamic json values which Map type

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-352319

### Before
- Code generation didn’t handle OpenAPI object maps with dynamic json as values, resulting in a panic during code generation.
```
objectOfDynamicValues:
  description: Object of dynamic values.
  type: object
  additionalProperties:
    type: object
```
- `search_index_api` couldn’t express `definition.mappings.fields` as a map of schemaless JSON values.

### After
- Codegen recognizes this pattern and generates a `MapValue[jsontypes.Normalized]` on the terraform side.
- `search_index_api` now exposes `definition.mappings.fields` enabling `jsonencode(...)` per field.
- Add unit testing to ensure Marshal/Unmarshal support dynamic JSON (`jsontypes.Normalized`) across map/list/set.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [ ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [ ] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [ ] I have added any necessary documentation (if appropriate)
- [ ] I have run make fmt and formatted my code
- [ ] If changes include deprecations or removals I have added appropriate changelog entries.
- [ ] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

PR #3884, opened 2025-11-14
chore: Generate and consume flattened multi-version API Spec in autogeneration

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-358039

### Motivation

Our autogeneration pipeline previously consumed a single‑version API spec, with all transformations associated to Go SDK. This change switches the pipeline to use a flattened, multi‑version API spec, allowing for generation of resources using older versions.

### What’s changing
- Define a new make command which fetches latest API Spec and applies flattening transformation using Go SDK transformation CLI (https://github.com/mongodb/atlas-sdk-go/pull/682).
- Start storing raw and flattened API Spec which are used for autogeneration. This will preserve additional metadata of how autogen resources are generated at any point in time.
- Adjust autogen logic to support processing API Spec which contains multiple versions. This implied adjusting logic on how we pick over the multiple versions available.
- Flattened API Spec does not incorporate specific Go SDK step of removing additional properties. As object types with dynamic values are supported (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3860) we can handle this case and align more closely to the original API Spec.

## Type of change:

## Required Checklist:


.github/workflows/generate-autogen-resources.yml
```diff
@@ -17,8 +17,18 @@ jobs:
...
+      - name: Check for changes in ./internal (resource implementations)
+        id: check_internal_changes
+        run: |
+          if git ls-files -m -o --exclude-standard -- internal | grep -q .; then
+            echo "internal_changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "internal_changed=false" >> "$GITHUB_OUTPUT"
+          fi
       - name: Create PR
+        if: steps.check_internal_changes.outputs.internal_changed == 'true'
```
EspenAlbert, line 31: [nit] Don't we use third person in PR title: Updates (notice the "s")

scripts/generate-autogen-api-spec.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SOURCE_SPEC_URL="${SOURCE_SPEC_URL:-https://raw.githubusercontent.com/mongodb/openapi/main/openapi/v2.yaml}"
```
EspenAlbert, line 4: I think we need to avoid committing these big files.
Instead we can just store the `info.xgen-sha: 1b6bd14effc25bc220b0f5b5a904b4ed4cf6ffb3` from the spec.

Not only will these files make a big diff but it will make our git commands slower and slower

tools/codegen/codespec/api_spec.go
```diff
@@ -41,16 +42,55 @@ func BuildSchema(proxy *base.SchemaProxy) (*APISpecSchema, error) {
 	return resp, nil
 }
 
-func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType]) (*APISpecSchema, error) {
+func getSchemaFromMediaType(mediaTypes *orderedmap.Map[string, *high.MediaType], configuredVersion *string) (*APISpecSchema, error) {
 	if mediaTypes == nil {
 		return nil, errSchemaNotFound
 	}
 
+	// If a configured version is provided, attempt to select the exact match or the closest older available version.
```
EspenAlbert, line 50: Why do we need the fuzzy matching?

tools/codegen/openapi/parser.go
```diff
@@ -25,31 +21,3 @@ func ParseAtlasAdminAPI(filePath string) (*libopenapi.DocumentModel[v3.Document]
 
 	return docModel, nil
 }
-
```
EspenAlbert, line 28: IMO: This is a downgrade. Now you have to run shell scripts+`go run`. I prefer having a standalone go run that is "smart" can support a file path specified or downloading if it doesn't exist

.github/workflows/check-autogen-resources.yml
```diff
@@ -1,6 +1,6 @@
 name: 'Check auto-generated resources'
 
-# Set of validations run on each pull request and merged commits to master.
+# Ensures that the auto-generated resources are up to date given current API Spec and autogen config.
```
manupedrozo, line 3: So we are not enforcing a bump of the API Spec even if it would cause new changes. We rely on the weekly update for that.
Ok with me, we can keep an eye on this.👌

manupedrozo (approved):

scripts/generate-autogen-api-spec.sh
manupedrozo: Good point, should have called this out in the proposal doc

AgustinBettati: This came from feedback in the eng proprosal.

> Not only will these files make a big diff but it will make our git commands slower and slower

I don't anticipate an impact, versioning the raw and transformed API specs is a common pattern in our other autogen tools ([Go SDK](https://github.com/mongodb/atlas-sdk-go/tree/main/openapi), [Atlas CLI](https://github.com/mongodb/mongodb-atlas-cli/tree/master/tools/internal/specs)).

Don't have a strong stance, found the benefit of keeping transparency of the source spec + explicit spec level diffs slightly outweighing the potential risk of slower git commands. Looping @ciprian-tibulca in case of additional inputs.

.github/workflows/generate-autogen-resources.yml
AgustinBettati: true, adjusted

tools/codegen/codespec/api_spec.go
AgustinBettati: TLDR: Allows defining a single version per resource in our autogen config.

In practice what can happen in our API is that different operation of a resource have different versions: E.g. to use latest cluster operations you have POST and PATCH in 2024-10-23 while GET and DELETE in 2024-08-05.

This allows defining autogen cluster config to use version 2024-10-23, and for the case of GET and DELETE it will parse from 2024-08-05. This is what the API also does when we send the header 2024-10-23 for all operations.

tools/codegen/openapi/parser.go
AgustinBettati: What we used to do of simply downloading the input API Spec (for Go SDK repo) is no longer possible because the API Spec we want to consume (flattened + multi-version) is not available as a direct download. The associated eng. proposal defines a long term plan in which we work towards having the input API Spec available for downloading directly, but as of now it will be generated as part of `make generate-autogen-api-spec`.

.github/workflows/check-autogen-resources.yml
AgustinBettati: Correct, what I want to avoid here is potentially blocking unrelated PRs due to changes in API Spec + autogenerated resources.

internal/serviceapi/streamprocessorapi/resource_schema.go
```diff
@@ -67,16 +67,21 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "The state of the stream processor. Commonly occurring states are 'CREATED', 'STARTED', 'STOPPED' and 'FAILED'.",
 			},
-			"stats": schema.StringAttribute{
+			"stats": schema.MapAttribute{
```
lantoli, line 70: search_index_api and stream_processor_api changes don't seem to be related to PR title

scripts/generate-autogen-api-spec.sh
lantoli: I agree that we don't need to store it here if it's already somewhere else and we can have the SHA in case we want to check if it's the file we used.

also will get worse if we want to store all API versions, not just the last one as in SDK and CLI

tools/codegen/openapi/parser.go
lantoli: don't know if it's a good idea, but having the [version list](https://github.com/mongodb/atlas-sdk-go/blob/79c329b70c7e6be757537cead3a821371fac2543/tools/scripts/fetch.sh#L35...L36) we could download OpenAPI spec files for all versions (e.g. suffix with version) and use instead of one apiSpec with type *libopenapi.DocumentModel[v3.Document],  a map from string (version) to apiSpec

scripts/generate-autogen-api-spec.sh
AgustinBettati: > if it's already somewhere

Currently it is not defined elsewhere, the resulting API Spec is obtained by downloading raw multi-version API Spec and applying the flattening transformation (does not include unrelated SDK transformations).

> will get worse if we want to store all API versions

In terms of file size it is not looking considerable, with multi-version [v2.yaml](https://github.com/mongodb/openapi/blob/main/openapi/v2.yaml) (3.25 MB) actually being more light compared to single version [openapi-2025-03-12.yaml](https://github.com/mongodb/openapi/blob/main/openapi/v2/openapi-2025-03-12.yaml) (4.13 MB). Suspect this difference is due to formatting differences when doing the splitting.
Can see [CLI is using the multi version spec](https://github.com/mongodb/mongodb-atlas-cli/blob/master/tools/internal/specs/spec.yaml) as well.

Just to be clear, we have an eng proposal that seeks longer term improvements (CLOUDP-358089). For now not seeing a risk and we can always come back to this decision.

tools/codegen/openapi/parser.go
AgustinBettati: As we already had some logic for handling header version, expanding for it to handle different versions was not complex. Keeping to a single file that is transformed also kept the process more simple.

internal/serviceapi/streamprocessorapi/resource_schema.go
AgustinBettati: Yeah true, this was a consequence of using the new API Spec which did not include some SDK specific transformations. Left details in the PR description:

- Flattened API Spec does not incorporate specific Go SDK step of removing additional properties. As object types with dynamic values are supported (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3860) we can handle this case and align more closely to the original API Spec.

scripts/generate-autogen-api-spec.sh
ciprian-tibulca: Adding my thoughts:
even with few ~3 MB specs (less for single version ones) the impact on clone, fetch etc should not be significant.
We already follow the same pattern in AtlasCLI and have not seen issues with git performance.
Keeping the specs in the repo has a few benefits:
1. changes related to the auto-generated resources can be directly linked with spec changes (in the same PR)
2. easier debugging, we can reproduce faster what the CLI used at a given commit
3. we keep a backup snapshot of the specs that the CLI depends on
4. the debugging process becomes independent of where the source specs are hosted (in the unlikely event we change the repo, or the storage entirely, e.g. start downloading from a API endpoint or from S3).

tools/codegen/codespec/api_spec.go
EspenAlbert: Nice, thank you for the explanation. Maybe add a comment? FFTR

tools/codegen/openapi/parser.go
EspenAlbert: I think we could have called the flattening command directly from the go code and avoided the shell script.
FFTR.

scripts/generate-autogen-api-spec.sh
EspenAlbert: Thanks for the discussion, not seeing this as a blocker.
I think the main point of:

> changes related to the auto-generated resources can be directly linked with spec changes (in the same PR)

Has the greatest value. Although this could probably be a link comparing the same file of the different hashes (harder to compare the flattened version). 

FFTR.
I think if you try to clone the openapi repo: `git clone https://github.com/mongodb/openapi` you can see the impact of including these specs, not saying it will be as bad 😅  but it takes more than 20 min on my machine 😱

tools/codegen/codespec/api_spec.go
AgustinBettati: Thanks, included

tools/codegen/openapi/parser.go
AgustinBettati: In practical terms I see separation of updating/generating spec and actual code generation can be useful. API Spec will be updated weekly by PRs, however when running locally having the option of only running the code generation is helpful because it runs faster (no need to download each time). All paths (only update API Spec, only generate resources, update API Spec + generate resources) are exposed by separate make commands.

EspenAlbert: I would rather add this as a feature to the go script, e.g., skip download if the file exists.
FFTR

PR #3895, opened 2025-11-18
chore: Add use_effective_fields to resource and data source schema and model

marcabreracast:
## Description

This PR adds the `use_effective_fields` boolean to the schema and model definition of the resource and data source for `advanced_cluster`. 

This change does not contain the implementation of the field, this will be addressed in a follow-up PR. 

Link to any related issue(s): CLOUDP-359712

## Type of change:

- [X] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.

## Required Checklist:


internal/service/advancedcluster/schema.go
```diff
@@ -313,6 +313,10 @@ func resourceSchema(ctx context.Context) schema.Schema {
 				Update: true,
 				Delete: true,
 			}),
+			"use_effective_fields": schema.BoolAttribute{
+				Optional:            true,
+				MarkdownDescription: "Flag that indicates whether to use the effective fields for the cluster.",
```
lantoli:
```suggestion
				MarkdownDescription: "Controls how hardware specification fields are returned in the response. When set to true, returns the original client-specified values and provides separate effective fields showing current operational values. When false (default), hardware specification fields show current operational values directly. Primarily used for autoscaling compatibility.",
```

I'd use the same description as in API: https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-getgroupcluster#operation-getgroupcluster-headers

also if we agree on the description, for example this proposed description, I can use the same in the documentation PR

marcabreracast: Yep, SGTM!

marcabreracast: Addressed in [7b5a7f5](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3895/commits/7b5a7f540f9ff61c92f5a36efa9377bfe17da10f)

marcabreracast:
`advanced_cluster` tests are also failing on nightly run with same error message: https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/19485228042/job/55766090036

internal/service/advancedcluster/resource_compatiblity.go
```diff
@@ -101,6 +101,8 @@ func overrideAttributesWithPrevStateValue(modelIn, modelOut *TFModel) {
 	if modelIn.DeleteOnCreateTimeout.ValueBoolPointer() != nil {
 		modelOut.DeleteOnCreateTimeout = modelIn.DeleteOnCreateTimeout
 	}
+
+	modelOut.UseEffectiveFields = modelIn.UseEffectiveFields
```
marcabreracast: This ensures that we're copying the value of what's on the state. Because the implementation on the resource and datasource is not done yet (will be done in [CLOUDP-360425](https://jira.mongodb.org/browse/CLOUDP-360425)). 
Once that's implemented, this line can be removed, for now it avoids encountering errors such as: 

```
│ 
│ When applying changes to mongodbatlas_advanced_cluster.cluster, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an
│ unexpected new value: .use_effective_fields: was cty.True, but now null.
│ 
╵
```

lantoli: I don't think it can be removed later

internal/service/advancedcluster/schema.go
```diff
@@ -335,6 +339,7 @@ func dataSourceOverridenFields() map[string]dsschema.Attribute {
 	return map[string]dsschema.Attribute{
 		"accept_data_risks_and_force_replica_set_reconfig": nil,
 		"delete_on_create_timeout":                         nil,
+		"use_effective_fields":                             dsschema.BoolAttribute{Optional: true},
```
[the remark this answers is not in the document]
lantoli: @marcabreracast this is true, Mardown must also be defined in overrides, but I'm doing it in a commit so no need for you to do it

internal/service/advancedcluster/resource_compatiblity.go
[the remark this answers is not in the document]
lantoli: changing approach

lantoli (approved):

internal/service/advancedcluster/resource_compatiblity.go
```diff
@@ -90,19 +90,21 @@ func resolveContainerIDs(ctx context.Context, projectID string, cluster *admin.C
 }
 
 func overrideAttributesWithPrevStateValue(modelIn, modelOut *TFModel) {
+	if modelIn == nil || modelOut == nil {
```
marcosuma: one difference I notice with this change is that this method was being called before if `modelIn == nil` - but now it won't since it returns. Do we think that's a problem?

cc @lantoli

internal/service/advancedcluster/resource_compatiblity.go
```diff
@@ -90,19 +90,21 @@ func resolveContainerIDs(ctx context.Context, projectID string, cluster *admin.C
...
 	if beforeVersion != nil && !modelIn.MongoDBMajorVersion.Equal(modelOut.MongoDBMajorVersion) {
 		modelOut.MongoDBMajorVersion = types.StringPointerValue(beforeVersion)
 	}
-	retainBackups := conversion.NilForUnknown(modelIn.RetainBackupsEnabled, modelIn.RetainBackupsEnabled.ValueBoolPointer())
-	if retainBackups != nil && !modelIn.RetainBackupsEnabled.Equal(modelOut.RetainBackupsEnabled) {
-		modelOut.RetainBackupsEnabled = types.BoolPointerValue(retainBackups)
-	}
-	if modelIn.DeleteOnCreateTimeout.ValueBoolPointer() != nil {
-		modelOut.DeleteOnCreateTimeout = modelIn.DeleteOnCreateTimeout
-	}
```
marcosuma, line 103: curious: why this is not needed anymore? and is it related to this PR that adds the effective_field boolean?

marcosuma (commented):
LGTM, just left some comments before I approve

internal/service/advancedcluster/resource_compatiblity.go
lantoli: no problem, we just really need to check modelOut nilness,   modelIn is always present (with current code), it would have panicked if not.       But checking modelIn and modelOut to be safe if we change the code later

internal/service/advancedcluster/resource_compatiblity.go
lantoli: Timeout has just been moved to the func, they were set in the callees before. 
About RetainBackupsEnabled, I realized that we don't need to do the complicated logic that we copied from MongoDBMajorVersion.  In MongoDBMajorVersion is necessary because it's Optional+Computed but RetainBackupsEnabled is only Optional. Similarly DeleteOnCreateTimeout is O+C but because the default logic (not because Atlas returns it) and can't be changed after creation, so no complex logic needed either.

Note that fields are still copied, but directly without complex logic

lantoli: looping @EspenAlbert in to confirm as he did the initial implementation, to make sure I'm not missing anything

EspenAlbert: Sounds good, thank you for the simplification!
I think it works for all fields the cannot be Unknown after the planning:
- Optional only (will never be unknown)
- O+C (will never be unknown if it uses UseStateForUnknown or has a default value, otherwise not safe)

internal/service/advancedcluster/schema.go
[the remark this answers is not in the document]
marcabreracast: Ack, thanks!

EspenAlbert (approved):
Nice!

internal/service/advancedcluster/schema.go
```diff
@@ -328,13 +336,23 @@ func pluralDataSourceSchema(ctx context.Context) dsschema.Schema {
...
+			},
+		},
 	})
 }
 
 func dataSourceOverridenFields() map[string]dsschema.Attribute {
 	return map[string]dsschema.Attribute{
 		"accept_data_risks_and_force_replica_set_reconfig": nil,
 		"delete_on_create_timeout":                         nil,
+		"use_effective_fields": dsschema.BoolAttribute{
```
EspenAlbert, line 352: not sure why this is not set by default? Why do we override it? Because it must be `optional`?

internal/service/advancedcluster/resource_compatiblity.go
lantoli: I've changed my mind and don't check modelIn, it would be a programming error so better to panic and detect it earlier while developing: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3895/commits/739b281686dbf8d48f34eaae140e8b6a504681fa

cc @EspenAlbert

internal/service/advancedcluster/resource_compatiblity.go
lantoli: thanks for the explanation, added comment: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3895/commits/dd69e68789ba153ce631c506dee1bb681b93836e

internal/service/advancedcluster/schema.go
lantoli: correct, DataSourceSchemaFromResource and PluralDataSourceSchemaFromResource change all params (except the ones defined in RequiredFields) to be Computed-only.  In this case we want it to be Optional-only.

manupedrozo (approved):

internal/service/advancedcluster/resource_compatiblity.go
```diff
@@ -90,19 +90,23 @@ func resolveContainerIDs(ctx context.Context, projectID string, cluster *admin.C
...
 	overrideMapStringWithPrevStateValue(&modelIn.Labels, &modelOut.Labels)
 	overrideMapStringWithPrevStateValue(&modelIn.Tags, &modelOut.Tags)
+
+	// Copy Terraform-only attributes which are not returned by Atlas API.
```
oarbusi, line 103: nice simplification!

oarbusi (approved):

marcosuma (approved):

PR #3901, opened 2025-11-19
chore: Add moved block support in autogen

manupedrozo:
## Description

Add support for moved blocks in autogen. 
Enabled it for `cluster_api`, can now move a `cluster_old_api` into a `cluster_api` resource.

Link to any related issue(s): CLOUDP-358052

## Type of change:

## Required Checklist:


internal/serviceapi/clusterapi/move_test.go
```diff
@@ -0,0 +1,213 @@
...
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroyDatabaseUser,
+		Steps: []resource.TestStep{
+			{
+				Config: configMoveFirstInvalid(projectID, clusterName),
+			},
+			{
+				Config:      configMoveSecondInvalid(projectID, clusterName),
+				ExpectError: regexp.MustCompile("Unable to Move Resource State"),
+			},
```
AgustinBettati, line 57: If we want to test the scenario of the source resource not existing I would avoid the first step create the unrelated database_user resource altogether

internal/common/autogen/move.go
```diff
@@ -0,0 +1,55 @@
...
+		return
+	}
+
+	var stateObj map[string]tftypes.Value
+	if err := rawStateValue.As(&stateObj); err != nil {
+		resp.Diagnostics.AddError("Unable to parse source state", err.Error())
+		return
+	}
+
+	for _, attrName := range idAttrs {
```
AgustinBettati, line 44: q: As we are only copying the id attributes (which seems reasonable), how is the UX for a user that defines timeouts? Do they have to do a followup apply after the initial move operation is complete?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -842,11 +863,67 @@ func TestConvertToProviderSpec_dynamicJSONProperties(t *testing.T) {
...
+						HTTPMethod: "PATCH",
+					},
+					Delete: &codespec.APIOperation{
+						Path:       "/api/atlas/v2/groups/{groupId}/testResourceMoveState",
+						HTTPMethod: "DELETE",
+					},
+					VersionHeader: "application/vnd.atlas.2023-01-01+json",
+				},
+				MoveState: &codespec.MoveState{SourceResources: []string{"test_resource"}},
+			}},
```
AgustinBettati, line 923: nit: As this test is only verifying the config is reflected in the intermediate model, I would consider reusing an existing basic test like `TestConvertToProviderSpec` (we avoid adding all the additional CRUDs in the API Spec which are not really relevant).

AgustinBettati (approved):
LGTM

internal/serviceapi/clusterapi/move_test.go
manupedrozo: Terraform ignores moved blocks when the source resource does not exist.
This test covers the case of attempting to move from an (existing) unsupported resource.

internal/common/autogen/move.go
manupedrozo: If the timeouts are defined for the target, those are saved to the state in the same apply that is doing the move without the need for an extra apply.

The plan looks like this:
```sh

  # mongodbatlas_cluster_api.test will be updated in-place
  # (moved from mongodbatlas_cluster_old_api.old)
  ~ resource "mongodbatlas_cluster_api" "test" {
      ~ advanced_configuration                                 = {
          + custom_openssl_cipher_config_tls12 = (known after apply)
          ~ minimum_enabled_tls_protocol       = "TLS1_2" -> (known after apply)
          ~ tls_cipher_config_mode             = "DEFAULT" -> (known after apply)
        } -> (known after apply)
      
      ...

      + timeouts                                               = {
          + create = "1h"
        }
      ~ use_aws_time_based_snapshot_copy_for_fast_initial_sync = false -> (known after apply)
      ~ version_release_system                                 = "LTS" -> (known after apply)
        # (2 unchanged attributes hidden)
    }
```

Regarding the copying of timeout values, I am on the side of letting the user define them explicitly in the target resource if desired as opposed to using values defined in the old resource implicitly, which is likely to cause confusion. Happy to discuss further.

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
manupedrozo: I will come back to this in a followup task. 
I do like having a separate test for this but +1 on the current approach not scaling as we add more features.
I'll define a simple test resource API with minimal attributes that can be reused for multiple tests. 
So we will have clearly defined tests while avoiding duplicating stuff in the spec.

internal/serviceapi/clusterapi/move_test.go
AgustinBettati: I see, might be good to add a comment just for future reference

internal/common/autogen/move.go
AgustinBettati: Interesting, so the moved plan shows any additional changes it would be making to the final state. I agree in this case moving the timeouts into the target state does not make much difference in the migration experience.

internal/serviceapi/clusterapi/move_test.go
manupedrozo: Changed test naming to "unsupported" in [916ce1a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3901/commits/916ce1aa5c8c66f7afdfb9cd789818e608f3b71f)

manupedrozo:
Test failures unrelated to changes - cloud-dev is flaky today.

PR #3908, opened 2025-11-20
chore: Support configuration for update marshalling to send null for list/set types

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-360429

### Context
We currently have a configuration `include_null_on_update` in our autogen tool that allows configuring specific properties to send null if there is no value present in the config (applies for update requests). This configuration however is not working for list and set types, for these types an empty array is always sent (default behaviour).

Sending null instead of empty array is a relevant use case for `type_sets` attribute in search index resource. Sending an empty array in all patch requests was leading to API validation errors in the PATCH request.

### Changes
- Change in marshalling logic to ensure `include_null_on_update` works for all attribute types.
- Update `search_index_api` with configuration for `type_sets`. Uncomment all update test steps which are now successful when sending `"typeSets": null` instead of `"typeSets": []`

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogen/marshal.go
+++ b/internal/common/autogen/marshal.go
@@ -81,13 +81,15 @@ func marshalAttr(attrNameModel string, attrValModel reflect.Value, objJSON map[s
 		return err
 	}
 
-	if val == nil && isUpdate {
+	// Emit empty array for null list/set attributes unless explicit configuration with includeNullOnUpdate
+	if val == nil && isUpdate && !includeNullOnUpdate {
 		switch obj.(type) {
 		case customtypes.ListValueInterface, customtypes.NestedListValueInterface, customtypes.SetValueInterface, customtypes.NestedSetValueInterface:
-			val = []any{} // Send an empty array if it's a null root list or set
+			val = []any{}
 		}
 	}
 
+	// Emit value if non-nil, or emit null on update when configured by includeNullOnUpdate
 	if val != nil || (isUpdate && includeNullOnUpdate) {
 		objJSON[attrNameJSON] = val
 	}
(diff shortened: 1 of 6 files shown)


internal/common/autogen/marshal_test.go
```diff
@@ -157,27 +157,70 @@ func TestMarshalOmitJSONUpdate(t *testing.T) {
...
+		AttrBoolIncludeNull:         types.BoolNull(),
 	}
 	// null list and set root elements are sent as empty arrays in update.
 	// fields with includenullonupdate tag are included even when null during updates.
 	const expectedJSON = `
 		{
-			"attrCustomList": [],
-			"attrCustomSet": [],
-			"attrIncludeString": null
+			"attrList": [],
```
AgustinBettati, line 210: marshal changes only impacted `attrListIncludeNull` and `attrSetIncludeNull`, added additional testing to ensure override works for all types

AgustinBettati:
Unrelated test failing: `TestAccClusterAPI_basic`

internal/common/autogen/marshal.go
```diff
@@ -81,13 +81,15 @@ func marshalAttr(attrNameModel string, attrValModel reflect.Value, objJSON map[s
 		return err
 	}
 
-	if val == nil && isUpdate {
+	// Emit empty array for null list/set attributes unless explicit configuration with includeNullOnUpdate
+	if val == nil && isUpdate && !includeNullOnUpdate {
```
manupedrozo, line 85: Ty for adding comments.
I think we should eventually come back to this and cleanup the different cases. 
Maybe make everything explicit instead of sending an empty array as the default behavior.

manupedrozo (approved):

internal/common/autogen/marshal.go
AgustinBettati: Yes agreed, will bring this topic up in a tech sync to go more in detail. 
Main gap I see is we are not consistent between list and map types:
- list: by default we send  empty value `[]`
- map: by default we omit the property, instead of sending an empty value `{}`
- for other types (string, int, nested objects) I can understand the default of not sending the property, a default value could have a particular meaning for certain APIs (e.g. 0 or "" != omitting property)

PR #3929, opened 2025-11-26
chore: Autogen API operations customization support

manupedrozo:
## Description

POC: Adding customization support in autogen.
- Added support for ignoring operation attributes for schema generation via the config file: `schema_ignore: true`.
- Enable custom implementation of API operations by overriding `PerformRead/Create/Update/Delete` functions.
- Autogen resources now embed `RSAutogen`, which provides the default implementation for API operations (maintaining existing behavior).
- Added support for `id_attributes` in the config file. This is necessary when the read path does not contain all the resource ID attributes (as is the case for `org_service_account_secrets`).
- Generated the `org_service_account_secrets` resource
  - Implemented the custom `PerformRead` operation which uses the [Org Service Account](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-getorgserviceaccount) API and searches the `secrets` array by id (see `orgserviceaccountsecretapi/custom_operations.go`).
  - Added a basic acc test.

As this is a POC, it is missing tests for the new autogen functionality.

Alternative version: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3933

Link to any related issue(s): WRITING-34227

## Type of change:

## Required Checklist:


tools/codegen/config.yml
```diff
@@ -761,6 +761,25 @@ resources:
         roles:
           type: set
 
+  org_service_account_secret_api:
+    read:
+      path: /api/atlas/v2/orgs/{orgId}/serviceAccounts/{clientId}
+      schema_ignore: true
```
manupedrozo, line 767: We don't want `org_service_account` attributes in the `org_service_account_secret` schema.

tools/codegen/config.yml
```diff
@@ -761,6 +761,25 @@ resources:
...
+      method: GET
+    create:
+      path: /api/atlas/v2/orgs/{orgId}/serviceAccounts/{clientId}/secrets
+      method: POST
+    delete:
+      path: /api/atlas/v2/orgs/{orgId}/serviceAccounts/{clientId}/secrets/{secretId}
+      method: DELETE
+    id_attributes: ["id", "org_id", "client_id"]
```
manupedrozo, line 775: Since the read operation path does not contain all the id attributes, we provide them manually.

internal/serviceapi/orgserviceaccountsecretapi/custom_operations.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/autogen"
+)
+
+type response struct {
+	Secrets []map[string]any `tfsdk:"secrets"`
+}
+
+func (r *rs) PerformRead(ctx context.Context, req *autogen.HandleReadReq) ([]byte, *http.Response, error) {
```
manupedrozo, line 16: Custom read implementation for `org_service_account_secret`.

internal/common/autogen/resource_base.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/config"
+)
+
+var _ ResourceAPIOperations = &RSAutogen{}
+
+// RSAutogen is used as an embedded struct for all autogen resources.
+type RSAutogen struct {
```
manupedrozo, line 14: We could have the default implementations at the RSCommon level but would rather add this layer to keep autogen-only stuff separated.

internal/common/autogen/resource_base.go
```diff
@@ -0,0 +1,60 @@
...
+	config.RSCommon
+}
+
+// ResourceAPIOperations defines the interface for API operations in autogen resources.
+// RSAutogen provides default implementations which can be overridden by individual resources.
+type ResourceAPIOperations interface {
+	PerformRead(ctx context.Context, req *HandleReadReq) ([]byte, *http.Response, error)
+	PerformCreate(ctx context.Context, req *HandleCreateReq) ([]byte, *http.Response, error)
+	PerformUpdate(ctx context.Context, req *HandleUpdateReq) ([]byte, *http.Response, error)
+	PerformDelete(ctx context.Context, req *HandleDeleteReq) error
```
ciprian-tibulca, line 24: [nit] `Perform` prefix is redundant + golang prefers simple verbs. If you drop the code is still self-explanatory: `...Operations.Delete`

manupedrozo: ~Agree, I'll remove the prefix.~ The names collide with the resource operations without the prefix.

internal/common/autogen/handle_operations.go
```diff
@@ -72,12 +69,12 @@ func HandleCreate(ctx context.Context, req HandleCreateReq) {
 		addError(d, opCreate, errResolvingResponse, err)
 		return
 	}
-	errWait := handleWaitCreateUpdate(ctx, req.Wait, req.Client, req.Plan)
+	errWait := handleWaitCreateUpdate(ctx, req.Wait, req.APIOperations, req.Client, req.Plan)
```
ciprian-tibulca, line 72: I see all `handleWait...` functions expect a full ResourceAPIOperations implementation (req.APIOperations) while the deep code (`refreshFunc`) only needs the PerformRead. 
It is preferable to depend on a minimal contract so the "deep code" is not coupled to all operations but only to the one it needs. This makes the code easier to test and it's intent is clear (that param will be used to read the resource, not read or create/update/delete)
I'd suggest having a `type ResourceAPIReader interface {` to be used here

internal/serviceapi/orgserviceaccountsecretapi/custom_operations.go
```diff
@@ -0,0 +1,47 @@
...
+	bodyResp, apiResp, err := autogen.CallAPIWithoutBody(ctx, req.Client, req.CallParams)
+	if err != nil {
+		tflog.Error(ctx, "Failed to read service account secret", map[string]any{
+			"error": err.Error(),
+		})
+		return nil, apiResp, err
+	}
+	if autogen.NotFound(bodyResp, apiResp) {
+		return bodyResp, apiResp, nil
+	}
```
ciprian-tibulca, line 22: if SA is not found, I assume `err` is set, right?
If this is tru, then the notfound block should be nested inside the err handling
```
	if err != nil {
	    if autogen.NotFound(bodyResp, apiResp) {
		    return bodyResp, apiResp, nil // I think we should return the err here
	    }
		tflog.Error(ctx, "Failed to read service account secret", map[string]any{
			"error": err.Error(),
		})
		return nil, apiResp, err
	}
```

internal/serviceapi/orgserviceaccountsecretapi/custom_operations.go
```diff
@@ -0,0 +1,47 @@
...
+
+	bodyResp, apiResp, err := autogen.CallAPIWithoutBody(ctx, req.Client, req.CallParams)
+	if err != nil {
+		tflog.Error(ctx, "Failed to read service account secret", map[string]any{
+			"error": err.Error(),
+		})
+		return nil, apiResp, err
+	}
+	if autogen.NotFound(bodyResp, apiResp) {
+		return bodyResp, apiResp, nil
```
ciprian-tibulca: Why not returning an err, the parent entity is not found, which is unexpected outcome?

internal/common/autogen/handle_operations.go
manupedrozo: I didn't see it as particularly harmful here to just pass around the APIOperations but yes we can have an interface per operation for these deeper calls.

internal/serviceapi/orgserviceaccountsecretapi/custom_operations.go
manupedrozo: Good catch, forgot to delete this check. We don't need this check at all since the not-found case falls under the err block above.
It should be:
```go
bodyResp, apiResp, err := autogen.CallAPIWithoutBody(ctx, req.Client, req.CallParams)
if err != nil {
	return nil, apiResp, fmt.Errorf("failed to read service account secret: %w", err)
}
```
The NotFound check on the `apiResp` is already handled by the caller.

internal/serviceapi/orgserviceaccountsecretapi/custom_operations.go
manupedrozo: Similar but we don't need the check, see previous comment. Ty!

PR #3930, opened 2025-11-26
chore: Supports aliases in the ImportState method implementation of auto-generated resources

oarbusi:
## Description

Supports aliases in the ImportState method implementation of auto-generated resources.

Link to any related issue(s): CLOUDP-361981

## Type of change:

## Required Checklist:


tools/codegen/gofilegen/resource/resource_file.go
```diff
@@ -85,11 +90,30 @@ func getPathParams(s string) []codetemplate.Param {
...
-func getIDAttributes(readPath string) []string {
+func getIDAttributes(readPath string, attributes codespec.Attributes) []string {
 	params := getPathParams(readPath)
 	result := make([]string, len(params))
+
 	for i, param := range params {
-		result[i] = stringcase.ToSnakeCase(param.CamelCaseName)
+		snakeCaseName := stringcase.ToSnakeCase(param.CamelCaseName)
+
+		// Find the matching attribute in the schema (which has aliases already applied)
```
manupedrozo: I think this won't work if the path param is aliased to a different name than the one in the TF model. 
It would be a rare case though as that's what we use these path params for.

tools/codegen/gofilegen/resource/resource_file.go
```diff
@@ -85,11 +90,30 @@ func getPathParams(s string) []codetemplate.Param {
...
+		// Path params are marked as Required with OmitAlways req body usage
+		found := false
+		for j := range attributes {
+			attr := &attributes[j]
+			if attr.ReqBodyUsage == codespec.OmitAlways && attr.ComputedOptionalRequired == codespec.Required {
+				// Check if this attribute's TFModelName matches the param name
+				if stringcase.ToSnakeCase(attr.TFModelName) == snakeCaseName {
```
manupedrozo: Can use PascalCase instead of snake_case here and avoid all the case conversions:
```
    if  attr.TFModelName == param.PascalCaseName {
```

tools/codegen/gofilegen/resource/resource_file.go
```diff
@@ -85,11 +90,30 @@ func getPathParams(s string) []codetemplate.Param {
...
+					result[i] = attr.TFSchemaName // Use the aliased TF schema name
+					found = true
+					break
+				}
+			}
+		}
+		if !found {
+			result[i] = snakeCaseName // Fallback to snake_case if not found
```
manupedrozo: Since this shouldn't happen, I think we should explicitly error the whole generation here. 
Would help quickly catch if we provide a path param alias that does not match any TF model field, which is always wrong.

tools/codegen/gofilegen/resource/testdata/path-params-with-aliases.golden.go
```diff
@@ -0,0 +1,149 @@
+// Code generated by terraform-provider-mongodbatlas using `make generate-resource`. DO NOT EDIT.
```
AgustinBettati, line 1: Feel free to merge this fix directly into master (I see it currently uses dev branch as base).
Additional would consider already using the alias against an existing autogen resource to showcase how the changes are working (e.g. applying alias to `project_settings_api` resource).

tools/codegen/gofilegen/resource/resource_file.go
```diff
@@ -85,11 +90,30 @@ func getPathParams(s string) []codetemplate.Param {
 	return params
 }
 
-func getIDAttributes(readPath string) []string {
+func getIDAttributes(readPath string, attributes codespec.Attributes) []string {
```
AgustinBettati, line 93: q: Instead of handling this logic during code generation, could we handle all alias management within model generation? Was suspecting a fix within https://github.com/mongodb/terraform-provider-mongodbatlas/blob/65b56ee6506127942a1b92868b1c15346326feca/tools/codegen/codespec/config.go#L20 but might be missing some details.

oarbusi: the problem of modifying applyAliasToPathParams is that then in the actual API calls we do in the resource, we would be generating: 
```
	pathParams := map[string]string{
		"projectId": m.GroupId.ValueString(),
		"id":      m.Id.ValueString(),
	}
```
and this would make the API calls wrong

tools/codegen/gofilegen/resource/testdata/path-params-with-aliases.golden.go
oarbusi: I have used this to generate the new PBLE resource and want this change in the dev branch for the following PR that will include the new resource. Will merge this into the dev branch to keep it simpler

tools/codegen/gofilegen/resource/resource_file.go
oarbusi: this should never happen right? when we define an alias we want all the occurances of that name to be changed to the same, so both path param and TF model should have the same name

tools/codegen/gofilegen/resource/resource_file.go
oarbusi: this is important for the non aliased id attributes, without it, we would be generating `idAttributes := []string{"project_id", ""}` instead of `idAttributes := []string{"project_id", "id"}`

tools/codegen/gofilegen/resource/resource_file.go
oarbusi: changed in [8cc5dae](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3930/commits/8cc5dae248b473ff41e0e6b7616602ceb289e4d7)

tools/codegen/gofilegen/resource/resource_file.go
manupedrozo: Shouldn't the non aliased attributes have a corresponding TF model field? In your example, `id` should be in the model as `Id`. 
I think that the `codespec.OmitAlways` check in the loop is skipping attributes that are both in the read path and sent in the create/update body.
This comment: `Path params are marked as Required with OmitAlways req body usage` is not always true.

tools/codegen/gofilegen/resource/resource_file.go
manupedrozo: I think we'll need to revisit aliasing in the future to make it more robust (and ideally move it to model generation) but would mark it as out of scope for this effort.

The current system mixes path param aliases and schema aliases which are used for completely different things:
- Path param aliases wire the path param name to a TF model field name so that the generated code works.
- Schema aliases are made to improve the UX.

tools/codegen/gofilegen/resource/resource_file.go
oarbusi: removing this makes it so the idAttributes in the import are wrong, see for example projectapi/resource.go
Screenshot 2025-11-26 at 16 02 21

> I think that the codespec.OmitAlways check in the loop is skipping attributes that are both in the read path and sent in the create/update body.

do we have an API where this is the case? seems like bad API design, usually in path params we have ids that would not be sent in the bodys of POST/PATCH

tools/codegen/gofilegen/resource/resource_file.go
AgustinBettati: > and this would make the API calls wrong

Why exactly? If the url is modified to "/api/atlas/v2/groups/**{projectId}**/logIntegrations" the request would be made correctly.

One additional thing we can revise is what format we expect in the `config.yml`, because in current branch we would have a mix of:

- `processorName: name` # renaming path param 
- `group_id: project_id` # for renaming schema attributes

This change in behaviour depending on formating can be confusing, thinking we can unify on a single string case and potentially have different entries `path_param_aliases` `schema_aliases` if necessary. cc @manupedrozo

oarbusi: yes if you modify the url it would work but IMO you are then using a path in the code different from the publicly documented API which can cause confusion when looking at the code. No strong opinion though

tools/codegen/gofilegen/resource/resource_file.go
manupedrozo: `Id` in projectapi is not required, that's why the check skips it.
I would remove the `attr.ReqBodyUsage == codespec.OmitAlways && attr.ComputedOptionalRequired == codespec.Required` check altogether.

My idea here is that if we don't have a TF Model attribute that matches the path param name at this stage, we know that the resource will not work, so we should error with a clear message.

> do we have an API where this is the case? seems like bad API design, usually in path params we have ids that would not be sent in the bodys of POST/PATCH

A good example is `username` in database_user_api, which is provided in the create body and used in all other operations as a path param.

tools/codegen/gofilegen/resource/resource_file.go
manupedrozo: We are already modifying the url, no?

+1 on the different entries @AgustinBettati, each using its own string case.

tools/codegen/gofilegen/resource/resource_file.go
AgustinBettati: > I would remove the attr.ReqBodyUsage == codespec.OmitAlways && attr.ComputedOptionalRequired == codespec.Required

I tend to lean on this direction, dont see much value in the additional check. Essentially we go over each path param and check in the schema if the associated attribute has a particular alias.

oarbusi: Done your suggestions in the last few commits

tools/codegen/gofilegen/resource/resource_file.go
oarbusi: we are using the path provided in config.yml directly as far as I can see [here](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/d0bd655dd1cc9fc6ee81c230ade34a2052505143/tools/codegen/gofilegen/codetemplate/resource-file.go.tmpl#L57)

manupedrozo: These are modified here: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/tools/codegen/codespec/config.go#L119

tools/codegen/gofilegen/resource/resource_file.go
manupedrozo: Thanks! Would be nice to have the error if not found as well.

manupedrozo (approved):

oarbusi:
Merging this for now, but will be doing some follow up improvements to aliasing in another PR

tools/codegen/gofilegen/resource/testdata/path-params-with-aliases.golden.go
AgustinBettati: Ok, just to consider we will have multiple changes (mainly data source generation and code customization support) going into master in next weeks. We will also be leveraging autogen for other service account resources. Whichever improvements we can do directly in master (e.g. improving alias management) the better, you can always rebase/merge pble dev branch with latest changes from master when needed.

oarbusi: will cherry-pick this change and do the followup improvements of aliasing in a PR that will go to master

AgustinBettati: sounds good, thanks!

PR #3934, opened 2025-11-28
chore: Improvements for alias in autogeneration 

oarbusi:
## Summary

This PR improves the alias handling mechanism in the code generation pipeline to ensure consistent and correct application of aliases across attributes and API path parameters. The changes fix how aliases are looked up and applied, ensuring that both Terraform schema names and API path parameters are correctly transformed when aliases are defined in the configuration.

## Background

The code generation system supports aliases to rename attributes when the API model name doesn't match the desired Terraform schema name. For example, when an API uses `GroupId` but we want the Terraform resource to use `ProjectId`, aliases allow this mapping. Previously, the alias lookup mechanism had inconsistencies that could lead to incorrect transformations.

This PR builds upon PR #3930, which added support for aliases in the `ImportState` method. Together, these changes ensure that aliases work consistently across the entire code generation pipeline, from schema definition through to import functionality.

Link to any related issue(s):

## Type of change:

## Required Checklist:


tools/codegen/config.yml
```diff
@@ -64,7 +64,7 @@ resources:
     schema:
       ignores: ["links", "mongo_db_employee_access_grant.links"]
       aliases:
-        clusterName: name # path param name does not match the API request property
```
AgustinBettati, line 67: very nit: Is there a reason we want the config to define values with `PascalCase` vs current `camelCase`

tools/codegen/config.yml
```diff
@@ -266,14 +266,14 @@ resources:
...
-        target_states: [ "DELETED" ] # DELETED is a special state value when API returns 404 or empty object
+        target_states: ["DELETED"] # DELETED is a special state value when API returns 404 or empty object
     move_state:
       source_resources: ["mongodbatlas_cluster_old_api"]
     version_header: application/vnd.atlas.2024-08-05+json
     schema:
-      ignores: [ "links", "mongo_db_employee_access_grant.links" ]
+      ignores: ["links", "mongo_db_employee_access_grant.links"]
       aliases:
-        clusterName: name # path param name does not match the API request property
```
AgustinBettati, line 276: As this PR is going against master we dont have an example of how GroupId -> ProjectId is defined and how it would work. Can we apply the alias override to one of the existing resources as part of this PR (any should work, e.g. project_settings_api)?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -479,7 +479,7 @@ func TestConvertToProviderSpec_nested_schemaOverrides(t *testing.T) {
 									Attributes: codespec.Attributes{
 										{
 											TFSchemaName:             "inner_num_attr_alias",
-											TFModelName:              "InnerNumAttr",
```
AgustinBettati, line 482: We can speak with @manupedrozo, but quite certain we currently want to preserve TFModelName with the name of the API (not the alias). This is the only way marshalling/unmarshalling can know a API property maps to an aliases tf attribute

tools/codegen/codespec/testdata/config-nested-schema-overrides.yml
```diff
@@ -15,10 +15,9 @@ resources:
     version_header: application/vnd.atlas.2035-01-01+json
     schema:
       aliases:
-        groupId: projectId
-        group_id: project_id
-        nested_list_array_attr.inner_num_attr: inner_num_attr_alias
-        outer_object.nested_level1.level_field1: level_field1_alias
```
AgustinBettati, line 21: Are we losing possibility of applying alias to a specific nested attribute?

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
manupedrozo: Yes, that is correct. Until (if) we add an additional struct tag for marshal/unmarshal mapping, we need the struct fields to preserve the API json property name (but capitalized).
Relevant lines:
- https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/common/autogen/unmarshal.go#L55-L56
- https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/common/autogen/marshal.go#L74

tools/codegen/config.yml
oarbusi: changed in [1c50883](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/1c508839c368d18b0e7a33d5e6e785d2a66c8909)

tools/codegen/config.yml
oarbusi: added an alias example in [df7b406](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/df7b4065a898d7668ff3c59d29a1769d52d99ec6)

tools/codegen/codespec/testdata/config-nested-schema-overrides.yml
oarbusi: changed in [113b0b9](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/113b0b9ab60ad4c867555aa6ec1538c38ad805fd). Please review again

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
oarbusi: Changed to preserve TFModelName to avoid issues with marshalling/unmarshalling

internal/common/autogen/stringcase/string_case.go
```diff
@@ -46,6 +46,30 @@ func ToSnakeCase(str string) string {
 	return builder.String()
 }
 
+// SnakeToCamel converts snake_case to camelCase
+// Example: "inner_num_attr" -> "innerNumAttr"
+func SnakeToCamel(str string) string {
```
manupedrozo: Careful here! 
We removed the snake_case to camelCase conversion from all autogen code because it causes mismatches with API properties due to the conversion between camelCase and snake_case being irreversible (more context in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3740 and https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3840).

A good example is the API field `MongoDBMajorVersion`:
  - To snake case (with the custom conversion in this file): `MongoDBMajorVersion` -> `mongo_db_major_version`
  - Back to camel case: `mongo_db_major_version` -> `MongoDbMajorVersion`
  - Mismatch:`MongoDbMajorVersion` != `MongoDBMajorVersion`

If we need to convert, try to always convert the value we have in camel (or pascal) case into snake case and not the other way around.

oarbusi: You are right that this could happen, but only if we have a nested attribute(that we want to alias) under mongo_db_major_version(or any other attribute with multiple capital letters per word):
- attrPathName = "mongo_db_major_version.some_child"
- buildAliasPathKey converts parent: "mongoDbMajorVersion.someChild" 
- But config alias would be: "MongoDBMajorVersion.someChild" (correct API casing)
- Alias lookup fails

As of now, we don't have any aliasing case that would encounter this issue, but I'll see what I can do to avoid this from happening

manupedrozo: Yes! exactly, I would rather not reintroduce this issue.

The good thing is that we never need the snake case to camel case conversion: the API uses camel case so we can always access the original camel case value of a property name (might require a little refactoring).
If we find ourselves needing to convert from snake case to camel case, then we did something wrong along the way.

Lmk if you want to sync offline on this particular case, happy to help.

oarbusi: No longer using SnakeToCamel. Changed in [d74d81b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/d74d81bea3ccc630481683d590d4e7bc930b3bb4), please let me know what you think on the approach, also added a test that covers this case. @AgustinBettati @manupedrozo

tools/codegen/codespec/config.go
```diff
@@ -23,51 +25,62 @@ func applyTransformationsWithConfigOpts(resourceConfig *config.Resource, resourc
...
+type AttributeTransformation func(attr *Attribute, paths *attrPaths, schemaOptions config.SchemaOptions) error
 
 var transformations = []AttributeTransformation{
 	aliasTransformation,
 	overridesTransformation,
 	createOnlyTransformation,
 }
 
-func applyAttributeTransformations(schemaOptions config.SchemaOptions, attributes *Attributes, parentName string) error {
+func applyAttributeTransformations(schemaOptions config.SchemaOptions, attributes *Attributes, parentSchemaPath, parentAPIPath string) error {
```
manupedrozo: Can take attrPaths here instead of schema and api separately, ok either way.

internal/common/autogen/marshal.go
```diff
@@ -21,6 +21,7 @@ const (
 	tagValOmitJSON            = "omitjson"
 	tagValOmitJSONUpdate      = "omitjsonupdate"
 	tagValIncludeNullOnUpdate = "includenullonupdate"
+	tagAPINamePrefix          = "apiname:" // e.g., autogen:"apiname:groupId" means JSON field is "groupId"
```
manupedrozo: Nice addition! 
- I'd consider having this as a separate tag (not within the `autogen` tag). It would make it simpler to check it via Tag.Get("apiname") as well.
- We may also want to always include it in the model.

tools/codegen/config/config_model.go
```diff
@@ -36,7 +36,7 @@ type MoveState struct {
 }
 
 type SchemaOptions struct {
-	Aliases   map[string]string   `yaml:"aliases"` // only supports modifying path param names, full alias support is not yet implemented in conversion logic for request/response bodies
+	Aliases   map[string]string   `yaml:"aliases"` // keys and values use camelCase (e.g., groupId: projectId), only supports modifying path param names, full alias support is not yet implemented in conversion logic for request/response bodies
```
manupedrozo: `full alias support is not yet implemented in conversion logic for request/response bodies` is this still true?

internal/common/autogen/unmarshal.go
```diff
@@ -39,31 +39,63 @@ func unmarshalAttrs(objJSON map[string]any, model any) error {
 	if valModel.Kind() != reflect.Struct {
 		panic("model must be pointer to struct")
 	}
+	// Build a map from API name (JSON field name) to struct field index
+	apiNameToFieldIdx := buildAPINameToFieldMap(valModel.Type())
```
manupedrozo: We should be able to avoid this mapping by traversing the TFModel fields instead of the JSON fields.
- So, do a lookup into the json properties based on the model fields instead of looking for a target model field based on the json property.

No need to do now but would leave a comment.

internal/serviceapi/projectsettingsapi/resource_schema.go
```diff
@@ -64,7 +64,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 }
 
 type TFModel struct {
-	GroupId                                         types.String `tfsdk:"group_id" autogen:"omitjson"`
```
AgustinBettati, line 67: q: Is `apiname:groupId` extension needed here? Or altogether adding `apiname` as part of this PR? 
From changes I see we now apply the alias `groupId: projectId` to the paths and ModelTFName so it is working correctly ([reference](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/files#diff-eafb506e0429411e736781c09f3ca96fe8ca5c8b6818042afce130a5944eee6bL85-L89)).

manupedrozo (approved):

tools/codegen/codespec/config.go
oarbusi: done in [574ef4c](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/574ef4cc65d01cf2a11a105bc0243b5f39956c7e)

internal/common/autogen/marshal.go
oarbusi: good point, changed so that apiname is an independent tag [a8a2d1e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/a8a2d1eca77ccf7bb13824da87e6c9b01ba0bbb1)

tools/codegen/config/config_model.go
oarbusi: good catch, changed the comment in [563dd91](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/563dd91c0395318fac82c8e3d85f07cbf32fee38)

internal/common/autogen/unmarshal.go
oarbusi: took a stab at improving it in [4eaf577](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/4eaf577a4f0367e358e3f3d40e8543c22a4eb995)

internal/serviceapi/projectsettingsapi/resource_schema.go
oarbusi: not strictly necessary but thought that it made sense to have it here sense I am improving the aliasing in general

internal/common/autogen/marshal.go
```diff
@@ -21,6 +21,7 @@ const (
 	tagValOmitJSON            = "omitjson"
 	tagValOmitJSONUpdate      = "omitjsonupdate"
 	tagValIncludeNullOnUpdate = "includenullonupdate"
+	tagAPIName                = "apiname" // e.g., apiname:"groupId" means JSON field is "groupId"
```
AgustinBettati: copying from comment in schema generation code, would adjust comment to mention `used if the API name is different from the uncapitalized model name`

tools/codegen/config.yml
```diff
@@ -461,6 +461,8 @@ resources:
     version_header: application/vnd.atlas.2023-01-01+json
     schema:
       ignores: ["links"]
+      aliases:
+        username: dbUser
```
AgustinBettati:
```suggestion
        username: dbUser # alias is only added to illustrate full aliasing support (username is both a path param and present in request body)
```
Can be helpful for future reference

tools/codegen/config/config_model.go
```diff
@@ -36,7 +36,7 @@ type MoveState struct {
 }
 
 type SchemaOptions struct {
-	Aliases   map[string]string   `yaml:"aliases"` // only supports modifying path param names, full alias support is not yet implemented in conversion logic for request/response bodies
+	Aliases   map[string]string   `yaml:"aliases"` // keys and values use camelCase (e.g., groupId: projectId, nestedObject.innerAttr: renamedAttr). Supports path params and request/response body fields via APIName preservation and apiname tag generation
```
AgustinBettati, line 39: nice!

AgustinBettati (approved):
Nice work!

internal/common/autogen/marshal.go
oarbusi: good point, added your suggestion in [5f9c781](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3934/commits/5f9c7814b17dfaa81baa773e1b4890f3c19cfad1)

PR #3969, opened 2025-12-09
feat: Adds tier attribute to stream processor resource and datasource

kpatel71716:
## Description

CLOUDP-359071

This PR does 2 things:

1. Switch the admin api for starting an Atlas Stream Processor to use the stream processor `startWith` endpoint instead of the `start` endpoint. This does not reduce any functionality for the user but allows us to do the next point.
2. Add a `tier` attribute for the stremprocessor resource and datasource. This attribute can be used to start a stream processor on a given tier rather than forcing the processor to default to the workspace setting

## Type of change:

## Required Checklist:


QuiqueSanMongo (approved):
LGTM!

internal/service/streamprocessor/resource.go
```diff
@@ -107,13 +107,12 @@ func (r *streamProcessorRS) Create(ctx context.Context, req resource.CreateReque
...
-				TenantName:    workspaceOrInstanceName,
-				ProcessorName: processorName,
-			},
-		).Execute()
+		startWithOptions := &admin.StreamsStartStreamProcessorWith{}
+		if plan.Tier.ValueString() != "" {
+			startWithOptions.SetTier(plan.Tier.ValueString())
+		}
+
+		_, err := connV2.StreamsApi.StartStreamProcessorWith(ctx, projectID, workspaceOrInstanceName, processorName, startWithOptions).Execute()
```
lantoli, line 115: Create function is getting very long, consider extracting some parts into functions

.changelog/3969.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:enhancement
+resource/mongodbatlas_stream_processor: Adds the `tier` attribute
+```
+
+```release-note:enhancement
+data-source/mongodbatlas_stream_processor: Adds the `tier` attribute
```
lantoli, line 6: changelog for plural datasource is missing

internal/service/streamprocessor/resource_schema.go
```diff
@@ -95,6 +95,14 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "The stats associated with the stream processor. Refer to the [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/atlas-stream-processing/manage-stream-processor/#view-statistics-of-a-stream-processor) for more information.",
 			},
+			"tier": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
```
lantoli, line 100: is it Computed, because Atlas will return the value if not provided by client?

internal/service/streamprocessor/resource_schema.go
```diff
@@ -95,6 +95,14 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+				MarkdownDescription: "Selected tier for the Stream Processor. Configures Memory / VCPU allowances. Valid options are SP2, SP5, SP10, SP30, and SP50.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
lantoli: q: why are we using UseStateForUnknown?

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +53,33 @@ func TestAccStreamProcessor_basic(t *testing.T) {
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroyStreamProcessor,
+		Steps: []resource.TestStep{
+			{
+				Config: configWithTier(t, projectID, workspaceName, processorName, tier),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "project_id", projectID),
+					resource.TestCheckResourceAttr(resourceName, "workspace_name", workspaceName),
```
lantoli: nit: as the resource is in TPF, no need to check attributes that are provided in the Terraform configuration, Terraform gives you that guarantee. Only need to check for Computed values or Optional+Computed that are not provided (so they're like Computed)

internal/service/streamprocessor/resource.go
```diff
@@ -237,13 +236,12 @@ func (r *streamProcessorRS) Update(ctx context.Context, req resource.UpdateReque
...
 	if plannedState == StartedState {
-		_, err := r.Client.AtlasV2.StreamsApi.StartStreamProcessorWithParams(ctx,
-			&admin.StartStreamProcessorApiParams{
-				GroupId:       projectID,
-				TenantName:    workspaceOrInstanceName,
-				ProcessorName: processorName,
-			},
-		).Execute()
+		startWithOptions := &admin.StreamsStartStreamProcessorWith{}
+		if plan.Tier.ValueString() != "" {
```
lantoli, line 240: It looks like you're only changing Create, can the attribute be Updated?

If it can't, it should have the customplanmodifier.CreateOnly() plan modifier.   With the current implementation, I think that if customer tries to change the tier value, they will get an "inconsistency error" because the updated value won't match the initial value

docs/data-sources/stream_processor.md
```diff
@@ -140,6 +141,7 @@ output "stream_processors_results" {
 
 **NOTE** When a Stream Processor is updated without specifying the state, it is stopped and then restored to previous state upon update completion.
 - `stats` (String) The stats associated with the stream processor. Refer to the [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/atlas-stream-processing/manage-stream-processor/#view-statistics-of-a-stream-processor) for more information.
+- `tier` (String) Selected tier for the Stream Processor. Configures Memory / VCPU allowances. Valid options are SP2, SP5, SP10, SP30, and SP50.
```
lantoli: can tier change at some moment, e.g. Atlas decides to scale it up/down?   Or it'll always have the value that client used to create the processor?  (or alternatively the value decided by Atlas if not sent by client in  Create)

docs/resources/stream_processor.md
```diff
@@ -144,6 +145,7 @@ output "stream_processors_results" {
 - `state` (String) The state of the stream processor. Commonly occurring states are 'CREATED', 'STARTED', 'STOPPED' and 'FAILED'. Used to start or stop the Stream Processor. Valid values are `CREATED`, `STARTED` or `STOPPED`. When a Stream Processor is created without specifying the state, it will default to `CREATED` state. When a Stream Processor is updated without specifying the state, it will default to the Previous state. 
 
 **NOTE** When a Stream Processor is updated without specifying the state, it is stopped and then restored to previous state upon update completion.
+- `tier` (String) Selected tier for the Stream Processor. Configures Memory / VCPU allowances. Valid options are SP2, SP5, SP10, SP30, and SP50.
```
AgustinBettati: I see this has the same API docs description as `tier` in [stream workspace](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-creategroupstreamworkspace#operation-creategroupstreamworkspace-body-application-vnd-atlas-2023-02-01-json-streamconfig-tier). Would adding a phrase similar to `start a stream processor on a given tier rather than forcing the processor to default to the workspace setting` help in both API docs and terraform docs?

internal/service/streamprocessor/resource.go
AgustinBettati: > can the attribute be Updated?

If it can, would also consider how it behaves if a stream processor is configured as started vs stopped when modifying `tier`.

docs/data-sources/stream_processor.md
kpatel71716: As of today, the tier cannot change. It is always the value that the client has set.

It is worth noting that in the future, this could not be the case when we add autoscaling. When that happens, the idea is that the flag to allow autoscaling will not be allowed to be used with this defined tier field. But I am still discussing that with the team

internal/service/streamprocessor/resource_schema.go
kpatel71716: Yes that is correct. Today each stream processor has a tier set even without user input. And we have made changes to return the tier. But now the client can provide the tier to override that

internal/service/streamprocessor/resource_schema.go
kpatel71716: Since this is optional to provide, my thought was that if a user removes the `tier` from the config, we should not show a plan change. If I have unnecessarily used or missed `UseStateForUnknown`, I can remove it

For our apis, when a user stops/starts a stream processor, we always use the last tier that the stream processor was running on unless the user explicitly provides an option to override it i.e. we won't go back to the workspace default unless the user explicitly specifies to do so

internal/service/streamprocessor/resource_test.go
kpatel71716: Ah thank you for that knowledge. Simplified the tests

.changelog/3969.txt
kpatel71716: Thanks, added

docs/resources/stream_processor.md
kpatel71716: that's a good idea. Changed it here and I'll update this in the api docs as well

internal/service/streamprocessor/resource.go
kpatel71716: I'll create a follow-up ticket to explore this. I believe we have an existing action item on our end to go through this api flow but I'll reply back here

EDIT: Created https://jira.mongodb.org/browse/CLOUDP-371011

internal/service/streamprocessor/resource.go
kpatel71716: The attribute can be updated. I had a couple of checks that I updated and added tests for the same. I believe the logic is complete for Create and Update now

internal/service/streamprocessor/resource_schema.go
AgustinBettati: > my thought was that if a user removes the tier from the config, we should not show a plan change

Since the attribute is Optional + Computed this will be the default behaviour. When the attribute is removed no terraform plan is displayed, preserving the existing value in the state.

`UseStateForUnknown` is for a different scenario, primarily to reduce plan verbosity of `known after apply`, used only for computed attributes which value is guaranteed to not change when not defined. If this assumption is not held users can receive errors after apply operations `produced inconsistent result after apply`.  If you believe we can hold this requirement we can keep it, otherwise would simple suggest to remove.

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +53,67 @@ func TestAccStreamProcessor_basic(t *testing.T) {
 	resource.ParallelTest(t, *basicTestCase(t))
 }
 
+func TestAccStreamProcessor_withTier(t *testing.T) {
```
AgustinBettati, line 56: `TestAccStreamProcessor_withTier` and `TestAccStreamProcessor_tierUpdate` seem to be doing the same steps right? Would leave only `TestAccStreamProcessor_withTier`.

AgustinBettati (commented):
LGTM, just want to confirm on remove/preserving `UseStateForUnknown`

internal/service/streamprocessor/resource_schema.go
kpatel71716: Ah thank you for cleaning that up! We should be good to remove and just did so after some testing

kpatel71716:
@lantoli @AgustinBettati are there any more outstanding items on your end or does this look good to merge?

docs/data-sources/stream_processor.md
```diff
@@ -140,6 +141,7 @@ output "stream_processors_results" {
 
 **NOTE** When a Stream Processor is updated without specifying the state, it is stopped and then restored to previous state upon update completion.
 - `stats` (String) The stats associated with the stream processor. Refer to the [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/atlas-stream-processing/manage-stream-processor/#view-statistics-of-a-stream-processor) for more information.
+- `tier` (String) Selected tier to start a stream processor on rather than defaulting to the workspace setting. Configures Memory / VCPU allowances. Valid options are SP2, SP5, SP10, SP30, and SP50.
```
lantoli, line 144: nit: I think the description is ok and clear for the customers. One con is that if there are some new options at some moment, we need to remember to update this doc

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +53,35 @@ func TestAccStreamProcessor_basic(t *testing.T) {
...
+		tier                     = "SP30"
+	)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroyStreamProcessor,
+		Steps: []resource.TestStep{
+			{
+				Config: configWithTier(t, projectID, workspaceName, processorName, tier),
```
lantoli: nit: in 2nd step we use constnt "SP50", we can use here constant "SP30" instead of var to make it similar

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +53,35 @@ func TestAccStreamProcessor_basic(t *testing.T) {
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttrSet(resourceName, "id"),
```
lantoli, line 71: consider adding more checks in both steps

internal/service/streamprocessor/resource_test.go
```diff
@@ -673,6 +702,43 @@ func config(t *testing.T, projectID, workspaceName, processorName, state, nameSu
...
+		pipeline = jsonencode([
+			{ "$source" = { "connectionName" = data.mongodbatlas_stream_connection.sample_stream_solar.connection_name } },
+			{ "$emit" = { "connectionName" = "__testLog" } }
+		])
+		state = "STARTED"
+		tier  = %[4]q
+	}
+
+	data "mongodbatlas_stream_processor" "test" {
+		project_id     = %[1]q
```
lantoli: nit: I prefer to set the external values in the main resources and reference them from the data sources or auxiliar resources, e.g.:  		project_id     = mongodbatlas_stream_processor.processor.project_id, etc.

but feel free to leave as it is

lantoli (approved):

internal/service/streamprocessor/resource_test.go
kpatel71716: Good point, changed!

internal/service/streamprocessor/resource_test.go
kpatel71716: Good idea, done!

internal/service/streamprocessor/resource_test.go
kpatel71716: agreed, done!

docs/data-sources/stream_processor.md
kpatel71716: agreed, for now I left it since we have no intention of adding new tiers in the near future

kpatel71716:
merging as dev tests were passing previously and dev env is inconsistent at the moment (latest commit should have no effect). We will use qa tests to verify

PR #3970, opened 2025-12-09
chore: Add singular data source support to codespec generation

oarbusi:
## Description

Adds support for codespec generation to the use it to support singular data source generation

Link to any related issue(s): CLOUDP-363703

## Type of change:

## Required Checklist:


tools/codegen/codespec/api_to_provider_spec_mapper_test.go
```diff
@@ -42,11 +42,11 @@ var (
 	}
 
 	simpleTestResourceOperations = codespec.APIOperations{
-		Create: codespec.APIOperation{
+		Create: &codespec.APIOperation{
```
AgustinBettati, line 45: would add a unit test in which a config defines data sources configuration and we can assert the resulting model

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -56,10 +57,22 @@ func ToCodeSpecModel(atlasAdminAPISpecFilePath, configPath string, resourceName
 		if err != nil {
 			return nil, fmt.Errorf("unable to map to code spec model for %s: %w", name, err)
 		}
-		results = append(results, *resource)
+
+		// Generate DataSources only when datasources block is defined in config
+		if resourceConfig.DataSources != nil {
+			// TODO: validateDataSourceOperations(resourceConfig.DataSources) - schemaIgnore not supported, staticRequestBody not supported
```
AgustinBettati: what is the intention of this TODO? I do see we handle validation within `validateRequiredOperations`

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -348,3 +364,172 @@ func extractCommonParameters(paths *high.Paths, path string) ([]*high.Parameter,
...
+// mergeDataSourceAttributes merges path parameters with response attributes.
+// Path params are enforced as required; response attributes are marked computed.
+// Aliases are applied to both path params and response attributes during merge to properly detect duplicates.
+// If duplicates exist (same TFSchemaName after aliasing), Required always wins over Computed.
+func mergeDataSourceAttributes(pathParams, responseAttrs Attributes, aliases map[string]string) Attributes {
+	merged := make(map[string]*Attribute) // key by TFSchemaName
+
+	// Add path params as required (they identify the data source)
+	// Apply aliases to path params during merge
+	for i := range pathParams {
```
AgustinBettati: q: For all merging logic, can we reuse existing `addOrUpdate` logic we use for resource model generation in `merge_attributes.go`?
We could also move data source merge function into this file.

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -348,3 +364,172 @@ func extractCommonParameters(paths *high.Paths, path string) ([]*high.Parameter,
...
+	}
+
+	return ds, nil
+}
+
+// mergeDataSourceAttributes merges path parameters with response attributes.
+// Path params are enforced as required; response attributes are marked computed.
+// Aliases are applied to both path params and response attributes during merge to properly detect duplicates.
+// If duplicates exist (same TFSchemaName after aliasing), Required always wins over Computed.
+func mergeDataSourceAttributes(pathParams, responseAttrs Attributes, aliases map[string]string) Attributes {
```
AgustinBettati: We do have a step during defined in `dataSourceTransformations` which calls `aliasTransformation`, I like this as it reuses all logic with resource model generation. Is there a reason why this function is also handling aliasing? Would strongly favour handling alias in a single place after all merging of attributes is done (similar to resource model generation).

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -348,3 +364,172 @@ func extractCommonParameters(paths *high.Paths, path string) ([]*high.Parameter,
...
+	for _, attr := range merged {
+		result = append(result, *attr)
+	}
+
+	sortAttributesRecursive(&result)
+
+	return result
+}
+
+func sortAttributesRecursive(attrs *Attributes) {
```
AgustinBettati: Any reason why we need this dedicated sorting step? For reference for resource generation we dont have a dedicated sorting step, `buildResourceAttrs` does a simple sorting when iterating over schema properties.

tools/codegen/codespec/config.go
```diff
@@ -43,7 +89,21 @@ var transformations = []AttributeTransformation{
...
+	// Note: createOnlyTransformation is excluded for data sources (read-only, no create operation)
+}
+
 func applyAttributeTransformations(schemaOptions config.SchemaOptions, attributes *Attributes, parentPaths *attrPaths) error {
+	return applyAttributeTransformationsList(schemaOptions, attributes, parentPaths, transformations)
+}
+
+func applyDataSourceAttributeTransformations(schemaOptions config.SchemaOptions, attributes *Attributes, parentPaths *attrPaths) error {
+	return applyAttributeTransformationsList(schemaOptions, attributes, parentPaths, dataSourceTransformations)
+}
```
AgustinBettati, line 105: nit: we can remove this function and simply call `applyAttributeTransformations` right?

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -348,3 +364,172 @@ func extractCommonParameters(paths *high.Paths, path string) ([]*high.Parameter,
...
+		// Extract list operation for description
+		if oasListOp, err := extractOp(spec.Paths, dsConfig.List); err == nil && oasListOp != nil {
+			pluralDescription = &oasListOp.Description
+		}
+
+		listOp = &APIOperation{
+			HTTPMethod: dsConfig.List.Method,
+			Path:       dsConfig.List.Path, // alias will be applied later by transformations helper
+		}
+	}
```
AgustinBettati, line 428: So right now we only process path params of the read, but not of the list right? Would make clear in PR description what we support related to singular data source model and what is currently pending for the plural (knowing path params of list operation + query params). cc @maastha

AgustinBettati (approved):
Thanks for splitting this PR, helped a lot for review.
Overall LGTM, leaving some suggestions and questions for improvements.

tools/codegen/models/alert_configuration_api.yaml
```diff
@@ -630,5 +630,632 @@ operations:
...
+              api_name: alertConfigId
+              req_body_usage: omit_always
+              sensitive: false
+              create_only: false
+            - list_nested:
+                nested_object:
+                    attributes:
+                        - string: {}
+                          description: Uniform Resource Locator (URL) that points another API resource to which this response has some relationship. This URL often begins with `https://cloud.mongodb.com/api/atlas`.
+                          computed_optional_required: optional
```
AgustinBettati: spotted in other PR (https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3971#discussion_r2605964887), nested attributes are being marked as `Optional`/`Required` in the data source model

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: This is to implement codespec generation for singular data source only. Changed the title to say singular data source explicitly

tools/codegen/models/alert_configuration_api.yaml
oarbusi: Nice catch! working on it

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: I think moving data source merge functions to `merge_attributes.go`  makes sense, but I would keep separate merge functions, because data source logic is much simpler, and the `addOrUpdate` function is designed for resource-specific complexity (POST/PATCH/response merging) that doesn't apply to data sources

tools/codegen/models/alert_configuration_api.yaml
oarbusi: fixed in [f7f90de](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3970/commits/f7f90deb6131a352cc9ad5f187d8b87006556186)

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: refactored this in [ca348a7](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3970/commits/ca348a7c07321d161c847900a47b9d9b217d7862)

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: moved in [ca348a7](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3970/commits/ca348a7c07321d161c847900a47b9d9b217d7862)

tools/codegen/codespec/config.go
oarbusi: the transformations are different in applyAttributeTransformations and in applyDataSourceAttributeTransformations. We could refactor but I like that the method explicitly says data source

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: I did this differently to be able to handle the case where we want to alias a path param and a response attribute that are the same but have different names in the path and in the response. For example:
```
schema:
        aliases:
          alertConfigId: integrationId
          id: integrationId
```
The problem with aliasing AFTER merge:
- Path param: alertConfigId → TFSchemaName: alert_config_id (Required)
- Response attr: id → TFSchemaName: id (Computed)
- After merge: both alert_config_id and id exist
- aliasTransformation changes alert_config_id → integration_id and id → integration_id
Result: TWO attributes with TFSchemaName = integration_id ❌

Why aliasing DURING merge works:
- Path param: alertConfigId → alias to integration_id (Required)
- Response attr: id → integration_id (Computed)
- During merge: duplicate detected → Required wins
- **Result:** ONE integration_id attribute (Required) ✅

The `aliasTransformation` doesn't handle duplicate detection, it just renames. So for data sources that we want to alias like in the example above, we need to handle it differently. 

Let me know your thoughts, i wouldn't block this PR for this, so if we come up with a different solution, we can do it in a follow-up PR

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: I think this was not meant to be here. Removed, thanks

tools/codegen/codespec/api_to_provider_spec_mapper.go
AgustinBettati: Is this more advanced aliasing feature only supported for data sources? My main doubts comes from diff between resource and data source generation, and hence the difference it what aliasing they support. 
I would keep both consistent, we can create a CLOUDP to support this more advanced alias in both implementations and prioritise as we see the need.

tools/codegen/codespec/config.go
AgustinBettati: I see, makes sense

tools/codegen/codespec/api_to_provider_spec_mapper.go
oarbusi: This case only happens in the data source because we have the same field as a path param(required input) and as a response attribute(computed output), and there we want to align the name of both fields while avoiding duplicates. In the case of the resource, we only have it in the response attribute, so a computed field, and there's no possibility of duplicates. Hope it makes sense why we need it for the data source only

tools/codegen/models/org_service_account_secret_api.yaml
```diff
@@ -92,7 +92,6 @@ operations:
     read:
         http_method: GET
         path: /api/atlas/v2/orgs/{orgId}/serviceAccounts/{clientId}
-    update: null
```
maastha, line 95: can you also update the model for org service account so we know it works as expected?

tools/codegen/codespec/config_test.go
```diff
@@ -15,28 +15,28 @@ func TestApplyTimeoutTransformation(t *testing.T) {
 	}{
 		"No wait blocks - no timeout attribute added": {
 			inputOperations: codespec.APIOperations{
-				Create: codespec.APIOperation{},
-				Read:   codespec.APIOperation{},
+				Create: &codespec.APIOperation{},
+				Read:   &codespec.APIOperation{},
```
maastha, line 22: can you add a test here for aliasing as well for resource & data source?

oarbusi: added in [2a1af42](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3970/commits/2a1af42297d081a22ee8a280c2d6569e88cad8eb)

tools/codegen/codespec/api_to_provider_spec_mapper_test.go
oarbusi: added in [a9bfc3d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3970/commits/a9bfc3dda5fbeee4174a81fe1bc9414b2898f8e2)

tools/codegen/models/org_service_account_secret_api.yaml
oarbusi: it's up to date, there are no changes because in that resource there is update operation (update is not null), which is why this file changed

PR #3972, opened 2025-12-10
chore: Add required-create only implementation

marcabreracast:
## Description

This PR identifies "required create-only attributes (`RequestOnlyRequiredOnCreate `) based on a set of conditions, and applies a transformation step where if those conditions are met, the computability will be modified to `Optional`. 

Based on the set of conditions, new property called `PresentInAnyResponse ` will also be added, as it allows to track if a field is present on a request. 

Link to any related issue(s): [CLOUDP-363708](https://jira.mongodb.org/browse/CLOUDP-363708)

## Type of change:

## Required Checklist:

## Further comments
- The rest of the implementation of this functionality (adding new custom plan modifier will be done in [CLOUDP-363710](https://jira.mongodb.org/browse/CLOUDP-363710). 
- Tested that `org_service_account_api` resource transforms the attribute `secret_expires_after_hours` from `Required` to `Optional` ([resource_schema.go](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-363708_RequiredCreateOnly_Attribute_Implementation/internal/serviceapi/orgserviceaccountapi/resource_schema.go))


tools/codegen/codespec/merge_attributes.go
```diff
@@ -51,6 +51,7 @@ func addOrUpdate(merged map[string]*Attribute, newAttr *Attribute, reqBodyUsage
 	} else {
 		if isFromResponse {
 			newAttr.ComputedOptionalRequired = Computed // setting as computed as attribute was defined only in response
+			newAttr.PresentInResponse = true
```
marcosuma: I might miss some context, but it looks like your implementation considers PresentInResponse false by default and in certain conditions it sets it to true - like here and below. 

IMO PresentInResponse should be true by default and it should set to false only in the exceptional case that is not returned.

Follow-up question: what do we exactly mean with `PresentInResponse`? is it any response? is it only GET? I think the name is not self-explanatory

marcosuma: actually now that I am thinking the name should be negated, like `MissingInResponse` so that the default is false and you set it to true only when it's needed

tools/codegen/codespec/model.go
```diff
@@ -103,29 +103,31 @@ type Attributes []Attribute
...
+	CustomType                *CustomType              `yaml:"custom_type,omitempty"`
+	ComputedOptionalRequired  ComputedOptionalRequired `yaml:"computed_optional_required"`
+	TFSchemaName              string                   `yaml:"tf_schema_name"`
+	TFModelName               string                   `yaml:"tf_model_name"`
+	APIName                   string                   `yaml:"api_name,omitempty"` // original API property name (camelCase), used for apiname tag when different from Uncapitalize(TFModelName)
+	ReqBodyUsage              AttributeReqBodyUsage    `yaml:"req_body_usage"`
+	Sensitive                 bool                     `yaml:"sensitive"`
+	CreateOnly                bool                     `yaml:"create_only"` // leveraged for defining plan modifier which avoids updates on this attribute
+	PresentInResponse         bool                     `yaml:"present_in_response"`
+	RequiredOnCreateInputOnly bool                     `yaml:"required_on_create_input_only"`
```
marcosuma: trying to stay consistent with the language, I see that in other parts we use the Request and Response vs Input and Output. In addition, I think it's granted that something can only be required for a request, not a response.

Can we rename this with `RequiredOnCreateOnly`?

marcosuma:
Based on the PR content, I believe we need to update the description since it's also setting another new field, "PresentInResponse" (which I suggested to rename)

tools/codegen/codespec/merge_attributes.go
marcabreracast: For some context regarding the implementation, we currently process a set of sources: path params, request bodies and response bodies. Depending on the source, we will capture different metadata (e.g. `presentInResponse`). 
With this current implementation, where `PresentInResponse` is `false` by default, it's very simple to confirm if the attribute is present in a response. If it has the source, then we know it's from the response. 

If we inverted the default, we’d be assuming fields are in responses and would need additional tracking to distinguish “assumed present” from “actually observed”. This is also more complicated in terms of implementation.

Mainly, the benefit of having this approach is that for any attribute that is created, we can safely have the value in `false`, and we only put this value to `true` when we verify that it comes from a source response. 

I believe that also the naming of the attribute is appropriate, as it keeps track if the field is present in any response (from your previous question, this is not limited to GET requests). Having a negated attribute brings some of the same concerns that we had on the `nonUpdatable` field and the double negation potential situation. 

My proposal is to leave the implementation as it is, and leave the name on the attribute to avoid potential double negations when `!MissingInResponse` is encountered.

tools/codegen/codespec/model.go
marcabreracast: The only thing that is that the `InputOnly` part was added to make explicit that the attribute will not be present in the response. If we are ok with that assumption as you mention, I see an issue renaming it.

tools/codegen/codespec/merge_attributes.go
marcosuma: ack! I was of the assumption that it's actually right to assume it's going to be in the response. In fact, I always expect a field to be in the response if it's in the request, and the fact that it won't be there is rather the exception.

On the negated: I agree, but IMO the confusion of negated is when the field starts with `not...`

I am fine to go with `presentInResponse` in any case, but if we are happy with any response (meaning either POST, PATCH or GET response), should that be represented in the name? `presentInAnyResponse`? 
We basically need to think: in 6 months time (or a new joiner), will it be easy to understand what is this field representing?

tools/codegen/codespec/model.go
marcosuma: ah! fair point, but as you can see the name is confusing. Because you are correctly saying that this param is true if it's required on create AND it doesn't show up on the response.

Instead when I read `RequiredOnCreateInputOnly` I think of: it's a required field only on the create input. What about `RequestOnlyRequiredOnCreate`? cc @AgustinBettati

AgustinBettati: leaving a comment of the `why` we have this property can help a lot for future reference. Thinking something like:

> Flags API property which is only present in create request body as required. These properties are intentionally modified to optional attributes with a plan modifier to preserve creation validation but allow for omitting value on imports.

AgustinBettati (approved):
LGTM, agree with Marco suggestions on renaming

tools/codegen/codespec/model.go
AgustinBettati: 👍 ,`RequestOnlyRequiredOnCreate` captures more clearly the `not present in responses` aspect.

tools/codegen/codespec/merge_attributes.go
marcabreracast: Yeah I believe that's a good point. Changed it to `presentInAnyResponse` so it's easier to understand and addressed it in [fea3780](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3972/commits/fea3780a6f48ab0a1fc4b367a621c8d419d6bd37).

tools/codegen/codespec/model.go
marcabreracast: Sounds good! Thank you both for the opinions. Renamed the attribute to `RequestOnlyRequiredOnCreate` on [fea3780](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3972/commits/fea3780a6f48ab0a1fc4b367a621c8d419d6bd37)

marcabreracast: Addressed in [3419293](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3972/commits/34192937a9605d0bda7080409a85129e2f2a90c5)

tools/codegen/codespec/config.go
```diff
@@ -88,6 +88,7 @@ var transformations = []AttributeTransformation{
 	aliasTransformation,
 	overridesTransformation,
 	createOnlyTransformation,
+	requiredOnCreateInputOnlyTransformation,
```
marcosuma, line 91: @marcabreracast in a follow-up PR we should also change the name of this transformation, for consistency.

marcabreracast: Ah yes sorry, I missed on that when doing the refactor, will address that in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3977 since it's the follow-up ticket on the implementation (still WIP)

PR #3977, opened 2025-12-12
chore: Add custom plan modifier for RequestOnlyRequiredOnCreate Attribute

marcabreracast:
## Description

This PR addresses the following: 
1. Follow-up from PR #3972, where due to a renaming decision on attribute `requestOnlyRequiredOnCreate`, the transformation step function naming needs to match for consistency purposes. 
2. Addition of new custom plan modifier `RequestOnlyRequiredOnCreateModifier`, which adds validation when trying to create a resource with an attribute marked as `RequestOnlyRequiredOnCreate = true`, an error message will be shown. 
3. This implementation allows to perform `import` without having to apply an additional step, improving the user experience. 

Link to any related issue(s): [CLOUDP-363710](https://jira.mongodb.org/browse/CLOUDP-363710)

## Type of change:

## Required Checklist:

## Further comments

Behaviour of `CreateOnly` attribute is preserved, where the attribute is required on creation, but when tried to be updated in the configuration, a validation error message will be shown.

### DIFF
--- a/internal/common/customplanmodifier/request_only_required_create.go
+++ b/internal/common/customplanmodifier/request_only_required_create.go
@@ -1,0 +1,118 @@
+package customplanmodifier
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+// RequestOnlyRequiredOnCreate returns a plan modifier that fails planning if the value is
+// missing (null/unknown) during create (i.e., when state is null), but allows omission on read/import.
+func RequestOnlyRequiredOnCreate() RequestOnlyRequiredOnCreateModifier {
+	return &requestOnlyRequiredOnCreateAttributePlanModifier{}
+}
+
+// Single interface so the modifier can be applied to any attribute type.
+type RequestOnlyRequiredOnCreateModifier interface {
+	planmodifier.String
+	planmodifier.Bool
+	planmodifier.Int64
+	planmodifier.Float64
+	planmodifier.Number
+	planmodifier.List
+	planmodifier.Map
+	planmodifier.Set
+	planmodifier.Object
+}
+
+type requestOnlyRequiredOnCreateAttributePlanModifier struct{}
+
+func (m *requestOnlyRequiredOnCreateAttributePlanModifier) Description(ctx context.Context) string {
+	return m.MarkdownDescription(ctx)
+}
+
+func (m *requestOnlyRequiredOnCreateAttributePlanModifier) MarkdownDescription(ctx context.Context) string {
+	return "Ensures that create operations fail when attempting to create a resource with a missing required attribute."
+}
+
+func (m *requestOnlyRequiredOnCreateAttributePlanModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
+	if isCreate(&req.State) && (req.PlanValue.IsNull() || req.PlanValue.IsUnknown()) {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("%s is required when creating this resource", req.Path),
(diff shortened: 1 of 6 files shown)


tools/codegen/gofilegen/schema/schema_attribute.go
```diff
@@ -199,21 +198,27 @@ func commonProperties(attr *codespec.Attribute, planModifierType string) []CodeS
...
+	}
+	if attr.RequestOnlyRequiredOnCreate {
+		customPlanModifiers = append(customPlanModifiers, "customplanmodifier.RequestOnlyRequiredOnCreate()")
+	}
 
+	if len(customPlanModifiers) > 0 {
 		result = append(result, CodeStatement{
-			Code:    code,
-			Imports: planModifierImports,
+			Code: fmt.Sprintf("PlanModifiers: []%s{%s}", planModifierType, strings.Join(customPlanModifiers, ", ")),
```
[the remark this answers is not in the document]
marcabreracast: That package was imported 14 months ago in 0f8b8f5baa58291dbea717091cf02113a6cdde5c 😅

internal/common/customplanmodifier/request_only_required_create.go
```diff
@@ -0,0 +1,136 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+// RequestOnlyRequiredOnCreate returns a plan modifier that fails planning if the value is
+// missing (null/unknown) during create (i.e., when state is null), but allows omission on read/import.
```
maastha, line 14: I believe such a value can only be null/non-null but never unknown right? Since these would be Optional attributes? If yes, I'd remove the Known check in the validation method & only check for null

internal/common/customplanmodifier/request_only_required_create.go
```diff
@@ -0,0 +1,136 @@
...
+		isCreate(&req.State),
+		req.PlanValue,
+		req.Path,
+		&resp.Diagnostics,
+	)
+}
+
+// validateRequestOnlyRequiredOnCreate checks that an attribute has a known, non-null
+// value during create and adds an error if it does not.
+func validateRequestOnlyRequiredOnCreate(isCreate bool, planValue attr.Value, attrPath path.Path, diagnostics *diag.Diagnostics) {
```
maastha, line 125: should we also throw an error if the user tries to update it? Or is the expectation to always use it in conjunction with customplanmodifier.CreateOnly()?

How is it handled during resource reads? I assume the attribute would still stay in the state even though the GET calls don't return it right?

maastha (approved):

internal/common/customplanmodifier/request_only_required_create.go
marcabreracast: If the user tries to update it, `customplanmodifier.CreateOnly()` is responsible for blocking the change and surfacing an error. `RequestOnlyRequiredOnCreate` only enforces that the attribute is present on create, so for attributes that are both required-on-create and immutable we expect to always use it in conjunction with CreateOnly(). However, an attribute could be `CreateOnly`, without `RequestOnlyRequiredOnCreate`.

Regarding your second question: you're correct. On resource reads, the attribute stays in the state, so it retains its last known value.

internal/common/customplanmodifier/request_only_required_create.go
marcabreracast: That is true yes, addressed in [7b3a175](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/3977/commits/7b3a1751f3f745e73ca0fd04e1d5cbdafef6899a)

PR #4017, opened 2025-12-22
feat: Adds `port_mapping_enabled` attribute to `privatelink_endpoint` and `privatelink_endpoint_service`

filipcirtog:
## Description

This PR implements support for the new GCP port-mapped architecture, which uses port mapping to reduce resource provisioning. Unlike the legacy architecture that requires dedicated resources for each Atlas node, the new design uses a single set of resources to support up to 150 nodes through port mapping, enabling direct targeting of specific nodes using only one customer IP address.

## Changes Overview

Note: An in-place update has been implemented for the `mongodbatlas_privatelink_endpoint resource`, for the `port_mapping_enabled` attribute. This is because all other attributes force a ForceNew operation. The in-place update is designed to prevent updates to `port_mapping_enabled`, which would otherwise require destroying child resource and recreating it after. This process is not feasible, as the child resource recreation would fail due to the new architecture requiring different fields. The in-place update will only throw an error if a user attempts to update the `mongodbatlas_privatelink_endpoint`.

### New Attributes

1. **`port_mapping_enabled`** (Optional, Computed)
   - Added to `mongodbatlas_privatelink_endpoint` resource (Optional, settable)
   - Added to `mongodbatlas_privatelink_endpoint` data source (Computed, read-only)
   - Added to `mongodbatlas_privatelink_endpoint_service` resource (Computed, read-only)
   - Added to `mongodbatlas_privatelink_endpoint_service` data source (Computed, read-only)
   - When set to `true` on the endpoint resource, enables the new port-mapped architecture for GCP
   - Defaults to `false` (legacy architecture)

2. **`gcp_endpoint_status`** (Computed)
   - Added to `mongodbatlas_privatelink_endpoint_service` resource and data source
   - Status of the individual GCP endpoint
   - Only populated for the new port-mapped architecture (when `port_mapping_enabled = true`)
   - Returns one of: `INITIATING`, `AVAILABLE`, `FAILED`, `DELETING`

### Reused Attributes

1. **`private_endpoint_ip_address`** (Optional)
   - Previously used for Azure, Now also used for the new GCP port-mapped architecture
   - For new architecture: Required and should be the IP address of the forwarding rule
   - For legacy architecture: Not used
   - Conflicts with `endpoints` attribute

2. **`endpoint_service_id`** (Required)
   - For legacy architecture: Can be any identifier string
   - For new architecture: Should be the forwarding rule name
   
- **Updated documentation** for all affected resources and data sources

- **Added test coverage**:
  - Test case for explicitly enabled port mapping
  - Test case for explicitly disabled port mapping
  
Link to any related issue(s): [CLOUDP-363082](https://jira.mongodb.org/browse/CLOUDP-363082)

## Type of change:

## Required Checklist:


internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,81 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+			},
+		},
+	})
+}
+
+func TestAccNetworkRSPrivateLinkEndpointGCP_basic_with_new_architecture_explicitly_enabled(t *testing.T) {
+	var (
+		resourceName       = "mongodbatlas_privatelink_endpoint.test"
+		orgID              = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName        = "test-acc-tf-p-gcp-port-based-routing-feature-flag-enabled"
```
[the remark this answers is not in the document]
filipcirtog: Note: This will be updated to use the random name once the feature is officially released (GA) and not gated by a feature-flag (before merging to master)

lizo-mdb (approved):
LGTM

docs/resources/privatelink_endpoint.md
```diff
@@ -47,6 +47,7 @@ resource "mongodbatlas_privatelink_endpoint" "test" {
...
 * `timeouts`- (Optional) The duration of time to wait for Private Endpoint to be created or deleted. The timeout value is defined by a signed sequence of decimal numbers with a time unit suffix such as: `1h45m`, `300s`, `10m`, etc. The valid time units are:  `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. The default timeout for Private Endpoint create & delete is `1h`. Learn more about timeouts [here](https://www.terraform.io/plugin/sdkv2/resources/retries-and-customizable-timeouts).
 * `delete_on_create_timeout`- (Optional) Indicates whether to delete the resource being created if a timeout is reached when waiting for completion. When set to `true` and timeout occurs, it triggers the deletion and returns immediately without waiting for deletion to complete. When set to `false`, the timeout will not trigger resource deletion. If you suspect a transient error when the value is `true`, wait before retrying to allow resource deletion to finish. Default is `true`.
+* `port_mapping_enabled` - (Optional) Flag that indicates whether this endpoint service uses PSC port-mapping.
```
EspenAlbert: [nit] Is there any example we can update also? I struggle a bit to understand this feature seeing only this variable description

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,81 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(resourceName, "region"),
 					resource.TestCheckResourceAttr(resourceName, "provider_name", providerName),
 					resource.TestCheckResourceAttr(resourceName, "region", region),
+					resource.TestCheckResourceAttr(resourceName, "port_mapping_enabled", "false"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
```
EspenAlbert, line 96: nice to add the import step 👍

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,81 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+			{
+				Config: configWithPortMapping(orgID, projectName, providerName, region, portMappingEnabled),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
+					resource.TestCheckResourceAttrSet(resourceName, "provider_name"),
+					resource.TestCheckResourceAttrSet(resourceName, "region"),
+					resource.TestCheckResourceAttr(resourceName, "provider_name", providerName),
+					resource.TestCheckResourceAttr(resourceName, "region", region),
+					resource.TestCheckResourceAttr(resourceName, "port_mapping_enabled", "true"),
```
EspenAlbert: [nit] not sure if it is worth testing the attribute changes:
`null -> true`
`true -> false`

Also, does this value have a default? (What is returned by the API if the field is not set?)
What is the difference between false and null? Will `null` -> `false` and ForceNew leading to an unnecessary DELETE+CREATE plan?

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: it's just git displaying the diff wrong as that was already there but it just got pushed back to the end :)

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: 1. Attribute changes (null -> true, true -> false):
These transitions were not tested since I thought it is not needed because of ForceNew. I will add those.

2. Default value:
Atlas defaults this attribute to false. Should we do the same? 

3. null vs false:
I will check this as I do not have a response now (I personally expect yes).

docs/resources/privatelink_endpoint.md
filipcirtog: This is the description available on Atlas and unfortunately we are not able to update it. (we need DELETE + CREATE)

EspenAlbert: I just wanted to know if we have an example in repo_root/examples/* .tf that we could update?

oarbusi (approved):
LGTM

docs/resources/privatelink_endpoint.md
filipcirtog: This makes total sense. I will add an example for this.

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,103 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+				),
+			},
+			{
```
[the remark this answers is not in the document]
filipcirtog: test added according to the previous PR review

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,103 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
+					resource.TestCheckResourceAttrSet(resourceName, "provider_name"),
+					resource.TestCheckResourceAttrSet(resourceName, "region"),
+					resource.TestCheckResourceAttr(resourceName, "provider_name", providerName),
+					resource.TestCheckResourceAttr(resourceName, "region", region),
+					resource.TestCheckResourceAttr(resourceName, "port_mapping_enabled", "false"),
+				),
+			},
+			{
+				Config: configWithPortMapping(orgID, projectName, providerName, region, true),
```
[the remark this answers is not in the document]
filipcirtog: test added according to the previous PR review

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/versions.tf
```diff
@@ -0,0 +1,17 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "~> 3.0"
+    }
```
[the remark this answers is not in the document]
filipcirtog: yes indeed. changed thx.

docs/resources/privatelink_endpoint_service.md
```diff
@@ -144,20 +144,92 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
 
 ```
 
+## Example with GCP (Port-Based Architecture)
+
+The new PSC port-based architecture simplifies setup by requiring only 1 endpoint instead of 50. Enable it by setting `port_mapping_enabled = true` on the endpoint resource.
+
+**Important:** For the new port-based architecture, the `endpoint_service_id` must match the `endpoint_name` in the `endpoints` block. Although the new API ignores the `endpoint_service_id` value, it is still required by the Terraform provider.
```
manupedrozo: Is endpoint_service_id still required just because it was already marked as required in the schema and we don't want to change that or is there some other reason?
  - Did we consider changing it to optional?
  - I am also not sure why endpoint_service_id it must match endpoint_name if it is ignored.

Would not mention: "Although the new API ignores the `endpoint_service_id` value, it is still required by the Terraform provider."

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/README.md
```diff
@@ -0,0 +1,100 @@
+# Example with GCP with Port-Based architecture and MongoDB Atlas Private Endpoint
+
+This project demonstrates the **new PSC port-based architecture** for setting up GCP Private Service Connect with MongoDB Atlas, which requires only 1 endpoint.
```
EspenAlbert: [nit] what is PSC?

Seems like the benefits are a bit spread and duplicated across this doc.
Maybe I would start with the benefit to motivate the user instead of leaving it for the end

docs/resources/privatelink_endpoint_service.md
filipcirtog: Since the GET API call relies on endpoint_service_id in all cases, we don’t have a direct flag for this (similar to port_mapping_enabled on this resource, which is only computed in this specific case i.e., it’s defined on the endpoint_service, not on the endpoint).

The only viable approach appears to be:
- Not set → Indicates new architecture
- Set → Indicates old architecture

Would this approach work for you?

manupedrozo: From offline discussion:
- endpoint_service_id is used by the provider to call the private endpoint APIs: [example](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/0fe3ec78873c05bda4cb86025210e59b6938cfbc/internal/service/privatelinkendpointservice/resource.go#L152)
- Could move to optional and if not present, use the endpoint_name

Keeping as is makes sense to me, but I would improve the wording in the documentation as mentioned in the previous comment.
We should bring this topic to the tech sync to get input from the team.

docs/data-sources/privatelink_endpoint_service.md
```diff
@@ -122,5 +122,6 @@ In addition to all arguments above, the following attributes are exported:
   * `endpoint_name` - Forwarding rule that corresponds to the endpoint you created in GCP.
   * `ip_address` - Private IP address of the network endpoint group you created in GCP.
   * `status` - Status of the endpoint. Atlas returns one of the [values shown above](https://docs.atlas.mongodb.com/reference/api/private-endpoints-endpoint-create-one/#std-label-ref-status-field).
+* `port_mapping_enabled` - Flag that indicates whether this endpoint service uses PSC port-mapping.
```
marcosuma: should we add a link from here to "## Example with GCP (Port-Based Architecture)" section?

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/README.md
```diff
@@ -0,0 +1,100 @@
+# Example with GCP with Port-Based architecture and MongoDB Atlas Private Endpoint
```
marcosuma: @filipcirtog we're missing the migration guide from "old" to "new" supported by an example on how to do this for module users. 
Is it planned? cc @lantoli

marcosuma: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/cluster-to-advanced-cluster-migration-guide is an example (a bit more complex than your case)

docs/resources/privatelink_endpoint_service.md
```diff
@@ -144,20 +144,92 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
...
+- [GCP Private Service Connect Endpoint and Service (Port-Based Architecture)](https://github.com/mongodb/terraform-provider-mongodbatlas/tree/v2.2.0/examples/mongodbatlas_privatelink_endpoint/gcp-port-based)
 
 ## Argument Reference
 
 * `project_id` - (Required) Unique identifier for the project.
 * `private_link_id` - (Required) Unique identifier of the `AWS` or `AZURE` PrivateLink connection which is created by `mongodbatlas_privatelink_endpoint` resource.
-* `endpoint_service_id` - (Required) Unique identifier of the interface endpoint you created in your VPC with the `AWS`, `AZURE` or `GCP` resource.
+* `endpoint_service_id` - (Required) Unique identifier of the interface endpoint you created in your VPC with the `AWS`, `AZURE` or `GCP` resource. **Note:** For GCP with the new port-based architecture (when `port_mapping_enabled = true` on the endpoint resource), this value must match the `endpoint_name` in the `endpoints` block. Although the new API ignores this value, it is still required by the Terraform provider.
```
marcosuma: related to our TD conversation: why don't we make it optional and add a validation error if port_mapping_enabled = false? (probably missing something)

filipcirtog: thank you! this was an initial iteration before flagging the decision, will change it as soon as we commit to one

internal/service/privatelinkendpointservice/data_source.go
```diff
@@ -103,8 +112,8 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 	projectID := d.Get("project_id").(string)
 	privateLinkID := conversion.GetEncodedID(d.Get("private_link_id").(string), "private_link_id")
-	endpointServiceID := conversion.GetEncodedID(d.Get("endpoint_service_id").(string), "endpoint_service_id")
 	providerName := d.Get("provider_name").(string)
```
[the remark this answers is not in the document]
filipcirtog: fixed

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -158,22 +162,39 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+	case providerName == "AZURE":
 		if !pEIAOk {
 			return diag.FromErr(errors.New("`private_endpoint_ip_address` must be set when `provider_name` is `AZURE`"))
 		}
 		createEndpointRequest.Id = &endpointServiceID
 		createEndpointRequest.PrivateEndpointIPAddress = conversion.Pointer(pEIA.(string))
-	case "GCP":
-		if !gPIOk || !eOk {
-			return diag.FromErr(errors.New("`gcp_project_id`, `endpoints` must be set when `provider_name` is `GCP`"))
+	case providerName == "GCP" && gPIOk && pEIAOk && !eOk:
```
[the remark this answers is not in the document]
filipcirtog: fixed

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/main.tf
```diff
@@ -0,0 +1,92 @@
...
+
+data "mongodbatlas_advanced_cluster" "cluster" {
+  count = var.cluster_name == "" ? 0 : 1
+  # Use endpoint service as source of project_id to gather cluster data after endpoint changes are applied
+  project_id = mongodbatlas_privatelink_endpoint_service.test.project_id
+  name       = var.cluster_name
+}
+
+locals {
+  endpoint_service_id = google_compute_network.default.name
```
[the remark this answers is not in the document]
filipcirtog: nice catch! ty

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,103 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
+					resource.TestCheckResourceAttrSet(resourceName, "provider_name"),
+					resource.TestCheckResourceAttrSet(resourceName, "region"),
+					resource.TestCheckResourceAttr(resourceName, "provider_name", providerName),
+					resource.TestCheckResourceAttr(resourceName, "region", region),
+					resource.TestCheckResourceAttr(resourceName, "port_mapping_enabled", "false"),
+				),
+			},
```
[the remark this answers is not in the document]
filipcirtog: yes. this should be documented.

docs/resources/privatelink_endpoint_service.md
filipcirtog: As agreed on the TD. Continue using `endpoint_service_id` as it is a PK within all providers. As a result, for GCP with port-based architecture, use `endpoint_service_id`, `private_endpoint_ip_address` and a new field `gcp_endpoint_status`.

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/README.md
filipcirtog: yes. a migration guide is planned for a separate PR

docs/resources/privatelink_endpoint_service.md
filipcirtog: As agreed on the TD: Continue using `endpoint_service_id` as it is a PK within all providers. As a result, for GCP with port-based architecture, use `endpoint_service_id`, `private_endpoint_ip_address` and a new field `gcp_endpoint_status` and not the `endpoints` array anymore.

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,103 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+		},
+	})
+}
+
+func TestAccNetworkRSPrivateLinkEndpointGCP_basic_with_new_architecture_explicitly_disabled(t *testing.T) {
+	var (
+		resourceName = "mongodbatlas_privatelink_endpoint.test"
+		orgID        = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName  = "test-acc-tf-p-gcp-port-based-routing-feature-flag-enabled"
```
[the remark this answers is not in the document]
filipcirtog: we will not need this when the FF gets disabled

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -151,29 +155,46 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	projectID := d.Get("project_id").(string)
 	privateLinkID := conversion.GetEncodedID(d.Get("private_link_id").(string), "private_link_id")
 	providerName := d.Get("provider_name").(string)
-	endpointServiceID := d.Get("endpoint_service_id").(string)
-	pEIA, pEIAOk := d.GetOk("private_endpoint_ip_address")
-	gPI, gPIOk := d.GetOk("gcp_project_id")
-	e, eOk := d.GetOk("endpoints")
+	endpointServiceID := conversion.GetEncodedID(d.Get("endpoint_service_id").(string), "endpoint_service_id")
```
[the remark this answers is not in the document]
filipcirtog: yes, I will revert everything back to how it was before. still need to check why GetEncodedID is used in `data-source` and not in `resource`

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -73,7 +73,7 @@ func Resource() *schema.Resource {
 				Type:          schema.TypeString,
 				Optional:      true,
 				Computed:      true,
-				ConflictsWith: []string{"gcp_project_id", "endpoints"},
+				ConflictsWith: []string{"endpoints"},
```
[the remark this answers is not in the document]
filipcirtog: the validation is there . i.e. hasGCPProjectID

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/README.md
filipcirtog: Thank you for the feedback. PSC stands for Private Service Connect but I've decided to stick with the following title `GCP with Port-Based architecture`, removing PSC . Additionally, I've reorganized the documentation: the benefits are now summarized and highlighted in the header of the README, instead of being spread throughout the content. Thank you again!

internal/service/privatelinkendpoint/resource_test.go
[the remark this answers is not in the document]
filipcirtog: it is now documented

docs/data-sources/privatelink_endpoint_service.md
filipcirtog: Should we? The example is just 50 lines above the attribute.

docs/data-sources/privatelink_endpoint_service.md
```diff
@@ -84,8 +112,8 @@ data "mongodbatlas_privatelink_endpoint_service" "test" {
 
 * `project_id` - (Required) Unique identifier for the project.
 * `private_link_id` - (Required) Unique identifier of the private endpoint service for which you want to retrieve a private endpoint.
-* `endpoint_service_id` - (Required) Unique identifier of the `AWS` or `AZURE` or `GCP` resource.
-* `provider_name` - (Required) Cloud provider for which you want to create a private endpoint. Atlas accepts `AWS` or `AZURE` or `GCP`.
+* `endpoint_service_id` - (Required) Unique identifier of the interface endpoint you created in your VPC with the `AWS`, `AZURE`, or `GCP` resource. For GCP legacy architecture, this can be any identifier string. For GCP port-based architecture (when `port_mapping_enabled = true` on the endpoint resource), this should be the forwarding rule name.
```
marcosuma: why should? do we need any confirmation to change the wording?

docs/resources/privatelink_endpoint_service.md
```diff
@@ -144,20 +144,86 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
...
 * `provider_name` - (Required) Cloud provider for which you want to create a private endpoint. Atlas accepts `AWS`, `AZURE` or `GCP`.
-* `private_endpoint_ip_address` - (Optional) Private IP address of the private endpoint network interface you created in your Azure VNet. Only for `AZURE`.
-* `gcp_project_id` - (Optional) Unique identifier of the GCP project in which you created your endpoints. Only for `GCP`.
-* `endpoints` - (Optional) Collection of individual private endpoints that comprise your endpoint group. Only for `GCP`. See below.
+* `private_endpoint_ip_address` - (Optional) Private IP address of the private endpoint network interface. **Required for `AZURE`.** For GCP port-based architecture (when `port_mapping_enabled = true` on the endpoint resource), this is required and should be the IP address of the forwarding rule. For GCP legacy architecture, this is not used.
```
marcosuma: remove should if possible

marcosuma (approved):

docs/data-sources/privatelink_endpoint_service.md
filipcirtog: thank you! no, this must be the forwarding rule name, I will change the wording

filipcirtog: updated. thank you!

docs/resources/privatelink_endpoint_service.md
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -95,15 +95,10 @@ func Resource() *schema.Resource {
...
-				Computed: true,
-			},
 			"gcp_project_id": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				ForceNew:      true,
-				ConflictsWith: []string{"private_endpoint_ip_address"},
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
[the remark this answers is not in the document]
filipcirtog: This is NOT a breaking change because it does not BREAK any existing configurations.

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -152,28 +156,45 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+		if !hasPrivateEndpointIP {
 			return diag.FromErr(errors.New("`private_endpoint_ip_address` must be set when `provider_name` is `AZURE`"))
 		}
 		createEndpointRequest.Id = &endpointServiceID
-		createEndpointRequest.PrivateEndpointIPAddress = conversion.Pointer(pEIA.(string))
-	case "GCP":
-		if !gPIOk || !eOk {
-			return diag.FromErr(errors.New("`gcp_project_id`, `endpoints` must be set when `provider_name` is `GCP`"))
+		createEndpointRequest.PrivateEndpointIPAddress = conversion.Pointer(privateEndpointIP.(string))
+	case providerName == "GCP" && hasGCPProjectID && hasPrivateEndpointIP && !hasEndpoints:
```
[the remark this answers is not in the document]
filipcirtog: updated. TY!

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -152,28 +156,45 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+		createEndpointRequest.Endpoints = expandGCPEndpoints(endpoints.([]any))
+	default:
+		if providerName == "GCP" {
+			if !hasGCPProjectID {
+				return diag.FromErr(errors.New("`gcp_project_id` must be set for GCP"))
+			}
+			if hasPrivateEndpointIP && hasEndpoints {
+				return diag.FromErr(errors.New("for GCP, provide either `private_endpoint_ip_address` (new architecture) or `endpoints` (legacy architecture), but not both"))
+			}
+			return diag.FromErr(errors.New("for GCP, provide either `private_endpoint_ip_address` (new architecture) or `endpoints` (legacy architecture)"))
```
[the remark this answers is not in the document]
filipcirtog: updated. TY!

.changelog/4017.txt
```diff
@@ -0,0 +1,15 @@
+```release-note:enhancement
+resource/mongodbatlas_privatelink_endpoint_service: Adds `port_mapping_enabled` attribute
+```
+
+```release-note:enhancement
```
lantoli, line 5: the only other resource and ds in Private Endpoint Services group is https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/private_endpoint_regional_mode, can you confirm nothing changes there?

docs/resources/privatelink_endpoint_service.md
```diff
@@ -144,20 +146,86 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
 
 ```
 
+## Example with GCP (Port-Based Architecture)
+
+The new port-based architecture uses port mapping to reduce resource provisioning. Unlike the legacy architecture that requires dedicated resources for each Atlas node, the new design uses a single set of resources to support up to 1000 nodes through a port mapping network endpoint group (NEG), enabling direct targeting of specific nodes using only one customer IP address. Enable it by setting `port_mapping_enabled = true` on the endpoint resource.
```
lantoli: similarly to calling port-based architecture to the new architecture, don't know if we can use a different name for old one instead of legacy

davidhou17 (approved):

internal/service/privatelinkendpoint/resource.go
```diff
@@ -110,6 +110,11 @@ func Resource() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"port_mapping_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
```
lantoli: discussed offline, to remove ForceNew and throw an error in Update func

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -106,6 +106,103 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: configWithPortMapping(orgID, projectName, providerName, region, true),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
```
lantoli: nit: consider using helper functions for checks

internal/service/privatelinkendpointservice/data_source.go
```diff
@@ -92,6 +92,15 @@ func DataSource() *schema.Resource {
...
 				Computed: true,
 			},
+			"gcp_endpoint_status": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Status of the GCP endpoint. Only populated for port-based architecture.",
+			},
+			"port_mapping_enabled": {
+				Type:     schema.TypeBool,
+				Computed: true,
```
lantoli: add description

internal/service/privatelinkendpointservice/data_source.go
```diff
@@ -123,20 +132,56 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		return diag.FromErr(fmt.Errorf(errorEndpointSetting, "aws_connection_status", endpointServiceID, err))
 	}
 
+	if err := d.Set("interface_endpoint_id", serviceEndpoint.GetInterfaceEndpointId()); err != nil {
```
lantoli: remove these sets, as they're provided by customer in the TF config

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -95,15 +95,10 @@ func Resource() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"endpoint_group_name": {
```
lantoli, line 98: don't remove as it'll be a breaking change if some customers has it in the config, that will fail.  I recommend to keep it, remove it from the documentation, and add a code comment saying that's it's not being used, maybe create a ticket to remove if TF 3.0.0

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -145,35 +149,58 @@ func Resource() *schema.Resource {
 	}
 }
 
+func isGCPPortBasedArchitectureInput(providerName string, hasGCPProjectID, hasPrivateEndpointIP, hasEndpoints bool) bool {
+	return providerName == "GCP" && hasGCPProjectID && hasPrivateEndpointIP && !hasEndpoints
```
lantoli: nit: being "GCP" we already know hasGCPProjectID is true so no needed

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -145,35 +149,58 @@ func Resource() *schema.Resource {
...
 		createEndpointRequest.Id = &endpointServiceID
-		createEndpointRequest.PrivateEndpointIPAddress = conversion.Pointer(pEIA.(string))
+		createEndpointRequest.PrivateEndpointIPAddress = conversion.Pointer(privateEndpointIP.(string))
 	case "GCP":
-		if !gPIOk || !eOk {
-			return diag.FromErr(errors.New("`gcp_project_id`, `endpoints` must be set when `provider_name` is `GCP`"))
+		if !hasGCPProjectID {
+			return diag.FromErr(errors.New("`gcp_project_id` must be set for GCP"))
+		}
+		if (hasPrivateEndpointIP && hasEndpoints) || (!hasPrivateEndpointIP && !hasEndpoints) {
```
lantoli:
```suggestion
		if (hasPrivateEndpointIP  ==  hasEndpoints) {
```

internal/service/privatelinkendpointservice/resource.go
lantoli: similarly we've already checked that hasPrivateEndpointIP and hasEndpoints are the opposite in GCP so only need to check one

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -145,35 +149,58 @@ func Resource() *schema.Resource {
 	}
 }
 
+func isGCPPortBasedArchitectureInput(providerName string, hasGCPProjectID, hasPrivateEndpointIP, hasEndpoints bool) bool {
+	return providerName == "GCP" && hasGCPProjectID && hasPrivateEndpointIP && !hasEndpoints
+}
+
+func isGCPLegacyArchitectureInput(providerName string, hasGCPProjectID, hasPrivateEndpointIP, hasEndpoints bool) bool {
```
lantoli: recommend to remove isGCPLegacyArchitectureInput and use !isGCPPortBasedArchitectureInput instead

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -145,35 +149,58 @@ func Resource() *schema.Resource {
...
+			createEndpointRequest.EndpointGroupName = &endpointServiceID
+			createEndpointRequest.GcpProjectId = conversion.Pointer(gcpProjectID.(string))
+			singleEndpoint := admin.CreateGCPForwardingRuleRequest{
+				IpAddress:    conversion.Pointer(privateEndpointIP.(string)),
+				EndpointName: &endpointServiceID,
+			}
+			endpointsList := []admin.CreateGCPForwardingRuleRequest{singleEndpoint}
+			createEndpointRequest.Endpoints = &endpointsList
+		}
+		if isGCPLegacyArchitectureInput(providerName, hasGCPProjectID, hasPrivateEndpointIP, hasEndpoints) {
```
lantoli:
```suggestion
		} else {
```

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -273,26 +294,48 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
 
-	if err := d.Set("private_endpoint_ip_address", privateEndpoint.GetPrivateEndpointIPAddress()); err != nil {
-		return diag.FromErr(fmt.Errorf(errorEndpointSetting, "private_endpoint_ip_address", endpointServiceID, err))
-	}
-
 	if err := d.Set("private_endpoint_resource_id", privateEndpoint.GetPrivateEndpointResourceId()); err != nil {
 		return diag.FromErr(fmt.Errorf(errorEndpointSetting, "private_endpoint_resource_id", endpointServiceID, err))
 	}
 
+	if strings.EqualFold(providerName, "azure") {
```
lantoli:
```suggestion
	if  providerName == constant.AZURE
```

or even better with switch(provideName) {
case constant.AZURE:
...
case constant.GCP:
.....

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -273,26 +294,48 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
+		if err := d.Set("port_mapping_enabled", privateEndpoint.GetPortMappingEnabled()); err != nil {
+			return diag.FromErr(fmt.Errorf(errorEndpointSetting, "port_mapping_enabled", privateLinkID, err))
+		}
 
-	if providerName == "GCP" {
 		if err := d.Set("gcp_status", privateEndpoint.GetStatus()); err != nil {
 			return diag.FromErr(fmt.Errorf(errorEndpointSetting, "gcp_status", endpointServiceID, err))
 		}
+
+		if privateEndpoint.GetPortMappingEnabled() && privateEndpoint.Endpoints != nil && len(*privateEndpoint.Endpoints) == 1 {
```
lantoli:
```suggestion
		if privateEndpoint.GetPortMappingEnabled() && len(privateEndpoint.GetEndpoints()) == 1 {
```

lantoli (approved):

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. TY!

internal/service/privatelinkendpointservice/data_source.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpoint/resource.go
filipcirtog: updated. thank you!

docs/data-sources/privatelink_endpoint_service.md
```diff
@@ -71,9 +72,36 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
...
+  project_id          = mongodbatlas_privatelink_endpoint_service.test.project_id
+  private_link_id     = mongodbatlas_privatelink_endpoint_service.test.private_link_id
+  endpoint_service_id = mongodbatlas_privatelink_endpoint_service.test.endpoint_service_id
+  provider_name       = "GCP"
+}
+```
+
+## Example with GCP (Port-Based Architecture)
+
+The port-based architecture uses port mapping to reduce resource provisioning. Unlike the legacy architecture that requires dedicated resources for each Atlas node, the port-based design uses a single set of resources to support up to 150 nodes, enabling direct targeting of specific nodes using only one customer IP address. Enable it by setting `port_mapping_enabled = true` on the `mongodbatlas_privatelink_endpoint` resource.
```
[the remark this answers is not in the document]
filipcirtog: updated PR description. the note is good. ty

docs/resources/privatelink_endpoint.md
```diff
@@ -19,6 +19,8 @@ ensure that you review the [Multi-Region Private Endpoints](https://www.mongodb.
 
 -> **NOTE:** A network container is created for a private endpoint to reside in if one does not yet exist in the project.  
 
+~> **IMPORTANT:** For GCP Private Service Connect, MongoDB encourages customers to use the port-based architecture by setting `port_mapping_enabled = true` on the `mongodbatlas_privatelink_endpoint` resource. The port-based architecture uses a single set of resources to support up to 150 nodes through port mapping, enabling direct targeting of specific nodes using only one customer IP address. In contrast, the legacy architecture requires dedicated resources for each Atlas node, which can lead to IP address exhaustion and requires full redeployment when changing the number of private service connections per region group. For migration guidance, see the [GCP Private Link Port Mapping Migration Guide](../guides/gcp-privatelink-port-mapping-migration.md).
```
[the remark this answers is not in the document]
filipcirtog: updated PR description. the note is good. ty

docs/resources/privatelink_endpoint_service.md
```diff
@@ -144,26 +146,92 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
 
 ```
 
+## Example with GCP (Port-Based Architecture)
+
+The port-based architecture uses port mapping to reduce resource provisioning. Unlike the legacy architecture that requires dedicated resources for each Atlas node, the port-based design uses a single set of resources to support up to 150 nodes, enabling direct targeting of specific nodes using only one customer IP address. Enable it by setting `port_mapping_enabled = true` on the `mongodbatlas_privatelink_endpoint` resource.
```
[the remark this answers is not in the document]
filipcirtog: updated PR description. the note is good. ty

docs/resources/privatelink_endpoint_service.md
```diff
@@ -16,6 +16,8 @@ The [private link Terraform module](https://registry.terraform.io/modules/terraf
 
 -> **NOTE:** Create and delete wait for all clusters on the project to IDLE in order for their operations to complete. This ensures the latest connection strings can be retrieved following creation or deletion of this resource. Default timeout is 2hrs.
 
+~> **IMPORTANT:** For GCP Private Service Connect, MongoDB encourages customers to use the port-based architecture by setting `port_mapping_enabled = true` on the `mongodbatlas_privatelink_endpoint` resource. The port-based architecture uses a single set of resources to support up to 150 nodes through port mapping, enabling direct targeting of specific nodes using only one customer IP address. In contrast, the legacy architecture requires dedicated resources for each Atlas node, which can lead to IP address exhaustion and requires full redeployment when changing the number of private service connections per region group. For migration guidance, see the [GCP Private Link Port Mapping Migration Guide](../guides/gcp-privatelink-port-mapping-migration.md).
```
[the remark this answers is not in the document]
filipcirtog: updated PR description. the note is good. ty

examples/mongodbatlas_privatelink_endpoint/gcp-port-based/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+This project demonstrates the **port-based architecture** for setting up GCP Private Service Connect with MongoDB Atlas. Unlike the legacy architecture that requires dedicated resources for each Atlas node, the port-based design uses a single set of resources to support up to 150 nodes, enabling direct targeting of specific nodes using only one customer IP address.
+
+## Architecture Comparison
+
+| Feature | Legacy Architecture | Port-Based Architecture (this example) |
+|---------|-------------------|---------------------------|
+| Resources per Atlas node | Dedicated forwarding rule, service attachment, and instance group | Single set of resources for up to 150 nodes |
```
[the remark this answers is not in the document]
filipcirtog: updated PR description. the note is good. ty

examples/mongodbatlas_privatelink_endpoint/gcp/Readme.md
```diff
@@ -1,7 +1,16 @@
...
+# Example with GCP with legacy architecture and MongoDB Atlas Private Endpoint
 
-This project aims to provide an example of setting up GCP Private Service Connect with MongoDB Atlas.
+This project demonstrates the **legacy GCP architecture** for setting up GCP Private Service Connect with MongoDB Atlas. The legacy architecture requires dedicated resources for each Atlas node. Unlike the port-based architecture that uses a single set of resources to support up to 150 nodes, the legacy design requires one customer IP address per Atlas node.
 
+## Architecture Comparison
+
+| Feature | Legacy Architecture (this example) | Port-Based Architecture |
+|---------|-----------------------------------|---------------------------|
+| Resources per Atlas node | Dedicated forwarding rule, service attachment, and instance group | Single set of resources for up to 150 nodes |
```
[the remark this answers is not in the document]
filipcirtog: updated PR description. the note is good. ty

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -95,15 +96,14 @@ func Resource() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"endpoint_group_name": {
+			"endpoint_group_name": { // This attribute is not used anywhere is the code. Refraining from removal because of potential breaking change.
```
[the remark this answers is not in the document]
filipcirtog: solved. thx!

internal/service/privatelinkendpointservice/data_source.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: updated. thank you!

.changelog/4017.txt
lantoli: discussed offline

docs/resources/privatelink_endpoint_service.md
lantoli: discussed offline

docs/data-sources/privatelink_endpoint.md
```diff
@@ -55,8 +55,9 @@ In addition to all arguments above, the following attributes are exported:
...
 * `private_link_service_name` - Name of the Azure Private Link Service that Atlas manages.
 * `private_link_service_resource_id` - Resource ID of the Azure Private Link Service that Atlas manages.
-* `endpoint_group_names` - GCP network endpoint groups corresponding to the Private Service Connect endpoint service.
-* `region_name` - GCP region for the Private Service Connect endpoint service.
-* `service_attachment_names` - Unique alphanumeric and special character strings that identify the service attachments associated with the GCP Private Service Connect endpoint service.
+* `endpoint_group_names` - For legacy architecture, this is a list of the endpoint group names associated with the private endpoint service. For port-based architecture, this is a list of private endpoint names associated with the private endpoint service.
+* `region_name` - Region for the Private Service Connect endpoint service.
+* `service_attachment_names` - For legacy architecture, this is a list of service attachments connected to the private endpoint service (one per Atlas node). For port-based architecture, this is a list containing one service attachment connected to the private endpoint service.
```
lantoli: for endpoint_group_names and service_attachment_names, shouldn't we start with new architecture?

same for all doc files

docs/data-sources/privatelink_endpoint_service.md
```diff
@@ -84,8 +112,8 @@ data "mongodbatlas_privatelink_endpoint_service" "test" {
 
 * `project_id` - (Required) Unique identifier for the project.
 * `private_link_id` - (Required) Unique identifier of the private endpoint service for which you want to retrieve a private endpoint.
-* `endpoint_service_id` - (Required) Unique identifier of the `AWS` or `AZURE` or `GCP` resource.
-* `provider_name` - (Required) Cloud provider for which you want to create a private endpoint. Atlas accepts `AWS` or `AZURE` or `GCP`.
+* `endpoint_service_id` - (Required) Unique identifier of the interface endpoint you created in your VPC with the `AWS`, `AZURE`, or `GCP` resource. For legacy architecture, this can be any identifier string. For port-based architecture, this is the forwarding rule name.
```
lantoli:
```suggestion
* `endpoint_service_id` - (Required) Unique identifier of the interface endpoint you created in your VPC with the `AWS`, `AZURE` or `GCP` resource. For legacy architecture, this can be any identifier string. For port-based architecture, this is the forwarding rule name.
```

nit: i think we don't normally use Oxford comma

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -65,12 +60,7 @@ func TestAccNetworkRSPrivateLinkEndpointAzure_basic(t *testing.T) {
 			{
 				Config: configBasic(orgID, projectName, providerName, region),
 				Check: resource.ComposeAggregateTestCheckFunc(
-					checkExists(resourceName),
-					resource.TestCheckResourceAttrSet(resourceName, "project_id"),
-					resource.TestCheckResourceAttrSet(resourceName, "provider_name"),
-					resource.TestCheckResourceAttrSet(resourceName, "region"),
-					resource.TestCheckResourceAttr(resourceName, "provider_name", providerName),
-					resource.TestCheckResourceAttr(resourceName, "region", region),
+					checkBasicAttributes(resourceName, providerName, region)...,
```
lantoli: nit: we normally name with config, e.g. configBasic and checkBasic

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -100,12 +90,73 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+	})
+}
+
+func TestAccNetworkRSPrivateLinkEndpointGCP_basic_with_new_architecture_explicitly_enabled(t *testing.T) {
+	var (
+		resourceName = "mongodbatlas_privatelink_endpoint.test"
+		orgID        = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName  = "test-acc-tf-p-gcp-port-based-routing-feature-flag-enabled"
+		region       = "us-west3"
+		providerName = "GCP"
```
lantoli: nit: there is already constant.GCP

internal/service/privatelinkendpointservice/data_source.go
```diff
@@ -123,20 +133,39 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+		if err := d.Set("port_mapping_enabled", serviceEndpoint.GetPortMappingEnabled()); err != nil {
+			return diag.FromErr(fmt.Errorf(errorEndpointSetting, "port_mapping_enabled", privateLinkID, err))
+		}
 
-	if strings.EqualFold(providerName, "gcp") {
 		if err := d.Set("gcp_status", serviceEndpoint.GetStatus()); err != nil {
 			return diag.FromErr(fmt.Errorf(errorEndpointSetting, "gcp_status", endpointServiceID, err))
 		}
+
+		if serviceEndpoint.GetPortMappingEnabled() && len(serviceEndpoint.GetEndpoints()) == 1 {
```
lantoli: should we separate the if ?    If not, we won't detect if port-based architecture with multiple endpoints, we'll assume it's legacy archicture

if serviceEndpoint.GetPortMappingEnabled() {
    if len(serviceEndpoint.GetEndpoints()) != 1 {
      throw error
   }
   ...

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -145,35 +155,53 @@ func Resource() *schema.Resource {
...
-			return diag.FromErr(errors.New("`gcp_project_id`, `endpoints` must be set when `provider_name` is `GCP`"))
+		createEndpointRequest.PrivateEndpointIPAddress = conversion.Pointer(privateEndpointIP.(string))
+	case constant.GCP:
+		if !hasGCPProjectID {
+			return diag.FromErr(errors.New("`gcp_project_id` must be set for GCP"))
+		}
+		if hasPrivateEndpointIP == hasEndpoints {
+			return diag.FromErr(errors.New("for GCP, you must provide exactly one of: `private_endpoint_ip_address` (new architecture) or `endpoints` (legacy architecture)"))
+		}
+		if isGCPPortBasedArchitectureInput(hasPrivateEndpointIP, hasEndpoints) {
```
lantoli: i think it's probably clearer to inline and remove isGCPPortBasedArchitectureInput

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -407,7 +450,8 @@ func resourceRefreshFunc(ctx context.Context, client *admin.APIClient, projectID
 			return nil, "", err
 		}
 
-		if strings.EqualFold(providerName, "azure") || strings.EqualFold(providerName, "gcp") {
+		switch providerName {
+		case constant.AZURE, constant.GCP:
```
lantoli: one-case switch is probably better as an `if`

internal/service/privatelinkendpointservice/resource.go
```diff
@@ -285,14 +303,39 @@ func resourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 		return diag.FromErr(fmt.Errorf(errorEndpointSetting, "endpoint_service_id", endpointServiceID, err))
 	}
 
-	if err := d.Set("endpoints", flattenGCPEndpoints(privateEndpoint.Endpoints)); err != nil {
-		return diag.FromErr(fmt.Errorf(errorEndpointSetting, "endpoints", endpointServiceID, err))
-	}
+	switch providerName {
+	case constant.AZURE:
```
lantoli, line 303: no need for AWS case?

lantoli (approved):

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpointservice/data_source.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: updated. thank you!

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: updated. thank you!

docs/data-sources/privatelink_endpoint.md
filipcirtog: updated. thank you!

docs/data-sources/privatelink_endpoint_service.md
filipcirtog: updated. thank you!

docs/resources/privatelink_endpoint.md
filipcirtog: > Flag that indicates whether this resource uses GCP port-mapping. When `true`, it uses the port-mapped architecture. When `false` or unset, it uses the legacy architecture. Only applicable for GCP provider.

Since this comment has been addressed, would it be okay if I resolve it? Thank you!

docs/data-sources/privatelink_endpoint.md
filipcirtog: Since this comment has been addressed, would it be okay if I resolve it? Thank you!

>   * `endpoint_group_names` - For port-mapped architecture, this is a list of private endpoint names associated with the private endpoint service. For legacy architecture, this is a list of the endpoint group names associated with the private endpoint service.
>   * `region_name` - Region for the Private Service Connect endpoint service.
>   * `service_attachment_names` - For port-mapped architecture, this is a list containing one service attachment connected to the private endpoint service. For legacy architecture, this is a list of service attachments connected to the private endpoint service (one per Atlas node). Returns an empty list while Atlas creates the service attachments.
> * `status` - Status of the AWS PrivateLink connection or Status of the Azure/GCP Private Link Service. Atlas returns one of the following values:

docs/data-sources/privatelink_endpoint_service.md
filipcirtog: Since this comment has been addressed, would it be okay if I resolve it? Thank you!

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: Since this comment has been addressed, would it be okay if I resolve it? Thank you!

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: Since this comment has been addressed, would it be okay if I resolve it? Thank you!

internal/service/privatelinkendpointservice/data_source.go
filipcirtog: Since this comment has been addressed, would it be okay if I resolve it? Thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: Since this comment has been addressed, would it be okay if I resolve it? Thank you!

internal/service/privatelinkendpointservice/resource.go
filipcirtog: Since there wasn’t an AWS case previously, I decided to leave this unchanged. WDYT?

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -210,3 +281,17 @@ func configBasic(orgID, projectName, providerName, region string) string {
 		}
 	`, orgID, projectName, providerName, region)
 }
+
+func configWithPortMapping(orgID, projectName, providerName, region string, portMappingEnabled bool) string {
+	return fmt.Sprintf(`  
+		data "mongodbatlas_project" "test" {        
+			name   = %[2]q      
+		}
```
manupedrozo: Can we use `acc.ProjectIDExecution(t)` instead?

internal/service/privatelinkendpoint/resource.go
```diff
@@ -165,6 +175,13 @@ func resourceCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	return resourceRead(ctx, d, meta)
 }
 
+func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if d.HasChange("port_mapping_enabled") {
+		return diag.FromErr(errors.New("`port_mapping_enabled` cannot be changed after resource creation"))
+	}
+	return resourceRead(ctx, d, meta)
+}
+
```
manupedrozo, line 190: Can we replace this with a CreateOnly plan modifier?

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: Hi Espen. As an update on this, we do not support updates anymore (update via ForceNew).

An in-place update has been implemented for the `mongodbatlas_privatelink_endpoint` resource, for the `port_mapping_enabled attribute`. This is because all other attributes force a ForceNew operation. The in-place update is designed to prevent updates to `port_mapping_enabled`, which would otherwise require destroying child resource and recreating it after. This process is not feasible, as the child resource recreation would fail due to the new architecture requiring different fields. The in-place update will only throw an error if a user attempts to update the `mongodbatlas_privatelink_endpoint`.

internal/service/privatelinkendpointservice/data_source.go
```diff
@@ -123,20 +133,42 @@ func dataSourceRead(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+		}
 
-	if strings.EqualFold(providerName, "gcp") {
 		if err := d.Set("gcp_status", serviceEndpoint.GetStatus()); err != nil {
 			return diag.FromErr(fmt.Errorf(errorEndpointSetting, "gcp_status", endpointServiceID, err))
 		}
+
+		if serviceEndpoint.GetPortMappingEnabled() {
+			if len(serviceEndpoint.GetEndpoints()) != 1 {
+				return diag.FromErr(fmt.Errorf("port-mapped architecture requires exactly one endpoint, but found %d endpoints", len(serviceEndpoint.GetEndpoints())))
```
manupedrozo: This would be an API inconsistency if it were to happen. 
Curious if we have other cases where we check and error like this, and what messaging we provide in the error. Customers should contact us if this were to happen.

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -100,12 +91,73 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+			},
+		},
+	})
+}
+
+func TestAccNetworkRSPrivateLinkEndpointGCP_basic_with_new_architecture_explicitly_enabled(t *testing.T) {
+	var (
+		resourceName = "mongodbatlas_privatelink_endpoint.test"
+		orgID        = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName  = "test-acc-tf-p-gcp-port-mapped-routing-feature-flag-enabled"
```
manupedrozo: Is there actually a feature flag enabled for this project?

internal/service/privatelinkendpoint/resource_test.go
```diff
@@ -100,12 +91,73 @@ func TestAccNetworkRSPrivateLinkEndpointGCP_basic(t *testing.T) {
...
+		},
+	})
+}
+
+func TestAccNetworkRSPrivateLinkEndpointGCP_basic_with_new_architecture_explicitly_disabled(t *testing.T) {
+	var (
+		resourceName = "mongodbatlas_privatelink_endpoint.test"
+		orgID        = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		projectName  = "test-acc-tf-p-gcp-port-mapped-routing-feature-flag-enabled"
+		region       = "us-west4"
```
manupedrozo: Any reason for using different regions?

docs/resources/privatelink_endpoint_service.md
```diff
@@ -144,26 +146,92 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
...
+# Create Google Address (1 address for port-mapped architecture)
+resource "google_compute_address" "default" {
+  project      = google_compute_subnetwork.default.project
+  name         = "tf-test-psc-endpoint"
+  subnetwork   = google_compute_subnetwork.default.id
+  address_type = "INTERNAL"
+  address      = "10.0.42.1"
+  region       = google_compute_subnetwork.default.region
+
+  depends_on = [mongodbatlas_privatelink_endpoint.test]
```
manupedrozo, line 187: Curious, do we need this dependency? Is it just to avoid creating GCP resources in case the privatelink_endpoint creation fails?

.changelog/4017.txt
```diff
@@ -0,0 +1,15 @@
...
+resource/mongodbatlas_privatelink_endpoint_service: Adds `port_mapping_enabled` attribute
```
manupedrozo: Should we communicate that we are releasing/enabling a new architecture with this enhancement?

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: Yes, I’ve updated this in the resource tests. Thank you!

internal/service/privatelinkendpoint/resource.go
filipcirtog: Since this is SDKv2, the `CreateOnly` attribute is not available. Instead, Cursor suggested using `CustomizeDiff`, which seems to work well. With this approach, we can now throw errors during the terraform plan phase rather than during terraform apply. Thank you!

internal/service/privatelinkendpointservice/data_source.go
filipcirtog: Yes, this is definitely not something we expect to happen. However, I wanted to account for it just in case. Do you think I should remove it? Again from the API owner TD:

> endpoints - new definition - A list of size one, representing the single consumer forwarding rule created.

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: Yes. I have however commented out the two tests until the API is GA so that I can use `acc.ProjectIDExecution(t)` , make the CI green and later merge with the dev branch.

internal/service/privatelinkendpoint/resource_test.go
filipcirtog: No reason. I have now changed them. Thanks!

docs/resources/privatelink_endpoint_service.md
filipcirtog: Yes. Do you think we should remove it? Thank you!

internal/service/privatelinkendpointservice/data_source.go
manupedrozo: No harm in keeping the check. Regarding the error message, I would aim to be consistent with similar cases.

docs/resources/privatelink_endpoint_service.md
manupedrozo: Ok to keep

.changelog/4017.txt
filipcirtog: I have updated this. Thank you!

> resource/mongodbatlas_privatelink_endpoint: Adds `port_mapping_enabled` attribute to enable the new port-mapped architecture for GCP Private Service Connect.

docs/resources/privatelink_endpoint_service.md
filipcirtog: Thanks!

internal/service/privatelinkendpointservice/data_source.go
filipcirtog: I have added this. Thanks!

> unexpected API response: port-mapped architecture requires exactly one endpoint, but found %d endpoints. This is an API inconsistency. Please contact MongoDB support

.changelog/4017.txt
```diff
@@ -0,0 +1,15 @@
+```release-note:enhancement
+data-source/mongodbatlas_privatelink_endpoint: Adds `port_mapping_enabled` attribute to enable the new port-mapped architecture for GCP Private Service Connect.
```
manupedrozo: Nice! Please remove the period at the end for consistency (same for others)
```suggestion
data-source/mongodbatlas_privatelink_endpoint: Adds `port_mapping_enabled` attribute to enable the new port-mapped architecture for GCP Private Service Connect
```

filipcirtog: TY!

internal/service/privatelinkendpointservice/resource.go
lantoli: ok, one good thing is that we could check if it's not a good provider and throw an error but ok to keep it

internal/service/privatelinkendpointservice/resource.go
filipcirtog: updated. thanks!

docs/data-sources/privatelink_endpoint.md
```diff
@@ -55,8 +55,9 @@ In addition to all arguments above, the following attributes are exported:
 * `private_endpoints` - All private endpoints that you have added to this Azure Private Link Service.
 * `private_link_service_name` - Name of the Azure Private Link Service that Atlas manages.
 * `private_link_service_resource_id` - Resource ID of the Azure Private Link Service that Atlas manages.
-* `endpoint_group_names` - GCP network endpoint groups corresponding to the Private Service Connect endpoint service.
-* `region_name` - GCP region for the Private Service Connect endpoint service.
-* `service_attachment_names` - Unique alphanumeric and special character strings that identify the service attachments associated with the GCP Private Service Connect endpoint service.
+* `endpoint_group_names` - For port-mapped architecture, this is a list of private endpoint names associated with the private endpoint service. For legacy architecture, this is a list of the endpoint group names associated with the private endpoint service.
```
jvincent-mongodb: Not sure if this applies, but thought I'd call out that per the upstream engineer ```it should be gcp:<GCP_PROJECT_ID>:<PRIVATE_ENDPOINT_GROUP_NANE> instead of VPC_NAME for the legacy architecture. We also have to add another section for our port based architecture, which would be gcp:<GCP_PROJECT_ID>:<PRIVATE_ENDPOINT_NAME>:<PRIVATE_ENDPOINT_IP_ADDRESS>```

Again, not sure how this maps to the TF module itself, but thought I'd call this out.

docs/data-sources/privatelink_endpoint_service.md
```diff
@@ -71,9 +72,36 @@ resource "mongodbatlas_privatelink_endpoint_service" "test" {
...
+  project_id          = mongodbatlas_privatelink_endpoint_service.test.project_id
+  private_link_id     = mongodbatlas_privatelink_endpoint_service.test.private_link_id
+  endpoint_service_id = mongodbatlas_privatelink_endpoint_service.test.endpoint_service_id
+  provider_name       = "GCP"
+}
+```
+
+## Example with GCP (Port-Mapped Architecture)
+
+The port-mapped architecture uses port mapping to reduce resource provisioning. Unlike the legacy architecture that requires dedicated resources for each Atlas node, the port-mapped architecture design uses a single set of resources to support up to 150 nodes, enabling direct targeting of specific nodes using only one customer IP address. Enable it by setting `port_mapping_enabled = true` on the `mongodbatlas_privatelink_endpoint` resource.
```
jvincent-mongodb: Nit: we might clarify that in the legacy arch service attachments (dedicated resources) were mapped 1:1 with Atlas nodes, whereas now one service attachment can be mapped to up to 150 nodes via ports designated per node.

docs/data-sources/privatelink_endpoint_service.md
```diff
@@ -84,8 +112,8 @@ data "mongodbatlas_privatelink_endpoint_service" "test" {
 
 * `project_id` - (Required) Unique identifier for the project.
 * `private_link_id` - (Required) Unique identifier of the private endpoint service for which you want to retrieve a private endpoint.
-* `endpoint_service_id` - (Required) Unique identifier of the `AWS` or `AZURE` or `GCP` resource.
-* `provider_name` - (Required) Cloud provider for which you want to create a private endpoint. Atlas accepts `AWS` or `AZURE` or `GCP`.
+* `endpoint_service_id` - (Required) Unique identifier of the interface endpoint you created in your VPC with the `AWS`, `AZURE` or `GCP` resource. For port-mapped architecture, this is the forwarding rule name. For legacy architecture, this can be any identifier string.
```
jvincent-mongodb: This may be worth clarifying as legacy architecture only exists for GCP?

docs/data-sources/privatelink_endpoint_service.md
jvincent-mongodb: If the description is updated with the above, the update should be applied to all instances of this explanatory text.

docs/resources/privatelink_endpoint.md
```diff
@@ -62,10 +65,10 @@ AZURE:
...
-* `status` - Status of the AWS PrivateLink connection or Status of the Azure Private Link Service. Atlas returns one of the following values:
+  * `endpoint_group_names` - For port-mapped architecture, this is a list of private endpoint names associated with the private endpoint service. For legacy architecture, this is a list of the endpoint group names associated with the private endpoint service.
+  * `region_name` - Region for the Private Service Connect endpoint service.
+  * `service_attachment_names` - For port-mapped architecture, this is a list containing one service attachment connected to the private endpoint service. For legacy architecture, this is a list of service attachments connected to the private endpoint service (one per Atlas node). Returns an empty list while Atlas creates the service attachments.
+* `status` - Status of the AWS PrivateLink connection or Status of the Azure/GCP Private Link Service. Atlas returns one of the following values:
```
jvincent-mongodb: I don't think the GCP connection is called "Private Link" but rather "PSC" or Private Service Connect

https://docs.cloud.google.com/vpc/docs/private-service-connect#types

filipcirtog: Updated. TY!

docs/data-sources/privatelink_endpoint_service.md
filipcirtog: Updated. TY!

docs/data-sources/privatelink_endpoint_service.md
filipcirtog: Updated. TY! Will do the same for the description in the migration guide.

docs/data-sources/privatelink_endpoint.md
filipcirtog: Should we? Neither the [OpenAPI preview](https://www.mongodb.com/docs/openapi/preview/?src=https://raw.githubusercontent.com/mongodb/atlas-sdk-go/dev-latest/openapi/atlas-api.yaml#tag/Private-Endpoint-Services/operation/getGroupPrivateEndpointEndpointService) includes this, and it appears to be quite technical. If I recall correctly, there should be documentation from the API team explaining these changes. Once available, we can include a link to it in the migration guide for those who want to explore the changes further.

marcosuma: > we can include a link to it in the migration guide for those who want to explore the changes further

only comment I have is that this would be easily missed. if this doc is actually going to exist, let's sync with @celinatala-1 and create a follow-up CLOUDP ticket

filipcirtog: Yes. I have written to here. If so, I will create a CLOUDP ticket for that.

PR #4026, opened 2025-12-23
chore: Support nested set and list override in autogen

manupedrozo:
## Description

We already support overrides between set <-> list. Adding support for nested set & list as well.

Overriding org_service_account_api.secrets to a list.

Link to any related issue(s): CLOUDP-368414

## Type of change:

## Required Checklist:

### DIFF
--- a/tools/codegen/codespec/model.go
+++ b/tools/codegen/codespec/model.go
@@ -222,6 +222,7 @@ const (
 
 type CustomType struct {
 	Package CustomTypePackage `yaml:"package"`
+	Name    string            `yaml:"name,omitempty"` // Nested object name without the "TF" & "Model" prefix and suffix. Used for type overrides.
 	Model   string            `yaml:"model"`
 	Schema  string            `yaml:"schema"`
 }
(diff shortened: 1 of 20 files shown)


tools/codegen/codespec/model.go
```diff
@@ -222,6 +222,7 @@ const (
 
 type CustomType struct {
 	Package CustomTypePackage `yaml:"package"`
+	Name    string            `yaml:"name,omitempty"` // Nested object name without the "TF" & "Model" prefix and suffix. Used for type overrides.
```
maastha, line 225: [q] why can't this be reused from the parent attribute?

maastha (approved):

tools/codegen/config.yml
```diff
@@ -760,6 +760,8 @@ resources:
       method: DELETE
     schema:
       overrides:
+        secrets:
```
AgustinBettati, line 763: q: Why do we prefer List over Set in this case? Is it to allow access by index? Good to add a comment on the why and make sure ordering is guaranteed in API responses

AgustinBettati (approved):

tools/codegen/codespec/model.go
manupedrozo: Good question. For datasources the child name includes extra prefixes (DS, PluralDSResults), these are included in the Model and Schema strings initially but we dont know what to use when applying overrides.

tools/codegen/config.yml
manupedrozo: Yes, accessing by index. And yes, its ordered. 
Good call out, I will add a comment in the SA dev branch on the actual resource (this one is already deleted there).

manupedrozo:
Merging to unblock work on dev branch. I can followup in case of further comments.

PR #4085, opened 2026-01-20
chore: Add -local-fs flag to changelog-build command to fix changelog generation

AgustinBettati:
### Problem statement

- As of 5 days the [changelog generation process](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/workflows/generate-changelog.yml) has not been working, with the script `./scripts/update-changelog-unreleased-section.sh` running indefinitely until the job is either cancelled or fails.

Gained context:
- When running locally confirmed that time was consumed when calling external tool `hashicorp/go-changelog/cmd/changelog-build` `changelog-build` command.
- A change was recently made (https://github.com/hashicorp/go-changelog/pull/49) which adds a new flag allowing the use of local file system vs having all git information loaded in memory. This change however caused a regression in performance when not using the flag, as it removed the concurrent processing of changelog entries.

### Solution

When making use of the new flag the generation process remains < 5s. As per PR description it does also optimize usage of ram which was the initial concern being solved. The only downside of using local-fs is that during execution of the command local git is used and leaves the repo checked out in a random SHA. Additional logic was used to preserve the original HEAD.

As a followup will leave a comment in related PR to mention performance regression and track if follow up adjustments will be made.

## Type of change:

## Required Checklist:

### DIFF
--- a/scripts/update-changelog-unreleased-section.sh
+++ b/scripts/update-changelog-unreleased-section.sh
@@ -36,12 +36,30 @@ then
     exit 1
 fi
 
+# changelog-build -local-fs performs internal git checkouts; restore caller's HEAD afterward.
+ORIGINAL_HEAD_REF=$(git -C "$__parent" symbolic-ref -q --short HEAD || true)
+ORIGINAL_HEAD_SHA=$(git -C "$__parent" rev-parse HEAD)
+
+restore_head() {
+  if [ -n "$ORIGINAL_HEAD_REF" ]; then
+    git -C "$__parent" checkout --quiet "$ORIGINAL_HEAD_REF"
+  else
+    git -C "$__parent" checkout --quiet "$ORIGINAL_HEAD_SHA"
+  fi
+}
+
+trap restore_head EXIT
+
 CHANGELOG=$("$(go env GOPATH)"/bin/changelog-build -this-release "$TARGET_SHA" \
                       -last-release "$PREVIOUS_RELEASE_SHA" \
                       -git-dir "$__parent" \
                       -entries-dir .changelog \
                       -changelog-template "$__dir/changelog/changelog.tmpl" \
-                      -note-template "$__dir/changelog/release-note.tmpl")
+                      -note-template "$__dir/changelog/release-note.tmpl" \
+                      -local-fs)
+
+restore_head
+trap - EXIT
 
 if [ -z "$CHANGELOG" ]
 then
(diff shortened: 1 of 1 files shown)


marcabreracast (approved):
Non-blocking nit: If the PR is labeled as `bug`, should the checkbox for the `Type of change` field in the PR description be marked? Just for consistency purposes. 
LGTM, thanks for fixing this!

manupedrozo (approved):
Oh boy. Nice fix!

scripts/update-changelog-unreleased-section.sh
```diff
@@ -36,12 +36,30 @@ then
...
                       -last-release "$PREVIOUS_RELEASE_SHA" \
                       -git-dir "$__parent" \
                       -entries-dir .changelog \
                       -changelog-template "$__dir/changelog/changelog.tmpl" \
-                      -note-template "$__dir/changelog/release-note.tmpl")
+                      -note-template "$__dir/changelog/release-note.tmpl" \
+                      -local-fs)
+
+restore_head
+trap - EXIT
```
lantoli: qq: what is the last trap for? I see other scripts where we just use trap and the func we want to execute

scripts/update-changelog-unreleased-section.sh
```diff
@@ -36,12 +36,30 @@ then
...
+trap restore_head EXIT
+
 CHANGELOG=$("$(go env GOPATH)"/bin/changelog-build -this-release "$TARGET_SHA" \
                       -last-release "$PREVIOUS_RELEASE_SHA" \
                       -git-dir "$__parent" \
                       -entries-dir .changelog \
                       -changelog-template "$__dir/changelog/changelog.tmpl" \
-                      -note-template "$__dir/changelog/release-note.tmpl")
+                      -note-template "$__dir/changelog/release-note.tmpl" \
+                      -local-fs)
```
lantoli, line 59: qq: aren't they planning to fix the performance regression so we wouldn't need to change anything?  I'm surprised they'll keep this tool so complicated to use

AgustinBettati: good point, left a comment directly in the original PR providing context of the regression https://github.com/hashicorp/go-changelog/pull/49#issuecomment-3772339163

scripts/update-changelog-unreleased-section.sh
AgustinBettati: `trap - EXIT` clears the EXIT trap defined in this script. Can see this trap removal is redundant so opted for simply deleting this line (restore_head will always be run if script exists, no harm)

oarbusi (approved):

AgustinBettati:
merging to have changelog updated before release, will keep an eye on related PR to see if regression gets fixed

PR #4122, opened 2026-01-28
feat: Adds streams connection state readonly field

jwongmongodb:
## Description

This adds the state field to the streams connection to allow creating and updating connections to be more reliable. Since not all connections are immediately available, customers that generates a connection and immediately creates and starts a stream processor but have it fail since the connection is not ready to be used. This was a bad user experience.

## Type of change:

## Required Checklist:


jwongmongodb:
I tagged you @oarbusi since this enhancement came from our discussion a while back to add a state field for our connections to poll on. [Here](https://docs.google.com/document/d/1Gxe_ysPMfcCC2Hw1i0OIPcEdAIeRFPn8j6HSVGQqsR4/edit?tab=t.0#heading=h.my8l8kwcujpo) is the relevant doc. I believe this should also help with the flaky streams test for kafkaSSL that we have been seeing.

oarbusi (approved):
Nice! LGTM

internal/service/streamconnection/state_transition.go
```diff
@@ -24,3 +38,55 @@ func DeleteStreamConnection(ctx context.Context, api admin.StreamsApi, projectID
...
+				// After notFoundChecks consecutive 404s (~1.4s), it will fail fast
+				return nil, StatePending, nil
+			}
+			return nil, "", err
+		}
+		state := model.GetState()
+		if state == "" {
+			// If state is not present in the response, assume the connection is ready
+			return model, StateReady, nil
+		}
```
AgustinBettati, line 89: q: Are there certain connection types that always return an empty string? Asking as the API docs only mentions 4 non-empty values.

internal/service/streamconnection/state_transition.go
```diff
@@ -2,13 +2,27 @@ package streamconnection
...
+	"errors"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/validate"
 	"go.mongodb.org/atlas-sdk/v20250312013/admin"
 )
 
+const (
+	defaultCreateUpdateTimeout = 20 * time.Minute // The amount of time to wait before timeout for create/update
```
AgustinBettati, line 14: When a resource has long-running operations we define a `timeouts` attribute in the resource in case customers want to configure custom timeouts for waiting on operation completion. Do you think it would make sense in this case? For consistency it would be good, but maybe creation/update times are very short.

internal/service/streamconnection/state_transition.go
jwongmongodb: There can only be 4 values for the state right now (PENDING, READY, DELETING, FAILED)

internal/service/streamconnection/state_transition.go
jwongmongodb: so the creates and updates can be quite long like 5-7 mins for private networking connections. Another example is a cluster connection on a cluster that just got created which can take 7-10 minutes. What do you propose as a good timeout time instead?

AgustinBettati: Thanks for the context. I am okay with the default timeout value, my question is related to supporting [standard `timeouts`](https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/timeouts#migrate-timeouts-in-nested-attributes) attribute in the resource so users can define custom timeout values if desired. This is relevant if this resources now has long running operations in create/update/delete.

internal/service/streamconnection/resource_stream_connection.go
```diff
@@ -248,12 +258,20 @@ func (r *streamConnectionRS) Update(ctx context.Context, req resource.UpdateRequ
...
-	apiResp, _, err := connV2.StreamsApi.UpdateStreamConnection(ctx, projectID, workspaceOrInstanceName, connectionName, streamConnectionReq).Execute()
+	_, _, err := connV2.StreamsApi.UpdateStreamConnection(ctx, projectID, workspaceOrInstanceName, connectionName, streamConnectionReq).Execute()
 	if err != nil {
 		resp.Diagnostics.AddError("error updating resource", err.Error())
 		return
 	}
 
+	// Wait for the connection to reach a ready state before returning
+	// This ensures the connection is fully provisioned and available for use with stream processors
+	apiResp, err := WaitStateTransition(ctx, projectID, workspaceOrInstanceName, connectionName, connV2.StreamsApi)
```
AgustinBettati: Should we wait in the delete operation as well? Considering we have `DELETING` status.

jwongmongodb: The current deleting logic is sufficient since the connection will still be there if it has a deleting status and eventually be deleted afterwards when successfully deleted (the current logic polls for it). There is still implementation needed on our side to do the DELETING state so it is not available.

internal/service/streamconnection/state_transition.go
jwongmongodb: interesting, I can see if I can incorporate this user defined timeout for CREATING and UPDATING.

jwongmongodb:
@AgustinBettati @oarbusi I've added the custom timeouts, can you take a look? Also, I noticed that there have been occasional failures where the atlas clusters in dev are not reliably provisioned, which causes the tests to run for 2-3 hours waiting.

I think the Atlas cluster provisioning could use a shorter timeout than 1 hour? Or for streams related tests, can we always point to an idle atlas cluster that is already provisioned? 

```
2026/02/02 03:04:40 [WARN] WaitForState timeout after 1h0m0s
2026/02/02 03:04:40 [WARN] WaitForState starting 30s refresh grace period
```

docs/resources/stream_connection.md
```diff
@@ -183,6 +183,56 @@ resource "mongodbatlas_stream_connection" "example-schema-registry-sasl" {
 }
 ```
 
+### Example Usage with Stream Processor
+
+When using a stream connection with a stream processor, the connection must be fully provisioned before the processor can be created. The provider automatically waits for connections to be ready after creation or updates. The example below shows a typical pattern:
+
+```terraform
+resource "mongodbatlas_stream_instance" "example" {
```
AgustinBettati:
```suggestion
resource "mongodbatlas_stream_workspace" "example" {
```

docs/resources/stream_connection.md
```diff
@@ -256,6 +306,22 @@ If `type` is of value `SchemaRegistry` the following additional arguments are de
 * `username` - Username for the Schema Registry. Required when `type` is `USER_INFO`.
 * `password` - Password for the Schema Registry. Required when `type` is `USER_INFO`.
 
+### Timeouts
+
+```terraform
+resource "mongodbatlas_stream_connection" "example" {
+  # ... other configuration ...
+
+  timeouts {
```
AgustinBettati:
```suggestion
  timeouts = {
```
Good to double check, but quite confident it uses the attribute syntax given how it is defined in the schema

internal/service/streamconnection/resource_stream_connection.go
```diff
@@ -34,6 +36,33 @@ type streamConnectionRS struct {
...
+	Headers types.Map    `tfsdk:"headers"`
+	URL     types.String `tfsdk:"url"`
+	// SchemaRegistry connection
+	SchemaRegistryProvider       types.String `tfsdk:"schema_registry_provider"`
+	SchemaRegistryURLs           types.List   `tfsdk:"schema_registry_urls"`
+	SchemaRegistryAuthentication types.Object `tfsdk:"schema_registry_authentication"`
+}
+
+// TFStreamConnectionDSModel is the data source model without timeouts (data sources don't support timeouts)
+type TFStreamConnectionDSModel struct {
```
AgustinBettati: nit: For common struct properties you can define a common struct and embed it in each one (e.g. [TFPushBasedLogExportCommonModel](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/e25ce2eac75991c763a56a8815d59f54e98a4487/internal/service/pushbasedlogexport/resource_schema.go#L89))

AgustinBettati (approved):
LGTM, thanks for follow ups. Leaving some smaller suggestions/considerations.

oarbusi (approved):
LGTM

lmkerbey-mdb (approved):
LGTM.

jwongmongodb:
Merging, the current failures in the acceptance tests is unrelated and being looked at by others in my team.

PR #4140, opened 2026-02-03
feat: Add role_id azure key vault encryption at rest

marcabreracast:
## Description

This PR adds an improvement for Azure Key Vault, by providing serverless authentication adding `role_id` support to `mongodbatlas_encryption_at_rest`, meaning that static credentiasl (`client_id/tenant_id/secret`) are no longer required. 

The PR also adds test coverage for the  workflow, please note that acceptance test is skipped in CI, matching the existing GCP test about that same field addition. Test has been run in local, and is passing. 

Link to any related issue(s): [CLOUDP-377432](https://jira.mongodb.org/browse/CLOUDP-377432)

## Type of change:

- [X] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.

## Required Checklist:

## Further comments

With this new auth workflow using `role_id`, the API responses will omit fields `client_id` and `tenant_id`. These attributes are optional + sensitive, and the provider maps those omitted values to `""` instead of `null`, which will trigger post-apply “inconsistent result” errors like the following:

```
╷
│ 
│ When applying changes to mongodbatlas_encryption_at_rest.test, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an
│ unexpected new value: .azure_key_vault_config[0].client_id: inconsistent values for sensitive attribute.
│ 
╵
╷
│ 
│ When applying changes to mongodbatlas_encryption_at_rest.test, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an
│ unexpected new value: .azure_key_vault_config[0].tenant_id: inconsistent values for sensitive attribute.
```

In order to avoid this, it's been agreed to change `client_id` and `tenant_id` to `Computed + Optional`, to avoid any potential breaking changes. 

## Manual Tests for Migration - From Static Credentials to Serverless

1. Defined TF config with static credentials (usage of `client_id`, `tenant_id` and `secret`):

```
resource "mongodbatlas_encryption_at_rest" "test" {
  project_id = "<PROJECT_ID>"

  azure_key_vault_config {
    enabled           = true
    azure_environment = "AZURE"

     tenant_id       = "<YOUR_TENANT_ID>"
     subscription_id = "<SUBSCRIPTION_ID>"
     client_id       = "<CLIENT_ID>"
     secret          = "<SECRET>"

    resource_group_name = "RESOURCE_GROUP_NAME"
    key_vault_name      = "KEY_VAULT_NAME"
    key_identifier      = "<YOUR_KEY_IDENTIFIER>"
  }
}

```

2. `tf plan/tf apply`: Configuration gets created, and state file populates the correct fields

```
            "azure_key_vault_config": {
              "azure_environment": "AZURE",
              "client_id": <REDACTED>,
              "enabled": true,
              "key_identifier": "<REDACTED>",
              "key_vault_name": "<REDACTED>,
              "require_private_networking": false,
              "resource_group_name": "<REDACTED>",
              "role_id": null,
              "secret": <REDACTED>,
              "subscription_id": "<REDACTED>",
              "tenant_id": <REDACTED>,
              "valid": true
            },
```

3. Update config file with new secretless auth (usage of `role_id`):

```
resource "mongodbatlas_encryption_at_rest" "test" {
  project_id = "<PROJECT_ID>"

  azure_key_vault_config {
    enabled           = true
    azure_environment = "AZURE"

     subscription_id = "<SUBSCRIPTION_ID>"

    resource_group_name = "RESOURCE_GROUP_NAME"
    key_vault_name      = "KEY_VAULT_NAME"
    key_identifier      = "<YOUR_KEY_IDENTIFIER>"
    role_id             = "<ROLE_ID>"
  }
}

```

4. `tf plan/apply`: Plan command shows changes to azure config, apply is successful.

```
            "azure_key_vault_config": {
              "azure_environment": "AZURE",
              "client_id": "",
              "enabled": true,
              "key_identifier": "<REDACTED>",
              "key_vault_name": "<REDACTED>,
              "require_private_networking": false,
              "resource_group_name": "<REDACTED>",
              "role_id": "<REDACTED>",
              "secret": null,
              "subscription_id": "<REDACTED>",
              "tenant_id": "",
              "valid": true
            },
```

### DIFF
--- a/docs/data-sources/encryption_at_rest.md
+++ b/docs/data-sources/encryption_at_rest.md
@@ -174,6 +174,7 @@ Read-Only:
 - `key_vault_name` (String) Unique string that identifies the Azure Key Vault that contains your key.
 - `require_private_networking` (Boolean) Enable connection to your Azure Key Vault over private networking.
 - `resource_group_name` (String) Name of the Azure resource group that contains your Azure Key Vault.
+- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
 - `secret` (String, Sensitive) Private data that you need secured and that belongs to the specified Azure Key Vault (AKV) tenant (**azureKeyVault.tenantID**). This data can include any type of sensitive data such as passwords, database connection strings, API keys, and the like. AKV stores this information as encrypted binary data.
 - `subscription_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies your Azure subscription.
 - `tenant_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies the Azure Active Directory tenant within your Azure subscription.
--- a/docs/data-sources/encryption_at_rest.md
+++ b/docs/data-sources/encryption_at_rest.md
@@ -174,7 +174,7 @@ Read-Only:
 - `key_vault_name` (String) Unique string that identifies the Azure Key Vault that contains your key.
 - `require_private_networking` (Boolean) Enable connection to your Azure Key Vault over private networking.
 - `resource_group_name` (String) Name of the Azure resource group that contains your Azure Key Vault.
-- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
+- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that Atlas uses to access the Azure Key Vault.
 - `secret` (String, Sensitive) Private data that you need secured and that belongs to the specified Azure Key Vault (AKV) tenant (**azureKeyVault.tenantID**). This data can include any type of sensitive data such as passwords, database connection strings, API keys, and the like. AKV stores this information as encrypted binary data.
 - `subscription_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies your Azure subscription.
 - `tenant_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies the Azure Active Directory tenant within your Azure subscription.
--- a/docs/resources/encryption_at_rest.md
+++ b/docs/resources/encryption_at_rest.md
@@ -223,6 +223,7 @@ Optional:
 - `key_vault_name` (String) Unique string that identifies the Azure Key Vault that contains your key.
 - `require_private_networking` (Boolean) Enable connection to your Azure Key Vault over private networking.
 - `resource_group_name` (String) Name of the Azure resource group that contains your Azure Key Vault.
+- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
 - `secret` (String, Sensitive) Private data that you need secured and that belongs to the specified Azure Key Vault (AKV) tenant (**azureKeyVault.tenantID**). This data can include any type of sensitive data such as passwords, database connection strings, API keys, and the like. AKV stores this information as encrypted binary data.
 - `subscription_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies your Azure subscription.
 - `tenant_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies the Azure Active Directory tenant within your Azure subscription.
--- a/docs/resources/encryption_at_rest.md
+++ b/docs/resources/encryption_at_rest.md
@@ -223,7 +223,7 @@ Optional:
 - `key_vault_name` (String) Unique string that identifies the Azure Key Vault that contains your key.
 - `require_private_networking` (Boolean) Enable connection to your Azure Key Vault over private networking.
 - `resource_group_name` (String) Name of the Azure resource group that contains your Azure Key Vault.
-- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
+- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that Atlas uses to access the Azure Key Vault.
 - `secret` (String, Sensitive) Private data that you need secured and that belongs to the specified Azure Key Vault (AKV) tenant (**azureKeyVault.tenantID**). This data can include any type of sensitive data such as passwords, database connection strings, API keys, and the like. AKV stores this information as encrypted binary data.
 - `subscription_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies your Azure subscription.
 - `tenant_id` (String, Sensitive) Unique 36-hexadecimal character string that identifies the Azure Active Directory tenant within your Azure subscription.
--- a/internal/service/encryptionatrest/data_source_schema.go
+++ b/internal/service/encryptionatrest/data_source_schema.go
@@ -75,7 +75,7 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
 					},
 					"role_id": schema.StringAttribute{
 						Computed:            true,
-						MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.",
+						MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that Atlas uses to access the Azure Key Vault.",
 					},
 					"key_vault_name": schema.StringAttribute{
 						Computed:            true,
--- a/internal/service/encryptionatrest/data_source_schema.go
+++ b/internal/service/encryptionatrest/data_source_schema.go
@@ -73,6 +73,10 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
 						Sensitive:           true,
 						MarkdownDescription: "Web address with a unique key that identifies for your Azure Key Vault.",
 					},
+					"role_id": schema.StringAttribute{
+						Computed:            true,
+						MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.",
+					},
 					"key_vault_name": schema.StringAttribute{
 						Computed:            true,
 						MarkdownDescription: "Unique string that identifies the Azure Key Vault that contains your key.",
--- a/internal/service/encryptionatrest/model.go
+++ b/internal/service/encryptionatrest/model.go
@@ -79,14 +79,14 @@ func NewTFAzureKeyVaultConfigItem(az *admin.AzureKeyVault) *TFAzureKeyVaultConfi
 
 	return &TFAzureKeyVaultConfigModel{
 		Enabled:                  types.BoolPointerValue(az.Enabled),
-		ClientID:                 types.StringValue(az.GetClientID()),
+		ClientID:                 conversion.StringNullIfEmpty(az.GetClientID()),
 		AzureEnvironment:         types.StringValue(az.GetAzureEnvironment()),
 		SubscriptionID:           types.StringValue(az.GetSubscriptionID()),
 		ResourceGroupName:        types.StringValue(az.GetResourceGroupName()),
 		KeyVaultName:             types.StringValue(az.GetKeyVaultName()),
 		KeyIdentifier:            types.StringValue(az.GetKeyIdentifier()),
 		RoleID:                   conversion.StringNullIfEmpty(az.GetRoleId()),
-		TenantID:                 types.StringValue(az.GetTenantID()),
+		TenantID:                 conversion.StringNullIfEmpty(az.GetTenantID()),
 		Secret:                   conversion.StringNullIfEmpty(az.GetSecret()),
 		RequirePrivateNetworking: types.BoolValue(az.GetRequirePrivateNetworking()),
 		Valid:                    types.BoolPointerValue(az.Valid),
--- a/internal/service/encryptionatrest/model.go
+++ b/internal/service/encryptionatrest/model.go
@@ -79,14 +79,14 @@ func NewTFAzureKeyVaultConfigItem(az *admin.AzureKeyVault) *TFAzureKeyVaultConfi
 
 	return &TFAzureKeyVaultConfigModel{
 		Enabled:                  types.BoolPointerValue(az.Enabled),
-		ClientID:                 conversion.StringNullIfEmpty(az.GetClientID()),
+		ClientID:                 types.StringPointerValue(az.ClientID),
 		AzureEnvironment:         types.StringValue(az.GetAzureEnvironment()),
 		SubscriptionID:           types.StringValue(az.GetSubscriptionID()),
 		ResourceGroupName:        types.StringValue(az.GetResourceGroupName()),
 		KeyVaultName:             types.StringValue(az.GetKeyVaultName()),
 		KeyIdentifier:            types.StringValue(az.GetKeyIdentifier()),
-		RoleID:                   conversion.StringNullIfEmpty(az.GetRoleId()),
-		TenantID:                 conversion.StringNullIfEmpty(az.GetTenantID()),
+		RoleID:                   types.StringPointerValue(az.RoleId),
+		TenantID:                 types.StringPointerValue(az.TenantID),
 		Secret:                   conversion.StringNullIfEmpty(az.GetSecret()),
 		RequirePrivateNetworking: types.BoolValue(az.GetRequirePrivateNetworking()),
 		Valid:                    types.BoolPointerValue(az.Valid),
--- a/internal/service/encryptionatrest/model.go
+++ b/internal/service/encryptionatrest/model.go
@@ -148,14 +148,14 @@ func NewAtlasAzureKeyVault(tfAzKeyVaultConfigSlice []TFAzureKeyVaultConfigModel)
 
 	return &admin.AzureKeyVault{
 		Enabled:                  v.Enabled.ValueBoolPointer(),
-		ClientID:                 v.ClientID.ValueStringPointer(),
+		ClientID:                 conversion.NilForUnknownOrEmptyString(v.ClientID),
 		AzureEnvironment:         v.AzureEnvironment.ValueStringPointer(),
 		SubscriptionID:           v.SubscriptionID.ValueStringPointer(),
 		ResourceGroupName:        v.ResourceGroupName.ValueStringPointer(),
 		KeyVaultName:             v.KeyVaultName.ValueStringPointer(),
 		KeyIdentifier:            v.KeyIdentifier.ValueStringPointer(),
 		Secret:                   v.Secret.ValueStringPointer(),
-		TenantID:                 v.TenantID.ValueStringPointer(),
+		TenantID:                 conversion.NilForUnknownOrEmptyString(v.TenantID),
 		RequirePrivateNetworking: v.RequirePrivateNetworking.ValueBoolPointer(),
 		RoleId:                   v.RoleID.ValueStringPointer(),
 	}
--- a/internal/service/encryptionatrest/model.go
+++ b/internal/service/encryptionatrest/model.go
@@ -85,6 +85,7 @@ func NewTFAzureKeyVaultConfigItem(az *admin.AzureKeyVault) *TFAzureKeyVaultConfi
 		ResourceGroupName:        types.StringValue(az.GetResourceGroupName()),
 		KeyVaultName:             types.StringValue(az.GetKeyVaultName()),
 		KeyIdentifier:            types.StringValue(az.GetKeyIdentifier()),
+		RoleID:                   conversion.StringNullIfEmpty(az.GetRoleId()),
 		TenantID:                 types.StringValue(az.GetTenantID()),
 		Secret:                   conversion.StringNullIfEmpty(az.GetSecret()),
 		RequirePrivateNetworking: types.BoolValue(az.GetRequirePrivateNetworking()),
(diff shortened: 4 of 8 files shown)


internal/service/encryptionatrest/model.go
```diff
@@ -79,13 +79,14 @@ func NewTFAzureKeyVaultConfigItem(az *admin.AzureKeyVault) *TFAzureKeyVaultConfi
 
 	return &TFAzureKeyVaultConfigModel{
 		Enabled:                  types.BoolPointerValue(az.Enabled),
-		ClientID:                 types.StringValue(az.GetClientID()),
+		ClientID:                 conversion.StringNullIfEmpty(az.GetClientID()),
```
marcabreracast: See `Further Comments` section of the PR for an explanation of this change

internal/service/encryptionatrest/resource_test.go
```diff
@@ -156,6 +156,28 @@ func TestAccEncryptionAtRest_basicAzure(t *testing.T) {
 	})
 }
 
+func TestAccEncryptionAtRest_basicAzureWithRole(t *testing.T) {
+	acc.SkipTestForCI(t) // needs Azure configuration
```
marcabreracast, line 160: Mirroring behaviour of what we do for GCP. Acceptance test has been executed in local with the proper Azure configuration and it works as expected

marcabreracast:
`TestAccEncryptionAtRestPrivateEndpoint_Azure_basic` is throwing a 500, but this test has not been modified in this PR, and has passed when [running the workflow on QA](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/21665574419)

manupedrozo (commented):
Nice!

Maybe these are already planned for later:
* We should clarify in the docs when to use role_id vs client_id/tenant_id/secret. Probably within the attribute descriptions.
* If role_id + client_id/tenant_id/secret are provided, should we fail at plan time? I.e. validate like: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/fad37b1c746e837916edee4bf3ceb90291464d78/internal/service/projectipaccesslist/resource_schema.go#L59

marcabreracast:
@manupedrozo thanks for the review! Regarding the comments:

> We should clarify in the docs when to use role_id vs client_id/tenant_id/secret. Probably within the attribute descriptions.

I can do that, but for consistency purposes, we're not giving any specific clarifications in the GCP provider at the moment. Tagging @mongodb/docs-cloud-team for opinion on this, happy to address it if needed. 

> If role_id + client_id/tenant_id/secret are provided, should we fail at plan time? I.e. validate like:

For GCP we don't have validation at plan time if a user defines both types of authentication. We're letting it fail at apply phase does throw a very detailed error from the API (here's what it shows for Azure): 

```
 PATCH: HTTP 400 Bad Request (Error code: "INVALID_AZURE_KEY_VAULT_AUTH_CONFIG")
│ Detail: Azure Key Vault authorization config should not specify both the credentials and the Cloud Provider Access roleId. Reason: Bad Request. Params: [], BadRequestDetail: 
```

docs/resources/encryption_at_rest.md
```diff
@@ -223,6 +223,7 @@ Optional:
 - `key_vault_name` (String) Unique string that identifies the Azure Key Vault that contains your key.
 - `require_private_networking` (Boolean) Enable connection to your Azure Key Vault over private networking.
 - `resource_group_name` (String) Name of the Azure resource group that contains your Azure Key Vault.
+- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
```
AgustinBettati: Can be good to adjust our examples section in the [resource docs](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/encryption_at_rest#configuring-encryption-at-rest-using-customer-key-management-in-azure) and [examples](https://github.com/mongodb/terraform-provider-mongodbatlas/tree/master/examples/mongodbatlas_encryption_at_rest/azure) with the new authentication mechanism. We can also cover how the migration for static credentials to role based works (similar to https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/encryption-at-rest-gcp-role-based-migration)

manupedrozo:
> @manupedrozo thanks for the review! Regarding the comments:
> 
> > We should clarify in the docs when to use role_id vs client_id/tenant_id/secret. Probably within the attribute descriptions.
> 
> I can do that, but for consistency purposes, we're not giving any specific clarifications in the GCP provider at the moment. Tagging @mongodb/docs-cloud-team for opinion on this, happy to address it if needed.
> 
> > If role_id + client_id/tenant_id/secret are provided, should we fail at plan time? I.e. validate like:
> 
> For GCP we don't have validation at plan time if a user defines both types of authentication. We're letting it fail at apply phase does throw a very detailed error from the API (here's what it shows for Azure):
> 
> ```
>  PATCH: HTTP 400 Bad Request (Error code: "INVALID_AZURE_KEY_VAULT_AUTH_CONFIG")
> │ Detail: Azure Key Vault authorization config should not specify both the credentials and the Cloud Provider Access roleId. Reason: Bad Request. Params: [], BadRequestDetail: 
> ```

Ty for the context! Maybe something to bring to the tech sync tomorrow? I think these would be good additions for UX.
Approving in the meantime as we can always follow-up.

manupedrozo (approved):

internal/service/encryptionatrest/resource.go
```diff
@@ -532,11 +537,15 @@ func HandleAzureKeyVaultConfigDefaults(ctx context.Context, earRSCurrent, earRSN
...
-	// that user provided. encryptionAtRestRSConfig is nil during Read(), so we use the current plan
+	// Some fields may be omitted in the response (especially sensitive fields like client_id, tenant_id, secret).
+	// To keep state stable after Create/Update, we need topreserve those values from config (or prior state).
 	if earRSConfig != nil && len(earRSConfig.AzureKeyVaultConfig) > 0 {
+		earRSNew.AzureKeyVaultConfig[0].ClientID = earRSConfig.AzureKeyVaultConfig[0].ClientID
+		earRSNew.AzureKeyVaultConfig[0].TenantID = earRSConfig.AzureKeyVaultConfig[0].TenantID
 		earRSNew.AzureKeyVaultConfig[0].Secret = earRSConfig.AzureKeyVaultConfig[0].Secret
 	} else {
+		earRSNew.AzureKeyVaultConfig[0].ClientID = earRSCurrent.AzureKeyVaultConfig[0].ClientID
+		earRSNew.AzureKeyVaultConfig[0].TenantID = earRSCurrent.AzureKeyVaultConfig[0].TenantID
```
AgustinBettati: I am suprised this is added now, is clientID and tenantID not returned in the API response when defined in the request? Or are we setting the config value for another reason?

internal/service/encryptionatrest/model.go
```diff
@@ -79,13 +79,14 @@ func NewTFAzureKeyVaultConfigItem(az *admin.AzureKeyVault) *TFAzureKeyVaultConfi
...
 		AzureEnvironment:         types.StringValue(az.GetAzureEnvironment()),
 		SubscriptionID:           types.StringValue(az.GetSubscriptionID()),
 		ResourceGroupName:        types.StringValue(az.GetResourceGroupName()),
 		KeyVaultName:             types.StringValue(az.GetKeyVaultName()),
 		KeyIdentifier:            types.StringValue(az.GetKeyIdentifier()),
-		TenantID:                 types.StringValue(az.GetTenantID()),
+		RoleID:                   conversion.StringNullIfEmpty(az.GetRoleId()),
+		TenantID:                 conversion.StringNullIfEmpty(az.GetTenantID()),
```
AgustinBettati: Is there any impact for users which are not leveraging "client_id", "role_id", "tenant_id"? When response is empty state used to have "", now it will store null.

internal/service/encryptionatrest/resource_test.go
```diff
@@ -156,6 +156,28 @@ func TestAccEncryptionAtRest_basicAzure(t *testing.T) {
 	})
 }
 
+func TestAccEncryptionAtRest_basicAzureWithRole(t *testing.T) {
```
AgustinBettati, line 159: Not sure if worth capturing the migration from static credentials to role-based depending on the complexity

AgustinBettati (commented):
Looks good, want to double check no impact given change in how we store empty attributes in the azure config

internal/service/encryptionatrest/model.go
marcabreracast: No impact, I tested the static credentials flow and it did not detect any issues. 
The only impact customers might see is after upgrading the provider, terraform would show a change in the plan like `client_id: "" -> null` for example

docs/resources/encryption_at_rest.md
marcabreracast: Yes, https://jira.mongodb.org/browse/DOCSP-57550 has been created to address this

internal/service/encryptionatrest/resource_test.go
marcabreracast: A migration “static creds → role_id” test would be fairly complex IMO because it requires authorizing Cloud Provider Access for Azure (to obtain a valid role_id) and having a real Key Vault/key with the right permissions. We also currently skip TestAccEncryptionAtRest_basicAzureWithRole in CI (same pattern as GCP).

I did a manual test of the migration, and it worked as expected with the setup Azure Config.

internal/service/encryptionatrest/resource.go
marcabreracast: They're returned, I double-checked because I couldn't find a circumstance where these fields are needed and not returned in the response. Reverted in [7f481f8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140/commits/7f481f8f046ae372108fe6eaf3a7344f9d857b50), FFTR!

internal/service/encryptionatrest/model.go
AgustinBettati: > The only impact customers might see is after upgrading the provider, terraform would show a change in the plan like client_id: "" -> null for example

If during an upgrade of the provider version customers start seeing a plan change in their ear resource, having done no changes on the config, I would consider this a breaking change as we are forcing an action on them (even if very subtle). Can you confirm if there is a scenario this can happen? Not fully clear when the `client_id: "" -> null` can appear. If we confirm the breaking change scenario we can look into solutions.

docs/data-sources/encryption_at_rest.md
```diff
@@ -174,6 +174,7 @@ Read-Only:
 - `key_vault_name` (String) Unique string that identifies the Azure Key Vault that contains your key.
 - `require_private_networking` (Boolean) Enable connection to your Azure Key Vault over private networking.
 - `resource_group_name` (String) Name of the Azure resource group that contains your Azure Key Vault.
+- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
```
lmkerbey-mdb:
```suggestion
- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that Atlas uses to access the Azure Key Vault.
```

We should probably do a sweep through the Terraform docs to make sure we don't say 'MongoDB Cloud' anywhere else; it's Atlas.

docs/resources/encryption_at_rest.md
lmkerbey-mdb:
```suggestion
- `role_id` (String) Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.
```
See above.

internal/service/encryptionatrest/data_source_schema.go
```diff
@@ -73,6 +73,10 @@ func DataSourceSchema(ctx context.Context) schema.Schema {
 						Sensitive:           true,
 						MarkdownDescription: "Web address with a unique key that identifies for your Azure Key Vault.",
 					},
+					"role_id": schema.StringAttribute{
+						Computed:            true,
+						MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.",
```
lmkerbey-mdb:
```suggestion
						MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.",
```
See above.

internal/service/encryptionatrest/resource.go
```diff
@@ -230,6 +231,10 @@ func (r *encryptionAtRestRS) Schema(ctx context.Context, req resource.SchemaRequ
 							Computed:            true,
 							MarkdownDescription: "Flag that indicates whether the Azure encryption key can encrypt and decrypt data.",
 						},
+						"role_id": schema.StringAttribute{
+							Optional:            true,
+							MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.",
```
lmkerbey-mdb:
```suggestion
							MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the Azure Service Principal that MongoDB Cloud uses to access the Azure Key Vault.",
```
See above.

internal/service/encryptionatrest/model.go
marcosuma: > The only impact customers might see is after upgrading the provider, terraform would show a change in the plan like client_id: "" -> null for example

Indeed IIUC that looks like a **breaking change**, which means we should not do that.
The idea is: if I am a customer who hasn't changed anything on my configuration and I go to 2.x.0 to 2.y.0 (where this code is released) there must be no plan changes.

lmkerbey-mdb (changes requested):

internal/service/encryptionatrest/resource_test.go
marcosuma: > I did a manual test of the migration,

@marcabreracast in case of manual tests we usually add the executed steps in the description of the summary, so the reviewer can look into it and make sure it is sound.

Also, for Agustin's request we're going to have an example (written by docs I presume), yes?

internal/service/encryptionatrest/model.go
marcabreracast: I've retested this to confim, and there were no changes in the plan: 

Performing a tf plan showed `No changes.`, but the `terraform.tfstate` file does modify the attributes from `""` to `null` silently.  I can bring this topic to our weekly so we can align on a decision for this? And give more details on the test I did

marcabreracast: Additionally, re-reading my first reply I can see where the problem was: I misused the term "plan", when I said "terraform would show a change in the plan", I meant in the state.

internal/service/encryptionatrest/resource_test.go
marcabreracast: Sure, I'll add the manual tests to the PR description.

Regarding Agustin's request, yes that's correct: in this [PR comment ](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140#discussion_r2764516755) I mention the DOCSP ticket I created for this, and I'm already in contact with Javier

internal/service/encryptionatrest/model.go
marcabreracast: For further context, here are the testing steps I performed, and some additional observations:

### Test setup
- Using the `mongodbatlas_encryption_at_rest` resource, I configured the `aws_kms_config` block.
- Since Azure Key Vault EAR requires static credentials for authentication, I used AWS as the configured provider to observe how the provider writes the unused Azure block into state when Azure is not configured.

### Baseline state (before the change)
With a basic AWS EAR config applied, the `azure_key_vault_config` object in state looked like this :

```
"azure_key_vault_config": {
  "azure_environment": "",
  "client_id": "",
  "enabled": false,
  "key_identifier": "",
  "key_vault_name": "",
  "require_private_networking": false,
  "resource_group_name": "",
  "role_id": null,
  "secret": null,
  "subscription_id": "",
  "tenant_id": "",
  "valid": false
}
```

### Change under test
I changed the mapping in `internal/service/encryptionatrest/model.go` (`NewTFAzureKeyVaultConfigItem`) from:

```
ClientID:         types.StringValue(az.GetClientID()),
TenantID:         types.StringValue(az.GetTenantID()),
```

to:

```
ClientID:         conversion.StringNullIfEmpty(az.GetClientID()),
TenantID:         conversion.StringNullIfEmpty(az.GetTenantID()),
```

### Behavior after the change
After rebuilding the provider, I ran `terraform plan` again with **no changes to the config**, and got:

```
No changes. Your infrastructure matches the configuration.
```

### State behavior after the change
After the `terraform plan`, I inspected the state file and saw the provider normalized those fields from `""` to `null`:

```
json
"azure_key_vault_config": {
"azure_environment": "",
"client_id": null,
"enabled": false,
"key_identifier": "",
"key_vault_name": "",
"require_private_networking": false,
"resource_group_name": "",
"role_id": null,
"secret": null,
"subscription_id": "",
"tenant_id": null,
"valid": false
}
```

### Conclusion
In this scenario, the provider silently normalizes state for `azure_key_vault_config.client_id` and `azure_key_vault_config.tenant_id` from `""` to `null`, and this does not force an apply (the plan remains empty).

### Additional observation
There could be an edge case is if a user is reading these attributes from state (e.g., via outputs)

But in practical terms, if Azure isn’t configured/enabled, it generally wouldn’t make sense to reference `azure_key_vault_config.client_id` / `tenant_id` downstream. Additionally, a customer cannot use Azure EAR without providing authentication credentials.

marcosuma: ack! this doesn't seem a breaking change then, but let's bring this up to tech sync

marcosuma (commented):
LGTM after we've confirmed behaviour during tech-sync and docs comments are addressed

internal/service/encryptionatrest/resource_test.go
marcabreracast: Follow-up: `Manual Tests for Migration - From Static Credentials to Serverless` section included in the PR description

docs/data-sources/encryption_at_rest.md
marcabreracast: Changes addressed in [f123874](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140/commits/f12387411bfc86952f1a67180b2171b0f422b963), however I did a quick search in the codebase (not filtering auto generated resources), and it found over 4K occurrences of "MongoDB Cloud" 😅

internal/service/encryptionatrest/resource.go
marcabreracast: Addressed in [f123874](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140/commits/f12387411bfc86952f1a67180b2171b0f422b963)

internal/service/encryptionatrest/data_source_schema.go
marcabreracast: Addressed in [f123874](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140/commits/f12387411bfc86952f1a67180b2171b0f422b963)

docs/resources/encryption_at_rest.md
marcabreracast: Addressed in [f123874](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140/commits/f12387411bfc86952f1a67180b2171b0f422b963)

internal/service/encryptionatrest/model.go
marcabreracast: Resolving comment in favor of different approach, this change has been reverted in [aa0726a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4140/commits/aa0726a47bffeb38ec7b51e31e330935d7d0f27b)

internal/service/encryptionatrest/model.go
marcabreracast: Based on latest alignment with the team, we've decided to go with the approach of `Computed + Optional` as it proposes no breaking changes in the data source. Resolving this conversation so we can focus on the updated changes

internal/service/encryptionatrest/resource_test.go
marcabreracast: Resolving conversation as this request has been addressed.

internal/service/encryptionatrest/model.go
```diff
@@ -147,15 +148,16 @@ func NewAtlasAzureKeyVault(tfAzKeyVaultConfigSlice []TFAzureKeyVaultConfigModel)
 
 	return &admin.AzureKeyVault{
 		Enabled:                  v.Enabled.ValueBoolPointer(),
-		ClientID:                 v.ClientID.ValueStringPointer(),
+		ClientID:                 conversion.NilForUnknownOrEmptyString(v.ClientID),
```
marcabreracast, line 151: Since client_id and tenant_id are Optional + Computed, they can be unknown at plan time. `ValueStringPointer()/ValueString()` return `&""` for unknown strings, so the provider sends "" for these fields and Atlas returns `INVALID_AZURE_KEY_VAULT_AUTH_CONFIG` when both explicit credentials (with "") and a Cloud Provider Access `roleId` are present.

Using `NilForUnknownOrEmptyString` converts unknown/empty values to nil, so these fields are omitted from the Azure config unless explicitly set. This is non-breaking and only affects request serialization: we no longer send empty strings for client_id/tenant_id in the API request, while existing configs with non-empty values continue to behave the same.

marcabreracast: CC: @maastha for context, as we discussed this behaviour and validated that there won't be breaking changes

elyse-mdb (approved):

marcosuma (approved):

AgustinBettati (approved):
LGTM

PR #4165, opened 2026-02-13
chore: Add discriminator metadata in intermediate model of autogenerated resources

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-379812

Implements the OpenAPI-to-intermediate-model mapping for polymorphic discriminator metadata. This work enables future plan-phase validation of polymorphic types by ensuring that discriminator information from the `x-xgen-discriminator` OpenAPI extension is captured, transformed, and persisted in the intermediate model used by codegen.

**Note**: Proper aliasing management within discriminator model is handled in separate PR.

### Key changes

- **Intermediate model types:** Introduced `Discriminator` and `DiscriminatorType` structs in `model.go`, carrying `property_name` and a `mapping` of each discriminator value to its `allowed` and `required` attribute lists. The `Discriminator` field is added to both `Schema` (root-level) and `NestedAttributeObject` (for nested polymorphic types).

- **Extraction logic (`discriminator.go`):**
  - Filters `readOnly` properties from `required` (they map to `Computed` in TF and cannot be set in configuration) while keeping them in `allowed`.
  - Returns `nil` when all variants have empty allowed lists (value-level polymorphism with no structural differences, e.g. enum-only discriminators).

- **Request/Response merging (`MergeDiscriminators`):** Discriminator data is extracted from both request and response schemas and merged: `allowed` is the union of both sources (to capture response-only computed properties), while `required` is sourced exclusively from the request discriminator (response properties are `Computed`).

- **Nested attribute propagation (`attribute.go`, `merge_attributes.go`):** Discriminator extraction and merging is wired into handling of nested attributes same as how root level discriminators are handled.

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
[...]

PR #4175, opened 2026-02-17
chore: Adding aliasing support in discriminator model

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-379812 (follow up from https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4165)

### Relevant changes:

- New applyAliasesToDiscriminators() function walks the schema tree after attribute transformations are applied and renames `discriminator` PropertyName and variant Allowed/Required entries to match any configured aliases. Supports both root-level and path-based nested aliases.
- Opportunistic refactor: Remove implicit root-level aliases propagating to nested-level attributes. Previously, non-dotted aliases (e.g. `name: connectionName`) applied at every nesting level. This was problematic because it could silently rename unrelated nested fields that happened to share a name with a root-level alias, while also introducing more complexity in the implementation. Now aliases are strictly path-scoped: a non-dotted alias like groupId: projectId only matches at root level (where apiPath == "groupId"), while nested attributes require an explicit dotted alias (e.g., results.id: integrationId). This aligns applyAliasesToDiscriminator with the same deterministic lookup pattern.
- Modify discriminator model to preserve orginal API Name values (camel case). This enables simplifying aliasing logic while keeping traceability of original names.

### Testing

Added unit testing for existing aliasing logic (we did not have before).

## Type of change: 

## Required Checklist:

### DIFF
--- a/tools/codegen/codespec/config.go
+++ b/tools/codegen/codespec/config.go
@@ -442,19 +442,12 @@ func applyAliasesToDiscriminator(disc *Discriminator, aliases map[string]string,
 	}
 }
 
-// resolveAliasAtLevel constructs the full alias key from an API name and the parent path,
-// then performs a direct lookup in the aliases map.
-func resolveAliasAtLevel(apiName string, aliases map[string]string, parentAPIPath string) (string, bool) {
-	fullPath := apiName
-	if parentAPIPath != "" {
-		fullPath = parentAPIPath + "." + apiName
-	}
-	alias, ok := aliases[fullPath]
-	return alias, ok
-}
-
 func applyAliasToAttributeName(name *AttributeName, aliases map[string]string, parentAPIPath string) {
-	if alias, ok := resolveAliasAtLevel(name.APIName, aliases, parentAPIPath); ok {
+	fullPath := name.APIName
+	if parentAPIPath != "" {
+		fullPath = parentAPIPath + "." + name.APIName
+	}
+	if alias, ok := aliases[fullPath]; ok {
 		name.TFSchemaName = stringcase.ToSnakeCase(alias)
 	}
 }
--- a/tools/codegen/codespec/config.go
+++ b/tools/codegen/codespec/config.go
@@ -22,7 +22,7 @@ func ApplyTransformationsToResource(resourceConfig *config.Resource, resource *R
 		return nil
 	}
 	parentPaths := &attrPaths{schemaPath: "", apiPath: ""}
-	if err := applyAttributeTransformationsList(resourceConfig.SchemaOptions, &resource.Schema.Attributes, parentPaths, transformations); err != nil {
+	if err := applyAttributeTransformationsList(resourceConfig.SchemaOptions, &resource.Schema.Attributes, parentPaths, resourceTransformations); err != nil {
 		return fmt.Errorf("failed to apply attribute transformations: %w", err)
 	}
 	applyCommonAliasTransformations(&resource.Operations, resourceConfig.SchemaOptions.Aliases, resource.Schema.Discriminator, &resource.Schema.Attributes)
@@ -109,7 +109,7 @@ type attrPaths struct {
 // Implementations may mutate the attribute in-place.
 type AttributeTransformation func(attr *Attribute, paths *attrPaths, schemaOptions config.SchemaOptions) error
 
-var transformations = []AttributeTransformation{
+var resourceTransformations = []AttributeTransformation{
 	aliasTransformation,
 	commonRSAndDSOverridesTransformation,
 	immutableComputedOverrideTransformation,
--- a/tools/codegen/codespec/config.go
+++ b/tools/codegen/codespec/config.go
@@ -191,7 +191,7 @@ func shouldIgnoreAttribute(attrPathName string, ignoredAttrs map[string]bool) bo
 	return ignoredAttrs[attrPathName]
 }
 
-func applyAliasToAttribute(attr *Attribute, paths *attrPaths, schemaOptions config.SchemaOptions) {
+func aliasTransformation(attr *Attribute, paths *attrPaths, schemaOptions config.SchemaOptions) error {
 	// Config uses full camelCase for aliases (e.g., groupId: projectId, nestedObject.innerAttr: renamedAttr)
 	// The apiPath is built from APIName values which preserve the original casing (e.g., "MongoDBMajorVersion")
 	// This avoids the lossy conversion from snake to camel case.
@@ -214,13 +214,6 @@ func applyAliasToAttribute(attr *Attribute, paths *attrPaths, schemaOptions conf
 		// Note: apiPath is not updated because it's only used for alias lookup,
 		// and aliases are defined using original API names
 	}
-}
-
-// Transformations
-func aliasTransformation(attr *Attribute, paths *attrPaths, schemaOptions config.SchemaOptions) error {
-	// Alias transformation runs first, updating TFSchemaName and paths.schemaPath.
-	// Subsequent overrides should use the aliased snake_case path (e.g., nested_list_array_attr.inner_num_attr_alias)
-	applyAliasToAttribute(attr, paths, schemaOptions)
 	return nil
 }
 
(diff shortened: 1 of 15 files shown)


tools/codegen/models/stream_connection_api.yaml
```diff
@@ -379,8 +379,8 @@ schema:
                                 - string: {}
                                   description: Reserved. Will be used by PRIVATE_LINK connection type.
                                   computed_optional_required: optional
-                                  tf_schema_name: connection_name
-                                  tf_model_name: ConnectionName
+                                  tf_schema_name: name
+                                  tf_model_name: Name
```
AgustinBettati, line 406: Previously root level alias `name: connectionName` was impacting nested `networking.access.name` attribute.

tools/codegen/codespec/config.go
```diff
@@ -412,6 +418,114 @@ func attrPathForOverrides(attrPathName string) string {
...
+			} else if parentAPIPath == "" {
+				// At root level, also check path-based aliases without parent prefix
+				apiName = original
+			}
+		}
+		if apiName != "" {
+			oldSnake := stringcase.ToSnakeCase(apiName)
+			newSnake := stringcase.ToSnakeCase(alias)
+			if oldSnake != newSnake {
+				renameMap[oldSnake] = newSnake
```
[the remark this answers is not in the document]
AgustinBettati: Provided good input on resolving the aliases deterministically per nesting level

tools/codegen/codespec/config.go
```diff
@@ -16,39 +17,51 @@ const DeleteOnCreateTimeoutDescription = "Indicates whether to delete the resour
...
 	applyAliasToPathParams(&ds.Operations, dsConfig.SchemaOptions.Aliases)
 	return nil
 }
 
+func applyAliasToDiscriminator(aliases map[string]string, rootDiscriminator *Discriminator, attributes *Attributes) {
+	applyAliasesToDiscriminator(rootDiscriminator, aliases, "")
+	if attributes == nil {
+		return
+	}
+	applyAliasesToNestedDiscriminators(*attributes, aliases, "")
```
manupedrozo: Just a suggestion
```suggestion
	if attributes != nil {
        applyAliasesToNestedDiscriminators(*attributes, aliases, "")
	}
```

tools/codegen/codespec/config.go
```diff
@@ -192,15 +197,10 @@ func applyAliasToAttribute(attr *Attribute, paths *attrPaths, schemaOptions conf
...
-
-	// First try path-based lookup for targeted aliasing of nested attributes
-	// apiPath already contains the correct camelCase path built from APIName values
-	if aliasCamel, found = schemaOptions.Aliases[paths.apiPath]; !found {
-		// Fall back to attribute-name only lookup (for path params like groupId: projectId)
-		aliasCamel, found = schemaOptions.Aliases[attr.APIName]
-	}
+	// Lookup by full apiPath only. At root level apiPath equals attr.APIName so non-dotted
+	// aliases like "groupId: projectId" still match. At nested levels only explicitly
+	// path-scoped aliases (e.g., "nestedObj.innerAttr: renamedAttr") apply.
```
manupedrozo: We probably don't need so many details now as I think this is the expected behavior.
Would simplify:
```suggestion
	// Lookup by full apiPath only. We do not want to implicitly apply aliases to nested-level attributes.
```

tools/codegen/codespec/config.go
```diff
@@ -412,6 +412,104 @@ func attrPathForOverrides(attrPathName string) string {
...
+			applyAliasesToDiscriminator(attr.MapNested.NestedObject.Discriminator, aliases, apiPath)
+			applyAliasesToNestedDiscriminators(attr.MapNested.NestedObject.Attributes, aliases, apiPath)
+		}
+	}
+}
+
+// applyAliasesToDiscriminator reconciles discriminator property names and variant attribute names
+// when aliases have been applied. It renames PropertyName and all entries in Allowed/Required lists
+// according to the aliases applicable at this nesting level.
+func applyAliasesToDiscriminator(disc *Discriminator, aliases map[string]string, parentAPIPath string) {
```
manupedrozo, line 431: I don't quite get why we need to build a map per nesting level for discriminators when we don't need it for aliasing attributes. 
Is it because it happens after the initial aliasing? If so, would it be simpler to do both together?

tools/codegen/codespec/config.go
AgustinBettati: Makes sense, applying suggestion directly

tools/codegen/codespec/config.go
AgustinBettati: > why we need to build a map per nesting level for discriminators when we don't need it for aliasing attributes

Main difference is that in common aliasing `Attribute` model has the different string cases available, so for applying aliasing it compares the config alias value with `attr.APIName`.
In case of discriminator model we are only storing snake_case values, so when it comes to applying aliasing the renamingMap stores the aliasing to apply using snake_case to snake_case.

manupedrozo: Discussed offline the following alternatives: 
A. Store the original discriminator names (camelCase) as well, so we can use them for these operations. 
B. Store the aliases in snake_case and use them instead.

I prefer A in order to avoid edge cases of the camelCase -> snake_case conversion E.g. two camelCase strings can convert into the same snake_case.

The goal is mainly simplicity which should also lead to a performance improvement.

tools/codegen/models/stream_connection_api.yaml
```diff
@@ -4,43 +4,66 @@ schema:
...
+                      tf_schema_name: aws
             Cluster:
                 allowed:
-                    - cluster_group_id
-                    - cluster_name
-                    - db_role_to_execute
+                    - api_name: clusterName
+                      tf_schema_name: cluster_name
+                    - api_name: clusterGroupId
+                      tf_schema_name: cluster_project_id
```
AgustinBettati, line 20: example where aliasing is being applied with discriminator model, api_name still preserves orginal camel case name from API Spec

tools/codegen/codespec/config.go
AgustinBettati: Went forward with option A, leaving `config.go` with simple logic for apply aliasing as original API Name is available

tools/codegen/codespec/config.go
```diff
@@ -412,6 +401,71 @@ func attrPathForOverrides(attrPathName string) string {
...
+	}
+	alias, ok := aliases[fullPath]
+	return alias, ok
+}
+
+func applyAliasToAttributeName(name *AttributeName, aliases map[string]string, parentAPIPath string) {
+	if alias, ok := resolveAliasAtLevel(name.APIName, aliases, parentAPIPath); ok {
+		name.TFSchemaName = stringcase.ToSnakeCase(alias)
+	}
+}
```
manupedrozo: If you want to inline resolveAliasAtLevel():
```suggestion
func applyAliasToAttributeName(name *AttributeName, aliases map[string]string, parentAPIPath string) {
    fullPath := name.APIName
    if parentAPIPath != "" {
        fullPath = parentAPIPath + "." + name.APIName
    }
    if alias, ok := aliases[fullPath]; ok {
        name.TFSchemaName = stringcase.ToSnakeCase(alias)
    }
}
```

tools/codegen/codespec/model.go
```diff
@@ -100,16 +100,23 @@ type Schema struct {
 	Attributes Attributes `yaml:"attributes"`
 }
 
+// AttributeName pairs the original API property name with the Terraform schema name.
+// TFSchemaName will have aliasing configurations applied.
+type AttributeName struct {
```
manupedrozo: May want to avoid using Attribute in the name (or add Discriminator prefix) to more clearly distinguish between discriminator properties and schema attributes. For example we are now introducing `applyAliasToAttributeName()` which is a bit confusing.

manupedrozo (approved):
Thanks for following up!

tools/codegen/codespec/config.go
AgustinBettati: good catch

tools/codegen/codespec/model.go
AgustinBettati: makes sense, adjusted struct name and associated functions

PR #4189, opened 2026-02-19
chore: Implement string validator for discriminator sibling attributes

AgustinBettati:
## Description

Link to any related issue(s): CLOUDP-381487

Implements a config-phase string validator for discriminator sibling attributes. Given a discriminator attribute (e.g., `type`) and a mapping of variant values to their allowed/required sibling attributes, the validator enforces:

- **Required sibling check**: attributes listed as `Required` for the active variant must be present (non-null) in the config.
- **Disallowed sibling check**: type-specific attributes belonging to *other* variants must not be set.
- If the discriminator value itself is unknown, null, or not found in the mapping, all checks are skipped.

### Key consideration: Handling of unknown values in config

During `ValidateResourceConfig`, the config reflects exactly what the practitioner wrote:
- **null** = attribute was not set
- **unknown** = attribute was set to an unresolved expression (e.g., `cluster_name = mongodbatlas_cluster.test.name`)
- **known** = attribute was set to a literal or resolved expression

Unset `Optional+Computed` attributes are **null** in the config at this phase — they only become unknown later during `PlanResourceChange` when the framework marks them so the provider can compute a value.

Given this, both checks use a null-only check (`isNullInConfig`):
- **Required siblings**: error only if null. Unknown means the user set it to a reference — no false positive.
- **Disallowed siblings**: error if non-null (known *or* unknown). Unknown means the user explicitly set a disallowed attribute — correctly flagged.

<details>
<summary><h3>Manual testing<h3></summary>

The validator was temporarily wired to the `mongodbatlas_stream_connection` resource's type attribute to verify behavior against real terraform plan runs. The discriminator was configured with Cluster (requiring cluster_name) and Kafka (requiring bootstrap_servers) variants.

#### Test 1: Unknown required sibling (false positive prevention)
Config where cluster_name references another resource, making it unknown at validation time:

```
resource "mongodbatlas_stream_connection" "test" {
  project_id      = mongodbatlas_project.test.id
  instance_name   = mongodbatlas_stream_instance.test.instance_name
  connection_name = "test-cluster-conn"
  type            = "Cluster"
  cluster_name    = mongodbatlas_cluster.test.name  # unknown at validation
}
```

Result: No false positive. The validator correctly accepts the unknown value as "user set this attribute".

#### Test 2: Unset Optional+Computed disallowed sibling
Config with type = "Cluster" where networking (an Optional+Computed attribute from the Kafka variant) is not set by the user:
```
resource "mongodbatlas_stream_connection" "test" {
  project_id      = mongodbatlas_project.test.id
  instance_name   = mongodbatlas_stream_instance.test.instance_name
  connection_name = "test-cluster-conn"
  type            = "Cluster"
  cluster_name    = "my-cluster"
  # networking is Optional+Computed, belongs to Kafka variant, not set here
}
```

Result: No false positive. Unset Optional+Computed attributes appear as null (not unknown) in the config during validation, so the disallowed check correctly ignores them.
</details>

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/common/autogen/validator/string_discriminator_validator.go
+++ b/internal/common/autogen/validator/string_discriminator_validator.go
@@ -26,7 +26,7 @@ type DiscriminatorDefinition struct {
 //   - If the discriminator value is unknown, null, or not found in Mapping all checks are skipped.
 //   - Required attributes for the active variant must be non-null (unknown is accepted as "set").
 //   - Type-specific attributes from other variants must be null.
-//   - Note: Unset Optional+Computed attributes are null in the config, they only become unknown later duringPlanResourceChange
+//   - Note: Unset Optional+Computed attributes are null in the config, they only become unknown later during PlanResourceChange
 func ValidateDiscriminator(def DiscriminatorDefinition) schemavalidator.String {
 	return discriminatorValidator{def: def}
 }
--- a/internal/common/autogen/validator/string_discriminator_validator.go
+++ b/internal/common/autogen/validator/string_discriminator_validator.go
@@ -21,11 +21,12 @@ type DiscriminatorDefinition struct {
 	Mapping map[string]VariantDefinition
 }
 
-// ValidateDiscriminator returns a plan-phase string validator that checks sibling
+// ValidateDiscriminator returns a config-phase string validator that checks sibling
 // attribute presence/absence based on the active discriminator value.
 //   - If the discriminator value is unknown, null, or not found in Mapping all checks are skipped.
-//   - Required attributes for the active variant must be non-null.
+//   - Required attributes for the active variant must be non-null (unknown is accepted as "set").
 //   - Type-specific attributes from other variants must be null.
+//   - Note: Unset Optional+Computed attributes are null in the config, they only become unknown later duringPlanResourceChange
 func ValidateDiscriminator(def DiscriminatorDefinition) schemavalidator.String {
 	return discriminatorValidator{def: def}
 }
@@ -61,7 +62,7 @@ func (v discriminatorValidator) ValidateString(_ context.Context, req schemavali
 
 	for _, name := range variant.Required {
 		siblingPath := parentPath.AtName(name)
-		if isNullOrUnknownInConfig(req.Config, siblingPath) {
+		if isNullInConfig(req.Config, siblingPath) {
 			resp.Diagnostics.AddAttributeError(
 				req.Path,
 				"Missing Required Attribute",
@@ -75,7 +76,7 @@ func (v discriminatorValidator) ValidateString(_ context.Context, req schemavali
 			continue
 		}
 		siblingPath := parentPath.AtName(name)
-		if !isNullOrUnknownInConfig(req.Config, siblingPath) {
+		if !isNullInConfig(req.Config, siblingPath) {
 			resp.Diagnostics.AddAttributeError(
 				req.Path,
 				"Invalid Attribute Combination",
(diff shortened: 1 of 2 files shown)


internal/common/autogen/validator/string_discriminator_validator.go
```diff
@@ -0,0 +1,149 @@
...
+
+	allTypeSpecific := allTypeSpecificAttrs(v.def)
+	activeAllowed := toSet(variant.Allowed)
+
+	for _, name := range variant.Required {
+		siblingPath := parentPath.AtName(name)
+		if isNullInConfig(req.Config, siblingPath) {
+			resp.Diagnostics.AddAttributeError(
+				req.Path,
+				"Missing Required Attribute",
```
marcosuma, line 76: just a note for confirmation: I am guessing if API changes a field from required to optional, we will continue to ask it as required for older TF versions, but this is also true for a manually generated resource when we set "required" in the schema. Am I right?

internal/common/autogen/validator/string_discriminator_validator.go
```diff
@@ -0,0 +1,149 @@
...
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	schemavalidator "github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// VariantDefinition describes which sibling attributes are allowed and required when a particular discriminator value is active.
```
manupedrozo, line 13: Curious on the "sibling" term, does it come from somewhere or are you introducing it?
- Do we need a term or is just "attributes" enough? I.e. "VariantDefinition describes which attributes are allowed and required when a particular discriminator value is active."

internal/common/autogen/validator/string_discriminator_validator.go
```diff
@@ -0,0 +1,149 @@
...
+	parentPath := req.Path.ParentPath()
+	discriminatorName := lastPathStepName(req.Path)
+
+	allTypeSpecific := allTypeSpecificAttrs(v.def)
+	activeAllowed := toSet(variant.Allowed)
+
+	for _, name := range variant.Required {
+		siblingPath := parentPath.AtName(name)
+		if isNullInConfig(req.Config, siblingPath) {
+			resp.Diagnostics.AddAttributeError(
```
manupedrozo, line 74: Nice that we accumulate errors

internal/common/autogen/validator/string_discriminator_validator.go
```diff
@@ -0,0 +1,149 @@
...
+	variant, ok := v.def.Mapping[discriminatorValue]
+	if !ok {
+		return
+	}
+
+	parentPath := req.Path.ParentPath()
+	discriminatorName := lastPathStepName(req.Path)
+
+	allTypeSpecific := allTypeSpecificAttrs(v.def)
+	activeAllowed := toSet(variant.Allowed)
```
manupedrozo, line 69: I think you can simplify a bit by creating a set of the attributes that are NOT allowed, and if the config has any of them set, then error.
Ok to keep as is, up to you.

internal/common/autogen/validator/string_discriminator_validator.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func pathToTFTypesPath(p path.Path) *tftypes.AttributePath {
+	result := tftypes.NewAttributePath()
+	for _, step := range p.Steps() {
+		switch s := step.(type) {
+		case path.PathStepAttributeName:
+			result = result.WithAttributeName(string(s))
+		case path.PathStepElementKeyInt:
+			result = result.WithElementKeyInt(int(s))
```
manupedrozo: Double checking: int() or int64()?

internal/common/autogen/validator/string_discriminator_validator.go
```diff
@@ -0,0 +1,149 @@
...
+		return tftypes.Value{}, false
+	}
+	val, ok := rawVal.(tftypes.Value)
+	return val, ok
+}
+
+func pathToTFTypesPath(p path.Path) *tftypes.AttributePath {
+	result := tftypes.NewAttributePath()
+	for _, step := range p.Steps() {
+		switch s := step.(type) {
```
manupedrozo: Just to make sure: are all types covered? Should we error on `default:`?

manupedrozo (approved):
Nice! Left some comments. None blocking.

marcosuma (approved):

internal/common/autogen/validator/string_discriminator_validator.go
AgustinBettati: > if API changes a field from required to optional, we will continue to ask it as required for older TF versions

Correct. This the same as how any **non-polymorphic resource** (autogenerated or manual) behaves, where required attribute validations are enforced in the schema definition directly.

With validator in place only UX difference between non-polymorphic resource required attributes and polymorphic resource type-specific required attributes will be in tooling (e.g. VSCode linting).

internal/common/autogen/validator/string_discriminator_validator.go
AgustinBettati: Sibling is coming from the fact that this validator is defined on a particular attribute (the discriminator, e.g. "type"), so the validation is effectively applied over sibling attributes. Don't see harm in leaving the additional detail.

internal/common/autogen/validator/string_discriminator_validator.go
AgustinBettati: `WithElementKeyInt` effectively receives an `int`, confirmed `AttributePath` doesn't have a function supporting int64

internal/common/autogen/validator/string_discriminator_validator.go
AgustinBettati: Found there was a missing type (specifically if nested set is used within the path of the nested attribute). 
Added additional unit testing around usage of validator nested within different paths, and also explicitly handle case in which validator fails to obtain attributes of the parent by logging details and not returning errors (for caution). To achieve this refactored so that parent attributes are all fetched in a single function call.
Thanks for bringing this up!

PR #4240, opened 2026-02-27
chore(provider): Uses env base URL and Realm base URL when unset in provider

EspenAlbert:
## Description

When `base_url` (or `realm_base_url`) is not set in the provider block or in the chosen credentials (e.g. from AWS Secrets Manager), the provider now fills it from the environment so that `MONGODB_ATLAS_BASE_URL` and `MONGODB_REALM_BASE_URL` apply to all API operations. This allows using a single env var (e.g. in CI or for dev/private endpoints) with provider blocks or aliases that only set credentials.

Provider- or AWS-set `base_url` and `realm_base_url` are never overwritten by env. `MCLI_OPS_MANAGER_URL` remains a fallback env var for `base_url` for compatibility.

Link to any related issue(s): CLOUDP-384262

## Type of change:

## Required Checklist:

## Design choice: where to merge base URL from env

**Chosen:** In `GetCredentials`, after `CoalesceCredentials` returns, call `mergeBaseURLFromEnv(c, envVars)` so empty `BaseURL`/`RealmBaseURL` are filled from env.

**Alternative:** Change `CoalesceCredentials` to merge unset variables (e.g. after picking the winning credential set, back-fill empty BaseURL/RealmBaseURL from another set such as env).

**Pros of current approach:** Coalesce stays single-purpose (pick first credential set by auth); “env backs fill when config base_url empty” stays an explicit provider rule in one place; no signature or semantic change to `CoalesceCredentials`; easy to extend (e.g. more env-backed fields) without touching coalesce.

**Cons of alternative:** Coalesce would need to know “which arg is env” and which fields to merge, mixing generic coalesce with provider-specific fallback; per-field coalesce (variant where every field is “first non-empty”) would change semantics and could mix auth from one source with URLs from another.

### DIFF
--- a/docs/guides/provider-configuration.md
+++ b/docs/guides/provider-configuration.md
@@ -143,8 +143,8 @@ provider "mongodbatlas" {
 * `access_token` - (Optional) SA Access Token (env: `MONGODB_ATLAS_ACCESS_TOKEN`). Instead of using Client ID and Client Secret, you can generate and use an SA token directly. See [Generate Service Account Token](https://www.mongodb.com/docs/atlas/api/service-accounts/generate-oauth2-token/#std-label-generate-oauth2-token-atlas) for details. Note: tokens have expiration times.
 * `public_key` - (Optional) PAK Public Key (env: `MONGODB_ATLAS_PUBLIC_API_KEY`).
 * `private_key` - (Optional) PAK Private Key (env: `MONGODB_ATLAS_PRIVATE_API_KEY`).
-* `base_url` - (Optional) MongoDB Atlas Base URL (env: `MONGODB_ATLAS_BASE_URL`). For advanced use cases, you can configure custom API endpoints.
-* `realm_base_url` - (Optional) MongoDB Realm Base URL (env: `MONGODB_REALM_BASE_URL`).
+* `base_url` - (Optional) MongoDB Atlas Base URL (env: `MONGODB_ATLAS_BASE_URL`, `MCLI_OPS_MANAGER_URL`). For advanced use cases, you can configure custom API endpoints. When not set in the provider block, the provider uses the value from the environment for all API operations. This allows using a single env var (e.g. `MONGODB_ATLAS_BASE_URL=https://cloud-dev.mongodb.com`) with provider blocks that only set credentials.
+* `realm_base_url` - (Optional) MongoDB Realm Base URL (env: `MONGODB_REALM_BASE_URL`). When not set in the provider block, the value from the environment is used for all Realm API operations.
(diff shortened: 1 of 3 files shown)


docs/guides/provider-configuration.md
```diff
@@ -143,8 +143,8 @@ provider "mongodbatlas" {
...
 * `public_key` - (Optional) PAK Public Key (env: `MONGODB_ATLAS_PUBLIC_API_KEY`).
 * `private_key` - (Optional) PAK Private Key (env: `MONGODB_ATLAS_PRIVATE_API_KEY`).
-* `base_url` - (Optional) MongoDB Atlas Base URL (env: `MONGODB_ATLAS_BASE_URL`). For advanced use cases, you can configure custom API endpoints.
-* `realm_base_url` - (Optional) MongoDB Realm Base URL (env: `MONGODB_REALM_BASE_URL`).
+* `base_url` - (Optional) MongoDB Atlas Base URL (env: `MONGODB_ATLAS_BASE_URL`, `MCLI_OPS_MANAGER_URL`). For advanced use cases, you can configure custom API endpoints. When not set in the provider block, the provider uses the value from the environment for all API operations. This allows using a single env var (e.g. `MONGODB_ATLAS_BASE_URL=https://cloud-dev.mongodb.com`) with provider blocks that only set credentials.
```
EspenAlbert, line 146: Not sure if we should document the `MCLI_OPS_MANAGER_URL`.
Also, can consider removing this extra docs update as mostly this is an internal use case

EspenAlbert:
Using `skip-changelog-check` as this is more an internal change

oarbusi:
nit: remove `(provider)` from PR title

oarbusi (commented):
I think we intentionally don't mix env and provider inputs here since we refactored this. Can cause unintended issues if you think are using an environment but due to an env var you are using another. Also I am not sure what real value we are getting out of this, the motivation of this is make some tests in a module work, I don't think we should change this just to make some tests work

internal/config/credentials.go
```diff
@@ -20,20 +20,35 @@ type Credentials struct {
 }
 
 // GetCredentials follows the order of AWS Secrets Manager, provider vars and env vars.
+// When the chosen credentials have empty BaseURL or RealmBaseURL, they are filled from env vars
+// so that MONGODB_ATLAS_BASE_URL (or MCLI_OPS_MANAGER_URL) and MONGODB_REALM_BASE_URL apply to all operations.
```
lantoli, line 24: we removed explicitly the possibility to combine values from different sources, I would keep it as it because it was confusing to get values from different sources (AWS Secret Manager, provider attributes, env vars), more info: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/provider-configuration

i recommend to discuss offline in more detail

you can something similar with the current approach passing a TF env variable in provider attributes for example

lantoli (changes requested):

docs/guides/provider-configuration.md
manupedrozo: Probably shouldn't mention https://cloud-dev.mongodb.com either. Would evaluate mentioning gov.

antellezr-mdb (approved):
LGTM

EspenAlbert:
Decided to not merge this in discussion offline

PR #4274, opened 2026-03-10
feat: Add mongodbatlas_service_account_jwt ephemeral resource

maastha:
## Description

Implement the `mongodbatlas_service_account_jwt` ephemeral resource for generating short-lived JWT access tokens from MongoDB Atlas Service Account credentials. Key features:

- **Credential resolution chain**: resource attributes take precedence over provider-level Service Account credentials, with clear error messages when credentials are incomplete or the provider uses a different auth method (PAK/Access Token).
- **Token revocation on closure**: when `revoke_on_closure` is set to true, the access token is revoked when the Terraform operation ends. Token data is passed via private data between Open and Close.
- **Schema**: exposes `client_id`, `client_secret` (optional inputs), `revoke_on_closure` (optional), and computed `access_token`, `token_type`, `expires_in`.
- **Unit tests**: cover credential resolution ordering, partial credential errors, and non-SA provider auth scenarios.
- Exports config.UserAgent() for reuse 
- adds a changelog entry for the new ephemeral resource.

Follow-ups: 
- [CLOUDP-387290](https://jira.mongodb.org/browse/CLOUDP-387290) will add acceptance tests
- If user-agent logic looks good, will look into centralizing it for the provider for calls to Oauth endpoints (token generation & revocation)

Link to related issue(s): CLOUDP-387287, CLOUDP-387288

## Type of change:

## Required Checklist:

## Further comments

This PR stacks on #4273 which adds the ephemeral resource provider infrastructure.

User agent testing:
```
2026/03/11 01:38:57 [DEBUG] Network Request Start: POST https://xxxx.mongodb.com/api/oauth/token (started at 2026-03-11T01:38:57.075476+01:00) User-Agent: terraform-provider-mongodbatlas/devel Terraform/1.10.0 Name/service_account_jwt Operation/open

2026/03/11 02:34:59 [DEBUG] Network Request Start: POST https://xxxx.mongodb.com/api/oauth/revoke (started at 2026-03-11T02:34:59.208455+01:00) User-Agent: terraform-provider-mongodbatlas/devel Terraform/1.10.0 Name/service_account_jwt Operation/close
```

Changelog test:
image


internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,163 @@
...
+
+	revokeData, err := json.Marshal(closeData{
+		AccessToken:   token.AccessToken,
+		ClientID:      clientID,
+		ClientSecret:  clientSecret,
+		BaseURL:       baseURL,
+		RevokeOnClose: model.RevokeOnClosure.ValueBool(),
+	})
+	if err == nil {
+		resp.Private.SetKey(ctx, closeDataKey, revokeData)
```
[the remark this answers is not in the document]
maastha: Fixed in 95b087d. Now handles marshal errors with a warning diagnostic and propagates SetKey diagnostics.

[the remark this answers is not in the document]
maastha: Fixed in 95b087d. Private data is now only written when revoke_on_closure is true.

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,163 @@
...
+	resp.Diagnostics.Append(resp.Result.Set(ctx, &model)...)
+}
+
+func (r *es) Close(ctx context.Context, req ephemeral.CloseRequest, resp *ephemeral.CloseResponse) {
+	raw, diags := req.Private.GetKey(ctx, closeDataKey)
+	resp.Diagnostics.Append(diags...)
+	if len(raw) == 0 {
+		log.Printf("[DEBUG] %s Close: no private state found (key=%q), skipping revocation", resourceTypeName, closeDataKey)
+		return
+	}
```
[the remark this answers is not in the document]
maastha: Fixed in 95b087d. Close() now returns early if GetKey diagnostics contain errors.

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,163 @@
...
+	// 1. Resource attributes (explicit client_id and client_secret on the ephemeral resource block).
+	id := strings.TrimSpace(model.ClientID.ValueString())
+	secret := strings.TrimSpace(model.ClientSecret.ValueString())
+	if id != "" && secret != "" {
+		return id, secret, providerBaseURL(erd), diags
+	} else if id != "" || secret != "" {
+		diags.AddError("Invalid Service Account credentials",
+			"When setting credentials on this ephemeral resource, both client_id and client_secret must be provided.")
+		return "", "", "", diags
+	}
```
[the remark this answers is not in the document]
maastha: Fixed in 95b087d. Added IsUnknown guard for client_id and client_secret before reading values.

internal/service/serviceaccountjwt/user_agent.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// withUserAgentClient injects an HTTP client into the context via
+// auth.HTTPClient so the atlas-sdk-go token exchange picks it up.
+//
+// Transport chain:
+//
+//	baseUserAgentTransport  (sets "terraform-provider-mongodbatlas/<ver> Terraform/<ver>")
+//	  → UserAgentTransport  (appends "Name/service_account_jwt Operation/..." from context)
+//	    → NetworkLoggingTransport  (logs the complete User-Agent)
```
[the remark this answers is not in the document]
maastha: Fixed in 95b087d. Updated comment to accurately describe what NetworkLoggingTransport logs.

internal/service/serviceaccountjwt/user_agent.go
```diff
@@ -0,0 +1,51 @@
...
+// withUserAgentClient injects an HTTP client into the context via
+// auth.HTTPClient so the atlas-sdk-go token exchange picks it up.
+//
+// Transport chain:
+//
+//	baseUserAgentTransport  (sets "terraform-provider-mongodbatlas/<ver> Terraform/<ver>")
+//	  → UserAgentTransport  (appends "Name/service_account_jwt Operation/..." from context)
+//	    → NetworkLoggingTransport  (logs method/URL/timing/status)
+//	      → http.DefaultTransport
+func (r *es) withUserAgentClient(ctx context.Context) context.Context {
```
maastha, line 29: will consider centralizing it for the provider for calls to Oauth endpoints (token generation & revocation) in a follow-up if no concerns.

This also requires a fix in the SDK to capture UA during revoke https://github.com/mongodb/atlas-sdk-go/pull/750

internal/provider/provider.go
```diff
@@ -385,7 +387,14 @@ func (p *MongodbatlasProvider) Resources(context.Context) []func() resource.Reso
 }
 
 func (p *MongodbatlasProvider) EphemeralResources(context.Context) []func() ephemeral.EphemeralResource {
-	return nil
+	ephemeralResources := []func() ephemeral.EphemeralResource{
+		serviceaccountjwt.New,
+	}
+	analyticsEphemeralResources := []func() ephemeral.EphemeralResource{}
```
marcosuma: what do we mean with analytics ?  is it the wrapping logic to send user agent info? if so can we call it ephemeralResourcesWithAnalytics

internal/config/client.go
```diff
@@ -89,7 +89,7 @@ type RealmClient struct {
 }
 
 func NewClient(c *Credentials, terraformVersion string) (*MongoDBClient, error) {
-	userAgent := userAgent(terraformVersion)
+	userAgent := UserAgent(terraformVersion)
```
marcosuma, line 92: not clear why this has been made a public method? but it hints at incapsulation violation (unless the existing code is not ideal)

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+
+	token, err := r.generateToken(ctx, clientID, clientSecret, baseURL)
+	if err != nil {
+		resp.Diagnostics.AddError("Error generating Service Account JWT", err.Error())
+		return
+	}
+
+	model.AccessToken = types.StringValue(token.AccessToken)
+	model.TokenType = types.StringValue(token.Type())
```
marcosuma, line 71: curious why Type() and the others are not functions

internal/service/serviceaccountjwt/user_agent.go
marcosuma: I am trying to understand why this new logic on user agents - can you expand?

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+			BaseURL:      baseURL,
+		})
+		if err != nil {
+			resp.Diagnostics.AddWarning(
+				"Failed to prepare Service Account token revoke payload",
+				"The token will not be automatically revoked when the ephemeral resource is closed: "+err.Error())
+		} else {
+			resp.Diagnostics.Append(resp.Private.SetKey(ctx, closeDataKey, revokeData)...)
+		}
+	}
```
marcosuma, line 88: this code and the close implementation is not clear to me. Can you confirm we're revoking the token in the close method only if the attribute is true?

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+
+	resp.Diagnostics.Append(resp.Result.Set(ctx, &model)...)
+}
+
+func (r *es) Close(ctx context.Context, req ephemeral.CloseRequest, resp *ephemeral.CloseResponse) {
+	raw, diags := req.Private.GetKey(ctx, closeDataKey)
```
marcosuma: raw? maybe a clearer name?

erabil-mdb (approved):
LGTM!

internal/service/serviceaccountjwt/resource.go
maastha: AccessToken and ExpiresIn are plain exported fields on the `type Token struct {}`, Type() is a method because it has normalization logic https://github.com/golang/oauth2/blob/4d954e69a88d9e1ccb8439f8d5b6cbef230c4ef9/token.go#L70

internal/service/serviceaccountjwt/user_agent.go
maastha: The provider's own HTTP client can't be reused because its transport already sets its own Authorization header on every request (Bearer for SA/access-token, or digest for PAK). The token endpoint expects `Authorization: Basic {base64(client_id:client_secret)}`, so two transports will conflict over the same header. `withUserAgentClient` builds a minimal client that mirrors the provider's observability (User-Agent + logging) without any auth transport.

maastha: Also requesting review from @EspenAlbert and @lantoli  on this PR for any feedback on the client and analytics

internal/config/client.go
maastha: because the serviceaccountjwt ephemeral resource needs to build the same User-Agent string for its standalone HTTP client. The ephemeral resource doesn't use `NewClient()` since it needs a separate token and transport chain (independent of the provider's cached token source)

internal/provider/provider.go
maastha: >is it the wrapping logic to send user agent info?

yes

>if so can we call it ephemeralResourcesWithAnalytics

yes will update

internal/config/client.go
EspenAlbert: I think we could have reused the `NewClient` instead? Similar to how org does it, passing it a set of credentials?
https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-387287_service-account-jwt/internal/service/organization/resource_organization.go#L465

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+			ClientID:     clientID,
+			ClientSecret: clientSecret,
+			BaseURL:      baseURL,
+		})
+		if err != nil {
+			resp.Diagnostics.AddWarning(
+				"Failed to prepare Service Account token revoke payload",
+				"The token will not be automatically revoked when the ephemeral resource is closed: "+err.Error())
+		} else {
+			resp.Diagnostics.Append(resp.Private.SetKey(ctx, closeDataKey, revokeData)...)
```
AgustinBettati, line 86: We are storing sensitive information within `Private` state. Is there any risk? Curious how other CSP providers are handling the authentication within the `Close` operation and if private state is used.

internal/service/serviceaccountjwt/resource.go
maastha: >Can you confirm we're revoking the token in the close method only if the attribute is true?


yes. We check for` if model.RevokeOnClosure.ValueBool() `. revokeData is required to store the details that would be required by the Close() to revoke the token. The oauth API requires the access_token to be passed in the request to revoke it, that's why if RevokeOnClosure is true we need to provide these to Close() via privateData. Also, Close() req object only has access to `Private` and not the schema that's why we need to set these attributes.

internal/service/serviceaccountjwt/resource.go
maastha: The issue is that our revoke API requires the token itself along with clientID, and secret to revoke the token, that's why we need to persist the token until close. Ideally, if the revoke API allowed to use some kind of ID of the token, that would have been better.
That said, Ephemeral resource data is never written to the .tfstate file or plan artifacts. The private state is held in memory by the Terraform process and passed between lifecycle phases (Open -> Renew -> Close) **within a single operation**, then discarded.

marcosuma: > Curious how other CSP providers are handling the authentication

I'd be interested in this part too

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+func (r *es) resolveCredentials(model *TFModel) (clientID, clientSecret, baseURL string, diags diag.Diagnostics) {
+	erd := r.EphemeralResourceData
+
+	// 1. Resource attributes (explicit client_id and client_secret on the ephemeral resource block).
+	if model.ClientID.IsUnknown() || model.ClientSecret.IsUnknown() {
+		diags.AddError("Unknown credentials",
+			"client_id and client_secret must be known at apply time to generate a token.")
+		return "", "", "", diags
+	}
+	id := strings.TrimSpace(model.ClientID.ValueString())
```
lantoli, line 124: why are we trimming spaces?

internal/service/serviceaccountjwt/resource_test.go
```diff
@@ -0,0 +1,112 @@
+//nolint:testpackage // White-box tests for internal helper behavior.
+package serviceaccountjwt
```
lantoli, line 2: I don't think we should skip the rule to have separate test packages (e.g., serviceaccountjwt_test). We want to test behavior, not internal workings. Tests shouldn't be coupled to implementation details but to expected behavior

internal/config/client.go
lantoli: +1 to explore reuse of NewClient

.changelog/4274.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:new-ephemeral-resource
+ephemeral/mongodbatlas_service_account_jwt
```
lantoli, line 2: qq: do other providers differentiate ephemeral resource from regular resources in the changelog?

internal/service/serviceaccountjwt/resource_test.go
maastha: yes this has been updated in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4290

internal/service/serviceaccountjwt/resource.go
maastha: Looking at [GCP ephemeral resources guide](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/using_ephemeral_resources): all five listed ephemeral resources are Open-only. They generate credentials, return them as ephemeral results, and let them expire naturally. None use Close, and none attempt revocation.
AWS has a wrapper [internal/provider/framework/wrap.go](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/provider/framework/wrap.go#L338) that implements Close but no actual resources use it at this point.
Same for Azure, no actual resources use Close() at this point.

IMO since we have `revoke_on_closure` as optional and false by default our current implementation is good. Regarding privateData, Close() is invoked within the same operation as Open() (i.e. within the same tf plan/apply) so the privateData is persisted only for that duration in memory.

.changelog/4274.txt
maastha: yes. Some example I found:

https://github.com/hashicorp/terraform-provider-random/blob/main/CHANGELOG.md#371-february-25-2025 

https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#6360-march-11-2026

https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#680-august-7-2025

internal/config/client.go
maastha: `NewClient` can't be reused here because the ephemeral resource needs the raw access_token string to return to the user, while `NewClient` consumes the token internally and only exposes API clients (AtlasV2, etc.). The ephemeral resource has a fundamentally different goal than the org resource.

There are also other considerations:
1. Singleton token source: `getTokenSource` caches the first SA credentials. The ephemeral resource supports per-resource credentials.
2. Auth conflict: NewClient's transport injects Authorization: Bearer on every request, but the token endpoint expects Authorization: Basic. 
3. Revocation: with `revoke_on_closure = true` we revoke the token on `Close`. 

For context, we don't capture user-agent details today for the token generation & revocation endpoints. That's why this user-agent logic is new.

That said, as mentioned in my comment here https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4274#discussion_r2915495807 some of the transport setup could be centralized with the provider's own `getTokenSource` such as, a shared helper for the (User-Agent + logging) could serve both. This can be explored in a follow-up. That will also address not having to export `UserAgent`.

internal/provider/provider.go
maastha: updated to ephemeralResourcesWithAnalytics

internal/service/serviceaccountjwt/resource.go
maastha: renamed to `revokeData`

internal/service/serviceaccountjwt/resource.go
maastha: no strong reason, just to avoid accidental whitespace in the HCL config

internal/service/serviceaccountjwt/resource.go
marcosuma: is `The oauth API requires the access_token to be passed in the request to revoke it` something that only our API requires? if not, how are other providers doing it?

internal/service/serviceaccountjwt/resource.go
marcosuma: I don't know if this is possible, but what I can think of me is a security risk: a bad actor, logged in the machine where the customer `tf apply` is ran, intercepts the data stored in privateData and access to the token. Do we know more about privateData and what does TF recommends it to use it for? Based on your answer, other customers are not offering the revoke option, which means we're doing something different (hence why we need to be more cautious)

internal/service/serviceaccountjwt/resource.go
marcosuma: I think you answered below, other providers are not offering the revoke at all. Going keep this open until we make a decision

internal/service/serviceaccountjwt/user_agent.go
EspenAlbert: Thanks for the explanations across threads.
I think the centralizing of the calls to the Oauth endpoints should be prioritized.
Could it be that we are missing the `tfLoggingInterceptor` and should replace `http.DefaultTransport` with `config.baseTransport`?

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (r *es) Close(ctx context.Context, req ephemeral.CloseRequest, resp *ephemeral.CloseResponse) {
+	revokeData, diags := req.Private.GetKey(ctx, closeDataKey)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if len(revokeData) == 0 {
+		log.Printf("[DEBUG] %s Close: no private state found (key=%q), skipping revocation", resourceTypeName, closeDataKey)
```
EspenAlbert: isn't this going to be noisy? I think by default we don't revoke so this will be logged every time?

internal/service/serviceaccountjwt/resource.go
EspenAlbert: I think this is ok.
The private data lives in the Go process memory, not in any file. An attacker with process memory access can also read the provider's own saInfo global which holds the token source.
Ephemeral resource private data is not persisted to .tfstate or plan files

internal/service/serviceaccountjwt/resource_schema.go
```diff
@@ -0,0 +1,50 @@
...
+				MarkdownDescription: "When true, the access token is revoked when the Terraform operation ends. Defaults to `false`.",
+			},
+			"access_token": schema.StringAttribute{
+				Computed:            true,
+				Sensitive:           true,
+				MarkdownDescription: "The service account access token for authenticating API requests.",
+			},
+			"token_type": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "The mechanism for token authorization, always `Bearer`.",
```
EspenAlbert, line 33: Do we risk `always Bearer` becoming outdated?

EspenAlbert (approved):
Good job :clap: 
LGTM but I think we should prioritize centralizing the oauth endpoints calls as described in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4274#discussion_r2915495807

internal/service/serviceaccountjwt/resource.go
maastha: Bringing this to the tech sync today

internal/service/serviceaccountjwt/user_agent.go
maastha: >I think the centralizing of the calls to the Oauth endpoints should be prioritized.

agreed, I will share a CLOUDP for this to take up as a follow-up to this

internal/service/serviceaccountjwt/resource.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+	var data closeData
+	if err := json.Unmarshal(revokeData, &data); err != nil {
+		resp.Diagnostics.AddWarning("Failed to read revoke payload", err.Error())
+		return
+	}
```
AgustinBettati, line 107: Could the warning include the raw sensitive value? While this code seem unreachable (we control how revokeData is serialized) would not include `err.Error()` just to avoid any risk here

AgustinBettati (commented):
LGTM, we can finalize conversation of Close in sync meeting

marcosuma (approved):

internal/config/client.go
marcosuma: ok thanks

marcosuma (approved):
few comments to be addressed but overall seems like we should "centralise the transport setup" for a better coding style and single-responsibility principle

internal/service/serviceaccountjwt/resource.go
maastha: As discussed in tech sync, proceeding with current approach and checking with MongoDB Security and [HashiCorp](https://discuss.hashicorp.com/t/ephemeral-resource-close-storing-sensitive-credentials-in-private-state-for-token-revocation/77247). Resolving this.

internal/service/serviceaccountjwt/resource.go
maastha: Replaced `err.Error()` with a static message in 0b054e4.

internal/service/serviceaccountjwt/resource.go
maastha: Removed the DEBUG log line in 0b054e4. When revoke_on_closure is false (default), no private data is stored, so Close() silently returns.

internal/service/serviceaccountjwt/user_agent.go
maastha: Created https://jira.mongodb.org/browse/CLOUDP-390064. Resolving this.

maastha:
>few comments to be addressed but overall seems like we should "centralise the transport setup" for a better coding style and single-responsibility principle



addressed comments, created https://jira.mongodb.org/browse/CLOUDP-390064 centralizing the transport setup.

PR #4277, opened 2026-03-10
fix: Fixes `max_tier_size` of `mongodbatlas_stream_workspace` being dropped on read causing inconsistent result after apply 

### LINKED ISSUE #4272: stream_workspace: FromInstanceModel drops max_tier_size (always null) causing inconsistent result after apply
When using mongodbatlas_stream_workspace with stream_config.max_tier_size, Terraform reports "inconsistent result after apply" because the provider's Read function always returns max_tier_size = null, regardless of the actual value set in
   Atlas.

  Affected Resource(s)

  - mongodbatlas_stream_workspace

  Terraform Configuration

  resource "mongodbatlas_stream_workspace" "example" {
    project_id     = var.atlas_project_id
    workspace_name = "my-workspace"
    data_process_region = {
      region         = "eastus2"
      cloud_provider = "AZURE"
    }
    stream_config = {
      tier          = "SP2"
      max_tier_size = "SP30"
    }
  }

  Expected Behavior

  After terraform apply, the state should reflect stream_config.tier = "SP2" and stream_config.max_tier_size = "SP30", matching the values successfully applied to Atlas.

  Actual Behavior

  │
  │ When applying changes to mongodbatlas_stream_workspace.example,
  │ provider produced an inconsistent result. The value for
  │ stream_config.max_tier_size was "SP30" in the plan but null after apply.

  The workspace IS correctly created in Atlas with SP2 tier and SP30 max tier (confirmed in Atlas UI). The error is only in the provider's read-back.

  Root Cause

  The bug is in internal/service/streamworkspace/model.go, in the FromInstanceModel() function.

  The Read path works in 3 steps:

  1. Atlas API returns StreamsTenant with StreamConfig containing both Tier and MaxTierSize correctly.
  2. streaminstance.NewTFStreamInstance() correctly maps both fields:
  TFInstanceStreamConfigSpecModel{
      MaxTierSize: types.StringPointerValue(apiStreamConfig.MaxTierSize),  // ✅ correct
      Tier:        types.StringPointerValue(apiStreamConfig.Tier),         // ✅ correct
  }
  3. FromInstanceModel() converts the instance model to the workspace model but hardcodes max_tier_size to types.StringNull():
  "tier":          tierValue,           // ✅ preserves tier
  "max_tier_size": types.StringNull(),  // ❌ BUG: always null

  The fix should extract max_tier_size from the instance model's stream_config (just like tier is extracted) and pass it through instead of hardcoding null.

  Workaround

  lifecycle {
    ignore_changes = [stream_config]
  }

  Environment

  - Terraform: v1.9.5
  - Provider: mongodb/mongodbatlas v2.7.0
  - OS: Windows 11

oarbusi:
## Description

`FromInstanceModel()` hardcoded `max_tier_size` to null instead of reading it from the instance model attributes, causing "Provider produced inconsistent result after apply" when `stream_config.max_tier_size` was set. This was not caught earlier because `streamworkspace` tests were not being run in the CI

Fixed by reading `max_tier_size` from the instance model the same way tier already was. Also added `streamworkspace` to the stream CI partition (change-detection filter and ACCTEST_PACKAGES) so these tests run going forward.

Link to any related issue(s): CLOUDP-387913 fixes https://github.com/mongodb/terraform-provider-mongodbatlas/issues/4272

## Type of change:

## Required Checklist:

### DIFF
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -417,7 +417,7 @@ jobs:
             - 'internal/service/streaminstance/*.go'
             - 'internal/service/streamprocessor/*.go'
             - 'internal/service/streamprivatelinkendpoint/*.go'
-            - 'internal/service/streamworkspace/*.go' 
+            - 'internal/service/streamworkspace/*.go'
 
 
   advanced_cluster:
--- a/.github/workflows/acceptance-tests-runner.yml
+++ b/.github/workflows/acceptance-tests-runner.yml
@@ -416,7 +416,8 @@ jobs:
             - 'internal/service/streamconnection/*.go'
             - 'internal/service/streaminstance/*.go'
             - 'internal/service/streamprocessor/*.go'
-            - 'internal/service/streamprivatelinkendpoint/*.go' 
+            - 'internal/service/streamprivatelinkendpoint/*.go'
+            - 'internal/service/streamworkspace/*.go' 
 
 
   advanced_cluster:
@@ -1406,6 +1407,7 @@ jobs:
               ./internal/service/streaminstance
               ./internal/service/streamprocessor
               ./internal/service/streamprivatelinkendpoint
+              ./internal/service/streamworkspace
           run: make testacc
 
   slack-notification-stream:
--- a/internal/service/streamworkspace/model.go
+++ b/internal/service/streamworkspace/model.go
@@ -74,6 +74,7 @@ func (m *TFModel) FromInstanceModel(instanceModel *streaminstance.TFStreamInstan
 	} else {
 		instanceStreamConfigAttrs := instanceModel.StreamConfig.Attributes()
 		tierValue := instanceStreamConfigAttrs["tier"]
+		maxTierSizeValue := instanceStreamConfigAttrs["max_tier_size"]
 
 		workspaceStreamConfig, _ := types.ObjectValue(
 			map[string]attr.Type{
@@ -81,7 +82,7 @@ func (m *TFModel) FromInstanceModel(instanceModel *streaminstance.TFStreamInstan
 				"tier":          types.StringType,
 			},
 			map[string]attr.Value{
-				"max_tier_size": types.StringNull(),
+				"max_tier_size": maxTierSizeValue,
 				"tier":          tierValue,
 			},
 		)
--- a/internal/service/streamworkspace/model.go
+++ b/internal/service/streamworkspace/model.go
@@ -74,7 +74,6 @@ func (m *TFModel) FromInstanceModel(instanceModel *streaminstance.TFStreamInstan
 	} else {
 		instanceStreamConfigAttrs := instanceModel.StreamConfig.Attributes()
 		tierValue := instanceStreamConfigAttrs["tier"]
-		maxTierSizeValue := instanceStreamConfigAttrs["max_tier_size"]
 
 		workspaceStreamConfig, _ := types.ObjectValue(
 			map[string]attr.Type{
@@ -82,7 +81,7 @@ func (m *TFModel) FromInstanceModel(instanceModel *streaminstance.TFStreamInstan
 				"tier":          types.StringType,
 			},
 			map[string]attr.Value{
-				"max_tier_size": maxTierSizeValue,
+				"max_tier_size": types.StringNull(),
 				"tier":          tierValue,
 			},
 		)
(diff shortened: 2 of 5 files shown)


erabil-mdb (approved):
LGTM!

internal/service/streamworkspace/resource_test.go
```diff
@@ -103,6 +103,10 @@ func checkStreamsWorkspaceImportStateIDFunc(resourceName string) resource.Import
 }
 
 func streamsWorkspaceConfig(projectID, workspaceName, region, cloudProvider string) string {
+	return streamsWorkspaceWithStreamConfigConfig(projectID, workspaceName, region, cloudProvider, "SP10", "SP30")
+}
+
+func streamsWorkspaceWithStreamConfigConfig(projectID, workspaceName, region, cloudProvider, tier, maxTierSize string) string {
```
lantoli, line 86: `TestAccStreamWorkspaceRS_withStreamConfig` is functionally identical to `TestAccStreamWorkspaceRS_basic` (same config, same checks, just missing the import step),  consider removing it

internal/service/streamworkspace/model.go
```diff
@@ -74,14 +74,15 @@ func (m *TFModel) FromInstanceModel(instanceModel *streaminstance.TFStreamInstan
...
+		maxTierSizeValue := instanceStreamConfigAttrs["max_tier_size"]
 
 		workspaceStreamConfig, _ := types.ObjectValue(
 			map[string]attr.Type{
 				"max_tier_size": types.StringType,
 				"tier":          types.StringType,
 			},
 			map[string]attr.Value{
-				"max_tier_size": types.StringNull(),
+				"max_tier_size": maxTierSizeValue,
```
lantoli, line 85: q:  should we remove UseStateForUnknown from `max_tier_size` to avoid problems like in alert configuration?

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -416,7 +416,8 @@ jobs:
             - 'internal/service/streamconnection/*.go'
             - 'internal/service/streaminstance/*.go'
             - 'internal/service/streamprocessor/*.go'
-            - 'internal/service/streamprivatelinkendpoint/*.go' 
+            - 'internal/service/streamprivatelinkendpoint/*.go'
+            - 'internal/service/streamworkspace/*.go'
```
lantoli: nit: trailing whitespace on this line

lantoli (approved):

.github/workflows/acceptance-tests-runner.yml
oarbusi: fixed in [534b214](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4277/commits/534b214f19bf1334faeef9d14a289241f3878e66)

internal/service/streamworkspace/resource_test.go
oarbusi: good catch, done in [534b214](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4277/commits/534b214f19bf1334faeef9d14a289241f3878e66)

internal/service/streamworkspace/model.go
oarbusi: good point. I was thinking if there was a risk of breaking change removing it, in theory, removing UseStateForUnknown changes plan behavior: users who don't set max_tier_size in config would see `(known after apply)` instead of the prior state value during plan but in practice, it doesn't matter here because max_tier_size was always returned as null due to the bug we're fixing. Nobody could have been relying on UseStateForUnknown working correctly for this attribute, it was preserving a null value from a null value. Changed in [534b214](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4277/commits/534b214f19bf1334faeef9d14a289241f3878e66)

internal/service/streamworkspace/resource_test.go
```diff
@@ -112,9 +93,9 @@ func streamsWorkspaceConfig(projectID, workspaceName, region, cloudProvider stri
 				cloud_provider = %[4]q
 			}
 			stream_config = {
-				max_tier_size = "SP30"
-				tier = "SP10"
```
AgustinBettati, line 116: q: The test previously defined `max_tier_size` but did not catch this error. I am not clear what change was made to the tests to ensure it is now capturing the initial bug.

AgustinBettati (approved):
LGTM. Could also be good to get a review from streams team, very surprised how read was left with the hardcoded null.

internal/service/streamworkspace/resource_test.go
AgustinBettati: disregard, saw the clarifying description `This was not caught earlier because streamworkspace tests were not being run in the CI`

PR #4290, opened 2026-03-13
test: Add acceptance tests for service_account_jwt ephemeral resource

maastha:
## Description

Add acceptance tests and unit tests for the mongodbatlas_service_account_jwt ephemeral resource.
Acceptance tests use the echo` provider from `terraform-plugin-testing to "echo" ephemeral resource outputs into managed resource state for assertion. 
Ephemeral resource outputs cannot be asserted via standard plan/state checks since they are not persisted. The echo provider bridges this gap by copying ephemeral data into a managed resource's state. See [testing ephemeral resources.](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/ephemeral-resources#testing-ephemeral-data-with-echo-provider) 

Test cases cover:

- Token generation using provider-level SA credentials.
- Token generation with revoke_on_closure = true.
- Explicit credentials passed on the ephemeral resource block.
- Invalid credentials (expects token exchange error).
- Partial credentials (expects validation error).

Link to any related issue(s): CLOUDP-387290

## Type of change:

## Required Checklist:


internal/service/serviceaccountjwt/resource.go
```diff
@@ -90,29 +90,30 @@ func (r *es) Open(ctx context.Context, req ephemeral.OpenRequest, resp *ephemera
...
+	log.Printf("[DEBUG] %s Close: revoking access token", ResourceTypeName)
 	conf := config.GetServiceAccountConfig(data.ClientID, data.ClientSecret, config.NormalizeBaseURL(data.BaseURL))
-	if err := conf.RevokeToken(r.withUserAgentClient(ctx), &oauth2.Token{AccessToken: data.AccessToken}); err != nil {
+	if err := conf.RevokeToken(r.WithUserAgentClient(ctx), &oauth2.Token{AccessToken: data.AccessToken}); err != nil {
 		resp.Diagnostics.AddWarning("Failed to revoke Service Account token on close", err.Error())
 	}
 }
 
-func (r *es) resolveCredentials(model *TFModel) (clientID, clientSecret, baseURL string, diags diag.Diagnostics) {
+func (r *ES) ResolveCredentials(model *TFModel) (clientID, clientSecret, baseURL string, diags diag.Diagnostics) {
```
marcosuma: double checking : have we made ResolveCredentials and GenerateToken public in order to test them?

maastha: right

internal/service/serviceaccountjwt/resource_test.go
```diff
@@ -1,41 +1,78 @@
...
+}
+
+func protoV6FactoriesWithEcho() map[string]func() (tfprotov6.ProviderServer, error) {
+	factories := make(map[string]func() (tfprotov6.ProviderServer, error))
+	maps.Copy(factories, acc.TestAccProviderV6Factories)
+	factories["echo"] = echoprovider.NewProviderServer()
+	return factories
+}
+
+func preCheckSACreds(t *testing.T) {
```
manupedrozo: We have some helpers that can be used in these tests for skipping when not using SA: [example](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/20fbbf2f13e4ad2a568cf1a44421577484937122/internal/provider/provider_authentication_test.go#L13-L15)
May be practical to add a SkipIfNotSA()

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -304,7 +304,8 @@ jobs:
             - 'internal/provider/*.go'
           authentication:
             - 'internal/config/*.go'
-            - 'internal/provider/*.go' 
+            - 'internal/provider/*.go'
+            - 'internal/service/serviceaccountjwt/*.go'
```
manupedrozo, line 310: Thinking on whether we should have these tests in `authentication` or under `service_account` or in a separate `ephemeral` group.
I think we use `authentication` for testing the provider auth so I see a point for **not** having the jwt resource under this group.
Not highly relevant, we can always move it later.

internal/service/serviceaccountjwt/resource_test.go
```diff
@@ -1,41 +1,78 @@
...
-				BaseURL:          "https://provider.example.com/",
-				TerraformVersion: "1.10.0",
-			},
+			EphemeralResourceData: erd,
+			ResourceName:          serviceaccountjwt.ResourceTypeName,
 		},
 	}
+}
+
+func protoV6FactoriesWithEcho() map[string]func() (tfprotov6.ProviderServer, error) {
```
manupedrozo: Nice, consider moving to testutil/acc/factory.go

internal/service/serviceaccountjwt/resource.go
marcosuma: Changing private methods to public solely for testing is generally considered bad practice as it breaks encapsulation and leaks implementation details.

What you're doing here is: some other class is allowed now to consume these methods which makes the code more and more (potentially) unreadable and hard to maintain.

Is there an alternative? are these methods worth it to be in a separate class that based on a single-responsibility principle offers the methods to be consumed? 

cc @lantoli

internal/service/serviceaccountjwt/resource_test.go
```diff
@@ -76,37 +109,167 @@ func TestResolveCredentials_NonSAProviderAuth(t *testing.T) {
...
+func TestAccServiceAccountJWT_providerCredentials(t *testing.T) {
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { preCheckSACreds(t) },
+		TerraformVersionChecks:   versionChecks,
+		ProtoV6ProviderFactories: protoV6FactoriesWithEcho(),
+		Steps: []resource.TestStep{
+			{
+				Config: configProviderCredentials(false),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("echo.jwt", tfjsonpath.New("data").AtMapKey("token_type"), knownvalue.StringExact("Bearer")),
```
manupedrozo, line 32: Nice to see `statecheck` being used!

internal/service/serviceaccountjwt/resource.go
lantoli: Agreed that making methods public solely for testing is generally bad practice, we should try to test behavior using the public interface. But sometimes there's a tradeoff where testing the behavior would be complex or impractical and it's better to unit test it directly.

In this case I'd keep `GenerateToken` private as it's just a thin wrapper. But `ResolveCredentials` has enough complexity that unit testing it directly is valuable, and testing through `Open` would require real credentials and a live connection.

A cleaner alternative would be to have `Open` accept an interface for token generation so a fake one can be used in tests, that way ResolveCredentials stays private. It's a bit more code but a better approach.

maastha: Adopted the interface approach in b5c5bfb52. `TokenGenerator` interface injected via `ES.TokenGen` field. `ResolveCredentials` stays exported for direct unit testing.

internal/service/serviceaccountjwt/resource_test.go
maastha: Added `acc.PreCheckSA(t)` in b5c5bfb52.

.github/workflows/acceptance-tests-runner.yml
maastha: Moved to its own `service_account_jwt` group

internal/service/serviceaccountjwt/resource_test.go
maastha: Moved to `acc.ProtoV6FactoriesWithEcho()` in b5c5bfb52.

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -304,7 +304,9 @@ jobs:
             - 'internal/provider/*.go'
           authentication:
             - 'internal/config/*.go'
-            - 'internal/provider/*.go' 
+            - 'internal/provider/*.go'
+          service_account_jwt:
+            - 'internal/service/serviceaccountjwt/*.go'
           autogen_fast:
             - 'internal/common/autogen/*.go'
```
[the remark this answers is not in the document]
maastha: Fixed in edcd6f054. Added `service_account_jwt` output to `change-detection` and wired it to the `authentication` job's `if` condition. The JWT acceptance tests already run in that job (ACCTEST_PACKAGES includes `./internal/service/serviceaccountjwt`).

internal/testutil/acc/skip.go
```diff
@@ -64,3 +64,12 @@ func SkipInAccessToken(tb testing.TB, description string) {
...
 	}
 }
+
+// PreCheckSA skips the test if Service Account credentials are not configured.
+func PreCheckSA(tb testing.TB) {
+	tb.Helper()
+	PreCheckBasic(tb)
+	if !HasSACreds() {
+		tb.Skip("MONGODB_ATLAS_CLIENT_ID and MONGODB_ATLAS_CLIENT_SECRET required")
+	}
```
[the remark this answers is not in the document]
maastha: Fixed in edcd6f054. Changed `||` to `&&` so both env vars must be set.

internal/service/serviceaccountjwt/user_agent.go
```diff
@@ -26,7 +26,7 @@ func (r *es) terraformVersion() string {
 //	  → UserAgentTransport  (appends "Name/service_account_jwt Operation/..." from context)
 //	    → NetworkLoggingTransport  (logs method/URL/timing/status)
 //	      → http.DefaultTransport
-func (r *es) withUserAgentClient(ctx context.Context) context.Context {
+func (r *ES) WithUserAgentClient(ctx context.Context) context.Context {
 	networkLog := config.NewTransportWithNetworkLogging(http.DefaultTransport, logging.IsDebugOrHigher())
```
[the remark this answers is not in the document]
maastha: This file is deleted entirely by PR #4314 (CLOUDP-390064), which centralizes the OAuth HTTP client into `internal/config/`. No action needed here.

internal/service/serviceaccountjwt/resource.go
```diff
@@ -17,15 +17,21 @@ import (
...
+
+var _ ephemeral.EphemeralResource = &ES{}
+var _ ephemeral.EphemeralResourceWithConfigure = &ES{}
+var _ ephemeral.EphemeralResourceWithClose = &ES{}
 
-type es struct {
+type ES struct {
+	TokenGen TokenGenerator
 	config.ESCommon
 }
```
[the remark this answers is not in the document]
maastha: Noted. Will address in a follow-up.

internal/service/serviceaccountjwt/resource.go
```diff
@@ -161,9 +169,9 @@ func providerBaseURL(providerData *config.EphemeralResourceData) string {
 	return ""
 }
 
-func (r *es) generateToken(ctx context.Context, clientID, clientSecret, baseURL string) (*auth.Token, error) {
+func (r *ES) GenerateToken(ctx context.Context, clientID, clientSecret, baseURL string) (*auth.Token, error) {
 	conf := config.GetServiceAccountConfig(clientID, clientSecret, config.NormalizeBaseURL(baseURL))
-	token, err := conf.Token(r.withUserAgentClient(ctx))
+	token, err := conf.Token(r.WithUserAgentClient(ctx))
 	if err != nil {
 		return nil, err
```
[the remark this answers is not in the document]
maastha: Noted. Will address in a follow-up.

internal/service/serviceaccountjwt/resource.go
maastha: Thanks for the feedback, @marcosuma , this is a valid concern and I appreciate you flagging it.

To share some context on the reasoning behind this change, since the repo enforces black-box testing via the testpackage linter, unexported symbols aren't accessible to tests. This is a tradeoff the codebase has made pretty broadly, and not just for model/conversion functions but also several business logic helpers are exported so tests can reach them.

I do agree with you and @lantoli that we can be more intentional about what we expose. Following that guidance, I've adopted the interface approach for GenerateToken (TokenGenerator injected via ES.TokenGen so tests can supply a mock). ResolveCredentials stays exported for the trade-off that Leo shared.

If we want to move away from exporting for testability more broadly, it might be worth aligning on a consistent approach across the repo. Happy to help drive that discussion. Definitely open to other ideas if you see a cleaner path for ResolveCredentials specifically.

marcosuma: ok thanks! might be a good tech sync chat

marcosuma: Few more thoughts:

> This is a tradeoff the codebase has made pretty broadly,

Unfortunately I haven't spotted it in other PRs, but I think you agree with me that if we're done it so far and it's not a good practice, we should reverse the trend

> But ResolveCredentials has enough complexity that unit testing it directly is valuable

this is the red flag for me. If a method has enough complexity it means it carries a lot more responsibility than it should. And probably it should be a separate class, `CredentialResolver` that vends one single method. Now you do have a lot more control of how to split that method into private ones and a public one. This is a much better way to design the code, IMO. With the advent of coding agents, it is even more important to be predictable and write code at the highest standard.

> ResolveCredentials stays exported for the trade-off that Leo shared

Not going to block the PR, but I think Leo also shared the `Open` redesign approach. I personally still find it violates the main principle: it is not resource.go's responsibility to vend a "ResolveCredentials" or "Open" method: that should be a totally separate object.
Why: What if we were to have two resources that now need a credential resolver? With current code, you either duplicate code, or point to serviceaccountjwt/resource.go or you take the cost of refactoring - it's either more cost for later or more tech debt.

marcosuma (approved):

.github/workflows/acceptance-tests-runner.yml
```diff
@@ -611,6 +614,33 @@ jobs:
           MONGODB_ATLAS_CLIENT_SECRET: ${{ secrets.mongodb_atlas_client_secret }}        
         run: make access-token-revoke token="${{ steps.create-token.outputs.access_token }}"
 
+  service_account_jwt:
+    needs: [ change-detection, get-provider-version ]
+    if: ${{ needs.change-detection.outputs.service_account_jwt == 'true' || inputs.test_group == 'service_account_jwt' }}
```
manupedrozo: Can we skip running this group if not using SAs? 
To avoid spinning up the gh action altogether when we run the test suite with PAKs (in addition to the skip logic that was added at the test level).

internal/testutil/acc/skip.go
```diff
@@ -64,3 +64,12 @@ func SkipInAccessToken(tb testing.TB, description string) {
 		tb.Skip(description)
 	}
 }
+
+// PreCheckSA skips the test if Service Account credentials are not configured.
+func PreCheckSA(tb testing.TB) {
```
manupedrozo: Thanks for moving, would rename to `SkipIfNotSA`. I would expect `PreCheck` to fail instead of skipping.
Non blocking!

manupedrozo (approved):

internal/service/serviceaccountjwt/resource.go
```diff
@@ -17,15 +17,21 @@ import (
...
+type TokenGenerator interface {
+	GenerateToken(ctx context.Context, clientID, clientSecret, baseURL string) (*auth.Token, error)
+}
+
+var _ ephemeral.EphemeralResource = &ES{}
+var _ ephemeral.EphemeralResourceWithConfigure = &ES{}
+var _ ephemeral.EphemeralResourceWithClose = &ES{}
 
-type es struct {
+type ES struct {
```
AgustinBettati: nit: We are using `rs`,`ds`, `pluralDs` in our codebase for struct name of resource/data sources. `es` looked confusing to me, would `er` (ephemeral resource) work better?

internal/service/serviceaccountjwt/resource_test.go
```diff
@@ -76,37 +91,167 @@ func TestResolveCredentials_NonSAProviderAuth(t *testing.T) {
...
+					statecheck.ExpectKnownValue("echo.jwt", tfjsonpath.New("data").AtMapKey("expires_in"), knownvalue.NotNull()),
+					statecheck.ExpectKnownValue("echo.jwt", tfjsonpath.New("data").AtMapKey("access_token"), knownvalue.NotNull()),
+				},
+			},
+		},
+	})
+}
+
+func TestAccServiceAccountJWT_revokeOnClosure(t *testing.T) {
```
AgustinBettati, line 41: q: Is there opportunity for a check destroy that verifies token is correctly revoked?

AgustinBettati (approved):
LGTM

internal/service/serviceaccountjwt/resource.go
lantoli: yes, agree to reverse the trend

maastha: sounds good. I will bring this up to the next tech sync

maastha: Updated the PR to have a `CredentialResolver`

internal/service/serviceaccountjwt/resource.go
```diff
@@ -17,15 +17,21 @@ import (
 )
 
 const (
```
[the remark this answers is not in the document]
maastha: updated

.github/workflows/acceptance-tests-runner.yml
maastha: updated

internal/service/serviceaccountjwt/resource.go
maastha: updated

internal/service/serviceaccountjwt/resource_test.go
maastha: not practically feasible for this PR, Atlas has no token introspection endpoint to verify revocation succeeded.

PR #4384, opened 2026-04-17
fix: Suppress perpetual plan drift on post_auth_role_grants and data_access_identity_provider_ids for mongodbatlas_federated_settings_org_config

marcabreracast:
## Description

Fixes perpetual plan drift on `post_auth_role_grants` and `data_access_identity_provider_ids` fields in `mongodbatlas_federated_settings_org_config`.

These fields are defined as `TypeList`, while the API returns these fields in a non-deterministic order (they're of `Set` type). This causes that each read produces a different order than what's in state, causing an in-place update diff on every plan even when nothing changed.

For example, applying a config with post_auth_role_grants and immediately running terraform plan would show:

```
~ post_auth_role_grants = [
    - "ORG_BILLING_READ_ONLY",
    - "ORG_GROUP_CREATOR",
      "ORG_MEMBER",
      "ORG_READ_ONLY",
    + "ORG_BILLING_READ_ONLY",
    + "ORG_GROUP_CREATOR",
  ]
```

This PR aims to fix that behaviour by adding a `DiffSuppressFunc` to those fields: that compares both lists as sets, suppressing diffs when the only difference is element order. Real changes (additions, removals, value changes) are still detected.

### Testing:

* Unit test for `StringSlicesEqualIgnoringOrder` covering same order, different order, different values, different lengths, empty lists, and nil slices. 
* Acceptance test `TestAccFederatedSettingsOrg_noDriftOnListFields` that checks the no-drift behaviour for both fields. These tests are skipped in CI, but they are currently passing locally:

```
go test ./internal/service/federatedsettingsorgconfig/ -run TestAccFederatedSettingsOrg_noDriftOnListFields -v -timeout 30m

=== RUN   TestAccFederatedSettingsOrg_noDriftOnListFields
=== PAUSE TestAccFederatedSettingsOrg_noDriftOnListFields
=== CONT  TestAccFederatedSettingsOrg_noDriftOnListFields
--- PASS: TestAccFederatedSettingsOrg_noDriftOnListFields (7.52s)
PASS
ok  	github.com/mongodb/terraform-provider-mongodbatlas/internal/service/federatedsettingsorgconfig	8.460s
```

Link to any related issue(s): [CLOUDP-397484](https://jira.mongodb.org/browse/CLOUDP-397484)

## Type of change:

## Required Checklist:

- [X] I have run make fix and verified my code

### DIFF
--- a/internal/common/schemafunc/string_list.go
+++ b/internal/common/schemafunc/string_list.go
@@ -23,6 +23,6 @@ func StringSlicesEqualIgnoringOrder(a, b []string) bool {
 // DiffSuppressFunc that suppresses diffs on a TypeList of strings when the only difference is element order.
 func EqualStringListsIgnoringOrder(k, oldValue, newValue string, d *schema.ResourceData) bool {
 	baseKey := k[:strings.LastIndex(k, ".")]
-	old, new := d.GetChange(baseKey)
-	return StringSlicesEqualIgnoringOrder(cast.ToStringSlice(old), cast.ToStringSlice(new))
+	oldVal, newVal := d.GetChange(baseKey)
+	return StringSlicesEqualIgnoringOrder(cast.ToStringSlice(oldVal), cast.ToStringSlice(newVal))
 }
--- a/internal/common/schemafunc/string_list.go
+++ b/internal/common/schemafunc/string_list.go
@@ -1,0 +1,28 @@
+package schemafunc
+
+import (
+	"slices"
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/spf13/cast"
+)
+
+// Returns true if both slices contain the same elements regardless of order.
+func StringSlicesEqualIgnoringOrder(a, b []string) bool {
+	sortedA := make([]string, len(a))
+	copy(sortedA, a)
+	sort.Strings(sortedA)
+	sortedB := make([]string, len(b))
+	copy(sortedB, b)
+	sort.Strings(sortedB)
+	return slices.Equal(sortedA, sortedB)
+}
+
+// DiffSuppressFunc that suppresses diffs on a TypeList of strings when the only difference is element order.
+func EqualStringListsIgnoringOrder(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	baseKey := k[:strings.LastIndex(k, ".")]
+	old, new := d.GetChange(baseKey)
+	return StringSlicesEqualIgnoringOrder(cast.ToStringSlice(old), cast.ToStringSlice(new))
+}
--- a/internal/common/schemafunc/string_list.go
+++ b/internal/common/schemafunc/string_list.go
@@ -11,6 +11,9 @@ import (
 
 // Returns true if both slices contain the same elements regardless of order.
 func StringSlicesEqualIgnoringOrder(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
 	sortedA := make([]string, len(a))
 	copy(sortedA, a)
 	sort.Strings(sortedA)
--- a/internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
+++ b/internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/schemafunc"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/validate"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/config"
 	"github.com/spf13/cast"
@@ -44,15 +45,17 @@ func Resource() *schema.Resource {
 				},
 			},
 			"data_access_identity_provider_ids": {
-				Type:     schema.TypeList,
-				Optional: true,
+				Type:             schema.TypeList,
+				Optional:         true,
+				DiffSuppressFunc: schemafunc.EqualStringListsIgnoringOrder,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
 			},
 			"post_auth_role_grants": {
-				Type:     schema.TypeList,
-				Optional: true,
+				Type:             schema.TypeList,
+				Optional:         true,
+				DiffSuppressFunc: schemafunc.EqualStringListsIgnoringOrder,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
--- a/internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
+++ b/internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
@@ -45,6 +45,7 @@ func Resource() *schema.Resource {
 				},
 			},
 			"data_access_identity_provider_ids": {
+				// CLOUDP-398552 will deal with update to TypeSet for this field
 				Type:             schema.TypeList,
 				Optional:         true,
 				DiffSuppressFunc: schemafunc.EqualStringListsIgnoringOrder,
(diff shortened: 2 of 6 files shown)


internal/common/schemafunc/string_list.go
```diff
@@ -0,0 +1,28 @@
...
+	sort.Strings(sortedA)
+	sortedB := make([]string, len(b))
+	copy(sortedB, b)
+	sort.Strings(sortedB)
+	return slices.Equal(sortedA, sortedB)
+}
+
+// DiffSuppressFunc that suppresses diffs on a TypeList of strings when the only difference is element order.
+func EqualStringListsIgnoringOrder(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	baseKey := k[:strings.LastIndex(k, ".")]
```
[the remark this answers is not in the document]
marcabreracast: In this case, `DiffSuppressFunc` should only be used `TypeList fields`, where Terraform always passes element keys with a dot (e.g., `field.0`, `field.#`). A dotless key can't happen in this context, in which case it could be considered a misuse

internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org_test.go
```diff
@@ -97,6 +98,51 @@ func basicTestCase(tb testing.TB) *resource.TestCase {
...
+	acc.SkipTestForCI(t)
+	var (
+		resourceName         = "mongodbatlas_federated_settings_org_config.test"
+		federationSettingsID = os.Getenv("MONGODB_ATLAS_FEDERATION_SETTINGS_ID")
+		orgID                = os.Getenv("MONGODB_ATLAS_FEDERATED_ORG_ID")
+		idpID                = os.Getenv("MONGODB_ATLAS_FEDERATED_IDP_ID")
+		associatedDomain     = os.Getenv("MONGODB_ATLAS_FEDERATED_SETTINGS_ASSOCIATED_DOMAIN")
+		oidcIDP1             = os.Getenv("MONGODB_ATLAS_FEDERATED_OIDC_IDP_ID_1")
+		oidcIDP2             = os.Getenv("MONGODB_ATLAS_FEDERATED_OIDC_IDP_ID_2")
+	)
```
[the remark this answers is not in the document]
marcabreracast: Addressed in [2ca95c4](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4384/commits/2ca95c47c6fcc642b5634ae387da69dd6324c888)

internal/common/schemafunc/string_list.go
AgustinBettati, line 25: **Suggestion:** Early-return on non-length keys so the set comparison runs exactly once per plan:

```go
if !strings.HasSuffix(k, ".#") { return true }
```

internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
```diff
@@ -44,15 +45,17 @@ func Resource() *schema.Resource {
 				},
 			},
 			"data_access_identity_provider_ids": {
-				Type:     schema.TypeList,
-				Optional: true,
+				Type:             schema.TypeList,
```
AgustinBettati, line 48: Would consider creating a followup CLOUDP ticket with `terraform-provider-3.x` label so we can adjust the schema to type Set and avoid this DiffSuppressFunc workaround

AgustinBettati (approved):
LGTM, would take a look at the optimization suggestion as it can make code more clear

xargom (approved):
LGTM

internal/common/schemafunc/string_list.go
marcabreracast: I tested this, but the optimization causes Terraform to suppress element-level diffs blindly, so the plan shows `null` instead of the actual values:

```
  # mongodbatlas_federated_settings_org_config.this will be created
  + resource "mongodbatlas_federated_settings_org_config" "this" {
      + data_access_identity_provider_ids = [
          + null,
          + null,
        ]
      + domain_allow_list                 = [
          + "<DOMAIN_ALLOW_LIST>",
        ]
      + domain_restriction_enabled        = false
      + federation_settings_id            = "<FEDERATION_SETTINGS_ID>"
      + id                                = (known after apply)
      + identity_provider_id              = "<IDENTITY_PROVIDER_ID>"
      + org_id                            = "<ORG_ID>"
      + post_auth_role_grants             = [
          + null,
          + null,
          + null,
        ]
      + user_conflicts                    = (known after apply)
    }
```

From what I can tell, when Terraform builds the plan it breaks a list into sub-keys and calls `DiffSuppressFunc `on each one. Because the optimization returns true on all element sub-keys, it never does the comparison and instead just displays `null.` Running the comparison on every sub-key is less efficient, but seems like the only way to achieve the desired behaviour.

internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
marcabreracast: Sure, created CLOUDP-398552 to address this

internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
```diff
@@ -44,15 +45,17 @@ func Resource() *schema.Resource {
...
+				Optional:         true,
+				DiffSuppressFunc: schemafunc.EqualStringListsIgnoringOrder,
```
AgustinBettati, line 51: nit: we can leave a comment with the ticket that would change this to a Set (for future context)

internal/common/schemafunc/string_list.go
AgustinBettati: Sorry, I believe my code snippet was incorrect with the returned value:

```suggestion
	if !strings.HasSuffix(k, ".#") { return false }
```
Similar to current implementation, I believe the only relevant comparison to be done is when the key is ".#" and `newVal` and `oldVal` contain the complete arrays (correct me if understanding is not correct).

internal/common/schemafunc/string_list.go
```diff
@@ -0,0 +1,28 @@
...
+	"sort"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/spf13/cast"
+)
+
+// Returns true if both slices contain the same elements regardless of order.
+func StringSlicesEqualIgnoringOrder(a, b []string) bool {
+	sortedA := make([]string, len(a))
```
manupedrozo, line 17:
```suggestion
func StringSlicesEqualIgnoringOrder(a, b []string) bool {
    if len(a) != len(b) {
        return false
    }
    sortedA := make([]string, len(a))
```

internal/common/schemafunc/string_list.go
marcabreracast: The thing is that `oldValue` and `newValue` at the `.#` call are just count strings, not the full arrays. 
We get the full arrays via `d.GetChange(baseKey)`.
Additionally, suppressing `.#` alone isn't enough: Terraform calls `DiffSuppressFunc` independently for each key (.0, .1, .2), and those won't be suppressed if we return `false` for non `.#` keys.

As an example, if the plan has 4 separate lines to hide: 
```
~ post_auth_role_grants.#  3 -> 3                    //  HIDDEN
~ post_auth_role_grants.0  "ORG_BILLING_READ_ONLY" -> "ORG_OWNER"   // VISIBLE
~ post_auth_role_grants.1  "ORG_READ_ONLY" -> "ORG_READ_ONLY"        // VISIBLE
~ post_auth_role_grants.2  "ORG_OWNER" -> "ORG_BILLING_READ_ONLY"    // VISIBLE
```
With the suggested implementation, the values will still show up at plan time.

manupedrozo (approved):

internal/common/schemafunc/string_list.go
marcabreracast: TY! Addressed in [324c72c](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4384/commits/324c72c50302a56545952e7710fda478a0796d33)

internal/service/federatedsettingsorgconfig/resource_federated_settings_connected_org.go
marcabreracast: Addressed in [9383726](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4384/commits/9383726540e4c98fa3178162f9303e182f4baeb0) for both fields

internal/common/schemafunc/string_list.go
AgustinBettati: Thanks for clarifying, make sense 👍

PR #4394, opened 2026-04-21
fix: Avoid plan churn when adding/removing role_assignments field on mongodbatlas_federated_settings_org_role_mapping

marcabreracast:
## Description

This PR fixes an issue for `mongodbatlas_federated_settings_org_role_mapping`, where performing a `terraform plan` showed all `role_assignments` elements as remove+add whenever the set size changed (adding or removing one block).

### Root cause & fix
There are two separate SDKv2 limitations:

1. **Hash instability from nested `TypeSet`**: `role_assignments` is a `TypeSet` whose elements contain a nested roles TypeSet. SDKv2's HashResource/SerializeValueForHash produces unstable hashes for the outer elements because the inner TypeSet serialization differs depending on whether the value came from state or config.
2. **`null` vs `""` mismatch**: `group_id` and `org_id` are mutually exclusive fields, one is always unset. The state stores unset fields as `""` (the flat map can't represent null), but Terraform's SDK proposed state has null for unset fields. normalizeNullValues in SDKv2 fixes this for lists and maps but explicitly skips sets during plan ([grpc_provider.go:2435](https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.40.0/helper/schema/grpc_provider.go#L2435-L2441)). This causes Terraform to see every element as different, even when the SDKv2 diff layer correctly identifies them as matching.

```
  # mongodbatlas_federated_settings_org_role_mapping.test will be updated in-place
  ~ resource "mongodbatlas_federated_settings_org_role_mapping" "test" {

      - role_assignments {
          - org_id   = "<ORG_ID>" -> null
          - roles    = [
              - "ORG_BILLING_READ_ONLY",
              - "ORG_MEMBER",
              - "ORG_READ_ONLY",
            ] -> null
        }
      - role_assignments {
          - group_id = "<GROUP_ID>" -> null
          - roles    = [
              - "GROUP_DATA_ACCESS_READ_ONLY",
              - "GROUP_READ_ONLY",
            ] -> null
        }
      + role_assignments {
          + group_id = "<GROUP_ID>"
          + roles    = [
              + "GROUP_DATA_ACCESS_READ_ONLY",
              + "GROUP_READ_ONLY",
            ]
        }
      + role_assignments {
          + group_id = "<GROUP_ID>"
          + roles    = [
              + "GROUP_READ_ONLY",
            ]
        }
      + role_assignments {
          + org_id = "<ORG_ID>"
          + roles  = [
              + "ORG_BILLING_READ_ONLY",
              + "ORG_MEMBER",
              + "ORG_READ_ONLY",
            ]
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

### Proposed fix
1. Custom Set hash function (`hashRoleAssignment`) on `role_assignments`: sorts roles before hashing, bypassing the nested TypeSet serialization instability. 
2. Define `Default: ""` on `group_id` and `org_id`: makes Terraform use `""` instead of `null` for unset fields, matching the state representation.

**Important**: without `Default: ""`, the hash mismatch between config (null) and state ("") causes the full remove+add behaviour **even with the custom hash function**. This approach does not introduce breaking changes, as  SDKv2 already stores `""` for unset `Optional` string fields in state, so existing state is unaffected. 

After the fix: Only the field that is added/removed is present, avoiding plan churn. 

```
  # mongodbatlas_federated_settings_org_role_mapping.test will be updated in-place
  ~ resource "mongodbatlas_federated_settings_org_role_mapping" "test" {

      + role_assignments {
          + group_id = "<GROUP_ID>"
          + roles    = [
              + "GROUP_READ_ONLY",
            ]
        }

        # (2 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

Acceptance test passing in local:

```
=== RUN   TestAccFederatedSettingsOrgRoleMapping_NoDriftOnRoleAssignments
=== PAUSE TestAccFederatedSettingsOrgRoleMapping_NoDriftOnRoleAssignments
=== CONT  TestAccFederatedSettingsOrgRoleMapping_NoDriftOnRoleAssignments
--- PASS: TestAccFederatedSettingsOrgRoleMapping_NoDriftOnRoleAssignments (7.49s)
PASS
ok  	github.com/mongodb/terraform-provider-mongodbatlas/internal/service/federatedsettingsorgrolemapping	8.085s
```

Link to any related issue(s): [CLOUDP-397797](https://jira.mongodb.org/browse/CLOUDP-397797)

## Type of change:

## Required Checklist:

- [X] I have run make fix and verified my code

### DIFF
--- a/internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
+++ b/internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
@@ -17,9 +17,13 @@ import (
 // to show as remove+add when the set size changes.
 // See CLOUDP-397797 for more details.
 func hashRoleAssignment(v any) int {
-	m := v.(map[string]any)
-	orgID := m["org_id"].(string)
-	groupID := m["group_id"].(string)
+	m, ok := v.(map[string]any)
+	if !ok {
+		return schema.HashString("")
+	}
+
+	orgID, _ := m["org_id"].(string)
+	groupID, _ := m["group_id"].(string)
 	var roles []string
 	if rolesSet, ok := m["roles"].(*schema.Set); ok {
 		for _, r := range rolesSet.List() {
--- a/internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
+++ b/internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
@@ -16,7 +16,7 @@ import (
 // serialization instability in SDKv2, which causes all outer elements
 // to show as remove+add when the set size changes.
 // See CLOUDP-397797 for more details.
-func hashRoleAssignment(v interface{}) int {
+func hashRoleAssignment(v any) int {
 	m := v.(map[string]any)
 	orgID := m["org_id"].(string)
 	groupID := m["group_id"].(string)
--- a/internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
+++ b/internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
@@ -3,11 +3,32 @@ package federatedsettingsorgrolemapping
 import (
 	"cmp"
 	"slices"
+	"sort"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/conversion"
 	"go.mongodb.org/atlas-sdk/v20250312018/admin"
 )
+
+// hashRoleAssignment hashes a role_assignments element by org_id, group_id, and
+// sorted roles. Sorting the roles before hashing bypasses the nested TypeSet
+// serialization instability in SDKv2, which causes all outer elements
+// to show as remove+add when the set size changes.
+// See CLOUDP-397797 for more details.
+func hashRoleAssignment(v interface{}) int {
+	m := v.(map[string]any)
+	orgID := m["org_id"].(string)
+	groupID := m["group_id"].(string)
+	var roles []string
+	if rolesSet, ok := m["roles"].(*schema.Set); ok {
+		for _, r := range rolesSet.List() {
+			roles = append(roles, r.(string))
+		}
+	}
+	sort.Strings(roles)
+	return schema.HashString(orgID + "|" + groupID + "|" + strings.Join(roles, ","))
+}
 
 func compareRoleAssignment(a, b admin.ConnectedOrgConfigRoleAssignment) int {
 	if c := cmp.Compare(a.GetOrgId(), b.GetOrgId()); c != 0 {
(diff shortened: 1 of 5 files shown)


internal/service/federatedsettingsorgrolemapping/model_federated_settings_org_role_mapping.go
```diff
@@ -3,12 +3,33 @@ package federatedsettingsorgrolemapping
...
+	m := v.(map[string]any)
+	orgID := m["org_id"].(string)
+	groupID := m["group_id"].(string)
+	var roles []string
+	if rolesSet, ok := m["roles"].(*schema.Set); ok {
+		for _, r := range rolesSet.List() {
+			roles = append(roles, r.(string))
+		}
+	}
+	sort.Strings(roles)
```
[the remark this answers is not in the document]
marcabreracast: Note that with this proposed option into the improvement, setting `Default: ""` on both fields would guarantee that these values are always stored as `""`, so panic would not happen. 

However, added some defensive safeguards in  [89ac91a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4394/commits/89ac91a21b40ebca72cf1395e17b47fb0319526e) for the scenario where users might upgrade from a provider version without `Default: ""`.

JuliaMongo (approved):
LGTM for copy

internal/service/federatedsettingsorgrolemapping/resource_federated_settings_org_role_mapping.go
```diff
@@ -47,15 +47,18 @@ func Resource() *schema.Resource {
 			"role_assignments": {
 				Type:     schema.TypeSet,
 				Required: true,
+				Set:      hashRoleAssignment,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"group_id": {
 							Type:     schema.TypeString,
 							Optional: true,
+							Default:  "",
```
AgustinBettati: I am not fully clear on the need for adding `Default`. Is this change required in order to make the hash function work correctly? If so, cant we adjuste the hash function to handle the null vs "" difference?

marcabreracast: That's what I thought initially, but after testing I realized that `Default: ""` is required. 

The issue is about what happens at the schema layer **before our hash function is even called**. Without `Default: ""`, SDKv2 computes the hash for config-side elements using `null` for unset fields (see point 2 of the `Root Cause` section of the PR). State-side elements always have `""` for unset fields. This mismatch produces different hashes, causing the full remove+add behaviour. 

`Default: ""` ensures both sides have `""` before the hash is computed, making them match.

AgustinBettati: synced offline: We preserve current approach taking action of 
- update PR description that assess Default:  "" has not impact
- inline code comment providing context on the "hack"

AgustinBettati (approved):
LGTM, favouring improved UX at the cost of some technical debt. Ideally removed when eventually moving to TPF/autogen.

internal/service/federatedsettingsorgrolemapping/resource_federated_settings_org_role_mapping.go
marcabreracast: For reference, addressed inline comment in [7446238](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4394/commits/74462387e4a10c134589d037606d5c2c227b0e32)

PR #4408, opened 2026-04-28
fix: Fix inconsistent apply for OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD alert

manupedrozo:
## Description

The alerts API returns both the `threshold` and `metricsThreshold` fields for the OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD event type due to backwards compatibility.

The API initially required `threshold` for this event type, which prevented it from being used in TF since the `threshold` block in the TF schema does not include the attribute needed for metrics (`metric_name` and `mode`).
The API now accepts `metricsThreshold`, allowing the creation of this event type via TF. But the `threshold` field is still returned on responses to avoid breaking changes, which causes inconsistent applies on TF.

Error:
```
│ 
│ When applying changes to mongodbatlas_alert_configuration.test, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new value: .threshold_config: block count changed from 0 to 1.
│ 
```

These changes skip setting the `threshold` in state if the `metric_threshold` has a non-nil value.
This is a safe change as there is no alert that uses both fields.

Link to any related issue(s): CLOUDP-393513

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/service/alertconfiguration/model.go
+++ b/internal/service/alertconfiguration/model.go
@@ -85,6 +85,10 @@ func NewMatcherList(list []TfMatcherModel) *[]admin.StreamsMatcher {
 }
 
 func NewTFAlertConfigurationModel(apiRespConfig *admin.GroupAlertsConfig, currState *TfAlertConfigurationRSModel) TfAlertConfigurationRSModel {
+	metricThresholdConfig, thresholdConfig := newTFThresholdModels(
+		apiRespConfig.MetricThreshold, apiRespConfig.Threshold, currState.MetricThresholdConfig, currState.ThresholdConfig,
+	)
+
 	return TfAlertConfigurationRSModel{
 		ID:                    currState.ID,
 		ProjectID:             currState.ProjectID,
@@ -93,8 +97,8 @@ func NewTFAlertConfigurationModel(apiRespConfig *admin.GroupAlertsConfig, currSt
 		Created:               types.StringPointerValue(conversion.TimePtrToStringPtr(apiRespConfig.Created)),
 		Updated:               types.StringPointerValue(conversion.TimePtrToStringPtr(apiRespConfig.Updated)),
 		Enabled:               types.BoolPointerValue(apiRespConfig.Enabled),
-		MetricThresholdConfig: NewTFMetricThresholdConfigModel(apiRespConfig.MetricThreshold, currState.MetricThresholdConfig),
-		ThresholdConfig:       NewTFThresholdConfigModel(apiRespConfig.Threshold, currState.ThresholdConfig),
+		MetricThresholdConfig: metricThresholdConfig,
+		ThresholdConfig:       thresholdConfig,
 		Notification:          NewTFNotificationModelList(apiRespConfig.GetNotifications(), currState.Notification),
 		Matcher:               NewTFMatcherModelList(apiRespConfig.GetMatchers(), currState.Matcher),
 		SeverityOverride:      types.StringPointerValue(apiRespConfig.SeverityOverride),
@@ -184,9 +188,6 @@ func NewTFNotificationModelList(n []admin.AlertsNotificationRootForGroup, currSt
 }
 
 func NewTFMetricThresholdConfigModel(t *admin.FlexClusterMetricThreshold, currStateSlice []TfMetricThresholdConfigModel) []TfMetricThresholdConfigModel {
-	if t == nil {
-		return []TfMetricThresholdConfigModel{}
-	}
 	if len(currStateSlice) == 0 { // metric threshold was created elsewhere from terraform, or import statement is being called
 		return []TfMetricThresholdConfigModel{
 			{
@@ -218,10 +219,6 @@ func NewTFMetricThresholdConfigModel(t *admin.FlexClusterMetricThreshold, currSt
 }
 
 func NewTFThresholdConfigModel(t *admin.StreamProcessorMetricThreshold, currStateSlice []TfThresholdConfigModel) []TfThresholdConfigModel {
-	if t == nil {
-		return []TfThresholdConfigModel{}
-	}
-
 	if len(currStateSlice) == 0 { // threshold was created elsewhere from terraform, or import statement is being called
 		return []TfThresholdConfigModel{
 			{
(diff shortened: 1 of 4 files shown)


internal/service/alertconfiguration/model.go
```diff
@@ -184,9 +188,6 @@ func NewTFNotificationModelList(n []admin.AlertsNotificationRootForGroup, currSt
 }
 
 func NewTFMetricThresholdConfigModel(t *admin.FlexClusterMetricThreshold, currStateSlice []TfMetricThresholdConfigModel) []TfMetricThresholdConfigModel {
-	if t == nil {
-		return []TfMetricThresholdConfigModel{}
-	}
 	if len(currStateSlice) == 0 { // metric threshold was created elsewhere from terraform, or import statement is being called
```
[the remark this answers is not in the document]
manupedrozo: nil check was moved, no other usage of this function.

internal/service/alertconfiguration/model.go
```diff
@@ -315,3 +316,21 @@ func NewTFAlertConfigurationDSModelList(alerts []admin.GroupAlertsConfig, projec
 
 	return results
 }
+
+// newTFThresholdModels returns metricThreshold and threshold models, giving precedence to metricThreshold when both are set.
+// This is necessary to avoid inconsistent applies for the OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD event type, for which
+// the API returns both metricThreshold and threshold in the response for backwards compatibility. See HELP-91242.
+func newTFThresholdModels(
```
manupedrozo, line 326: We could branch and only do this for OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD type.
Given that `threshold` and `metrics_threshold` are mutually exclusive in the API, I see no issue in avoiding the branching and doing this for all types.

lmkerbey-mdb (approved):
LGTM.

internal/service/alertconfiguration/model.go
AgustinBettati: > We could branch and only do this for OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD type

Im inclined on adding the if to be on the safer side. For case of `OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD` we explicitly know we favour `metricThreshold` over `threshold`, but who knows if in the future a new inconsistency appears in a different type and in that case `threshold` needs precedence over `metricThreshold`, its essentially a type-by-type criteria.

internal/service/alertconfiguration/model.go
```diff
@@ -315,3 +316,21 @@ func NewTFAlertConfigurationDSModelList(alerts []admin.GroupAlertsConfig, projec
...
+	metricThreshold *admin.FlexClusterMetricThreshold,
+	threshold *admin.StreamProcessorMetricThreshold,
+	currMetricThreshold []TfMetricThresholdConfigModel,
```
AgustinBettati, line 329: Any reason TF models are being passed? Thinking function should just do a post processing over SDK models keeping all other existing logic unchanged.

internal/service/alertconfiguration/model.go
```diff
@@ -184,9 +188,6 @@ func NewTFNotificationModelList(n []admin.AlertsNotificationRootForGroup, currSt
 }
 
 func NewTFMetricThresholdConfigModel(t *admin.FlexClusterMetricThreshold, currStateSlice []TfMetricThresholdConfigModel) []TfMetricThresholdConfigModel {
-	if t == nil {
```
AgustinBettati, line 187: double checking why this is needed

AgustinBettati (commented):
Aligned with changes, believe we should add specific `if` by type to be future proof

internal/service/alertconfiguration/model.go
manupedrozo: Yes, I understand your point. Following the same logic, going with the specific check removes the chance of other inconsistencies being catched retroactively, meaning that any new case would require a new provider version to work.

I prefer the approach of making this the only path and documenting it clearly in a comment.
Lmk what you think and I can change if you insist.

internal/service/alertconfiguration/model.go
manupedrozo: `NewTFMetricThresholdConfigModel` and `NewTFThresholdConfigModel` use them (logic unchanged).

internal/service/alertconfiguration/model.go
manupedrozo: We always go through the new function which does the nil checking. t cannot be nil here now.

internal/service/alertconfiguration/model.go
AgustinBettati: Alright, only additional side effect I see of keeping it generic to all types is that we discard possibility that in the future a notification can define both `threshold` and `metricThreshold` with different values, but does seems unlikely.
Im okay with keeping current implementation and capture assumption in comment. Current logic holds `metricThreshold` favoured over `threshold` specifically because of `OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD`, if in the future a different type requires other logic (`metric` favoured over `metricThreshold`) the changes will have to ensure both the new type and existing `OUTSIDE_STREAM_PROCESSOR_METRIC_THRESHOLD` are handled correctly.

manupedrozo: Expanded comment here: [c9faa4b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4408/commits/c9faa4b68e638ee3512f06e04b9302e2c205c683)

internal/service/alertconfiguration/model.go
```diff
@@ -315,3 +316,24 @@ func NewTFAlertConfigurationDSModelList(alerts []admin.GroupAlertsConfig, projec
 
 	return results
 }
+
+// newTFThresholdModels returns metricThreshold and threshold models, giving precedence to metricThreshold when both are set.
```
AgustinBettati, line 320: Should we add an acceptance test?

AgustinBettati (approved):
LGTM

internal/service/alertconfiguration/model.go
manupedrozo: Added here [15a5ba8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4408/commits/15a5ba8b2e8a18e5539f9821526f54cc665e3f7b)

PR #4455, opened 2026-05-22
feat: Mitigate plan verbosity of `region_usage_restrictions` in mongodbatlas_project

AgustinBettati:
### Context

The `region_usage_restrictions` attribute on `mongodbatlas_project` is declared `Optional + Computed`, which is the correct schema choice: on Atlas for Government projects the API returns a default value (`COMMERCIAL_FEDRAMP_REGIONS_ONLY`) when the field is omitted from the request.

The side effect is that on every plan that re-evaluates the resource the attribute renders as `(known after apply)` when omitted in the config. This was reported in HELP-94010, where the customer (a Commercial Atlas user) was confused by the line appearing on what should have been a reduced plan — and on Commercial it's especially misleading because the API does not accept the field on input and does not return it in responses, so the value the attribute "could be after apply" is always `null`.

### Solution

Add `stringplanmodifier.UseStateForUnknown()` to the attribute. When state holds a known value, Terraform reuses it during planning instead of marking the field unknown.

### Why this is safe

The attribute is set once at project creation and is immutable thereafter — the project update API does not expose the field. Verified end-to-end against Commercial and Atlas for Government environments:

| Deployment | Config | POST response | Follow-up GET | Stable on re-plan? |
|---|---|---|---|---|
| Commercial | omitted | field absent | field absent | yes |
| Gov | omitted | `COMMERCIAL_FEDRAMP_REGIONS_ONLY` | identical | yes |
| Gov | `GOV_REGIONS_ONLY` | `GOV_REGIONS_ONLY` | identical | yes |
| Gov | `COMMERCIAL_FEDRAMP_REGIONS_ONLY` | `COMMERCIAL_FEDRAMP_REGIONS_ONLY` | identical | yes |

In every case the value returned at Create matches every subsequent Read — the precondition for `UseStateForUnknown` to be sound. The user-set value is preserved in state and a no-op plan is genuinely empty.

Link to any related issue(s): [CLOUDP-407011](https://jira.mongodb.org/browse/CLOUDP-407011), HELP-94010

## Type of change:

## Required Checklist:

## Further comments

Verified manually against Commercial and Atlas for Government dev environments. The fix was tested by reproducing the `(known after apply)` line on master, then confirming it disappears with the modifier in place across all three Gov scenarios above.


manupedrozo:
> The attribute is set once at project creation and is immutable thereafter — the project update API does not expose the field. 

Double checking this since I do see the field in the update spec: https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-updategroup#operation-updategroup-200-body-application-vnd-atlas-2023-01-01-json-regionusagerestrictions

We should be ok to use `useStateForUnknown` regardless, since Atlas shouldn't modify the value once known (please verify).

AgustinBettati:
@manupedrozo you will see that it is not present in the update request, the link provided points to the response schema of the udpate.
But yes, the important aspect is that when the user does not define the attribute it will not be modified by Atlas.

manupedrozo (approved):

internal/service/project/resource_project_schema.go
```diff
@@ -116,6 +116,9 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"region_usage_restrictions": schema.StringAttribute{
 				Computed: true,
 				Optional: true,
+				PlanModifiers: []planmodifier.String{
```
lantoli, line 119: nit: the only existing test that checks `region_usage_restrictions` is `TestAccGovProject_withProjectOwner`, which is skipped, consider adding a check in a CI acc test

lantoli (approved):

internal/service/project/resource_project_schema.go
AgustinBettati: This will require more setup effort because we need to create specific gov orgs (`cloud.mongodbgov.com`) for dev and qa. Not sure if worth the setup/maintenance for covering this attribute, but I did test locally.

AgustinBettati:
Failing tests are not related, looking into address the flakiness found in the CI execution in https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4456

internal/service/project/resource_project_schema.go
lantoli: ok, thanks

QuiqueSanMongo (approved):
LGTM!

PR #4473, opened 2026-05-28
feat: Adds mongodbatlas_stream_connection_failover resource

jandersongo:
https://jira.mongodb.org/browse/CLOUDP-407496

Implements create, update, and delete of failover connections via the MMS failover connections API endpoints.


jandersongo:
bump

jandersongo:
bump

Edmondton:
bump

docs/resources/stream_connection_failover.md
```diff
@@ -0,0 +1,189 @@
+---
+subcategory: "Streams"
+---
+
+# Resource: mongodbatlas_stream_connection_failover
+
+`mongodbatlas_stream_connection_failover` provides a Stream Failover Connection resource. The resource lets you create, update, delete, and import a failover (regional-alternate) connection for an existing stream connection.
+
+A failover connection shares the primary connection's `connection_name` and is created for one of the workspace's failover regions (see the `failover_regions` argument of `mongodbatlas_stream_workspace`). It is uniquely identified by its region and its computed `id` (the failover connection id).
```
jromo-mdb:
```suggestion
A failover connection shares the primary connection's `connection_name` and is created for one of the workspace's failover regions (see the `failover_regions` argument of `mongodbatlas_stream_workspace`). It is uniquely identified by its region and its computed `id` (the failover connection ID).
```

docs/resources/stream_connection_failover.md
```diff
@@ -0,0 +1,189 @@
...
+<a id="nestedatt--security"></a>
+### Nested Schema for `security`
+
+Optional:
+
+- `broker_public_certificate` (String)
+- `protocol` (String)
+
+## Import
+You can import a stream failover connection using the workspace name, project id, connection name, and failover connection id, in the format `WORKSPACE_NAME-PROJECT_ID-CONNECTION_NAME-FAILOVER_CONNECTION_ID`. For example:
```
jromo-mdb:
```suggestion
You can import a stream failover connection using the workspace name, project ID, connection name, and failover connection ID, in the format `WORKSPACE_NAME-PROJECT_ID-CONNECTION_NAME-FAILOVER_CONNECTION_ID`. For example:
```

templates/resources/stream_connection_failover.md.tmpl
```diff
@@ -0,0 +1,23 @@
+---
+subcategory: "Streams"
+---
+
+# {{.Type}}: {{.Name}}
+
+`{{.Name}}` provides a Stream Failover Connection resource. The resource lets you create, update, delete, and import a failover (regional-alternate) connection for an existing stream connection.
+
+A failover connection shares the primary connection's `connection_name` and is created for one of the workspace's failover regions (see the `failover_regions` argument of `mongodbatlas_stream_workspace`). It is uniquely identified by its region and its computed `id` (the failover connection id).
```
jromo-mdb:
```suggestion
A failover connection shares the primary connection's `connection_name` and is created for one of the workspace's failover regions (see the `failover_regions` argument of `mongodbatlas_stream_workspace`). It is uniquely identified by its region and its computed `id` (the failover connection ID).
```

templates/resources/stream_connection_failover.md.tmpl
```diff
@@ -0,0 +1,23 @@
...
+
+## Example Usages
+
+{{ tffile (printf "examples/%s/main.tf" .Name )}}
+
+{{ .SchemaMarkdown | trimspace }}
+
+## Import
+You can import a stream failover connection using the workspace name, project id, connection name, and failover connection id, in the format `WORKSPACE_NAME-PROJECT_ID-CONNECTION_NAME-FAILOVER_CONNECTION_ID`. For example:
```
jromo-mdb:
```suggestion
You can import a stream failover connection using the workspace name, project ID, connection name, and failover connection ID, in the format `WORKSPACE_NAME-PROJECT_ID-CONNECTION_NAME-FAILOVER_CONNECTION_ID`. For example:
```

templates/resources/stream_connection_failover.md.tmpl
```diff
@@ -0,0 +1,23 @@
...
+```
+$ terraform import mongodbatlas_stream_connection_failover.test "yourWorkspaceName-6117ac2fe2a3d04ed27a9871-KafkaConnection-671f9c2fe2a3d04ed27a9880"
+```
+
+For more information see: [MongoDB Atlas API - Streams Failover Connection](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/group/endpoint-streams) Documentation.
```
jromo-mdb, line 24:
```suggestion
For more information see [MongoDB Atlas API - Streams Failover Connection](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/group/endpoint-streams) documentation.
```

jromo-mdb (approved):
LGTM with some style suggestions that you can easily commit

jromo-mdb (approved):
LGTM

jandersongo:
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # mongodbatlas_stream_connection.primary will be created
  + resource "mongodbatlas_stream_connection" "primary" {
      + authentication    = {
          + mechanism = "PLAIN"
          + password  = (sensitive value)
          + username  = (sensitive value)
        }
      + bootstrap_servers = "lkc-6kdyn2q.dom4g20rd0p.us-east-1.aws.confluent.cloud:9092"
      + connection_name   = "KafkaFailoverConn"
      + id                = (known after apply)
      + networking        = {
          + access = {
              + connection_id = "6a272a58d8176729234844a0"
              + type          = "PRIVATE_LINK"
            }
        }
      + project_id        = "685d759376489629c36fcd18"
      + security          = {
          + protocol = "SASL_SSL"
        }
      + type              = "Kafka"
      + workspace_name    = "failover-test-ws"
    }

  # mongodbatlas_stream_connection_failover.dublin will be created
  + resource "mongodbatlas_stream_connection_failover" "dublin" {
      + authentication            = {
          + mechanism = "PLAIN"
          + password  = (sensitive value)
          + username  = (sensitive value)
        }
      + bootstrap_servers         = "pkc-z9doz.eu-west-1.aws.confluent.cloud:9092"
      + connection_name           = "KafkaFailoverConn"
      + delete_on_create_timeout  = true
      + failover_connection_id    = (known after apply)
      + networking                = {
          + access = {
              + connection_id = "6a272b96ed33c1e2a6cb6a73"
              + type          = "PRIVATE_LINK"
            }
        }
      + project_id                = "685d759376489629c36fcd18"
      + public_private_networking = (known after apply)
      + region                    = "DUBLIN_IRL"
      + security                  = {
          + protocol = "SASL_SSL"
        }
      + state                     = (known after apply)
      + type                      = "Kafka"
      + workspace_name            = "failover-test-ws"
    }

  # mongodbatlas_stream_workspace.this will be created
  + resource "mongodbatlas_stream_workspace" "this" {
      + data_process_region = {
          + cloud_provider = "AWS"
          + region         = "VIRGINIA_USA"
        }
      + failover_regions    = [
          + {
              + cloud_provider = "AWS"
              + region         = "DUBLIN_IRL"
            },
        ]
      + hostnames           = (known after apply)
      + id                  = (known after apply)
      + project_id          = "685d759376489629c36fcd18"
      + stream_config       = {
          + max_tier_size = (known after apply)
          + tier          = "SP10"
        }
      + workspace_name      = "failover-test-ws"
    }

Plan: 3 to add, 0 to change, 0 to destroy.
mongodbatlas_stream_workspace.this: Creating...
mongodbatlas_stream_workspace.this: Creation complete after 1s [id=6a4ea9e8675bf4b3220f436e]
mongodbatlas_stream_connection.primary: Creating...
mongodbatlas_stream_connection.primary: Still creating... [00m10s elapsed]
...
mongodbatlas_stream_connection.primary: Still creating... [14m40s elapsed]
mongodbatlas_stream_connection.primary: Creation complete after 14m40s [id=failover-test-ws-685d759376489629c36fcd18-KafkaFailoverConn]
mongodbatlas_stream_connection_failover.dublin: Creating...
mongodbatlas_stream_connection_failover.dublin: Still creating... [00m10s elapsed]
...
mongodbatlas_stream_connection_failover.dublin: Still creating... [07m40s elapsed]
mongodbatlas_stream_connection_failover.dublin: Creation complete after 7m49s

Apply complete! Resources: 3 added, 0 changed, 0 destroyed.
```

.changelog/4473.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:new-resource
+resource/mongodbatlas_stream_connection_failover
```
AgustinBettati, line 2: will we want singular and plural data sources for this new resource?

docs/resources/stream_connection_failover.md
```diff
@@ -0,0 +1,212 @@
+# Resource: mongodbatlas_stream_connection_failover
+
+<!-- schema generated by tfplugindocs -->
```
AgustinBettati, line 73: Could we add a section that describe the resoure, as well as examples? Please reference existing resources that leverage autogenerated docs as reference, a template has to be defined.
We also need to add the import section.

docs/resources/stream_connection_failover.md
```diff
@@ -0,0 +1,212 @@
...
+## Schema
```
AgustinBettati, line 74: lets add examples for this resource.

tools/codegen/config.yml
```diff
@@ -871,6 +871,116 @@ resources:
...
+  stream_connection_failover:
+    read:
+      path: /api/atlas/v2/groups/{groupId}/streams/{tenantName}/connections/{connectionName}/failoverConnections/{failoverConnectionId}
+      method: GET
+    create:
+      path: /api/atlas/v2/groups/{groupId}/streams/{tenantName}/connections/{connectionName}/failoverConnections
+      method: POST
+      wait: &stream-connection-failover-create-wait
+        state_property: "state"
+        pending_states: ["PENDING"]
```
AgustinBettati: Using manual stream_connection resource as reference, it explicitly treats 404 as a pending state on create, with a comment saying the Streams GET is eventually consistent (resource_stream_connection.go:246).
Is it the same case for this API? Would we need to consider factoring that in as a pending state?

internal/serviceapi/streamconnectionfailover/resource_schema.go
```diff
@@ -0,0 +1,444 @@
...
+					},
+				},
+			},
+			"project_id": schema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "Unique 24-hexadecimal digit string that identifies your project, also known as `groupId` in the official documentation.",
+				PlanModifiers:       []planmodifier.String{customplanmodifier.CreateOnly()},
+			},
+			"headers": schema.MapAttribute{
+				Optional:            true,
```
AgustinBettati: Can we double check what happens when this attribute is unset on an update?
With current implementation I suspect removing the headers block from config produces a PATCH without the field, Atlas keeps the old headers, and apply fails with "Provider produced inconsistent result after apply".
Autogen has a configuration for sending null on PATCH request when the attribute is absent, but I am not certain if the API properly support handling an explicit null value for unsetting.

internal/serviceapi/streamconnectionfailover/resource_schema.go
```diff
@@ -0,0 +1,444 @@
...
+							"type": schema.StringAttribute{
+								Computed:            true,
+								Optional:            true,
+								MarkdownDescription: "Selected networking type. Either `PUBLIC` or `PRIVATE_LINK`. Defaults to `PUBLIC`. For AWS, Azure, and GCP connections, use `PRIVATE_LINK` for AWS PrivateLink, Azure Private Link, or GCP Private Service Connect (PSC) respectively.",
+							},
+						},
+					},
+				},
+			},
+			"region": schema.StringAttribute{
```
AgustinBettati: are `type` and `region` updateable? If not, we could look at defining a requires replace modifier following the manual implementation of stream_connection (we can sync on how to do this)

internal/serviceapi/streamconnectionfailover/resource_schema.go
```diff
@@ -0,0 +1,444 @@
...
+					},
+					"scope": schema.StringAttribute{
+						Optional:            true,
+						MarkdownDescription: "OIDC scope parameter defining the access permissions requested.",
+					},
+					"ssl_certificate": schema.StringAttribute{
+						Optional:            true,
+						MarkdownDescription: "SSL certificate for client authentication to Kafka.",
+					},
+					"ssl_key": schema.StringAttribute{
```
AgustinBettati, line 60: to confirm, this attribute is not sensitive?

internal/testutil/acc/shared_resource.go
```diff
@@ -62,8 +63,8 @@ func cleanupSharedResources() error {
 			hasError = true
 		}
 	}
-	if sharedInfo.streamInstanceName != "" {
-		fmt.Printf("Deleting execution stream instance: %s, project id: %s\n", sharedInfo.streamInstanceName, firstProjectID)
+	if sharedInfo.streamInstanceName != "" || sharedInfo.streamInstanceFailoverName != "" {
+		fmt.Printf("Deleting execution stream instances in project id: %s\n", firstProjectID)
 		if _, err := clean.RemoveStreamInstances(context.TODO(), false, ConnV2(), firstProjectID); err != nil {
```
AgustinBettati, line 67: q: does clean.RemoveStreamInstances delete both stream instance and stream instance failovers?

tools/codegen/config.yml
jandersongo: nice catch! 
confirmed it's the same eventual-consistency case here, and it would have caused flaky create failures.

The autogen wait polls GET until state is READY, and it maps a 404 during polling to the sentinel state DELETED. Since the create wait only had PENDING as a pending state, a transient post-create 404 (the read side not yet caught up) would show up as an unexpected DELETED state and fail the apply.

Fixed by mirroring the manual resource: DELETED is now a pending state on create only.
```
create:
  wait:
    pending_states: ["PENDING", "DELETED"]  # DELETED = 404 sentinel; Streams GET is eventually consistent after create
    target_states: ["READY"]
update:
  wait:
    pending_states: ["PENDING"]  # connection already exists on update, so a 404 is unexpected
```

delete is unchanged (there DELETED is the target). Left update without DELETED to match the manual resource's create-only handling (resource_stream_connection.go:246).

internal/serviceapi/streamconnectionfailover/resource_schema.go
jandersongo: Good catch — it should be sensitive, and it wasn't. ssl_key is the client's SSL private key, so I've marked it sensitive.

Root cause: the autogen generator only derives sensitivity from OpenAPI format: password (IsSensitive() → Schema.Format == "password"). In the spec, sslKey and sslKeyPassword are marked writeOnly: true (so the API redacts them from responses) but are missing format: password — unlike password/clientSecret, which have both. So they weren't auto-marked.

Fixed with an override in config.yml:
```
authentication.ssl_key:
  sensitive: true
```
(ssl_key_password was already overridden; ssl_certificate left as-is since it's public cert material.)

I've also filed ticket to add format: password to sslKey/sslKeyPassword upstream, matching password/clientSecret — once that lands these overrides can be dropped. https://jira.mongodb.org/browse/CLOUDP-421890

internal/testutil/acc/shared_resource.go
jandersongo: Yes. `clean.RemoveStreamInstances` lists all stream workspaces/instances in the project (ListStreamWorkspaces) and deletes each one — it's not scoped to a single instance name. So the one call in cleanup removes both streamInstanceName and streamInstanceFailoverName (both live in the same firstProjectID), plus any others.

That's why we have a single call and why the guard is widened to streamInstanceName != "" || streamInstanceFailoverName != "" — it just needs to fire if either shared instance was created; the function then sweeps all instances in the project regardless of which field(s) are set. It also handles the `STREAM_TENANT_HAS_STREAM_PROCESSORS` case by deleting processors first, so a failover instance with a processor still cleans up.

internal/serviceapi/streamconnectionfailover/resource_schema.go
jandersongo: I tried it with the `config` field first and it failed. TF produced inconsistent result after apply … `.config:` was `null`, but ended up being `cty.MapVal(...)`. The PATCH omitted the field, Atlas kept the old value, and the read-back no longer matched the planned null.

On the null-vs-empty question you raised: I went with empty, not null — `request_body_usage: send_null_as_empty_on_update` on both config and headers. Two reasons it's the safe choice here:

- The existing `mongodbatlas_stream_connection` resource already clears headers on this same PATCH endpoint by sending an empty map (`NewStreamConnectionUpdateReq` → "the headers will be removed by sending an empty headers map to the PATCH endpoint"), so empty-clears is proven-tolerated by the API; an explicit null is not.

- Autogen's read side normalizes an empty/absent map back to null (unmarshal.go: "Keep current map if both model and JSON maps are zero-len … avoids inconsistent result after apply"), so we don't just trade the error for the reverse {} vs null mismatch.

Coverage: the acc test now removes config on update, and since these are Kafka connections with headers unset, every update step also sends headers: {} — so the empty-map path is exercised for both.

I scoped this to `config/headers` (the reachable maps). `schema_registry_urls` and the nested-object blocks have the same theoretical limitation, but **they only apply to connection types that don't yet support failover**, and the manual resource doesn't clear those either — I'd rather wire clearing for them alongside acc coverage when those types become reachable, than ship unverifiable behavior now.

jandersongo: On reflection, headers (and the other non-Kafka/Cluster fields) don't apply to failover connections at all, so I've dropped them from the schema entirely rather than special-casing their update behavior. config (Kafka) keeps the send_null_as_empty_on_update fix.

.changelog/4473.txt
jandersongo: since this pr is already quite large, would it be ok if we implemented this in a follow up pr?

internal/testutil/acc/shared_resource.go
```diff
@@ -228,6 +229,28 @@ func ProjectIDExecutionWithStreamInstance(tb testing.TB) (projectID, streamInsta
 	return projectID, sharedInfo.streamInstanceName
 }
 
+// ProjectIDExecutionWithStreamInstanceWithFailover returns the project ID and the name of a stream
```
lantoli: shared create helpers should be used sparingly and when used by multiple tests and multiple resources as they add global complexity. In this case `ProjectIDExecutionWithStreamInstanceWithFailover` is only used in a resource (and just one test), so I would keep it in the resource test package.

(`ProjectIDExecutionWithStreamInstance` earns its place in shared_resource.go because it's used by 3 resource packages and ~25 call sites)

docs/resources/stream_connection_failover.md
```diff
@@ -0,0 +1,195 @@
+---
+subcategory: "Streams"
+---
+
+# Resource: mongodbatlas_stream_connection_failover
+
+`mongodbatlas_stream_connection_failover` provides a Stream Failover Connection resource. It lets you create, update, delete, and import a failover (regional-alternate) connection for an existing stream connection.
+
+A failover connection shares its primary connection's `connection_name` and is created for one of the workspace's failover regions (configured via the `failover_regions` argument of `mongodbatlas_stream_workspace`). It is uniquely identified by its `region` and its computed `failover_connection_id`, and carries its own regional connection configuration. Only `Kafka` and `Cluster` connection types support failover.
```
lantoli: the resource description says "uniquely identified by its region and its computed failover_connection_id", but the import format below uses only failover_connection_id (plus parent identifiers), not region. consider dropping region from the identification sentence, or rewording to say it's identified by failover_connection_id and carries a region attribute.

.changelog/4473.txt
AgustinBettati: sounds good, lets just make sure before we release next week (~wednesday) we have both resource and data sources merged into master

internal/serviceapi/streamconnectionfailover/resource_schema.go
AgustinBettati: Good to know, so the schema of stream_connection_failover is a subset of stream_connection? Would be good to have this reflected in the upstream API Spec, this way we dont need any config in the TF to ignore these fields (and avoid confusion on the API docs side)

docs/resources/stream_connection_failover.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Optional
+
+- `delete_on_create_timeout` (Boolean) Indicates whether to delete the resource being created if a timeout is reached when waiting for completion. When set to `true` and timeout occurs, it triggers the deletion and returns immediately without waiting for deletion to complete. When set to `false`, the timeout will not trigger resource deletion. If you suspect a transient error when the value is `true`, wait before retrying to allow resource deletion to finish. Default is `true`.
+- `timeouts` (Attributes) (see [below for nested schema](#nestedatt--timeouts))
+
+<!-- polymorphic attributes restructured by docpostprocess -->
+The following attributes depend on the value of `type`:
+
+#### `AWSKinesisDataStreams`
```
AgustinBettati: I think we have contradicting docs given we say `Only `Kafka` and `Cluster` connection types support failover.`.
API Spec should only have the correct types, but we can also assess mitigations on the generation side here if needed.

AgustinBettati (commented):
Thanks for addressing comments, have one remaining aspect to adjust

internal/serviceapi/streamconnectionfailover/resource_schema.go
jandersongo: agreed this belongs in the upstream spec! The failover endpoints currently reuse the shared StreamsConnection schema (discriminator covers all 9 connection types), which is why the provider has to carry the ignores list and why the generated docs still show AWSKinesisDataStreams/S3 groupings.

I created an MMS ticket to give the failover operations their own Kafka/Cluster schema (either a dedicated StreamsFailoverConnection model or a discriminator scoped to those two types). Once that lands and we re-fetch the spec, we drop the ignores block, regenerate, and the doc noise clears up for free. Then, the public API reference + SDK stop advertising inapplicable fields too.

https://jira.mongodb.org/browse/CLOUDP-421992

.changelog/4473.txt
jandersongo: great! thank you, i'll follow up after this pr is merged https://jira.mongodb.org/browse/CLOUDP-421994

docs/resources/stream_connection_failover.md
jandersongo: I added a small, general codegen capability — a discriminator_types allow-list in SchemaOptions — that prunes the discriminator mapping to the supported types before descriptions/validation are generated. On the failover resource:
`discriminator_types: [Kafka, Cluster]`

Now the docs show only Kafka/Cluster sections, networking's prefix reads "Optional for type: Kafka", and the generated ValidateDiscriminator only accepts those two types (a nice validation tightening too). It's opt-in, so no other resource's output changes, and it's covered by a codespec unit test.

tools/codegen/config.yml
```diff
@@ -871,6 +871,118 @@ resources:
...
+          # shared DTO covering every stream connection type).
+          "aws", # AWSKinesisDataStreams, AWSLambda, S3
+          "azure", # AzureBlobStorage
+          "gcp", # GCPPubSub
+          "headers", # Https
+          "url", # Https
+          "public_private_networking", # AzureBlobStorage, GCPPubSub
+          "provider", # SchemaRegistry (aliased to schema_registry_provider; ignore the source name)
+          "schema_registry_authentication", # SchemaRegistry
+          "schema_registry_urls", # SchemaRegistry
```
AgustinBettati, line 933: Highly recommend we align the schema of our API Spec (https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-creategroupstreamconnectionfailoverconnection) to only contain the types which are actually supported. This will fix inconsistent API docs, while greatly simplifying the config we have for this resource.
Would also consider that any new property added in stream connection schema (which is not Kafka or Cluster related) it will force us to add new ignore entries here.

AgustinBettati (approved):
LGTM, leaving a comment which I see relevant for future maintenance of this autogen resource

tools/codegen/config.yml
jandersongo: i will address this as part of this ticket https://jira.mongodb.org/browse/CLOUDP-421992  from this thread https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4473#discussion_r3559190336

thank you!

PR #4499, opened 2026-06-09
chore: Surface acceptance-test client init errors instead of nil panic

AgustinBettati:
## Description

Improves observability when the acceptance-test Atlas client fails to initialize.

**Problem (today's unclear error):** In the scheduled run [27176866675](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/27176866675), `TestAccFlexClusterRS_basic` failed with an opaque `SIGSEGV: nil pointer dereference` deep in test setup (`factory.go:34`). The panic gave no hint of the real cause: `acc.MongoDBClient` was nil because `init()` discarded the error from `config.NewClient` (`MongoDBClient, _ = config.NewClient(...)`), so every later access nil-paniced instead of reporting *why* the client wasn't built.

**Confirmed flaky:** A re-run on the same commit, [27206933653](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/27206933653), passed `flex_cluster` (✓ 4m17s) with no code change — so the failure was not a regression.

**Likely cause of the flaky failure:** The discarded error means we can only infer it, which is exactly the gap this PR closes. `NewClient`'s only non-deterministic, network-dependent step under service-account auth is the fail-fast OAuth token fetch (`service_account.go`, `tokenSource.Token()`). The rest of the run showed widespread `cloud-dev` instability (hundreds of `groups POST: HTTP 500`), so a transient token-endpoint failure during that window is the most plausible trigger. It surfaced as a nil-pointer panic only because the error was swallowed.

**Change:** `init()` now fails fast via `log.Fatalf` when `config.NewClient` returns an error (and `ConnV2UsingGov` does the same), so a failed client build aborts the test binary with the underlying cause — e.g. a service-account token fetch failure — instead of leaving a nil client that later panics with an opaque nil pointer dereference. Test-infrastructure only, no end-user impact.

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/testutil/acc/factory.go
+++ b/internal/testutil/acc/factory.go
@@ -1,6 +1,8 @@
 package acc
 
 import (
+	"fmt"
+	"log"
 	"maps"
 	"os"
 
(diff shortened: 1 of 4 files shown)


internal/testutil/acc/factory.go
```diff
@@ -76,5 +93,8 @@ func init() {
 		BaseURL:      os.Getenv("MONGODB_ATLAS_BASE_URL"),
 		RealmBaseURL: os.Getenv("MONGODB_REALM_BASE_URL"),
 	}
-	MongoDBClient, _ = config.NewClient(c, "")
+	sharedClient, clientInitErr = config.NewClient(c, "")
+	if clientInitErr != nil {
+		log.Printf("[ERROR] failed to initialize Atlas client for acceptance tests: %v", clientInitErr)
+	}
```
manupedrozo: We should probably fail here instead. Either panic() or log.Fatal() - same for Gov client. 
This would fail fast and remove the need for the MongoDBClient() to wrap every access.

Alternatively, we can check and fail in the Run() function which all acc tests call ([ptr](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/a98730f8b5d5e3689725ab5ed2d65ae020199c3b/internal/testutil/acc/shared_resource.go#L42))

EspenAlbert (approved):
Thank you!

internal/testutil/acc/factory.go
AgustinBettati: Good call, switched to failing fast in init() via log.Fatalf, the change is now essentially just the two error checks.

One aspect I doubted and checked: init() runs for unit tests too, but with no credentials config.NewClient makes no network call and returns no error (only the service-account path eagerly fetches a token), so credential-less unit runs are unaffected — the Fatalf only triggers when SA creds are present and the token endpoint fails.

Pushed as a follow-up commit.

PR #4502, opened 2026-06-12
feat: Adds `roles` attribute to service_account_project_assignments data source

AgustinBettati:
## Description

This PR incorporates the change surfaced by autogen visibility PR #4484: each `results` element of the
`mongodbatlas_service_account_project_assignments` data source now exposes the project `roles`.

- `roles` is generated as a **Set** to align with the resource and singular data source, via a `type: set` override in `tools/codegen/config.yml` (the API spec lacks `uniqueItems: true` on `ServiceAccountGroup.roles`).
- Codegen: collection type overrides are now a no-op when the attribute already has the target type. Needed because override keys are matched after stripping the `results.` prefix, so the `roles` override also reaches the singular data source where `roles` is already a Set (previously a hard error).
- Acceptance tests assert `results.*.roles` in the plural data source checks; regenerated docs and added changelog entry.

Note: a follow-up PR can align `overrides` with how `aliases`/`ignores` handle plural paths by requiring full paths (e.g. `results.roles`) instead of stripping the prefix; today `results.`-prefixed override keys are silently ignored.

Link to any related issue(s): CLOUDP-413373

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/serviceapi/serviceaccountprojectassignment/resource_test.go
+++ b/internal/serviceapi/serviceaccountprojectassignment/resource_test.go
@@ -24,6 +24,7 @@ func TestAccServiceAccountProjectAssignment_singleAssignment(t *testing.T) {
 		orgID         = os.Getenv("MONGODB_ATLAS_ORG_ID")
 		projectIDs    = acc.MultipleProjectIDsExecution(t, 1)
 		resourceName0 = resourceName + "_0"
+		roles         = []string{"GROUP_OWNER", "GROUP_READ_ONLY"}
 	)
 
 	resource.ParallelTest(t, resource.TestCase{
@@ -32,8 +33,8 @@ func TestAccServiceAccountProjectAssignment_singleAssignment(t *testing.T) {
 		CheckDestroy:             checkDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: configBasic(orgID, projectIDs, []string{"GROUP_OWNER", "GROUP_READ_ONLY"}),
-				Check:  checkBasic(projectIDs),
+				Config: configBasic(orgID, projectIDs, roles),
+				Check:  checkBasic(projectIDs, roles),
 			},
 			{
 				ResourceName:                         resourceName0,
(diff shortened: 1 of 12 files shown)


tools/codegen/codespec/config.go
```diff
@@ -273,6 +273,9 @@ func requestOnlyRequiredOnCreateTransformation(attr *Attribute, _ *attrPaths, _
 func applyTypeOverride(override *config.Override, attr *Attribute) error {
 	switch *override.Type {
 	case config.Set:
+		if attr.Set != nil || attr.SetNested != nil { // no-op if the attribute already has the target type, e.g. an override applying to both the singular and plural data sources
```
lizo-mdb:
```suggestion
		if attr.Set != nil || attr.SetNested != nil { // no-op if the attribute already has the target type, for example, an override applying to both the singular and plural data sources.
```

internal/serviceapi/serviceaccountprojectassignment/resource_test.go
```diff
@@ -137,6 +137,14 @@ func checkBasic(projectIDs []string) resource.TestCheckFunc {
 		))
 	}
 	checks = append(checks, resource.TestCheckResourceAttr(dataSourcePluralName, "results.#", strconv.Itoa(len(projectIDs))))
+	for i := range projectIDs { // all assignments use the same roles so index-based checks are order-independent
```
lizo-mdb:
```suggestion
	for i := range projectIDs { // all assignments use the same roles, so index-based checks are order-independent.
```

lizo-mdb (approved):
LGTM with two micro-suggestions.

internal/serviceapi/serviceaccountprojectassignment/resource_test.go
```diff
@@ -137,6 +137,14 @@ func checkBasic(projectIDs []string) resource.TestCheckFunc {
...
+		checks = append(checks,
+			resource.TestCheckResourceAttrSet(dataSourcePluralName, fmt.Sprintf("results.%d.project_id", i)),
+			resource.TestCheckResourceAttr(dataSourcePluralName, fmt.Sprintf("results.%d.roles.#", i), "2"),
+			resource.TestCheckTypeSetElemAttr(dataSourcePluralName, fmt.Sprintf("results.%d.roles.*", i), "GROUP_OWNER"),
+			resource.TestCheckTypeSetElemAttr(dataSourcePluralName, fmt.Sprintf("results.%d.roles.*", i), "GROUP_READ_ONLY"),
```
manupedrozo: Non-blocking: I'd pass in the config roles set to `checkBasic()` instead of harcoding here.

manupedrozo (approved):
Great!

PR #4517, opened 2026-06-18
chore: Support x-xgen-array-semantic extension in terraform autogen for list vs set typing

AgustinBettati:
## Description

Consumes the new `x-xgen-array-semantic` OpenAPI extension in the autogen pipeline (`tools/codegen`): `set` generates a set-typed attribute, `list` (or absent) a list-typed one. Precedence is config.yml `type` override > extension > the `uniqueItems` heuristic. An invalid value fails generation via a sentinel error (`ErrInvalidArraySemantic`) wherever it appears — request, response, or query-param schema.

Also drops the previous `format: set` heuristic: it never appears in the Atlas spec, was never documented as a set signal, and is now superseded by this extension. `uniqueItems` remains the implicit fallback (could be removed in future given we now have formal guidance on how to represent a set type).

Part of the IPA-131 declarative-tooling-extensions work (epic CLOUDP-356095). Tooling-only: no runtime resource or docs change, and existing generated resources are unaffected since no current spec carries the extension.

Note: the query-param invalid-value path uses the same sentinel wiring as request/response but isn't covered by a dedicated fixture — array-typed query params don't occur in the Atlas spec today.

Link to any related issue(s): CLOUDP-412579

## Type of change:

## Required Checklist:


tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -334,22 +345,27 @@ func opRequestToAttributes(op *high.Operation, configuredVersion *string) (Attri
...
 		} else {
 			log.Printf("[WARN] Operation response body schema could not be mapped (OperationId: %s): %s", op.OperationId, err)
 		}
-		return nil, nil
+		return nil, nil, nil
 	}
 	responseAttributes, err := buildResourceAttrs(responseSchema, "", false)
 	if err != nil {
+		// A malformed declarative-tooling extension must fail generation wherever it appears; other
+		// mapping failures stay tolerant (warn and drop) to preserve existing behavior.
```
manupedrozo: Would remove/adjust this comment (reads a bit too LLM-y).
```suggestion
		// Hard-fail generation on malformed extensions.
```

tools/codegen/codespec/api_spec_schema.go
```diff
@@ -10,6 +12,18 @@ import (
...
+const arraySemanticExtensionKey = "x-xgen-array-semantic"
+
+const (
+	arraySemanticList = "list"
+	arraySemanticSet  = "set"
+)
+
+// ErrInvalidArraySemantic is returned when the x-xgen-array-semantic extension carries a value
+// other than "list" or "set". It is a sentinel so callers can distinguish a malformed extension
+// (which must fail generation) from a response/parameter schema that simply could not be mapped.
```
manupedrozo: "It is a sentinel so callers can distinguish a malformed extension" - Didn't get this part. Is it just explaining the case of: we return an error and a nil value if the extension value is invalid? Would omit if so.

manupedrozo (approved):

tools/codegen/codespec/api_spec_schema.go
AgustinBettati: Its explaining that the error is referenced through `errors.Is` branches. Removing comment altogether as it is not adding much value

tools/codegen/codespec/attribute.go
```diff
@@ -240,17 +240,22 @@ func (s *APISpecSchema) buildArrayAttr(name, ancestorsName string, computability
 
 	itemSchema, err := BuildSchema(s.Schema.Items.A)
 	if err != nil {
-		return nil, fmt.Errorf("error while building nested schema: %s", name)
+		return nil, fmt.Errorf("error while building nested schema %s: %w", name, err)
 	}
 
-	isSet := s.Schema.Format == OASFormatSet || (s.Schema.UniqueItems != nil && *s.Schema.UniqueItems)
+	isSet := s.Schema.UniqueItems != nil && *s.Schema.UniqueItems
```
lantoli, line 246: is there any current autogen resource that could benefit from this so we can use it now?

PR #4518, opened 2026-06-18
chore: Support x-xgen-server-computed-when-client-omitted extension in terraform autogen

AgustinBettati:
## Description

Consumes the new `x-xgen-server-computed-when-client-omitted` OpenAPI extension in the autogen pipeline (`tools/codegen`): when set on an optional request property, the generated attribute becomes Optional + Computed, preventing spurious drift when the client omits the field. The extension is ignored (with a warning) when present on a required property. config.yml computability overrides still take precedence.

Boolean placement is prohibited by the extension's IPA-131 registration and enforced upstream by the validator, so the pipeline treats booleans uniformly rather than special-casing them.

Part of the IPA-131 declarative-tooling-extensions work (epic CLOUDP-356095). Tooling-only: no runtime resource or docs change, and existing generated resources are unaffected since no current spec carries the extension.

Link to any related issue(s): CLOUDP-412580

## Type of change:

## Required Checklist:


tools/codegen/codespec/api_spec_schema.go
```diff
@@ -85,3 +87,25 @@ func (s *APISpecSchema) GetXGenDiscriminator() *DiscriminatorExtension {
...
+
+// GetXGenServerComputedWhenClientOmitted reports whether the property is annotated with the
+// x-xgen-server-computed-when-client-omitted extension set to true. Returns false if the extension
+// is absent or cannot be decoded.
+func (s *APISpecSchema) GetXGenServerComputedWhenClientOmitted() bool {
+	if s.Schema.Extensions == nil {
+		return false
+	}
+
+	node, ok := s.Schema.Extensions.Get(serverComputedWhenClientOmittedExtensionKey)
```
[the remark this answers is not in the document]
AgustinBettati: Leaving this as a known limitation rather than addressing it here:

The unwrap is a targeted fix, not a general handler. It was added in [#4444](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4444) to recover from a libopenapi v0.36.4 change that wrapped {$ref + description} into a synthetic allOf and broke type resolution (alert_configuration_api). description is the only sibling-override that appears in the flattened Atlas spec, so the unwrap only ever needed to preserve it. Extensions aren't used as $ref siblings. x-xgen-* extensions live on the inline property or the referenced component schema (per IPA-131), both of which BuildSchema handles normally.

tools/codegen/codespec/attribute.go
```diff
@@ -26,7 +27,23 @@ func buildResourceAttrs(s *APISpecSchema, ancestorsName string, isFromRequest bo
...
+		if isFromRequest && schema.GetXGenServerComputedWhenClientOmitted() {
+			switch computability {
+			case Optional:
+				computability = ComputedOptional
+			case Required:
+				qualifiedName := name
+				if ancestorsName != "" {
+					qualifiedName = ancestorsName + "." + name
+				}
+				log.Printf("[WARN] Ignoring %s on required property %q (kept required)", serverComputedWhenClientOmittedExtensionKey, qualifiedName)
```
manupedrozo, line 40: Is this something we should catch at the API level and ignore here? 
That is: do not allow setting this extension for required fields.
Or is there any case where `x-xgen-server-computed-when-client-omitted` applies to required fields?

manupedrozo (approved):

tools/codegen/codespec/attribute.go
AgustinBettati: > Is this something we should catch at the API level and ignore here?

Yes that would be ideal case, would leave for now as we dont have any spectral rule enforcing. Will add this to CLOUDP capturing improvements we can make to spectral validations.

tools/codegen/codespec/attribute.go
```diff
@@ -26,7 +27,23 @@ func buildResourceAttrs(s *APISpecSchema, ancestorsName string, isFromRequest bo
 			continue
 		}
 
-		attribute, err := schema.buildResourceAttr(name, ancestorsName, s.GetComputability(name), isFromRequest)
+		computability := s.GetComputability(name)
+		if isFromRequest && schema.GetXGenServerComputedWhenClientOmitted() {
```
lantoli, line 31: is there any current autogen resource that could benefit from this so we can use it now?

PR #4536, opened 2026-06-26
chore: Improvements on mongodbatlas_org_maintenance_settings resource after API improvements

marcabreracast:
## Description

Adapts the `mongodbatlas_org_maintenance_settings` resource to two API improvements:

**CLOUDP-414476 - Dedicated reset endpoint**: The API now exposes `POST /api/atlas/v2/orgs/{orgId}/maintenanceSettings:reset` for resetting org maintenance settings to defaults. The provider's destroy operation now calls this endpoint. 

**CLOUDP-415703 - Correct null handling**: The API no longer rewrites `waveAssignmentMode` to `MANUAL` in responses when the client sends `null`. The field is now omitted when no explicit preference is set, with `effectiveWaveAssignmentMode` reflecting the system value. The attribute `wave_assignment_mode` will only be `Optional` as a consequence, as it's a client-owned field. 

**NOTE**: dev OpenAPI spec is added in order to gain access to `/maintenanceSettings` endpoint. Once changes are public, this will be reverted to use the prod one. 

Link to any related issue(s): [CLOUDP-416869](https://jira.mongodb.org/browse/CLOUDP-416869)

## Type of change:

## Required Checklist:

## Further comments

### DIFF
--- a/internal/serviceapi/orgmaintenancesettings/resource_test.go
+++ b/internal/serviceapi/orgmaintenancesettings/resource_test.go
@@ -23,16 +23,16 @@ func TestAccOrgMaintenanceSettings_basic(t *testing.T) {
 		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
 		Steps: []resource.TestStep{
 			{
-				Config: configWithMode(orgID, "MANUAL"),
-				Check:  checkWithMode(orgID, "MANUAL"),
+				Config: configWithMode(orgID, "ENV_TAG_MAPPING"),
+				Check:  checkWithMode(orgID, "ENV_TAG_MAPPING"),
 			},
 			{
-				Config:   configWithMode(orgID, "MANUAL"),
+				Config:   configWithMode(orgID, "ENV_TAG_MAPPING"),
 				PlanOnly: true,
 			},
 			{
-				Config: configWithMode(orgID, "ENV_TAG_MAPPING"),
-				Check:  checkWithMode(orgID, "ENV_TAG_MAPPING"),
+				Config: configWithMode(orgID, "MANUAL"),
+				Check:  checkWithMode(orgID, "MANUAL"),
 			},
 			{
 				ResourceName:                         resourceName,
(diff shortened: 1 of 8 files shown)


internal/serviceapi/orgmaintenancesettings/resource_test.go
```diff
@@ -23,16 +23,16 @@ func TestAccOrgMaintenanceSettings_basic(t *testing.T) {
...
 		Steps: []resource.TestStep{
 			{
-				Config: configWithMode(orgID, "MANUAL"),
-				Check:  checkWithMode(orgID, "MANUAL"),
+				Config: configWithMode(orgID, "ENV_TAG_MAPPING"),
+				Check:  checkWithMode(orgID, "ENV_TAG_MAPPING"),
 			},
 			{
-				Config:   configWithMode(orgID, "MANUAL"),
+				Config:   configWithMode(orgID, "ENV_TAG_MAPPING"),
```
EspenAlbert: maybe a comment about all these different steps? Why `PlanOnly: true` steps?

internal/serviceapi/orgmaintenancesettings/resource_test.go
```diff
@@ -23,16 +23,16 @@ func TestAccOrgMaintenanceSettings_basic(t *testing.T) {
...
 				PlanOnly: true,
 			},
 			{
-				Config: configWithMode(orgID, "ENV_TAG_MAPPING"),
-				Check:  checkWithMode(orgID, "ENV_TAG_MAPPING"),
+				Config: configWithMode(orgID, "MANUAL"),
+				Check:  checkWithMode(orgID, "MANUAL"),
 			},
```
EspenAlbert, line 31: any reason for swapping the values from before?

EspenAlbert (approved):
Great work on cleaning up the API 👏

internal/serviceapi/orgmaintenancesettings/resource_test.go
marcabreracast: Unfortunately, the first step on the test is broken after the implementation of [CLOUDP-415703](https://jira.mongodb.org/browse/CLOUDP-415703). 

I detected that if a user that has never tried to set `maintenance_wave_mode` in their org defines `maintenance_wave_mode = "MANUAL"`, the API does not return the set attribute, which is unexpected behaviour 🙁 I've reached out to the upstream team about this topic since it's an important use case, but in the meantime I figured that since it does not cause issues for `ENV_TAG_MAPPING`, I could swap the order. There's not a CLOUDP for this yet, but not sure if we want the tests to fail on purpose, no strong preference.

internal/serviceapi/orgmaintenancesettings/resource_test.go
marcabreracast: We're using in `PlanOnly` in `maintenance_window` on the unset field step for the API after the checkEmpty step, but you're right it does not make much sense to have an additional check in the first one. Addressed in [10e888d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4536/commits/10e888ddbc9e5d43e59a9030a47c2936fc6f18d6)

internal/serviceapi/orgmaintenancesettings/resource_test.go
EspenAlbert: Ohh, I would keep the test as is as it was (to showcase the error), users will feel this pain if we merge

marcabreracast: Yeah makes sense. Reverted it in [10e888d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4536/commits/10e888ddbc9e5d43e59a9030a47c2936fc6f18d6) and left a comment so I can follow-up

marcabreracast: Funny observation, tests are passing now because the org had been set to `ENV_TAG_MAPPING`, which removes the issue. However, this still will be a failure for users that set their org to `MANUAL` for the first time. Added a CLOUDP to keep track of this

PR #4564, opened 2026-07-10
feat: Supports is_mongodbgov_cloud in AWS Secrets Manager payload

AgustinBettati:
## Description

Makes `is_mongodbgov_cloud` resolve consistently across credential sources: the AWS Secrets Manager secret payload and provider-block attributes. This mirrors how `base_url` already behaves.

Motivation: on the AWS Secrets Manager auth path, the secret payload is the single source for the provider configuration. Since v2.1.0, config attributes set outside the secret are ignored, and `is_mongodbgov_cloud` had no secret-key form at all, so gov could only be selected via a provider attribute. This surfaced in HELP-96418, where a provider-block `base_url` was silently dropped on the Secrets Manager path.

Changes:
- Move gov-to-`base_url` resolution from the provider package into `config.GetCredentials`, applied once to the single winning credential source. This keeps resolution within one source (no cross-source mixing) and makes the three sources symmetric.
- Add `is_mongodbgov_cloud` as a key on the `Credentials` struct (readable from the secret JSON).
- Provider entry points (Plugin Framework and SDKv2) now pass the raw `base_url` and the gov flag through, and the duplicated provider-side gov helper is removed.
- Document the supported secret-payload config keys.

Additional identified inconsistency: unlike `base_url` and other config attributes, `is_mongodbgov_cloud` cannot be set through an environment variable. This PR does not close that gap (no `MONGODB_ATLAS_IS_MONGODBGOV_CLOUD` variable is added). A separate ticket will track adding env-variable support for parity (ensure alignment with team).

Link to any related issue(s): CLOUDP-421952

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [x] This change requires a documentation update
- [x] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
[...]

PR #4581, opened 2026-07-22
feat: Warn when base_url, realm_base_url, or is_mongodbgov_cloud are set alongside AWS Secrets Manager credentials

AgustinBettati:
## Description

On the AWS Secrets Manager auth path (`assume_role.role_arn` set), the provider builds its client exclusively from the secret payload. `base_url`, `realm_base_url`, and `is_mongodbgov_cloud` set as provider attributes or env vars are silently ignored, which can route requests to the default control plane and produce a hard-to-diagnose 401 (surfaced by HELP-96418).

This behavior is already documented in the provider configuration guide. This PR adds an explicit warning so users are notified at plan/apply time rather than having to discover it in the docs. The warning is emitted during provider configuration when any of the three attributes is set while AWS Secrets Manager credentials are in use. It names the ignored attribute(s) and points to the secret payload as the fix. Behavior is otherwise unchanged.

The check is a pure function `config.AWSSecretsManagerIgnoredWarning`, wired into the framework provider's `Configure` only. The provider is a mux of the SDKv2 and framework providers, and both `Configure` functions run on every configuration, so emitting the warning from both would surface it twice. Following the existing convention (see the comment in `providerConfigure` in `provider_sdk2.go`), the SDKv2 provider does not surface credential warnings or errors; the framework provider owns them.

Out of scope: warning when configuration values are mixed across provider attributes and environment variables. That case is intentionally left out here, per the decision recorded in the related design doc (linked in CLOUDP-421953).

Jira: CLOUDP-421953
Related: #4564 (support `is_mongodbgov_cloud` in the secret payload)

## Type of change:

## Required Checklist:


internal/config/credentials.go
```diff
@@ -54,6 +55,31 @@ func GetCredentials(ctx context.Context, providerVars, envVars *Vars, getAWSCred
...
 
+// AWSSecretsManagerIgnoredWarning returns a warning when base_url, realm_base_url or
+// is_mongodbgov_cloud are set (as provider attributes or env vars) while AWS Secrets Manager
+// credentials are in use, since those values are ignored on that path and must be defined in
+// the secret payload instead. It returns an empty string when there is nothing to warn about.
+func AWSSecretsManagerIgnoredWarning(providerVars, envVars *Vars) string {
+	if CoalesceAWSVars(providerVars.GetAWS(), envVars.GetAWS()) == nil {
+		return ""
+	}
+	var ignored []string
```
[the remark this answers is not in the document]
AgustinBettati: Callers always pass non-nil (`NewEnvVars` is never nil; the `getProviderVars` nil path returns before this call), and this matches `GetCredentials` above. By-value would avoid the concern but `gocritic`'s `hugeParam` flags `Vars` (232 bytes), so pointer stays.

internal/config/credentials.go
```diff
@@ -54,6 +55,31 @@ func GetCredentials(ctx context.Context, providerVars, envVars *Vars, getAWSCred
...
+	if providerVars.BaseURL != "" || envVars.BaseURL != "" {
+		ignored = append(ignored, "base_url")
+	}
+	if providerVars.RealmBaseURL != "" || envVars.RealmBaseURL != "" {
+		ignored = append(ignored, "realm_base_url")
+	}
+	if providerVars.IsMongodbGovCloud || envVars.IsMongodbGovCloud {
```
lantoli, line 73: nit: explicit false is_mongodbgov_cloud won't be caught (e.g. having a true in some source and pretending to disable it for some specific cases) but i don't think that'll be a real issue

lantoli (approved):

elyse-mdb (approved):

internal/config/credentials.go
AgustinBettati: True, dont believe it is worth catching as an explicit `false` has nothing being ignored to warn about

PR #4530, opened 2026-06-23
feat: Add failover_regions to stream_workspace and failover_enabled to stream_processor

jandersongo:
⚠️ do not merge until after the TF release on June 30th ⚠️ 

https://jira.mongodb.org/browse/CLOUDP-405521

https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4441

Here's a summary of what was done:

## mongodbatlas_stream_workspace: add failover_regions

- New `failover_regions` attribute. Each element has `cloud_provider` and `region` (both required).
- `failover_regions` is **write-once**: it can be set via PATCH exactly once (null → value). Once configured it cannot be modified — any change triggers workspace replacement (`RequiresReplace`).
- `data_process_region` and `failover_regions` are **mutually exclusive** in the backend PATCH body. The provider enforces this: if both fields change in the same `terraform apply`, an error is returned before any API call is made.
- The PATCH body correctly sends either `failoverRegions` or `cloudProvider`+`region`, never both.
- `mongodbatlas_stream_instance` gains a computed-only `failover_regions` that reads back any failover regions returned by the API.
- Docs updated with a "With Failover Regions" example, argument reference, and nested attribute table.

test cases: 
1. create a workspace with failover regions: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#issuecomment-4795752711
2. create a workspace without failover region then update it with a failover region: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#issuecomment-4795805461
3. update failover region on workspace https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#issuecomment-4795841636
4. attempt to update data process region and failover region https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#issuecomment-4795921107
5. create processor with failover enabled https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#issuecomment-4795955856
6. update processor by disabling failover https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#issuecomment-4795968417

PR #4597, opened 2026-07-27
fix: Caches Service Account tokens per client_id

EspenAlbert:
## Description

Fixes org create with a nested `service_account` block when the provider is itself authenticated with a Service Account.

**Before:** `getTokenSource` kept a single process-wide SA token source. After `CreateOrg` returned the new org SA, switching credentials failed with `service account credentials changed`. `newSAClient` swallowed that error and fell back to the org-creator client, so the post-create settings `PATCH` returned `USER_CANNOT_ACCESS_ORG` and rollback `DELETE` failed the same way (dangling org).

**After:** Token sources are cached per `client_id`, so one provider process can hold both the org-creator SA and the newly created org SA. Same `client_id` with a different secret or `base_url` still errors. `CloseTokenSource` revokes all cached sources.

Link to related issue(s): [CLOUDP-428116](https://jira.mongodb.org/browse/CLOUDP-428116)

Test run: https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/30368242613

## Type of change

## Required Checklist

- [x] I have run `make fix` / lint and verified my code
- [x] I have added a changelog entry (`.changelog/4597.txt`) for this bug fix

### DIFF
--- a/internal/config/export_test.go
+++ b/internal/config/export_test.go
@@ -30,7 +30,7 @@ func ResetRevokeTokenForTest() {
 func ResetSAInfoForTest() {
 	saTokenSourceCache.mu.Lock()
 	defer saTokenSourceCache.mu.Unlock()
-	saTokenSourceCache.sources = make(map[string]*saTokenSourceEntry)
+	saTokenSourceCache.sources = nil
 	saTokenSourceCache.closed = false
 }
 
--- a/internal/config/export_test.go
+++ b/internal/config/export_test.go
@@ -30,7 +30,7 @@ func ResetRevokeTokenForTest() {
 func ResetSAInfoForTest() {
 	saTokenSourceCache.mu.Lock()
 	defer saTokenSourceCache.mu.Unlock()
-	saTokenSourceCache.sources = nil
+	saTokenSourceCache.sources = make(map[string]*saTokenSourceEntry)
 	saTokenSourceCache.closed = false
 }
 
--- a/internal/config/export_test.go
+++ b/internal/config/export_test.go
@@ -16,15 +16,26 @@ func ResetCreateTokenSourceForTest() {
 	createTokenSourceFn = defaultCreateTokenSource
 }
 
+// SetRevokeTokenForTest records which clientIDs get revoked without hitting the network.
+func SetRevokeTokenForTest(fn func(clientID string)) {
+	revokeTokenFn = func(clientID string, _ *saTokenSourceEntry) {
+		fn(clientID)
+	}
+}
+
+func ResetRevokeTokenForTest() {
+	revokeTokenFn = defaultRevokeToken
+}
+
 func ResetSAInfoForTest() {
-	saInfo.mu.Lock()
-	defer saInfo.mu.Unlock()
-	saInfo.sources = nil
-	saInfo.closed = false
+	saTokenSourceCache.mu.Lock()
+	defer saTokenSourceCache.mu.Unlock()
+	saTokenSourceCache.sources = nil
+	saTokenSourceCache.closed = false
 }
 
 func SetSAInfoClosedForTest(closed bool) {
-	saInfo.mu.Lock()
-	defer saInfo.mu.Unlock()
-	saInfo.closed = closed
+	saTokenSourceCache.mu.Lock()
+	defer saTokenSourceCache.mu.Unlock()
+	saTokenSourceCache.closed = closed
 }
--- a/internal/config/export_test.go
+++ b/internal/config/export_test.go
@@ -27,14 +27,16 @@ func ResetRevokeTokenForTest() {
 	revokeTokenFn = defaultRevokeToken
 }
 
-func ResetSAInfoForTest() {
+func ResetSATokenSourceCacheForTest() {
 	saTokenSourceCache.mu.Lock()
 	defer saTokenSourceCache.mu.Unlock()
 	saTokenSourceCache.sources = nil
+	saTokenSourceCache.baseURL = ""
+	saTokenSourceCache.terraformVersion = ""
 	saTokenSourceCache.closed = false
 }
 
-func SetSAInfoClosedForTest(closed bool) {
+func SetSATokenSourceCacheClosedForTest(closed bool) {
 	saTokenSourceCache.mu.Lock()
 	defer saTokenSourceCache.mu.Unlock()
 	saTokenSourceCache.closed = closed
--- a/internal/config/export_test.go
+++ b/internal/config/export_test.go
@@ -1,0 +1,30 @@
+package config
+
+import "github.com/mongodb/atlas-sdk-go/auth"
+
+// Test helpers exported only for package config_test (see service_account_test.go).
+
+func GetTokenSourceForTest(clientID, clientSecret, baseURL, terraformVersion string) (auth.TokenSource, error) {
+	return getTokenSource(clientID, clientSecret, baseURL, terraformVersion)
+}
+
+func SetCreateTokenSourceForTest(fn func(clientID, clientSecret, baseURL, terraformVersion string) (auth.TokenSource, error)) {
+	createTokenSourceFn = fn
+}
+
+func ResetCreateTokenSourceForTest() {
+	createTokenSourceFn = defaultCreateTokenSource
+}
+
+func ResetSAInfoForTest() {
+	saInfo.mu.Lock()
+	defer saInfo.mu.Unlock()
+	saInfo.sources = nil
+	saInfo.closed = false
+}
+
+func SetSAInfoClosedForTest(closed bool) {
+	saInfo.mu.Lock()
+	defer saInfo.mu.Unlock()
+	saInfo.closed = closed
+}
--- a/internal/config/export_test.go
+++ b/internal/config/export_test.go
@@ -30,7 +30,7 @@ func ResetRevokeTokenForTest() {
 func ResetSATokenSourceCacheForTest() {
 	saTokenSourceCache.mu.Lock()
 	defer saTokenSourceCache.mu.Unlock()
-	saTokenSourceCache.sources = nil
+	saTokenSourceCache.sources = make(map[string]*saTokenSourceEntry)
 	saTokenSourceCache.baseURL = ""
 	saTokenSourceCache.terraformVersion = ""
 	saTokenSourceCache.closed = false
--- a/internal/config/service_account.go
+++ b/internal/config/service_account.go
@@ -30,7 +30,7 @@ var saTokenSourceCache = struct {
 	terraformVersion string
 	mu               sync.Mutex
 	closed           bool
-}{}
+}{sources: make(map[string]*saTokenSourceEntry)}
 
 // createTokenSourceFn is the OAuth token-source factory; overridden in unit tests.
 var createTokenSourceFn = defaultCreateTokenSource
@@ -60,10 +60,10 @@ func defaultRevokeToken(clientID string, entry *saTokenSourceEntry) {
 }
 
 func getTokenSource(clientID, clientSecret, baseURL, terraformVersion string) (auth.TokenSource, error) {
-	// Hold the cache lock through token creation so concurrent callers with the same
-	// client_id cannot create duplicate, untracked tokens. This serializes SA init
-	// across client_ids (e.g. concurrent org creates with nested SAs); that path is
-	// rare enough that we prefer this over a finer-grained lock.
+	// Hold the cache lock through token creation so map updates stay synchronous when
+	// concurrent callers use different client_ids (multi-org create with nested SAs).
+	// That is the only path impacted by locking across the network call; we accept that
+	// serialization over a finer-grained lock.
 	saTokenSourceCache.mu.Lock()
 	defer saTokenSourceCache.mu.Unlock()
 
@@ -72,9 +72,6 @@ func getTokenSource(clientID, clientSecret, baseURL, terraformVersion string) (a
 	}
 
 	baseURL = NormalizeBaseURL(baseURL)
-	if saTokenSourceCache.sources == nil {
-		saTokenSourceCache.sources = make(map[string]*saTokenSourceEntry)
-	}
 	if len(saTokenSourceCache.sources) > 0 && saTokenSourceCache.baseURL != baseURL {
 		return nil, fmt.Errorf("service account credentials changed")
 	}
--- a/internal/config/service_account.go
+++ b/internal/config/service_account.go
@@ -16,19 +16,20 @@ import (
 const saTokenExpiryBuffer = 10 * time.Minute
 
 type saTokenSourceEntry struct {
-	tokenSource      auth.TokenSource
-	clientSecret     string
-	baseURL          string
-	terraformVersion string
+	tokenSource  auth.TokenSource
+	clientSecret string
 }
 
 // saTokenSourceCache caches token sources per service account (keyed by clientID) so a single
 // provider process can authenticate as more than one SA (e.g. org-creator SA, then the SA
-// created with a new org).
+// created with a new org). baseURL and terraformVersion are process-wide: provider config and
+// nested SA clients share them.
 var saTokenSourceCache = struct {
-	sources map[string]*saTokenSourceEntry
-	mu      sync.Mutex
-	closed  bool
+	sources          map[string]*saTokenSourceEntry
+	baseURL          string
+	terraformVersion string
+	mu               sync.Mutex
+	closed           bool
 }{}
 
 // createTokenSourceFn is the OAuth token-source factory; overridden in unit tests.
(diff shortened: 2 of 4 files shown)


internal/config/service_account.go
```diff
@@ -69,7 +96,7 @@ func GetServiceAccountConfig(clientID, clientSecret, baseURL string) *clientcred
 	return config
 }
 
-// CloseTokenSource is called just before the provider finishes, it does a best-effort try to revoke the Service Access token.
+// CloseTokenSource is called just before the provider finishes, it does a best-effort try to revoke all cached Service Account tokens.
 // It sets saInfo.closed = true to avoid future calls to getTokenSource, that should't happen as the provider is exiting.
```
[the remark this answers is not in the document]
EspenAlbert: Fixed. Corrected `should't` to `shouldn't` in the `CloseTokenSource` doc comment (`internal/config/service_account.go:95`). — 🤖 claude-opus-4-8

internal/config/service_account.go
```diff
@@ -15,16 +15,40 @@ import (
...
+	mu      sync.Mutex
+	closed  bool
 }{}
 
+// createTokenSourceFn is the OAuth token-source factory; overridden in unit tests.
+var createTokenSourceFn = defaultCreateTokenSource
+
+func saCacheKey(clientID, baseURL string) string {
+	return clientID + "\x00" + NormalizeBaseURL(baseURL)
+}
```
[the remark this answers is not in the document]
EspenAlbert: Outdated, as the comment itself notes. The current implementation has no `saCacheKey` function. The cache is keyed on `clientID` directly, and `baseURL` is normalized once in `getTokenSource` before the map lookup. There is no duplicate normalization to remove.

elyse-mdb (approved):

internal/config/service_account_test.go
```diff
@@ -0,0 +1,116 @@
...
+	})
+
+	_, err := config.GetTokenSourceForTest("client-a", "secret-a", "https://cloud-qa.mongodb.com", "1.0.0")
+	require.NoError(t, err)
+	_, err = config.GetTokenSourceForTest("client-a", "secret-b", "https://cloud-qa.mongodb.com", "1.0.0")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "service account credentials changed")
+}
+
+func TestGetTokenSource_RejectsBaseURLChangeForSameClientID(t *testing.T) {
```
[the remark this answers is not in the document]
EspenAlbert: added in 7da84f9

internal/config/service_account.go
```diff
@@ -69,21 +91,20 @@ func GetServiceAccountConfig(clientID, clientSecret, baseURL string) *clientcred
...
 	}
 	saInfo.closed = true
-	if saInfo.tokenSource == nil { // No need to do anything if SA was not initialized.
-		return
-	}
-	if token, err := saInfo.tokenSource.Token(); err == nil {
-		conf := GetServiceAccountConfig(saInfo.clientID, saInfo.clientSecret, saInfo.baseURL)
-		ctx := context.WithValue(context.Background(), auth.HTTPClient, NewOAuthHTTPClient(saInfo.terraformVersion))
-		_ = conf.RevokeToken(ctx, token) // Best-effort, no need to do anything if it fails.
+	for _, entry := range saInfo.sources {
```
lantoli: the new multi-entry revocation path isn't covered by the added tests, so a regression could leave one or more Service Account tokens active until expiry. Consider adding a test that caches two sources, calls `CloseTokenSource`, and verifies both revoke requests

internal/config/export_test.go
```diff
@@ -0,0 +1,30 @@
+package config
```
lantoli, line 1: similar to [this comment in #4530](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4530#discussion_r3507374575), `export_test.go` lets `config_test` reach private implementation details and bypasses the package boundary. Consider testing through observable behavior or a small stable abstraction instead, and removing these test-only exports

lantoli (approved):

internal/config/service_account.go
```diff
@@ -15,16 +15,36 @@ import (
 // Renew token if it expires within 10 minutes to avoid authentication errors during Atlas API calls.
 const saTokenExpiryBuffer = 10 * time.Minute
 
-var saInfo = struct {
+type saCacheEntry struct {
 	tokenSource      auth.TokenSource
 	clientID         string
```
lantoli: `clientID` duplicates the `sources` map key, and `saInfo`/`saCacheEntry` are vague now that this is a token-source cache. Consider dropping `clientID`, using the key in `CloseTokenSource`, and renaming these to something like `saTokenSourceCache` and `saTokenSourceEntry`

internal/config/service_account.go
```diff
@@ -15,16 +15,36 @@ import (
...
+// saInfo caches token sources per service account so a single provider process can
+// authenticate as more than one SA (e.g. org-creator SA, then the SA created with a new org).
+var saInfo = struct {
+	sources map[string]*saCacheEntry
+	mu      sync.Mutex
+	closed  bool
 }{}
 
+// createTokenSourceFn is the OAuth token-source factory; overridden in unit tests.
+var createTokenSourceFn = defaultCreateTokenSource
```
oarbusi, line 36: since you are doing this to unit test the token creation, could we do it as well for revoking token/CloseTokenSource?

oarbusi (approved):
LGTM

internal/config/service_account.go
EspenAlbert: Added `TestCloseTokenSource_RevokesAllCachedSources`: it caches `client-a` and `client-b`, calls `CloseTokenSource`, and asserts (via an injected `revokeTokenFn`) that both client IDs are revoked. Revocation is now behind `revokeTokenFn` so the test observes the calls without hitting the network. This also addresses oarbusi's request in thread 06.

internal/config/service_account.go
EspenAlbert: Done. Dropped the `clientID` field from the entry, renamed `saCacheEntry` -> `saTokenSourceEntry` and `saInfo` -> `saTokenSourceCache`, and `CloseTokenSource` now ranges with the map key (`for clientID, entry := range ...`) and passes it to the revoke helper.

internal/config/service_account.go
EspenAlbert: Done. Extracted the revocation into `defaultRevokeToken`/`revokeTokenFn` (mirroring `createTokenSourceFn`), and `CloseTokenSource` now calls `revokeTokenFn(clientID, entry)`. New test `TestCloseTokenSource_RevokesAllCachedSources` injects the revoke fn and verifies both cached client IDs get revoked (also covers lantoli's thread 03).

internal/config/service_account_test.go
```diff
@@ -0,0 +1,157 @@
+package config_test
```
AgustinBettati, line 1: even those it is skipped in our CI could we adjust `TestAccConfigRSOrganization_ServiceAccount` to capture the case of a follow-up update to org settings? This should capture the e2e bug.

AgustinBettati (approved):
LGTM

internal/config/export_test.go
EspenAlbert: I prefer being pragmatic and exposing them.`export_test.go` (compiled only under `_test`) is the standard Go escape hatch for exactly this.

lantoli: fair enough, and I'm okay keeping it here. `export_test.go` is an escape hatch for this, but ideally we avoid needing the escape hatch in the first place.

it bypasses the black-box boundary and couples tests to internal structure, so a behavior-preserving production refactor can still require changing both the test exports and the tests.

specifically here, the tests reset the cache and replace global create/revoke functions, which also introduces the global-state concerns described in [Google's Go guidance](https://google.github.io/styleguide/go/best-practices.html#global-state).

internal/config/service_account_test.go
EspenAlbert: This test when run with SA in paying org is actually failing already. Creating a follow-up task to provision a paying org with SA credentials that we can use to trigger this.

Test run error:
```sh
Step 1/1 error: Error running apply: exit status 1
        
        Error: an error occurred when updating Organization settings: https://cloud-qa.mongodb.com/api/atlas/v2/orgs/6a68c9fbf60a14b762f3130d/settings PATCH: HTTP 401 Unauthorized (Error code: "USER_CANNOT_ACCESS_ORG") Detail: User cannot access this organization. Reason: Unauthorized. Params: [6a68c9fbf60a14b762f3130d], BadRequestDetail: .
         Unable to delete organization, there may be dangling resources: https://cloud-qa.mongodb.com/api/atlas/v2/orgs/6a68c9fbf60a14b762f3130d DELETE: HTTP 401 Unauthorized (Error code: "USER_CANNOT_ACCESS_ORG") Detail: User cannot access this organization. Reason: Unauthorized. Params: [6a68c9fbf60a14b762f3130d], BadRequestDetail: 
        
          with mongodbatlas_organization.test,
          on terraform_plugin_test.tf line 12, in resource "mongodbatlas_organization" "test":
          12: 	  resource "mongodbatlas_organization" "test" {
```

EspenAlbert: added: [\[CLOUDP-428709\] TF: Add a test setup that allows running `TestAccConfigRSOrganization_ServiceAccount` in the CI](https://jira.mongodb.org/browse/CLOUDP-428709)

internal/config/service_account.go
```diff
@@ -15,45 +15,79 @@ import (
...
-		return saInfo.tokenSource, nil
+		return entry.tokenSource, nil
 	}
 
-	// Use a new context to avoid "context canceled" errors as the token source is reused and can outlast the callee context.
-	ctx := context.WithValue(context.Background(), auth.HTTPClient, NewOAuthHTTPClient(terraformVersion))
-	conf := GetServiceAccountConfig(clientID, clientSecret, baseURL)
-	tokenSource := oauth2.ReuseTokenSourceWithExpiry(nil, conf.TokenSource(ctx), saTokenExpiryBuffer)
-	if _, err := tokenSource.Token(); err != nil { // Retrieve token to fail-fast if credentials are invalid.
+	tokenSource, err := createTokenSourceFn(clientID, clientSecret, baseURL, terraformVersion)
```
manupedrozo, line 85: We are holding the mutex through the http call that fetches the token here which is not great, it would only impact the multiple-org creation case, but we should avoid it:
* We should only acquire the lock when reading/writing to the cache. So: `aquire lock for reading the cache -> read from cache (assuming a cache miss) -> release the lock -> perform the http call -> re-aquire lock -> check we still get a cache miss for the clientID -> write to cache -> release lock`
* We could use an RWLock to optimize further, not a huge win.

lantoli: also noticed this, but decided not to raise it. Releasing the lock around the HTTP call is possible, but the read/call/recheck sequence alone allows concurrent cache misses for the same client ID to create multiple tokens, leaving the losing token active until expiry. Avoiding that requires revoking the unused token or adding per-client coordination, plus handling shutdown during creation.

given this is initialization-only and affects a narrow path, I don't think the additional complexity is justified and would leave the current locking as is, maybe adding a comment like: "Hold the cache lock through token creation to avoid duplicate, untracked token sources. This intentionally serializes initialization across client IDs."

manupedrozo: Just pushed a change (agreed with Espen offline), please both take a look: [e9b7149](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4597/commits/e9b71499e946fb6cb0d1299b00c6184bdd1da57d)

On "...create multiple tokens, leaving the losing token active until expiry.": this is not possible since we would never get the same SA (clientID) in 2 different orgs.

internal/config/service_account.go
```diff
@@ -15,45 +15,98 @@ import (
 // Renew token if it expires within 10 minutes to avoid authentication errors during Atlas API calls.
 const saTokenExpiryBuffer = 10 * time.Minute
 
-var saInfo = struct {
+type saTokenSourceEntry struct {
 	tokenSource      auth.TokenSource
-	clientID         string
 	clientSecret     string
 	baseURL          string
```
lantoli: q: can baseURL differ between entries in one provider process? All resources and data sources using that provider instance share its baseURL, and the nested SA client inherits it from currentClient, so it looks process-wide. If so, consider storing it once in saTokenSourceCache instead of every entry

internal/config/service_account.go
lantoli: thanks for taking this, concurrent organization creations use different client IDs so I'm okay with your change.

(the hypothetical issue would require the same client ID to be initialized concurrently, creating two tokens and leaving the uncached one active until expiry)

EspenAlbert: I don't think this extra complexity is worth the performance benefit. I'm not sure I understand the performance benefit. From what I can tell the only performance benefit:

*concurrent org creates with different nested SAs. Holding the process-wide mutex through the token HTTP call serializes unrelated client_ids*

> the hypothetical issue would require the same client ID to be initialized concurrently, creating two tokens and leaving the uncached one active until expiry

I think this is a real issue. The call chain looks like both `internal/provider/provider.go#L228` and `provider/internal/provider/provider_sdk2.go#L312` might ~~invoke this at the same time.~~
\*Looks like these are done sequentially: https://github.com/hashicorp/terraform-plugin-mux/blob/main/tf6muxserver/mux_server_ConfigureProvider.go#L24

internal/config/service_account.go
manupedrozo: +1 (but `baseURL` may need a little refactoring to move), `terraformVersion` can also be global and should be more straightforward.

internal/config/service_account.go
manupedrozo: > concurrent org creates with different nested SAs. Holding the process-wide mutex through the token HTTP call serializes unrelated client_ids

Yes, that's it. 
I don't see the complexity argument, we are just not holding the lock through the expensive call. 
Happy to discuss in a call today to close on this if we are still on the fence, it has limited impact either way.

internal/config/export_test.go
```diff
@@ -0,0 +1,41 @@
...
+	revokeTokenFn = func(clientID string, _ *saTokenSourceEntry) {
+		fn(clientID)
+	}
+}
+
+func ResetRevokeTokenForTest() {
+	revokeTokenFn = defaultRevokeToken
+}
+
+func ResetSAInfoForTest() {
```
lantoli: nit: these helpers still use the old `saInfo` name after the rename to `saTokenSourceCache` (`ResetSAInfoForTest`, `SetSAInfoClosedForTest`, `resetSAInfo` in the test file), consider renaming for consistency

internal/config/service_account.go
lantoli: i was thinking about alias cases such as:

```hcl
provider "mongodbatlas" {
  alias         = "first"
  client_id     = "same-id"
  client_secret = "same-secret"
}

provider "mongodbatlas" {
  alias         = "second"
  client_id     = "same-id"
  client_secret = "same-secret"
}
```

but I don't think this causes the race. It looks like each provider configuration gets its own process, so the aliases don't share package globals such as `saTokenSourceCache`

EspenAlbert: thanks for the discussion, added comment to clarify 🚀

internal/config/service_account.go
EspenAlbert: Done in d1f84b534: `baseURL` and `terraformVersion` now live on `saTokenSourceCache` (process-wide). Entries only keep `tokenSource` + `clientSecret`. A changed `baseURL` is rejected for any client_id once the cache has entries.

internal/config/export_test.go
EspenAlbert: Renamed in d1f84b534: `ResetSATokenSourceCacheForTest` / `SetSATokenSourceCacheClosedForTest` (and the test helper).

internal/config/service_account.go
```diff
@@ -15,45 +15,87 @@ import (
...
+	_ = conf.RevokeToken(ctx, token) // Best-effort, no need to do anything if it fails.
+}
+
 func getTokenSource(clientID, clientSecret, baseURL, terraformVersion string) (auth.TokenSource, error) {
-	saInfo.mu.Lock()
-	defer saInfo.mu.Unlock()
+	// Hold the cache lock through token creation so concurrent callers with the same
+	// client_id cannot create duplicate, untracked tokens. This serializes SA init
+	// across client_ids (e.g. concurrent org creates with nested SAs); that path is
+	// rare enough that we prefer this over a finer-grained lock.
```
manupedrozo: I would reword a bit:
* The issue is not callers with the same client_id, since we know that does not happen.
* The issue is that callers with different client_ids may also update the map. Updates to the map must be synchronous.
* Good to clarify that the multi-org case is the only one impacted, which we are ok with.

internal/config/service_account.go
```diff
@@ -15,45 +15,87 @@ import (
...
+	if saTokenSourceCache.closed {
 		return nil, fmt.Errorf("service account token source already closed")
 	}
 
 	baseURL = NormalizeBaseURL(baseURL)
-	if saInfo.tokenSource != nil { // Token source in cache.
-		if saInfo.clientID != clientID || saInfo.clientSecret != clientSecret || saInfo.baseURL != baseURL {
+	if saTokenSourceCache.sources == nil {
+		saTokenSourceCache.sources = make(map[string]*saTokenSourceEntry)
+	}
```
manupedrozo: Can we keep the eager map creation?

lantoli (approved):
thanks for the follow-ups

internal/config/service_account.go
EspenAlbert: good idea, kept in 0db146c

internal/config/service_account.go
EspenAlbert: reworded in https://github.com/mongodb/terraform-provider-mongodbatlas/commit/0db146ce208a2c70a518ead5e43c09c0cd070311 WDYT?

PR #4608, opened 2026-07-31
feat: Add options.resume_from_checkpoint to mongodbatlas_stream_processor

jandersongo:
https://jira.mongodb.org/browse/STREAMS-1948


jromo-mdb (approved):
LGTM

docs/resources/stream_processor.md
```diff
@@ -102,6 +102,31 @@ resource "mongodbatlas_stream_processor" "stream-processor-kafka-to-cluster-exam
...
 
+# Modifying the $source stage of an existing processor requires discarding the current checkpoint.
+# Without resume_from_checkpoint = false, the Atlas Admin API rejects the change with HTTP 400.
+resource "mongodbatlas_stream_processor" "stream-processor-source-change-example" {
+  project_id     = var.project_id
+  workspace_name = mongodbatlas_stream_instance.example.instance_name
+  processor_name = "sourceChangeProcessorName"
+  pipeline = jsonencode([
+    { "$source" = {
+      "connectionName" = resource.mongodbatlas_stream_connection.example-cluster.connection_name
```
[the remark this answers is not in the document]
oarbusi: not sure why we have "resource." prefix but it's wrong and should be removed. If you could remove them, it would be great

internal/service/streamprocessor/resource_test.go
```diff
@@ -92,6 +92,127 @@ func TestAccStreamProcessor_withFailoverEnabled(t *testing.T) {
 		}})
 }
 
+// TestAccStreamProcessor_resumeFromCheckpoint verifies that modifying the $source stage of a running
+// processor fails with HTTP 400 unless resume_from_checkpoint is false, because the Atlas Admin API
```
oarbusi: I don't see the test checking that the update fails with HTTP 400 when resume_from_checkpoint is true. See this test as reference where we check API errors in the test: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/ed218feab961eef26ffa03f11e6d153fc137a6ed/internal/service/privatelinkendpoint/resource_test.go#L272

internal/service/streamprocessor/resource_test.go
```diff
@@ -388,13 +509,37 @@ func TestAccStreamProcessor_createErrors(t *testing.T) {
 				Config:      config(t, projectID, workspaceName, processorName, streamprocessor.StoppedState, randomSuffix, invalidJSONConfig, testLogDestConfig, "", nil),
 				ExpectError: regexp.MustCompile("Invalid JSON String Value"),
 			},
+			{
+				// dlq and resume_from_checkpoint are both optional, but an empty options block is not valid.
+				// Kept before the final step: the post-test destroy reuses the last step's config, and a
+				// config rejected at validation time makes that destroy fail.
```
oarbusi: nit: no need for this kind fo detail in the comment IMO

oarbusi (approved):
a few comments but overall LGTM

docs/data-sources/stream_processor.md
```diff
@@ -148,6 +173,9 @@ output "stream_processors_results" {
 Read-Only:
 
 - `dlq` (Attributes) Dead letter queue for the stream processor. Refer to the [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/reference/glossary/#std-term-dead-letter-queue) for more information. (see [below for nested schema](#nestedatt--options--dlq))
+- `resume_from_checkpoint` (Boolean) When `true`, the stream processor resumes from its last checkpoint after being modified. Set to `false` to discard the existing checkpoint, which is required when modifying the `$source` stage of the `pipeline`. Defaults to `true` when not set.
```
AgustinBettati: Would consider removing from data source docs and ideally schema as it will never have a value given current state of API.

.changelog/4608.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:enhancement
+resource/mongodbatlas_stream_processor: Adds `options.resume_from_checkpoint` to allow modifying the `$source` stage of a `pipeline`, which the Atlas Admin API rejects while resuming from an existing checkpoint. The attribute only applies to updates and is not returned by the API, so it is always `null` in data sources
```
AgustinBettati: `not returned by the API` would note this is clear IPA violation for resource oriented design (https://mongodb.github.io/ipa/101#resources point 4), forcing more complexity of the tf resource as the read operation has to fetch the value from the config (and not API response).

Was any API champion involved in reviewing the API changes? Lets make sure to incorporate into the process if currently not being done.

internal/service/streamprocessor/resource_test.go
```diff
@@ -92,6 +92,127 @@ func TestAccStreamProcessor_withFailoverEnabled(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "state", streamprocessor.StartedState),
+					// The API never returns this value, so data sources always report null.
+					resource.TestCheckNoResourceAttr(dataSourceName, "options.resume_from_checkpoint"),
+				),
+			},
+			{
+				// The value is write-only, so it must be preserved from state rather than nulled on refresh.
+				Config:   configWithResumeFromCheckpoint(t, projectID, workspaceName, clusterName, processorName, "update", new(false)),
+				PlanOnly: true,
+			},
```
AgustinBettati: would double check but I believe this step is redundant, a non empty plan after apply would be caught in the prior step.

internal/service/streamprocessor/resource_schema.go
```diff
@@ -86,8 +92,17 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+						Optional:            true,
 						MarkdownDescription: "Dead letter queue for the stream processor. Refer to the [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/reference/glossary/#std-term-dead-letter-queue) for more information.",
+						// Attached here rather than to options so that options itself is not part of the
+						// set, which would make the validator always pass when the block is present.
+						// Uses an absolute path so the error message reads `options.resume_from_checkpoint`
+						// rather than a relative `options.dlq.<.resume_from_checkpoint`.
+						Validators: []validator.Object{
+							objectvalidator.AtLeastOneOf(path.Expressions{
+								path.MatchRoot("options").AtName("resume_from_checkpoint"),
+							}...),
```
AgustinBettati: Is there any API validation in this case? Would avoid the tf side validation an reply on the API error directly.

AgustinBettati (approved):
LGTM

internal/service/streamprocessor/resource_test.go
jandersongo: added tests!

internal/service/streamprocessor/resource_schema.go
jandersongo: Dropped the validator entirely. It only existed because nesting the flag under options forced `dlq` to become `Optional`; with the attribute at top level, `dlq` is `Required` again and `options = {}` is unreachable, so there's nothing to validate.

For the record on the original question: there was no API validation to rely on. An empty options meant the provider sent no options at all, so the API never saw anything to reject.

internal/service/streamprocessor/model.go
```diff
@@ -86,19 +101,33 @@ func NewStreamProcessorUpdateReq(ctx context.Context, plan *TFStreamProcessorRSM
...
-			},
+		apiOptions.Dlq = &admin.StreamsDLQ{
+			Coll:           dlqModel.Coll.ValueStringPointer(),
+			ConnectionName: dlqModel.ConnectionName.ValueStringPointer(),
+			Db:             dlqModel.DB.ValueStringPointer(),
 		}
+		optionsSet = true
+	}
+
+	// resume_from_checkpoint is a top-level attribute in Terraform but nested under options in the API.
```
AgustinBettati: Is there a strong reason why deviate from the API schema? For reference, we push strongly for terraform resources to have a 1:1 alignment with API schema. Having custom abstractions on the terraform side leads to customer confusion, and more maintenance burden on our side (due to additional mapping logic).

jandersongo: I had it nested originally (2ef3affb3) and moved it after the earlier comments. I think the problem is it can't really be 1:1 either way — resumeFromCheckpoint is in the modify request body but not in any response type, StreamsOptions only has dlq and links. Same thing you pointed out on the changelog re: IPA.

Nesting it meant:

- dlq had to go from Required to Optional, otherwise you can't set options without also specifying a DLQ
- that made options = {} valid, which is why the [validator you suggested we drop](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4608#discussion_r3727076743) was there
- options gets rebuilt from the API response in the data sources, so it showed up there always null (the [removal was suggested](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4608#discussion_r3726996880)). Fixing that properly needed either a DS-specific object type + converter, or an OmitFields change in common/conversion

Top level is 6 schema lines, leaves options/dlq identical to master, no validator, and one nil override keeps it out of both data sources.

Happy to move it back if you'd rather keep the alignment, i think it just means the dlq change, the validator and the data source handling all come back with it.

docs/data-sources/stream_processor.md
```diff
@@ -83,19 +83,42 @@ resource "mongodbatlas_stream_processor" "stream-processor-kafka-to-cluster-exam
...
     dlq = {
       coll            = "exampleColumn"
-      connection_name = resource.mongodbatlas_stream_connection.example-cluster.connection_name
+      connection_name = mongodbatlas_stream_connection.example-cluster.connection_name
       db              = "exampleDb"
     }
   }
 }
 
+# Modifying the $source stage of an existing processor requires discarding the current checkpoint.
```
AgustinBettati: I believe this example is not required in the data source page

docs/data-sources/stream_processors.md
```diff
@@ -83,19 +83,42 @@ resource "mongodbatlas_stream_processor" "stream-processor-kafka-to-cluster-exam
...
     dlq = {
       coll            = "exampleColumn"
-      connection_name = resource.mongodbatlas_stream_connection.example-cluster.connection_name
+      connection_name = mongodbatlas_stream_connection.example-cluster.connection_name
       db              = "exampleDb"
     }
   }
 }
 
+# Modifying the $source stage of an existing processor requires discarding the current checkpoint.
```
AgustinBettati: (same here) I believe this example is not required in the data source page

examples/mongodbatlas_stream_processor/main.tf
```diff
@@ -50,8 +50,8 @@ resource "mongodbatlas_stream_processor" "stream-processor-sample-example" {
   workspace_name = mongodbatlas_stream_instance.example.instance_name
   processor_name = "sampleProcessorName"
   pipeline = jsonencode([
-    { "$source" = { "connectionName" = resource.mongodbatlas_stream_connection.example-sample.connection_name } },
```
AgustinBettati: curious, was the example not working before?

internal/service/streamprocessor/model.go
AgustinBettati: > resumeFromCheckpoint is in the modify request body but not in any response type

I would say this is a separate issue which ideally should never be the case for a resource oriented design (https://mongodb.github.io/ipa/101#resources points 2 & 3). For now will focus on TF support given current API state.

Focusing on customer UX I still believe matching closer to the API will provide a better experience in the long run. Looking wider I do see we also have resumeFromCheckPoint in start operation:

> POST /api/atlas/v2/groups/{groupId}/streams/{tenantName}/processor/{processorName}:startWith
{ "resumeFromCheckpoint": false, "tier": "...", "startAtOperationTime": "..." }

If that ever gets supported in the future, placing that attribute would will be hard if we already have current attribute placed at the root level. This is speculative but serves as example of problem which can appear once diverging from the API schema.

Happy to get more input from the team if we feel strongly about having this API/TF schema difference.

jandersongo: thank you! i agree the nested version is more aligned with the api and looks cleaner. i can revert to that original commit. 
its just that esting means `dlq` has to go from Required to Optional, which then makes `options = {} `valid, which is what the validator was there for, and i believe you [suggested dropping that](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4608#discussion_r3727076743). And because options gets rebuilt from the API response in the data sources, the attribute showed up there always null, [which you asked me to remove](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4608#discussion_r3726996880). Moving it to the root happened to resolve both at once. 

Your `startWith` point is a better reason than either of them though, so happy to take the cost.

One thing I'd like your call on before I do it. The `nil` trick in `OverridenFields` only deletes top-level attributes, so once it's nested I can't drop it from the data sources that way. Options as I see them:

1. Add dotted-path support `tocommon/conversion` so the data sources can do `OmitFields: []string{"options.resume_from_checkpoint"}`. Nicest at the call site and reusable, but it's shared code every resource goes through, roughly 60 lines plus tests. I had this in an earlier revision and took it back out.
2. Override the whole `options` attribute in both DS schemas with a `dlq`- only copy. No shared-code change, but it duplicates the `dlq` fields and their descriptions so the two can drift, and it needs a DS-specific object type and converter since the state object type has to match the schema exactly.
3. Leave it in the data sources as always-null. No code, but it undoes your earlier comment.

Only thing worth flagging is that 1 is also what made this PR feel heavy earlier on, so if we go that way it's coming back in.

examples/mongodbatlas_stream_processor/main.tf
jandersongo: i removed them bc of this comment
https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4608#discussion_r3726438461

internal/service/streamprocessor/model.go
AgustinBettati: > dlq has to go from Required to Optional

`dlq` has only been required because it was a single attribute within a nested property `options`, as soon as options object get a new attribute added it will unavoidably become optional.

> Options as I see them:

I am inclining towards option 2 to keep this PR more simple. We can leave a clear comment on both duplicated schemas that they must be kept in sync, and why we have them separated.

jandersongo: went with option 2! 
`options.resume_from_checkpoint` is nested to match the API, and the data sources define their own copy of the options attribute instead of touching common/conversion. Both copies have a comment pointing at the other. 
Separately, there's a newer commit on here from a [Slack thread](https://mongodb.slack.com/archives/C08CTUS1U2C/p1787767797936599) with Joe. His point was that resumeFromCheckpoint is a one-shot instruction but Terraform config is persistent: once you set false to change a $source stage, it stays in the config, and the gate at the time only checked whether the pipeline had changed at all. So a later unrelated edit (to a downstream $match for ex.) would re-send it and discard the checkpoint a second time, silently.
The gate now only fires for the changes the API actually rejects: the $source stage, or a window's type/interval/hopSize/allowedLateness, including adding or removing a window. A window's inner pipeline is excluded since the planner accepts those. So a stale false can't clear the checkpoint on an unrelated update. It's still only ever sent when explicitly configured. Nothing is inferred.

internal/service/streamprocessor/resource_test.go
```diff
@@ -92,6 +92,134 @@ func TestAccStreamProcessor_withFailoverEnabled(t *testing.T) {
...
+				ExpectError: checkpointRejected,
+			},
+			{
+				// resume_from_checkpoint = true is rejected the same way as omitting it.
+				Config:      configWithResumeFromCheckpoint(t, projectID, workspaceName, clusterName, processorName, "update", new(true)),
+				ExpectError: checkpointRejected,
+			},
+			{
+				// Setting it to false discards the checkpoint and allows the $source change.
+				Config: configWithResumeFromCheckpoint(t, projectID, workspaceName, clusterName, processorName, "update", new(false)),
```
oarbusi, line 141: I would be nice to add another step in this test unsetting the resume_from_checkpoint

internal/service/streamprocessor/resource_schema.go
```diff
@@ -68,13 +68,20 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
 				Optional: true,
 				Validators: []validator.Object{
 					OptionsValidator(),
 				},
 				MarkdownDescription: "Optional configuration for the stream processor. Empty `options` objects are not supported.",
 				Attributes: map[string]schema.Attribute{
+					"resume_from_checkpoint": schema.BoolAttribute{
+						Optional: true,
+						MarkdownDescription: "Controls checkpoint behavior when the `$source` stage or a window stage of the `pipeline` changes. When `true`, the stream processor resumes from its last checkpoint. Set to `false` to discard the existing checkpoint, which is necessary for those changes because the API rejects them while resuming from an incompatible checkpoint." +
+							" Defaults to `true` when not set.\n\n**NOTE** This attribute is only applied to updates that change the `$source` stage, or a window's type, `interval`, `hopSize` or `allowedLateness`. It is ignored on create and on any other update, such as a `tier` change or an edit to a `$match` stage, so leaving it set does not discard the checkpoint again. It cannot be imported.",
```
oarbusi: "leaving it set does not discard the checkpoint again" any future `$source` change while `false` remains in config will discard the checkpoint without the user re-acknowledging it, right?

internal/service/streamprocessor/model.go
```diff
@@ -60,6 +62,93 @@ func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel)
...
+// checkpointIncompatibleChange reports whether the plan changes something the API refuses to modify
+// while resuming from a checkpoint: the $source stage, or a window's type, interval, hopSize or
+// allowedLateness, including adding or removing a window. Any other edit, for example to a $match or
+// $emit stage, can resume from the existing checkpoint, so resume_from_checkpoint is not sent for it
+// and a value left in the configuration cannot discard the checkpoint again.
+//
+// When the change cannot be determined it reports true, so that an explicitly configured
+// resume_from_checkpoint is still sent rather than silently dropped. This is defensive: the plan
+// pipeline is already valid JSON by the time it gets here, both from the jsontypes.Normalized
+// attribute type and from convertPipelineToSdk above.
```
AgustinBettati, line 88: just double checking, is there some of this validation logic that is worth delegating to the API (just return API validation on the tf side)

internal/service/streamprocessor/model.go
```diff
@@ -60,6 +62,93 @@ func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel)
...
+	if len(stages) == 0 {
+		return nil
+	}
+	return stages[0]
+}
+
+// windowStages projects the window stages down to the fields the API rejects a checkpoint resume
+// for, so that changes to a window's inner pipeline are not mistaken for incompatible ones. Window
+// operators are matched by name suffix to cover $tumblingWindow, $hoppingWindow and any later
+// addition.
```
AgustinBettati, line 128: similar, this logic seems quite complex and could age over time, if possible would delegate fully to API

AgustinBettati (approved):
LGTM

internal/service/streamprocessor/model.go
jandersongo: Renamed it to `shouldSendResumeFromCheckpoint`. the old name made it look like validation, which it isn't. The API accepts `resumeFromCheckpoint` on any modify and never rejects it; mms just passes it to SPM, which discards the checkpoint. So there's no API error to lean on: leaving it out doesn't get us a rejection, it gets us a destroyed checkpoint on unrelated updates like a tier bump or a $match edit.

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -771,9 +771,10 @@ func TestAccStreamRSStreamConnection_SchemaRegistry(t *testing.T) {
 	var (
 		projectID, instanceName = acc.ProjectIDExecutionWithStreamInstance(t)
 		connectionName          = acc.RandomName()
-		schemaRegistryURLs      = []string{"https://schemaregistry.example.com", "https://schemaregistry2.example.com"}
-		username                = "user"
-		password                = "password"
+		// The API rejects URLs whose host does not resolve, so these must be real public hostnames.
```
jandersongo: drive-by to fix failing test. failure is unrelated to this pr

.changelog/4608.txt
jandersongo: i dropped the "not returned by the API". 
`resumeFromCheckpoint` is accepted in the modify request body but isn't in any response type, and it isn't sensitive. That's exactly what forces the extra plumbing: the read can't reconstruct it, so it's carried over from plan or prior state and can't be imported. 
technically, i don't think there are API changes. The field has been in the modify request body since the endpoint was introduced in CLOUDP-264678 (Nov 2024). This PR only consumes the existing field, no SDK bump. 
i agree on building it into the process for Streams API changes going forward.

oarbusi (approved):
LGTM, thanks

jandersongo:
## Local verification against cloud-dev

Ran this end to end with a locally-built provider against a real Atlas project, on top of an Atlas-to-Atlas processor in `STARTED` state (`$source` with an `operationType` filter → downstream `$match` → `$merge`), so it holds an actual checkpoint.

**Results**

- **1. Create processor** → ✅ runs, `STARTED`
- **2. `$source` change, `resume_from_checkpoint` unset** → ✅ API rejects, HTTP 400
- **3. `$source` change, `resume_from_checkpoint = true`** → ✅ API rejects, HTTP 400
- **4. `$source` change, `resume_from_checkpoint = false`** → ✅ in-place success, `0 added, 1 changed, 0 destroyed`
- **5. Downstream `$match` change, `false` left in config** → ✅ flag not sent, `sending=false`
- **6. Remove `options`** → ✅ clears the one-off, back to null

Steps 2 and 3 both returned the planner's rejection, confirming the problem the PR fixes:

```
PATCH .../processor/tf-test-resume-checkpoint
HTTP 400 (BAD_REQUEST) Detail: resumeFromCheckpoint must be false to modify a stream processor's $source stage
```

**Step 4 is the fix.** The processor id was unchanged across the whole run (`6a984f8391e4b345d24340a4`), so the `$source` stage was modified in place — no destroy/recreate, no `terraform state rm`, which is the actual ask in STREAMS-1948 and the thing the customer's `replace_triggered_by` workaround was forced into. It took 3s, and the processor came back `STARTED` on its own via the existing stop → modify → start flow.

State carry-over also behaved: `resume_from_checkpoint = false` was present in state afterwards despite the API never returning it, with no "provider produced inconsistent result after apply".

**Step 5 is the gate**, and it's the one worth calling out. With `resume_from_checkpoint = false` still in config and only a downstream `$match` edited, the provider did not send the field:

```
[INFO] resume_from_checkpoint decision:
  configured=false  checkpoint_incompatible_change=false  sending=false
  tf_rpc=ApplyResourceChange
```

That's the scenario raised in #[streams-uxdx](https://mongodb.slack.com/archives/C08CTUS1U2C/p1787767797936599) — a stale `false` re-clearing the checkpoint on an unrelated pipeline edit. Before the narrowing to `$source`/window stages, that run would have sent `false`.

Steps 2, 3, 4 and 6 are also covered by `TestAccStreamProcessor_resumeFromCheckpoint` in CI. Step 5 is covered by the unit case `downstream match change does not send a stale flag`; the log above confirms the same decision at runtime.

PR #4617, opened 2026-08-05
chore: Refetches API spec and removes Streams failover schema workarounds

jandersongo:
https://jira.mongodb.org/browse/CLOUDP-425590

Refetches the OpenAPI spec to pick up upstream Streams spec fixes and removes the provider-side workarounds they make unnecessary.

Upstream changes:
- [CLOUDP-421992](https://jira.mongodb.org/browse/CLOUDP-421992): failover endpoints now use a dedicated Kafka/Cluster StreamsFailoverConnection schema instead of the shared StreamsConnection
- [CLOUDP-421890](https://jira.mongodb.org/browse/CLOUDP-421890): sslKey/sslKeyPassword now carry format: password
- [CLOUDP-422582](https://jira.mongodb.org/browse/CLOUDP-422582): StreamsConnection field description cleanup

Removes from the stream_connection_failover autogen config, in both the resource and datasource blocks:
- the nine type-specific ignores (aws, azure, gcp, headers, url, public_private_networking, provider, schema_registry_*), which the new schema never exposes
- the discriminator_types: [Kafka, Cluster] overrides, now native to the spec
- the manual authentication.ssl_key / ssl_key_password sensitive overrides, now derived from format: password

Generated output is unchanged by the config cleanup itself, verified by regenerating with the new spec before and after the edits.

Side effect: ssl_key and ssl_key_password are now sensitive in the failover data sources too. The old override was resource-only, so the data sources had been exposing them in plaintext.

### DIFF
--- a/.changelog/4617.txt
+++ b/.changelog/4617.txt
@@ -1,0 +1,7 @@
+```release-note:bug
+data-source/mongodbatlas_stream_connection_failover: Marks `authentication.ssl_key` and `authentication.ssl_key_password` as sensitive so they are no longer shown in plaintext
+```
+
+```release-note:bug
+data-source/mongodbatlas_stream_connection_failovers: Marks `results.authentication.ssl_key` and `results.authentication.ssl_key_password` as sensitive so they are no longer shown in plaintext
+```
--- a/.changelog/4617.txt
+++ b/.changelog/4617.txt
@@ -1,7 +1,7 @@
 ```release-note:bug
-data-source/mongodbatlas_stream_connection_failover: Marks `authentication.ssl_key` and `authentication.ssl_key_password` as sensitive so they are no longer shown in plaintext
+data-source/mongodbatlas_stream_connection_failover: Marks `authentication.ssl_key` and `authentication.ssl_key_password` as sensitive
 ```
 
 ```release-note:bug
-data-source/mongodbatlas_stream_connection_failovers: Marks `results.authentication.ssl_key` and `results.authentication.ssl_key_password` as sensitive so they are no longer shown in plaintext
+data-source/mongodbatlas_stream_connection_failovers: Marks `results.authentication.ssl_key` and `results.authentication.ssl_key_password` as sensitive
 ```
--- a/docs/data-sources/stream_connection_failover.md
+++ b/docs/data-sources/stream_connection_failover.md
@@ -34,9 +34,9 @@ output "failover_bootstrap_servers" {
 
 ### Read-Only
 
-- `region` (String) The connection region.
-- `state` (String) The connection state.
-- `type` (String) The connection type.
+- `region` (String) Connection region.
+- `state` (String) Connection state.
+- `type` (String) Connection type.
 
 <!-- polymorphic attributes restructured by docpostprocess -->
 The following attributes depend on the value of `type`:
@@ -45,13 +45,13 @@ The following attributes depend on the value of `type`:
 
 - `cluster_name` (String) Name of the cluster configured for this connection.
 - `cluster_project_id` (String) Unique 24-hexadecimal digit string that identifies the project that contains the configured cluster. Required if the ID does not match the project containing the streams workspace. You must first enable the organization setting.
-- `db_role_to_execute` (Attributes) The name of a built-in or custom DB Role to connect to an Atlas Cluster. (see [below for nested schema](#nestedatt--db_role_to_execute))
+- `db_role_to_execute` (Attributes) Name of a built-in or custom DB Role to connect to a MongoDB Cloud Cluster. (see [below for nested schema](#nestedatt--db_role_to_execute))
 
 #### `Kafka`
 
 - `authentication` (Attributes) User credentials required to connect to a Kafka Cluster. Includes the authentication type, as well as the parameters for that authentication mode. (see [below for nested schema](#nestedatt--authentication))
 - `bootstrap_servers` (String) Comma separated list of server addresses.
-- `config` (Map of String) A map of Kafka key-value pairs for optional configuration. This is a flat object, and keys can have '.' characters.
+- `config` (Map of String) Map of Kafka key-value pairs for optional configuration. This object is flat, and keys can have '.' characters.
 - `networking` (Attributes) Networking configuration for Streams connections. (see [below for nested schema](#nestedatt--networking))
 - `security` (Attributes) Properties for the secure transport connection to Kafka. For SSL, this can include the trusted certificate to use. (see [below for nested schema](#nestedatt--security))
 
--- a/docs/data-sources/stream_connection_failover.md
+++ b/docs/data-sources/stream_connection_failover.md
@@ -34,9 +34,9 @@ output "failover_bootstrap_servers" {
 
 ### Read-Only
 
-- `region` (String) The connection's region.
-- `state` (String) The state of the connection.
-- `type` (String) Type of the connection.
+- `region` (String) The connection region.
+- `state` (String) The connection state.
+- `type` (String) The connection type.
 
 <!-- polymorphic attributes restructured by docpostprocess -->
 The following attributes depend on the value of `type`:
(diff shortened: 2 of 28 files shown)


jandersongo:
augment review

docs/data-sources/stream_connection_failover.md
```diff
@@ -34,9 +34,9 @@ output "failover_bootstrap_servers" {
 
 ### Read-Only
 
-- `region` (String) The connection's region.
-- `state` (String) The state of the connection.
-- `type` (String) Type of the connection.
+- `region` (String) The connection region.
+- `state` (String) The connection state.
+- `type` (String) The connection type.
```
corryroot:
```suggestion
- `region` (String) Connection region.
- `state` (String) Connection state.
- `type` (String) Connection type.
```

docs/data-sources/stream_connection_failover.md
```diff
@@ -45,7 +45,7 @@ The following attributes depend on the value of `type`:
 
 - `cluster_name` (String) Name of the cluster configured for this connection.
 - `cluster_project_id` (String) Unique 24-hexadecimal digit string that identifies the project that contains the configured cluster. Required if the ID does not match the project containing the streams workspace. You must first enable the organization setting.
-- `db_role_to_execute` (Attributes) The name of a Built in or Custom DB Role to connect to an Atlas Cluster. (see [below for nested schema](#nestedatt--db_role_to_execute))
+- `db_role_to_execute` (Attributes) The name of a built-in or custom DB Role to connect to an Atlas Cluster. (see [below for nested schema](#nestedatt--db_role_to_execute))
```
corryroot:
```suggestion
- `db_role_to_execute` (Attributes) Name of a built-in or custom DB Role to connect to a MongoDB Cloud Cluster. (see [below for nested schema](#nestedatt--db_role_to_execute))
```

tools/codegen/atlasapispec/multi-version-api-spec.flattened.yml
```diff
@@ -23760,6 +23760,87 @@ components:
...
+        dbRoleToExecute:
+          $ref: "#/components/schemas/DBRoleToExecute"
+        authentication:
+          $ref: "#/components/schemas/StreamsKafkaAuthentication"
+        bootstrapServers:
+          description: Comma separated list of server addresses.
+          type: string
+        config:
+          additionalProperties:
+            description: A map of Kafka key-value pairs for optional configuration. This is a flat object, and keys can have '.' characters.
```
corryroot, line 23816:
```suggestion
            description: Map of Kafka key-value pairs for optional configuration. This object is a flat, and keys can have '.' characters.
```

corryroot (approved):
LGTM! I commented with some copy nits. For repeated descriptions, the comment applies to all cases.

docs/data-sources/stream_connection_failover.md
jandersongo: upstream changes will be addressed in this ticket https://jira.mongodb.org/browse/CLOUDP-431787
and this pr https://github.com/10gen/mms/pull/179993

.changelog/4617.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
+data-source/mongodbatlas_stream_connection_failover: Marks `authentication.ssl_key` and `authentication.ssl_key_password` as sensitive so they are no longer shown in plaintext
+```
```
AgustinBettati: Given these are sensitive value, I would expect they are never returned in the API response. Can you confirm this is the case? 
If so I believe we could remove these changelog entries as has no difference on users side (for data sources), those properties are never returned.

Also just to be super safe can you confirm these where marked correctly as sensitive in the resource?

tools/codegen/config.yml
```diff
@@ -928,28 +942,12 @@ resources:
...
-          # shared DTO covering every stream connection type).
-          "aws", # AWSKinesisDataStreams, AWSLambda, S3
-          "azure", # AzureBlobStorage
-          "gcp", # GCPPubSub
-          "headers", # Https
-          "url", # Https
-          "public_private_networking", # AzureBlobStorage, GCPPubSub
-          "provider", # SchemaRegistry (aliased to schema_registry_provider; ignore the source name)
-          "schema_registry_authentication", # SchemaRegistry
-          "schema_registry_urls", # SchemaRegistry
```
AgustinBettati, line 942: nice! Thanks for the follow up here

AgustinBettati (commented):
Would like to confirm the behaviour of the sensitive fields, other changes LGTM and thanks!

.changelog/4617.txt
jandersongo: - [`password`](https://github.com/10gen/mms/blob/a3176d4b1cd8644e8a44029d18c01262c53eff6a/server/src/main/com/xgen/cloud/streams/_public/model/view/ApiStreamsKafkaAuthenticationView.java#L53) and [`clientSecret`](https://github.com/10gen/mms/blob/a3176d4b1cd8644e8a44029d18c01262c53eff6a/server/src/main/com/xgen/cloud/streams/_public/model/view/ApiStreamsKafkaAuthenticationView.java#L87) are not returned bc they carry Jackson's `access = WRITE_ONLY`, 
 - [`sslKey`](https://github.com/10gen/mms/blob/a3176d4b1cd8644e8a44029d18c01262c53eff6a/server/src/main/com/xgen/cloud/streams/_public/model/view/ApiStreamsKafkaAuthenticationView.java#L67) and [`sslKeyPassword`](https://github.com/10gen/mms/blob/a3176d4b1cd8644e8a44029d18c01262c53eff6a/server/src/main/com/xgen/cloud/streams/_public/model/view/ApiStreamsKafkaAuthenticationView.java#L74) only have the `@Schema(accessMode = WRITE_ONLY)` annotation — that emits `writeOnly: true` into the spec, but without the Jackson annotation the field still serializes into responses. [CLOUDP-411827](https://jira.mongodb.org/browse/CLOUDP-411827) calls this out verbatim and is still In Progress.

So there is a real user-visible difference: before this PR, a failover data source reading an mTLS Kafka connection receives the real PEM private key and displays it in plaintext. The entries deliberately cover only `ssl_key`/`ssl_key_password` — ones for `password`/`client_secret` would have been no-ops, exactly as you suspected.

AgustinBettati: I understand this analysis is done by looking at the API Spec, but this might not be accurate with actual runtime behaviour of the API. Could you double check?

jandersongo: Good idea!
I verified against the API on cloud-dev and it looks like the fields are returned
Created a Kafka connection with mTLS auth, then listed connections in the workspace:
```
 curl --user "<public key>:<private key>" --digest \
  -H "Accept: application/vnd.atlas.2023-02-01+json" \
  "https://cloud-dev.mongodb.com/api/atlas/v2/groups/685d759376489629c36fcd18/streams/wps1/connections?itemsPerPage=500" \
| jq '.results[] | select(.type=="Kafka") | {name, mechanism: .authentication.mechanism, authKeys: (.authentication | keys)}'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   142  100   142    0     0    991      0 --:--:-- --:--:-- --:--:--   993
100  4945  100  4945    0     0  13973      0 --:--:-- --:--:-- --:--:-- 13973
{
  "name": "kafkaConn",
  "mechanism": "PLAIN",
  "authKeys": [
    "mechanism",
    "username"
  ]
}
{
  "name": "kafkaConn",
  "mechanism": "PLAIN",
  "authKeys": [
    "mechanism",
    "username"
  ]
}
{
  "name": "mtlsRedactionTest",
  "mechanism": "",
  "authKeys": [
    "mechanism",
    "sslCertificate",
    "sslKey",
    "sslKeyPassword"
  ]
}
```

.changelog/4617.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
+data-source/mongodbatlas_stream_connection_failover: Marks `authentication.ssl_key` and `authentication.ssl_key_password` as sensitive so they are no longer shown in plaintext
```
AgustinBettati:
```suggestion
data-source/mongodbatlas_stream_connection_failover: Marks `authentication.ssl_key` and `authentication.ssl_key_password` as sensitive
```
Would keep both changelog entries more concise, the reality is the true fix will come on the API side mitigating current state.

AgustinBettati (approved):

PR #4620, opened 2026-08-06
feat: Add Cluster Adaptive Settings resource and data source

lantoli:
## Description

Adds the `mongodbatlas_cluster_adaptive_settings` resource and data source for managing Adaptive Settings overrides on an existing Atlas cluster.

Uses the stable `2025-03-12` API with replacement PATCH semantics: the configured map is sent directly to Atlas. Removing a key restores its Atlas-managed default. Configuring `{}` or removing the attribute after it was set resets all overrides. Destroy resets overrides without deleting the cluster. A PostRead hook detects drift when overrides are reset outside Terraform.

The `online-changed` documentation-link check is expected to fail until the [Intelligent Workload Management documentation page](https://www.mongodb.com/docs/atlas/intelligent-workload-management/#load-shedding) is live. The link currently returns HTTP 404.

Link to any related issue(s): CLOUDP-431564

## Type of change:

## Required Checklist:


internal/serviceapi/clusteradaptivesettings/resource_custom_hooks.go
```diff
@@ -0,0 +1,66 @@
...
+	if !ok || bytes.Equal(bytes.TrimSpace(rawOverrides), []byte("null")) {
+		request[adaptiveSettingsOverridesAPIField] = json.RawMessage("null")
+		return marshalRequestOrOriginal(request, bodyReq)
+	}
+
+	var overrides map[string]json.RawMessage
+	if err := json.Unmarshal(rawOverrides, &overrides); err != nil {
+		return bodyReq
+	}
+	for _, key := range adaptiveSettingKeys {
```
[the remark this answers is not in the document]
lantoli: the provider intentionally supports the two keys listed in `adaptiveSettingKeys`; a new API setting needs a provider update to add reset semantics and coverage. I changed the docs from “initial supported keys” to “supported keys” so this release doesn't promise transparent support for future keys

tools/codegen/models/cluster_adaptive_settings.yaml
```diff
@@ -0,0 +1,134 @@
...
+                schema: jsontypes.NormalizedType{}
+              computed_optional_required: computed
+              tf_schema_name: adaptive_settings_overrides
+              tf_model_name: AdaptiveSettingsOverrides
+              api_name: adaptiveSettingsOverrides
+              req_body_usage: omit_always
+              sensitive: false
+              create_only: false
+              present_in_any_response: false
+              request_only_required_on_create: false
```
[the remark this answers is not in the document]
lantoli: `present_in_any_response` records whether a resource attribute was merged from a response and is used by the request-only-required-on-create transformation. Singular data-source attributes are built directly as computed, so generated data-source model entries keep this internal flag false by design; no change is needed

internal/serviceapi/clusteradaptivesettings/resource_custom_hooks.go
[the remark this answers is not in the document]
lantoli: this supersedes my earlier reply: the request hook now reads `effectiveAdaptiveSettings`, initializes every returned key to `null`, and overlays the planned overrides before PATCH, so future setting keys don't require provider changes

corryroot (approved):
LGTM!

internal/serviceapi/clusteradaptivesettings/resource_custom_hooks.go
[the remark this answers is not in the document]
lantoli: This is outdated. The resource now uses replacement PATCH semantics instead of JSON Merge Patch. The configured map is sent directly to Atlas, which replaces the entire override map. Omitted keys are removed by Atlas, not preserved. The custom hooks file has been removed.

tools/codegen/models/cluster_adaptive_settings.yaml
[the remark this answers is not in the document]
lantoli: Already addressed in a previous reply. `present_in_any_response` is intentionally `false` for the data source model because the data source builds its model differently from the resource.

internal/serviceapi/clusteradaptivesettings/resource.go
```diff
@@ -0,0 +1,160 @@
...
+		Method:        "PATCH",
+	}
+	reqHandle := autogen.HandleCreateReq{
+		Hooks:      r,
+		Resp:       resp,
+		Client:     r.Client,
+		Plan:       &plan,
+		CallParams: &callParams,
+	}
+	autogen.HandleCreate(ctx, reqHandle)
```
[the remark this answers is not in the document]
lantoli: This is intentional. When the user omits adaptive_settings_overrides, they are choosing not to manage overrides for this cluster. The resource should not reset existing overrides that the user did not configure. This matches the API contract where an omitted field is a no-op.

internal/serviceapi/clusteradaptivesettings/resource.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	reqHandle := autogen.HandleReadReq{
+		Hooks:      r,
+		RespDiags:  &resp.Diagnostics,
+		RespState:  &resp.State,
+		Client:     r.Client,
+		State:      &state,
+		CallParams: readAPICallParams(&state),
+	}
+	autogen.HandleRead(ctx, reqHandle)
```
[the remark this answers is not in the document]
lantoli: This is a known trade-off. We removed the PostRead hook that handled this edge case to keep the resource simple. If a user reports drift detection issues after external resets, we will revisit. The acceptance test covers the standard lifecycle.

internal/serviceapi/clusteradaptivesettings/resource_schema.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/customplanmodifier"
+)
+
+func ResourceSchema(ctx context.Context) schema.Schema {
+	return schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"adaptive_settings_overrides": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "JSON-encoded map of customer-specified Adaptive Settings overrides, applied on a best-effort basis. Use `jsonencode` to configure this attribute. Each supported entry that you specify takes precedence over the corresponding Atlas-managed default.",
+				CustomType:          jsontypes.NormalizedType{},
```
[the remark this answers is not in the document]
lantoli: We intentionally removed the custom validator. The jsontypes.Normalized type already validates that the value is valid JSON. For semantic validation (object shape, non-null entries), Atlas returns clear 400 errors at apply time. This matches the pattern used by other autogen resources with JSON fields (e.g., streamprocessorapi, searchindexapi).

tools/codegen/config.yml
```diff
@@ -145,6 +145,51 @@ resources:
...
+      aliases:
+        groupId: projectId
+      overrides:
+        project_id:
+          description: *project_id_description
+        adaptive_settings_overrides:
+          # OpenAPI models this field as an object, but Terraform exposes it as a normalized JSON string.
+          description: &adaptive_settings_overrides_description "JSON-encoded map of customer-specified Adaptive Settings overrides, applied on a best-effort basis. Use `jsonencode` to configure this attribute. Each supported entry that you specify takes precedence over the corresponding Atlas-managed default."
+          # Preserve explicit null during update so removing the attribute clears all overrides through replacement PATCH semantics.
+          request_body_usage: "send_null_as_null_on_update"
```
EspenAlbert, line 176: `send_null_as_null_on_update` only affects `Update`; `Create` still marshals an omitted `adaptive_settings_overrides` as `{}`. The checked-in API contract says that omitting `adaptiveSettingsOverrides` is a no-op, so creating this resource without the optional attribute leaves pre-existing cluster overrides in place rather than resetting them as the resource documentation and PR description promise. Send `adaptiveSettingsOverrides: null` on create when the Terraform value is null, and add coverage that starts with an existing override.

internal/serviceapi/clusteradaptivesettings/resource_test.go
```diff
@@ -0,0 +1,199 @@
...
+					},
+					PostApplyPreRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+				Check: checkResourceAndDataSource(
+					`{}`,
+				),
+			},
+			{
```
EspenAlbert, line 108: Can you confirm this is not happening? 🤖 assisted review concern

# Out-of-band Adaptive Settings reset is not detected

This concerns Terraform drift detection after another actor resets a cluster’s Adaptive Settings outside of Terraform.

## Expected flow

Suppose Terraform previously applied:

```hcl
adaptive_settings_overrides = jsonencode({
  LOAD_SHEDDING = true
})
```

Terraform state then contains:

```json
{
  "adaptive_settings_overrides": "{\"LOAD_SHEDDING\":true}"
}
```

Later, an Atlas user, API client, or another tool clears those overrides. Atlas can report that reset as either:

```json
{
  "adaptiveSettingsOverrides": null
}
```

or by omitting `adaptiveSettingsOverrides` entirely.

On the next `terraform plan` or `terraform refresh`, the provider must update state to show that the configured override is no longer present. Terraform can then plan an update to restore the requested `LOAD_SHEDDING = true` setting.

## Current behavior

The resource read implementation passes the current Terraform state into the generic read handler:

```go
State: &state,
...
autogen.HandleRead(ctx, reqHandle)
```

`HandleRead` deliberately unmarshals the API response into that same current-state model rather than a blank model.

The shared unmarshaler has these rules:

```go
if !ok {
    continue // property absent: retain current model field
}

if attrObjJSON == nil {
    continue // property null: retain current model field
}
```

Therefore, if the pre-refresh state is:

```json
{"LOAD_SHEDDING": true}
```

and Atlas responds with either an absent or `null` `adaptiveSettingsOverrides`, the model keeps the old value. Terraform still believes Atlas has the override enabled.

The next plan sees no difference between configuration and state, so it does not send the PATCH that would restore the override. The real Atlas configuration and Terraform state have drifted, but the provider hides that drift.

## Why the removed hook mattered

Earlier versions of this PR included a `PostReadAPICall` hook that reset the field before normal response decoding:

```go
req.State.(*TFModel).AdaptiveSettingsOverrides = jsontypes.NewNormalizedNull()
```

Its placement is important:

- If Atlas returns an actual overrides object, normal unmarshaling replaces the temporary null with that object.
- If Atlas omits the property or returns `null`, the shared unmarshaler skips that field, leaving the temporary null in state.

That gave the resource the desired behavior despite the generic unmarshaler’s behavior of preserving existing values for absent or null fields.

The refactor removed both that hook and the tests that covered the reset case.

## Fix shape

The provider should restore equivalent resource-specific behavior, alongside a regression test:

1. Apply a non-empty override through the resource.
2. Reset overrides outside Terraform, such as by PATCHing `{"adaptiveSettingsOverrides": null}`.
3. Refresh or plan with the original Terraform configuration.
4. Assert Terraform observes the reset and plans an update.
5. Apply and confirm the configured override is restored.

This is specifically a resource-state issue. A data source does not start from an existing managed state in the same way, so preserving an old value is primarily problematic for the resource refresh path.

internal/serviceapi/clusteradaptivesettings/resource_schema.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/common/customplanmodifier"
+)
+
+func ResourceSchema(ctx context.Context) schema.Schema {
+	return schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"adaptive_settings_overrides": schema.StringAttribute{
```
EspenAlbert, line 18: why not make the attribute required? Does this resource make sense without it?

.changelog/4620.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:new-resource
```
AgustinBettati: (related to PR description) `Omitting the attribute or configuring {} resets all overrides`

For the case of omitting the property entirely, is this causing a reset? I would expect the only request that case a reset of the property are when sending explicit null, or setting value to to `{}`.

tools/codegen/config.yml
```diff
@@ -145,6 +145,51 @@ resources:
...
+      path: /api/atlas/v2/groups/{groupId}/clusters/{clusterName}/adaptiveSettings
+      method: PATCH
+    update:
+      path: /api/atlas/v2/groups/{groupId}/clusters/{clusterName}/adaptiveSettings
+      method: PATCH
+    delete:
+      path: /api/atlas/v2/groups/{groupId}/clusters/{clusterName}/adaptiveSettings
+      method: PATCH
+      # Adaptive Settings belong to the cluster, so destroy clears overrides instead of deleting the API object.
+      static_request_body: '{"adaptiveSettingsOverrides": null}'
```
AgustinBettati, line 164: This has the risk of getting stale over time?
Has https://mongodb.github.io/ipa/113#resetting-singleton-resources been considered? 

I would consider we only add the resetting functionality if it is exposed through a dedicated `:reset` operation as per our standards.

docs/resources/cluster_adaptive_settings.md
```diff
@@ -0,0 +1,52 @@
+---
+subcategory: "Clusters"
+---
+
+# Resource: mongodbatlas_cluster_adaptive_settings
+
+`mongodbatlas_cluster_adaptive_settings` provides a resource for managing Adaptive Settings overrides for one MongoDB Atlas cluster.
+
+-> **NOTE:** Set `adaptive_settings_overrides` with `jsonencode`. Use `LOAD_SHEDDING` for Load Shedding and `SEARCH_LOAD_SHEDDING` for Search Load Shedding. Both settings accept Boolean values. Removing a key from the configured object resets that override. Removing the attribute or configuring an empty object resets all overrides. Destroying this resource also resets all overrides to Atlas-managed defaults and does not delete the cluster. Manage Adaptive Settings for a cluster with only one `mongodbatlas_cluster_adaptive_settings` resource. If multiple resources target the same cluster, each resource can overwrite the others, causing Terraform to repeatedly report and reapply changes.
```
AgustinBettati, line 9: q: Is there any external docs we can reference? These key values will evolve and docs will quickly become outdated here.

AgustinBettati, line 9: (related to PR description) `Omitting the attribute or configuring {} resets all overrides`

For the case of omitting the attribute during creation causing a full reset, should that be called out as well?

AgustinBettati (commented):
changes LGTM, just want to confirm and document the create behaviour asked by @EspenAlbert

tools/codegen/config.yml
EspenAlbert: still present at head `36ba6b9a9`: `Create` marshals the omitted attribute into an empty request body, and the pinned API contract treats that omission as a no-op, so pre-existing overrides survive create. Making the attribute required (OPEN 05) or sending `adaptiveSettingsOverrides: null` on create when the Terraform value is null would close this

internal/serviceapi/clusteradaptivesettings/resource_test.go
EspenAlbert: verified fixed at `36ba6b9a9`: the PostRead hook clears `adaptive_settings_overrides` before response decoding, the omitted and null response cases have unit coverage, and the acceptance test now resets overrides out of band and asserts a corrective update. I also checked the SDK helper: `AdaptiveSettingsUpdateRequest.NullFields` marshals `{"adaptiveSettingsOverrides":null}` (admin/model_adaptive_settings_update_request.go), and `go test ./internal/serviceapi/clusteradaptivesettings/...` passes locally

internal/serviceapi/clusteradaptivesettings/resource_schema.go
EspenAlbert: making it required would also resolve the create-time omission ambiguity in OPEN 03: the resource then always sends the full map, `{}` becomes an explicit reset, and the sole-manager note in the docs holds without a null-omission special case. Tradeoff to weigh: `auditing_api` keeps its `enabled` toggle optional (internal/serviceapi/auditingapi/resource_schema.go:23), so both patterns exist for singleton PATCH resources. My take: required fits better here because this attribute is the resource's entire payload

tools/codegen/config.yml
EspenAlbert: checked the pinned API spec: the cluster adaptiveSettings path exposes only GET and PATCH (tools/codegen/atlasapispec/raw-multi-version-api-spec.yml:50051), so a dedicated `:reset` operation is not available for this API today. Repo precedent goes both ways: `ai_model_rate_limit` deletes via the API's `rateLimits:reset` POST (tools/codegen/config.yml:111-113), while `auditing_api` resets with a static PATCH body plus `resets_to_defaults` (tools/codegen/config.yml:141-145). If IPA 113 prefers `:reset`, that looks like an Atlas API follow-up rather than something the provider can do on this endpoint

docs/resources/cluster_adaptive_settings.md
EspenAlbert: verified what exists today: the spec's own externalDocs pointer (`https://dochub.mongodb.org/core/adaptive-settings`) redirects to `https://www.mongodb.com/docs/core/adaptive-settings/` and returns 404, and the only live public reference I found is the API resources spec page (HTTP 200). Worth linking that page for the operation contract and adding the docs page once it publishes; the key list itself is `additionalProperties` in the spec, so no public page enumerates the supported keys yet

EspenAlbert: the deeper issue is that the current head does not reset on create-omission at all: `Create` sends an empty body and the API contract defines omission as a no-op, so the docs sentence is inaccurate for create today (see OPEN 03). Once OPEN 03 is fixed or the attribute becomes required (OPEN 05), state the create behavior explicitly in both the NOTE and the PR description

internal/serviceapi/clusteradaptivesettings/resource_test.go
lantoli: you're right, I initially had this covered but wasn't sure about the complexity vs frequency tradeoff. Restored in 36ba6b9a9, thanks for the detailed analysis

tools/codegen/config.yml
lantoli: yes, this is intentional: when the user omits adaptive_settings_overrides, they choose not to manage overrides for this cluster.

the field is optional because the API may add more attributes to this endpoint in the future, and users may want to configure only some of them without being forced to set overrides. Keeping it optional aligns with the Atlas SDK and allows the resource to grow without breaking existing configurations.

the resource should not reset overrides the user did not configure. This matches the API contract where an omitted field is a no-op. If the user wants a clean slate, they explicitly set {}. Happy to discuss further if you see it differently.

see also the new clarifications in the doc.

internal/serviceapi/clusteradaptivesettings/resource_schema.go
lantoli: same reasoning as in the comment above: the field is optional because the API may add more attributes to this endpoint in the future, and users may want to configure only some of them without being forced to set overrides

tools/codegen/config.yml
lantoli: the API doesn't expose a dedicated :reset operation: PATCH with {"adaptiveSettingsOverrides": null} is the reset mechanism. The delete path is a single static request body with no dynamic content, so I think staleness risk is minimal.

an alternative would be a no-op on delete, leaving the overrides configured in Atlas. That would avoid the reset entirely, but it means destroying the Terraform resource doesn't clean up what it managed, the overrides would persist even though the user asked to remove them.

the current approach matches how other singleton resources in this provider handle destroy (e.g., project settings, maintenance window), but happy to discuss if you see it differently

docs/resources/cluster_adaptive_settings.md
EspenAlbert: the new wording still directs users to Atlas API documentation without a link, and the available API specification does not enumerate supported keys. Either link a published page that lists the keys, or remove this sentence until that documentation exists

EspenAlbert: the resource documentation now correctly distinguishes create omission from update removal, but the PR description still says omission resets all overrides. Update the PR description to match the shipped behavior

lantoli: sorry, the PR description was wrong: omitting the attribute on create is a no-op per the API contract, it does not reset overrides. Only configuring {} or removing the attribute after it was set resets all overrides.

fixed the description and clarified in the docs in c0fc1a634.

lantoli: good point, updated in c0fc1a634 to reference the Atlas documentation instead of enumerating keys inline. 

I'll follow up with the specific link once I have it,  so this PR is not blocked

EspenAlbert (approved):
Thank you for addressing the comments

tools/codegen/config.yml
AgustinBettati: > the current approach matches how other singleton resources in this provider handle destroy (e.g., project settings, maintenance window)

Those are internal resources, I would look into prod resources. `ai_model_rate_limit` and `org_maintenance_settings` use `:reset`, `org_delegation_settings` does a no-op on delete.

> The delete path is a single static request body with no dynamic content

This is assuming no new root properties will be added in the resource (might be the case for this resource). Not blocking but as a standard singleton resources I would push for resetting logic to be exposed by the API and not hardcoded client side.

AgustinBettati (approved):
LGTM

tools/codegen/config.yml
lantoli: thanks, will check with Atlas team if a :reset custom method would be possible

docs/resources/cluster_adaptive_settings.md
xargom: PR [#23477](https://github.com/10gen/docs-mongodb-internal/pull/23477) is creating the Atlas doc. The link will be: https://www.mongodb.com/docs/atlas/intelligent-workload-management/#load-shedding.

lantoli: thanks, adding it in this PR

PR #4623, opened 2026-08-06
chore: Skips TestAccProjectIPAccessList_settingMultiple when using PAK credentials

AgustinBettati:
## Description

Skips `TestAccProjectIPAccessList_settingMultiple` when the acceptance suite runs with programmatic API key credentials.

The test manages 20 access list entries and deletes them concurrently. Under digest auth those concurrent deletes intermittently return `401 Unauthorized` with a replay rejection, even though each nonce is sent exactly once.

This is not a non-prod artifact. PAK token exchange is off in all environments, so prod PAK traffic already authenticates with digest and is reachable by the same failure. The server behaviour did not change. CLOUDP-428083 disabled PAK token exchange in non-prod, moving non-prod PAK off bearer tokens and onto digest, which aligned non-prod with prod and exposed the failure to our nightly test runs. That is why the PAK nightly has failed on this test since 2026-07-29.

The upstream fix is tracked in CLOUDP-432227, owned by the IAM Workload Identity team. This PR only stops a known server-side failure from red-lining the PAK nightly. Remove the skip once that ticket is resolved. Service Account and access token runs are unaffected and continue to cover the test.

## Type of change:

## Required Checklist:

## Further comments

Test-only change, no changelog entry needed. Verified locally: with PAK credentials the test skips before any API call; with Service Account credentials it runs as before.

### DIFF
--- a/internal/service/projectipaccesslist/resource_project_ip_access_list_test.go
+++ b/internal/service/projectipaccesslist/resource_project_ip_access_list_test.go
@@ -113,6 +113,8 @@ func TestAccProjectIPAccessList_settingAWSSecurityGroup(t *testing.T) {
 }
 
 func TestAccProjectIPAccessList_settingMultiple(t *testing.T) {
+	// Concurrent access list deletes get spurious 401 replay rejections under digest auth, related ticket: CLOUDP-432227
+	acc.SkipInPAK(t, "concurrent deletes fail with 401 under digest auth, see CLOUDP-432227")
 	var (
 		projectID        = acc.ProjectIDExecution(t)
 		resourceFmt      = "mongodbatlas_project_ip_access_list.test_%d"


marcabreracast (approved):
LGTM!

manupedrozo:
We can consider putting the delete calls behind the MutexKV instead. 
Since this is not affecting prod, ok with skipping while upstream works on a fix.
See: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/ed218feab961eef26ffa03f11e6d153fc137a6ed/internal/service/projectipaccesslist/resource_project_ip_access_list.go#L32

manupedrozo (approved):

lantoli (approved):

AgustinBettati:
> We can consider putting the delete calls behind the MutexKV instead.
Since this is not affecting prod, ok with skipping while upstream works on a fix.

@manupedrozo Would note this issue does impact prod, and has for some time. It started impacting non-prod because non-prod aligned with how prod works (CLOUDP-428083).

Reason why I did not go for a workaround TF side:

- The error being surfaced (intermittent `401 Unauthorized`) only happens when requests are authorized with PAKs, with SA authorization concurrent request are handled with no issues. This implies the workaround would only be relevant for PAKs use case.
- No customer issues reported. If they do appear, it would be a good forcing function for prioritizing CLOUDP-432227.

Continuing with merge, with this context let me know if you prefer we discuss further potential changes TF side.

manupedrozo:
> @manupedrozo Would note this issue does impact prod, and has for some time. It started impacting non-prod because non-prod aligned with how prod works ([CLOUDP-428083](https://jira.mongodb.org/browse/CLOUDP-428083)).
> 
> Reason why I did not go for a workaround TF side:
> 
>     * The error being surfaced (intermittent `401 Unauthorized`) only happens when requests are authorized with PAKs, with SA authorization concurrent request are handled with no issues. This implies the workaround would only be relevant for PAKs use case.
> 
>     * No customer issues reported. If they do appear, it would be a good forcing function for prioritizing [CLOUDP-432227](https://jira.mongodb.org/browse/CLOUDP-432227).
> 
> 
> Continuing with merge, with this context let me know if you prefer we discuss further potential changes TF side.

In that case I think we should (in addition to the upstream improvement). I believe SAs can also fail in this scenario, but the error is different and retries handle it. If we serialize the calls, we'd avoid retries altogether. Let's discuss offline.

PR #4631, opened 2026-08-10
fix: Preserves stream processor pipeline field order (CLOUDP-433594)

jwongmongodb:
## Problem

`mongodbatlas_stream_processor` sends the aggregation pipeline to Atlas with the keys of every stage subdocument alphabetically reordered, regardless of how the pipeline is authored. Terraform keeps the authored string in state and reports a clean plan, so nothing signals that the pipeline Atlas actually stores and runs differs from the one that was written.

MongoDB documents are ordered, so this silently changes behaviour. Sort specifications lose their precedence (window-stage inner pipelines, `$top`/`$bottom` `sortBy`), equality comparisons against document literals (`$match`, `$expr $eq`, `$in`) match nothing or the wrong documents, and the field order of a literal subdocument in `$addFields`/`$project` propagates into the documents the processor writes, breaking order-sensitive matching downstream. Pipelines with no order-sensitive literal construct — most of them — are unaffected; the severity here is the silence.

This affects every provider version shipping the resource. It is the client-side counterpart of CLOUDP-421031 (Atlas alphabetizing pipelines in *responses*); that server-side fix does not help on its own, because order is destroyed before the request leaves the machine.

## Cause

`convertPipelineToSdk` (`internal/service/streamprocessor/model.go`) unmarshals the pipeline string into `[]any`, which binds each stage to `map[string]interface{}`. Go's `encoding/json` sorts map keys on marshal, so the outgoing request body is alphabetized.

## What changed

- **`model.go`** — decode stages as `[]json.RawMessage` and box them into the existing `[]any`. `json.Marshal` emits `json.RawMessage` verbatim even inside `[]any`, so no SDK change is needed and the function signature is unchanged. The slice has two consumers (create and update request builders) and is only assigned to `Pipeline`, never inspected.
- **`resource_schema.go`, `templates/`, `examples/`, `docs/`** — these are **required, not optional polish**. The schema description and all three examples currently recommend `jsonencode`, which serializes object keys in lexicographic order during HCL evaluation — i.e. order is already gone before the provider sees the value, so no code change can recover it. The examples are converted to heredocs, and the docs gain a single "Pipeline field ordering" section explaining that field order matters and that the pipeline should be authored as a raw JSON string: a heredoc, `file("pipeline.json")`, or `templatefile("pipeline.json", { ... })` when values need interpolating. The section also warns against `jsonencode(jsondecode(...))`, a natural way to lift a pipeline out of a larger JSON document, which sorts keys for exactly the same reason `jsonencode` does.
- **`.changelog/4631.txt`** — `release-note:bug` entry for the resource.

## Verification

- `TestPipelineFieldOrderPreserved` covers both the create and update request builders, marshalling the **complete** request struct and asserting **byte** equality on the body, so the assertion also exercises the SDK structs' own `MarshalJSON` (`assert.JSONEq` is order-blind and would pass against an alphabetized pipeline). Confirmed to fail when the `model.go` change is reverted.
- Live cloud-dev: the same `pipeline.json` supplied via `file()` was sent to the same endpoint and workspace with only the binary differing. Registry **v2.15.0** put the alphabetically-first key first on the wire; **this build** preserved the authored order. Create returned HTTP 200 and `mongosh` confirmed the stored pipeline carries the authored order.
- `make build`, `make test`, `golangci-lint run`, `gofmt -s -l .`, `make tflint`, and `terraform fmt -check` all pass, and the generated docs are in sync with `make generate-doc`.

## Known limitation (Terraform core, not fixable here)

`terraform plan` still *displays* JSON-string attributes with keys alphabetized, and renders them as `jsonencode(...)` even when the config used `file()` or a heredoc. Terraform's structured plan renderer decodes JSON-string attributes into an unordered map before rendering. The value in state and on the wire is correct. Reproducible with no providers at all:

```hcl
resource "terraform_data" "json_string" {
  input = "{\"region\": 1, \"city\": 1}"
}

# terraform plan renders:
#   + input = jsonencode({ + city   = 1
#                          + region = 1 })
```

Expect this to be reported as "the fix did not work". The docs note added here directs users to verify with `TF_LOG=DEBUG` or `sp.listStreamProcessors()`.

## Scope

In scope: the request path — what Atlas stores and executes. Out of scope: `jsonencode`-authored pipelines, which are unfixable in code, since HCL evaluation sorts the keys before the provider is called. Response-side ordering is handled separately by CLOUDP-421031, which has since shipped; typing responses as `json.RawMessage` in `mongodb/atlas-sdk-go` would be a breaking codegen change and is unnecessary here, as `jsontypes.Normalized` sources state from config.

Ticket: CLOUDP-433594


docs/resources/stream_processor.md
```diff
@@ -131,7 +143,11 @@ output "stream_processors_results" {
...
+- `pipeline` (String) Stream aggregation pipeline you want to apply to your streaming data, as a JSON string. [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/atlas-stream-processing/stream-aggregation/#std-label-stream-aggregation) contain more information. For more details see the [Aggregation Pipelines Documentation](https://www.mongodb.com/docs/atlas/atlas-stream-processing/stream-aggregation/).
+
+**Field order matters.** MongoDB documents are ordered, and some pipeline constructs depend on the order of keys within a document: sort specifications, where key order is sort precedence, and equality comparisons against a document literal, which match by exact field order. The order of fields in a `$addFields`/`$project` specification also becomes the field order of the documents the processor writes.
+
+Do not build this attribute with [jsonencode](https://developer.hashicorp.com/terraform/language/functions/jsonencode): it serializes object keys in lexicographic order, so the pipeline Atlas receives will not match what you wrote. Author the pipeline as a raw JSON string instead, using a heredoc or `file("pipeline.json")`, which Terraform passes through unchanged.
```
maastha: I think we can have both these points in one sub-bullet/"NOTE" callout

docs/resources/stream_processor.md
```diff
@@ -11,6 +11,10 @@ subcategory: "Streams"
 2. The update will be performed while the processor is in `STOPPED` state
 3. If the processor was originally in `STARTED` state, it will be restarted after the update
 
+-> **NOTE:** `terraform plan` displays the `pipeline` attribute with its keys in alphabetical order, and renders it as `jsonencode(...)` even when the configuration uses a heredoc or `file()`. This is how Terraform renders any JSON-string attribute and does not reflect what is sent to Atlas: the pipeline is transmitted and stored in the order you authored it. To confirm the order that was applied, inspect the request body with `TF_LOG=DEBUG`, or run `sp.listStreamProcessors()` against the workspace.
+
+~> **IMPORTANT:** Do not use `jsonencode()` to build the `pipeline` attribute. `jsonencode()` emits object keys in lexicographic order, so a pipeline written as `{"orderType": ..., "orderSubtype": ...}` is sent to Atlas as `{"orderSubtype": ..., "orderType": ...}`. MongoDB documents are ordered, and sort specifications and equality comparisons against document literals are order-sensitive, so this can silently change what your processor does. Author `pipeline` as a raw JSON string (heredoc or `file()`) instead.
```
maastha: same as below, we should combine these two ordering points under one short "Pipeline field ordering" section or one consolidated warning and remove the repetition below.

internal/service/streamprocessor/model_test.go
```diff
@@ -556,3 +556,30 @@ func TestGetWorkspaceOrInstanceName(t *testing.T) {
...
+	model := &streamprocessor.TFStreamProcessorRSModel{
+		ProjectID:     types.StringValue(projectID),
+		WorkspaceName: types.StringValue(workspaceName),
+		ProcessorName: types.StringValue(processorName),
+		Pipeline:      jsontypes.NewNormalizedValue(pipeline),
+	}
+
+	createReq, diags := streamprocessor.NewStreamProcessorReq(t.Context(), model)
+	require.False(t, diags.HasError(), "unexpected diagnostics: %v", diags)
+	createBody, err := json.Marshal(createReq.GetPipeline())
```
maastha: Should we marshal the complete create/update request—rather than only  GetPipeline() ? That would verify the ordering survives the final SDK serialization. Non-blocking.

maastha (approved):

jromo-mdb (approved):
LGTM, pending to address Copilot's suggestions.

docs/data-sources/stream_processor.md
```diff
@@ -135,7 +143,11 @@ output "stream_processors_results" {
...
 - `id` (String) Unique 24-hexadecimal character string that identifies the stream processor.
 - `options` (Attributes) Optional configuration for the stream processor. (see [below for nested schema](#nestedatt--options))
-- `pipeline` (String) Stream aggregation pipeline you want to apply to your streaming data. [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/atlas-stream-processing/stream-aggregation/#std-label-stream-aggregation) contain more information. Using [jsonencode](https://developer.hashicorp.com/terraform/language/functions/jsonencode) is recommended when setting this attribute. For more details see the [Aggregation Pipelines Documentation](https://www.mongodb.com/docs/atlas/atlas-stream-processing/stream-aggregation/)
+- `pipeline` (String) Stream aggregation pipeline you want to apply to your streaming data, as a JSON string. [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/atlas-stream-processing/stream-aggregation/#std-label-stream-aggregation) contain more information. For more details see the [Aggregation Pipelines Documentation](https://www.mongodb.com/docs/atlas/atlas-stream-processing/stream-aggregation/).
```
jwongmongodb: @maastha after sleeping on it, I think we can remove the datasource doc for this pipeline since datasources are read only right? putting jsonencode() here and the config specific guide is not relevant. Let me know if you agree, if so, I believe I need to create an override to autogenerate these docs separately from the resource.

I can also just keep this aligned with the resource docs for now and add a commit to change if necessary.

docs/resources/stream_processor.md
jwongmongodb: Done. Both callouts are now a single "Pipeline field ordering" section, and the pipeline schema description is one sentence instead of three paragraphs.

One thing worth flagging: I kept a short version of the warning in the schema description rather than deleting it. The data source templates only render SchemaMarkdown, so removing it would leave both data source pages with no ordering guidance at all.

While in there I also added templatefile as an authoring option and a warning about jsonencode(jsondecode(...)), which sorts keys the same way jsonencode does. That pattern came up in a real config.

internal/service/streamprocessor/model_test.go
jwongmongodb: Good call, done. It now marshals the whole create and update request and asserts byte equality on the body.

It covers more than I expected. The SDK request structs have their own MarshalJSON, which the GetPipeline() version was skipping entirely.

docs/data-sources/stream_processor.md
```diff
@@ -59,34 +59,42 @@ resource "mongodbatlas_stream_processor" "stream-processor-sample-example" {
...
-  ])
-  state = "STARTED"
-  tier  = "SP30"
+  # Authored as a raw JSON string so the field order is preserved exactly. Do not use jsonencode,
+  # which emits object keys in lexicographic order.
+  pipeline = <<-JSON
+    [
+      {"$source": {"connectionName": "${resource.mongodbatlas_stream_connection.example-sample.connection_name}"}},
+      {"$emit": {"connectionName": "${resource.mongodbatlas_stream_connection.example-cluster.connection_name}", "db": "sample", "coll": "solar", "timeseries": {"timeField": "_ts"}}}
+    ]
```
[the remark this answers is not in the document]
jwongmongodb: Applied, thanks. Dropped the resource. prefix here and in the other examples on
this page.

For the record it was not actually broken. terraform validate passes either way,
and the prefix was already on master before this PR. But bare references match
what the rest of the repo does, so this is the better form.

docs/data-sources/stream_processor.md
maastha: since there's a separate data source template for this I'd say yes we can leave it out

maastha (approved):
LGTM! Thanks for this one!

PR #4649, opened 2026-08-17
feat: Add mcp config org resources

laylaazouz:
## Description

Adds the `mongodbatlas_mcp_config` resource and its singular/plural data sources (`mongodbatlas_mcp_config`, `mongodbatlas_mcp_configs`). 

Generated via the existing autogen tooling, with two schema overrides on top of the standard CRUDL mapping:

- `ip_access_list`'s `ip_address`/`cidr_block` fields are modeled as optional+computed, since the API always echoes back both fields regardless of which one was submitted.

- `roles` is overridden from a list to a Set, since the API spec doesn't declare `uniqueItems` and role order isn't meaningful.

Includes acceptance tests, a working example, and generated docs.

Known CI failure: the resource doc's cross-link to `mongodbatlas_mcp_config_secret` points at `mcp_config_secret.md`. This will resolve once both PRs are merged into the shared dev branch.

Link to any related issue(s):

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [x] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [ ] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run make fix and verified my code
- [x] If changes include deprecations or removals I have added appropriate changelog entries.
- [x] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

PR #4650, opened 2026-08-17
feat: Add mcp config project resources

laylaazouz:
## Description

**_This PR mirrors #4649 (mongodbatlas_mcp_config), just scoped to the project-level path._** 

Adds the mongodbatlas_project_mcp_config resource and its singular/plural data sources (mongodbatlas_project_mcp_config, mongodbatlas_project_mcp_configs).

Generated via the existing autogen tooling, with two schema overrides on top of the standard CRUDL mapping:

ip_access_list's ip_address/cidr_block fields are modeled as optional+computed, since the API always echoes back both fields regardless of which one was submitted.
roles is overridden from a list to a Set, since the API spec doesn't declare uniqueItems and role order isn't meaningful.
Includes acceptance tests, a working example, and generated docs.

Known CI failure: the resource doc's cross-link to mongodbatlas_project_mcp_config_secret points at project_mcp_config_secret.md. This will resolve once both PRs are merged into the shared dev branch.

Link to any related issue(s):

## Type of change:

## Required Checklist:


QuiqueSanMongo (approved):
LGTM

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,155 @@
...
+		if projectID == "" || mcpConfigID == "" {
+			return fmt.Errorf("checkDestroy, attributes not found for: %s", resourceName)
+		}
+
+		_, _, err := acc.ConnPreview().RemoteMCPConfigurationsAPI.GetGroupMcpConfig(context.Background(), projectID, mcpConfigID).Execute()
+		if err == nil {
+			return fmt.Errorf("mcp config (%s/%s) still exists", projectID, mcpConfigID)
+		}
+	}
+	return nil
```
manupedrozo: We should delete the org MCP config here to avoid leaving orphan resources. 
We usually do not do this but for these resources (MCP project config and project service accounts), we do it to get around the API limitation and complete the test cleanup.
See: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4663

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,155 @@
...
+			{
+				Config: configBasic(projectID, name2, []string{"GROUP_OWNER"}, `ip_address = "203.0.113.10"`),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "ip_access_list.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "ip_access_list.0.ip_address", "203.0.113.10"),
+				),
+			},
+			{
+				Config: configBasic(projectID, name2, []string{"GROUP_OWNER"}, `cidr_block = "203.0.113.0/24"`),
```
EspenAlbert: q: Does it make sense to test more than 1 role?

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,155 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "ip_access_list.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "ip_access_list.0.cidr_block", "203.0.113.0/24"),
```
EspenAlbert: can ip_address and cidr_block be used together? should we test that?

tools/codegen/config.yml
```diff
@@ -1319,6 +1319,57 @@ resources:
...
+      schema:
+        aliases:
+          groupId: projectId
+        overrides:
+          project_id:
+            description: *project_id_description
+    version_header: application/vnd.atlas.2025-03-12+json
+    schema:
+      aliases:
+        groupId: projectId # Align with project_id naming used by other project-scoped resources (e.g. project_service_account).
```
manupedrozo: Nice, we do this for all resources since we don't use the term `group` in the TF provider. Ok to skip the comment.
```suggestion
        groupId: projectId
```

examples/mongodbatlas_project_mcp_config/README.md
```diff
@@ -0,0 +1,19 @@
+# MongoDB Atlas Provider -- Project MCP Config
+
+This example shows how to create a project-level Remote MCP configuration in MongoDB Atlas.
+
+## Important Notes
+
+This resource does not manage ingress secrets. Use `mongodbatlas_project_mcp_config_secret` to create one.
```
EspenAlbert: should we help users with a link here?

internal/serviceapi/projectmcpconfig/data_source_schema.go
```diff
@@ -0,0 +1,96 @@
...
+				Computed:            true,
+				MarkdownDescription: "Human-readable name that identifies this MCP configuration.",
+			},
+			"project_id": dsschema.StringAttribute{
+				Required:            true,
+				MarkdownDescription: "Unique 24-hexadecimal digit string that identifies your project, also known as `groupId` in the official documentation.",
+			},
+			"roles": dsschema.ListAttribute{
+				Computed:            true,
+				MarkdownDescription: "List of project roles associated with this MCP configuration.",
```
EspenAlbert, line 72: [nit] should we name this `Set` since we are using that?

tools/codegen/config.yml
```diff
@@ -1319,6 +1319,57 @@ resources:
...
+    create:
+      path: /api/atlas/v2/groups/{groupId}/mcpConfigs
+      method: POST
+    update:
+      path: /api/atlas/v2/groups/{groupId}/mcpConfigs/{mcpConfigId}
+      method: PATCH
+    delete:
+      path: /api/atlas/v2/groups/{groupId}/mcpConfigs/{mcpConfigId}
+      method: DELETE
+    datasources:
```
manupedrozo: Can you move the `datasources` key after the resource `schema` key? Just for consistency in the order.

templates/resources/project_mcp_config.md.tmpl
```diff
@@ -0,0 +1,23 @@
+---
+subcategory: "Remote MCP"
+---
+
+# {{.Type}}: {{.Name}}
+
+`{{.Name}}` provides a Project Remote MCP Configuration resource. The resource lets you create, update, delete, and import a Remote MCP configuration for the specified Project.
+
+~> **IMPORTANT:** This resource does not manage ingress secrets. Use [`mongodbatlas_project_mcp_config_secret`](project_mcp_config_secret.md) to create one.
```
manupedrozo, line 9: Should we also mention the `delete` limitation which only removed the MCP config from the project but keeps it at the org level? 
See: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/project_service_account

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,155 @@
...
+			{
+				Config: configBasic(projectID, name1, []string{"GROUP_READ_ONLY"}, ""),
+				Check:  checkBasic(),
+			},
+			{
+				Config: configBasic(projectID, name2, []string{"GROUP_OWNER"}, ""),
+				Check:  checkBasic(),
+			},
+			{
+				Config: configBasic(projectID, name2, []string{"GROUP_OWNER"}, `ip_address = "203.0.113.10"`),
```
manupedrozo: Can we apply the same change as in the org mcp config tests? Using a struct for passing in ip/cidr to both config and check.

internal/serviceapi/projectmcpconfig/resource_test.go
laylaazouz: this is rejected by the API. following our sync discussion, I will address this for both resources in a follow up pr since org is already merged

internal/serviceapi/projectmcpconfig/resource_test.go
laylaazouz: added a test for this!

examples/mongodbatlas_project_mcp_config/README.md
laylaazouz: yep, added a link, thanks!

tools/codegen/config.yml
laylaazouz: makes sense, removed comment

templates/resources/project_mcp_config.md.tmpl
laylaazouz: yep, added a note to both the resource template and the example README

internal/serviceapi/projectmcpconfig/resource_test.go
laylaazouz: refactored - same as the org branch

docs/resources/project_mcp_config.md
```diff
@@ -0,0 +1,83 @@
...
+subcategory: "Remote MCP"
+---
+
+# Resource: mongodbatlas_project_mcp_config
+
+`mongodbatlas_project_mcp_config` provides a Project Remote MCP Configuration resource. The resource lets you create, update, delete, and import a Remote MCP configuration for the specified Project.
+
+~> **IMPORTANT:** This resource does not manage ingress secrets. Use [`mongodbatlas_project_mcp_config_secret`](project_mcp_config_secret.md) to create one.
+
+-> **NOTE:** Deleting this resource only unlinks the MCP configuration from the project. The underlying MCP configuration is not deleted. To fully delete it, use the org-level [`mongodbatlas_mcp_config`](mcp_config.md) resource.
```
manupedrozo: Sorry for the churn. Lets try to keep consistency on the wording and guidance here with the project SA:
`Deleting a mongodbatlas_project_service_account resource unassigns the associated Service Account from the project, but doesn't delete it from the organization.`

examples/mongodbatlas_project_mcp_config/README.md
```diff
@@ -0,0 +1,21 @@
+# MongoDB Atlas Provider -- Project MCP Config
+
+This example shows how to create a project-level Remote MCP configuration in MongoDB Atlas.
+
+## Important Notes
+
+This resource does not manage ingress secrets. Use [`mongodbatlas_project_mcp_config_secret`](../mongodbatlas_project_mcp_config_secret/README.md) to create one.
+
+This resource only unlinks the MCP configuration from the project. the underlying MCP configuration is not deleted. To fully delete it, use the org-level [`mongodbatlas_mcp_config`](../mongodbatlas_mcp_config/README.md) resource.
```
manupedrozo: See comment on the docs above.

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,190 @@
...
+			{
+				Config: configBasic(projectID, name2, []string{"GROUP_OWNER"}, ipAccessListEntry{}),
+				Check:  checkBasic(ipAccessListEntry{}),
+			},
+			{
+				Config: configBasic(projectID, name2, []string{"GROUP_OWNER", "GROUP_READ_ONLY"}, ipAccessListEntry{}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					checkBasic(ipAccessListEntry{}),
+					resource.TestCheckResourceAttr(resourceName, "roles.#", "2"),
+				),
```
manupedrozo: Can we do this in `checkBasic` instead?

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,190 @@
...
+		return fmt.Sprintf("ip_address = %q", e.ip)
+	}
+	return ""
+}
+
+func (e ipAccessListEntry) attrMap() map[string]string {
+	if e.cidr == "" && e.ip == "" {
+		return nil
+	}
+	result := map[string]string{"ip_access_list.#": "1"}
```
manupedrozo: I see we only test 1 entry. Do you see value in testing multiple?
I am thinking that it would help verify that the API does not re-order entries, which would cause drifts.

internal/serviceapi/projectmcpconfig/resource_test.go
```diff
@@ -0,0 +1,190 @@
...
+			break
+		}
+
+		if _, _, err := acc.ConnPreview().RemoteMCPConfigurationsAPI.GetGroupMcpConfig(context.Background(), projectID, mcpConfigID).Execute(); err == nil {
+			errs = append(errs, fmt.Errorf("mcp config (%s/%s) still exists", projectID, mcpConfigID))
+			break
+		}
+		if _, err := acc.ConnPreview().RemoteMCPConfigurationsAPI.DeleteOrgMcpConfig(context.Background(), orgID, mcpConfigID).Cascading(true).Execute(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to delete org mcp config %s: %w", mcpConfigID, err))
+		}
```
manupedrozo, line 178: Can we delete the `org` level one even if the group one still exists? (currenlty that does not happen given the `break` in line 168). 
We know we'll need to do this for the project mcp secret as well, so perhaps its worth moving this logic to a helper like for SAs.

One other detail: `rs.Type != "mongodbatlas_project_mcp_config"` also matches the singular data source, so I am pretty sure we are making this delete call for that one as well, weird that you are not seeing errors. I may be missing something.

docs/resources/project_mcp_config.md
laylaazouz: updated to match!

internal/serviceapi/projectmcpconfig/resource_test.go
laylaazouz: added a step with two entries and refactored so it verifies the API doesnt reorder the two

AgustinBettati (approved):
LGTM, good to align checkDestroy with https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4663

internal/serviceapi/projectmcpconfig/resource_test.go
laylaazouz: fixed the break so the org-level delete always runs now 

also tested the data source point and I didn't get an error, but still switched to using state address prefix since thats more correct

 moved all into a shared helper

manupedrozo: Thanks! We can keep the `break` given that we delete the Org config outside the for loop now.

examples/mongodbatlas_project_mcp_config/README.md
```diff
@@ -0,0 +1,21 @@
+# MongoDB Atlas Provider -- Project MCP Config
+
+This example shows how to create a project-level Remote MCP configuration in MongoDB Atlas.
+
+## Important Notes
+
+This resource does not manage ingress secrets. Use [`mongodbatlas_project_mcp_config_secret`](../mongodbatlas_project_mcp_config_secret/README.md) to create one.
+
+Deleting this resource unlinks the MCP configuration from the project, but doesn't delete it from the organization. To fully delete it, use the org-level [`mongodbatlas_mcp_config`](../mongodbatlas_mcp_config/README.md) resource.
```
manupedrozo: Good if we could follow the same wording as in the registry docs for this one.

manupedrozo (approved):

manupedrozo:
Lets make sure to have tests enabled in the CI before merging.

tools/codegen/config.yml
```diff
@@ -794,6 +783,19 @@ resources:
...
+        method: GET
+      list:
+        path: /api/atlas/v2/orgs/{orgId}/mcpConfigs
+        method: GET
+      schema:
+        overrides:
+          org_id:
+            description: *org_id_description
+          roles:
+            type: set
```
oarbusi, line 798: I would expect a change in mcp_config data sources if we have this here. I would regenerate the (org) mcp_config  DS schemas in this PR or move this chunk of config to a follow-up PR so config and generated code don't drift

laylaazouz: good catch! regenerated, thanks :)

oarbusi (approved):
LGTM, docs need to be regenerated after the code regeneration

laylaazouz:
Note on enabling CI, I will merge this branch as is for now and enable CI on the dev branch before merging to master due to the rate limits issue affecting CI atm. https://mongodb.slack.com/archives/C08RN6TUT9P/p1787315895494349

PR #4653, opened 2026-08-18
chore: Adds Autogen emit `jsontypes` import for list/map JSON attributes

EspenAlbert:
## Description

Autogen emitted `customtypes.NewListType[jsontypes.Normalized]` (and map/set) without the `jsontypes` import. That only showed up when a schema had JSON as a collection element and no sibling scalar JSON attribute: `go build` failed with `undefined: jsontypes` unless `goimports` patched the file. `failed_indexes` on the collection restore-job data source is that shape.

Codegen now adds the import whenever `ElemType` is JSON. A missing or failing `goimports`/`fieldalignment` run still only logs a warning, so generate is not blocked.

Jira: [CLOUDP-436516](https://jira.mongodb.org/browse/CLOUDP-436516)

## Type of change:

## Required Checklist:

## Further comments


tools/codegen/gofilegen/schema/schema_attribute.go
```diff
@@ -228,10 +228,13 @@ func commonProperties(attr *codespec.Attribute, typeStatements *TypeSpecificStat
 		var imports []string
 		switch attr.CustomType.Package {
```
manupedrozo: Instead of resolving the extra import here, can we:
- Change `CustomType.Package` to a slice.
- For nested Custom types, we add the nested package (when needed) as well as the custom type package in `NewCustomListType`, `NewCustomSetType` & `NewCustomMapType`.
- Here we loop over packages and map them to the corresponding import.

So in custom types:
```
type CustomType struct {
    Packages []CustomTypePackage `yaml:"packages"`
    Name     string              `yaml:"name,omitempty"`
    Model    string              `yaml:"model"`
    Schema   string              `yaml:"schema"`
}

func NewCustomListType(elemType ElemType) *CustomType {
    elemTypeStr := ElementTypeToModelString[elemType]
    pkgs := []CustomTypePackage{CustomTypesPkg}
    if elemType == CustomTypeJSON {
        pkgs = append(pkgs, JSONTypesPkg)
    }
    return &CustomType{
        Packages: pkgs,
        Model:    fmt.Sprintf("customtypes.ListValue[%s]", elemTypeStr),
        Schema:   fmt.Sprintf("customtypes.NewListType[%s](ctx)", elemTypeStr),
    }
}

...
```
And here:
```
for _, pkg := range attr.CustomType.Packages {
    switch pkg {
    case codespec.JSONTypesPkg:
        imports = append(imports, jsontypesImportStatement)
    case codespec.CustomTypesPkg:
        imports = append(imports, "github.com/mongodb/terraform-provider-mongodbatlas/internal/common/autogen/customtypes")
    }
}
```

manupedrozo (commented):
Thank you for the fix! Suggested a slightly different approach.

tools/codegen/gofilegen/schema/schema_attribute.go
```diff
@@ -228,10 +228,13 @@ func commonProperties(attr *codespec.Attribute, typeStatements *TypeSpecificStat
...
 		case codespec.JSONTypesPkg:
-			imports = append(imports, "github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes")
+			imports = append(imports, jsontypesImportStatement)
 		case codespec.CustomTypesPkg:
 			imports = append(imports, "github.com/mongodb/terraform-provider-mongodbatlas/internal/common/autogen/customtypes")
 		}
+		if elemType, ok := collectionElemType(attr); ok && elemType == codespec.CustomTypeJSON {
+			imports = append(imports, jsontypesImportStatement)
+		}
```
AgustinBettati: I believe this addition is redundant (and hence also the definition of `collectionElemType`), collection types will end up calling `ElementTypeProperty` which will add the correct import. Can you verify?

AgustinBettati (approved):
LGTM, thanks! Would double check on possible redundant logic

tools/codegen/codespec/model.go
```diff
@@ -257,76 +259,106 @@ const (
...
-	Name    string            `yaml:"name,omitempty"` // Nested object name without the "TF" & "Model" prefix and suffix. Used for type overrides.
-	Model   string            `yaml:"model"`
-	Schema  string            `yaml:"schema"`
+	Name     string              `yaml:"name,omitempty"` // Nested object name without the "TF" & "Model" prefix and suffix. Used for type overrides.
+	Model    string              `yaml:"model"`
+	Schema   string              `yaml:"schema"`
+	Packages []CustomTypePackage `yaml:"packages"`
+}
+
+func (c *CustomType) UnmarshalYAML(value *yaml.Node) error {
```
manupedrozo: Checking in, sorry if changes are in-progress.
We'll need to re-generate the models with the new `Packages` as part of this PR, no need to add backwards compatibility.

On another note @AgustinBettati: These `codegen` changes should trigger a model diff check in CI. But we are only checking the code-gen step ([run](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/32121395504/job/95662393763?pr=4653)). Do you remember if there is a specific reason for this? Can we change it?

AgustinBettati: > Do you remember if there is a specific reason for this? Can we change it?

Known limitation we currently have, syncing offline

EspenAlbert:
Thank you! LGTM.

PR #4655, opened 2026-08-18
fix: Require replacement when networking changes

Edmondton:
## Description

Stream connections require replacement when networking changes

https://jira.mongodb.org/browse/CLOUDP-436423

Link to any related issue(s):

## Type of change:

## Required Checklist:

### DIFF
--- a/.changelog/4655.txt
+++ b/.changelog/4655.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Require replacement when networking changes
+```
--- a/.changelog/4655.txt
+++ b/.changelog/4655.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-Require replacement when networking changes
+resource/mongodbatlas_stream_connection: Require replacement when networking changes
 ```
--- a/internal/service/streamconnection/resource_schema.go
+++ b/internal/service/streamconnection/resource_schema.go
@@ -149,6 +149,7 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed: true,
 				PlanModifiers: []planmodifier.Object{
 					objectplanmodifier.UseStateForUnknown(),
+					objectplanmodifier.RequiresReplaceIfConfigured(),
 				},
 				Attributes: map[string]schema.Attribute{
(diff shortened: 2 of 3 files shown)


oarbusi:
Thanks for the PR. I see it comes from your fork. Could you do this into the repo directly so we can run all CI checks? 
Changes look good, but the issue I see is that it will now be easier to delete the stream connection without the user being aware of it(they update the networking and it drops and recreates the connection, maybe causing some downtime). Current behavior surfaces the API error which IMO is quite clear ("STREAM_NETWORKING_CANNOT_BE_MODIFIED") and forces the deletion of the stream connection to be intentional

Edmondton:
> Thanks for the PR. I see it comes from your fork. Could you do this into the repo directly so we can run all CI checks? Changes look good, but the issue I see is that it will now be easier to delete the stream connection without the user being aware of it(they update the networking and it drops and recreates the connection, maybe causing some downtime). Current behavior surfaces the API error which IMO is quite clear ("STREAM_NETWORKING_CANNOT_BE_MODIFIED") and forces the deletion of the stream connection to be intentional

Hi @oarbusi I can definitely make the change directly to the repo.  You bring up a good point. To make a networking change to a stream connection, the associated stream processor already needs to be stopped. The expected workflow would be to stop the processor, make the networking change (which would replace the connection), and then restart the processor. So there will be downtime associated with this change regardless. I can also update the documentation to make the replacement behavior and expected downtime clear to users.  The replacement will also be visible in the Terraform plan before apply, so the user will see that the networking change requires the connection to be replaced.  If the user wanted to avoid deleting the stream connection, then the user could make a new stream connection, stop the stream processor, update the stream processor to reference the new connection, then start the stream processor.

Edmondton:
@oarbusi - will close this PR in favor for https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4659, which is a PR directly against the repo.  i will also address feedback in PR #4659 going forward

oarbusi:
Okay, if we are not increasing the chance of downtime or introducing issues related to deletion of the stream connection, its fine by me.

PR #4665, opened 2026-08-21
feat: Add vertical autoscaling configuration for stream processors

jwongmongodb:
## Summary

Adds Phase 1 vertical autoscaling support to `mongodbatlas_stream_processor` and its singular and plural data sources:

- optional `options.autoscaling` block with optional server-defaulted `min_tier` and `max_tier`
- computed, read-only `effective_tier`, which reports the processor's current running tier
- `tier` remains the user-configured baseline tier; it does not autoscale or drift
- `options.dlq` and `options.autoscaling` are independently optional; an explicitly empty `options` object is rejected

## Included

- Resource and data-source schema, model, conversion, and CRUD updates.
- Declarative disable behavior: removing `options.autoscaling` sends an explicit `options.autoscaling: null` PATCH. If autoscaling is the only option, omit `options` entirely; if a DLQ remains, remove only `options.autoscaling`.
- Correct API shape for `:startWith`, where autoscaling is top-level rather than nested in `options`.
- Unit and acceptance coverage for conversions, `effective_tier`, create/update requests, optional DLQ, block-removal disablement, server-default bounds, and empty-options validation.
- Regenerated documentation, example configuration, and release-note entries.

## Validation

- Targeted unit tests and provider verification pass locally.
- Cloud Dev E2E validation confirmed the explicit PATCH payload:

  ```json
  {
    "options": {
      "autoscaling": null
    }
  }
  ```

  Cloud Dev accepted the update and cleared autoscaling.

## Draft status

This PR remains a draft while the final review and full CI run complete.

**Expected Terraform release: August 27, 2026.**


docs/data-sources/stream_processor.md
```diff
@@ -95,12 +95,19 @@ resource "mongodbatlas_stream_processor" "stream-processor-kafka-to-cluster-exam
...
+  # `tier` is the baseline tier; the current autoscaled tier is `effective_tier`.
+  tier = "SP10"
   options = {
     dlq = {
       coll            = "exampleColumn"
       connection_name = mongodbatlas_stream_connection.example-cluster.connection_name
       db              = "exampleDb"
     }
+    autoscaling = {
+      enabled  = true
```
jwongmongodb: One question on the Terraform shape: do we need to expose `enabled` in the `autoscaling` block? I know in the design/td, we have `enabled` to track if autoscaling is enabled or disabled. But given the API Contract [doc](https://docs.google.com/document/d/1q5mmlV63bEk006CXpDEWM3nXDa41USzgvE_nvHkZGGs/edit?tab=t.9igxibskrbi1#heading=h.yy6aqzg8yrfc) scenario, when an autoscaling config is disabled, the whole object is omitted from the response and subsequent GET calls for the stream processor.

Based on the API behavior we tested in Cloud Dev, the presence of `options.autoscaling` enables autoscaling, and removing the block disables it—the provider sends `enabled: false` on the PATCH, and the backend clears the configuration. Since `enabled = false` cannot round-trip and we reject it today, `enabled` seems redundant as a Terraform argument.

Could we instead make the schema:

```hcl
options = {
  autoscaling = {
    min_tier = "SP10"
    max_tier = "SP50"
  }
}
```

Then the provider can continue translating block presence/removal to `enabled: true` / `enabled: false` in the API request. This feels more idiomatic for Terraform and avoids requiring users to set a field that only has one valid value.

internal/service/streamprocessor/resource_schema.go
```diff
@@ -90,6 +94,35 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 						Required:            true,
 						MarkdownDescription: "Dead letter queue for the stream processor. Refer to the [MongoDB Atlas Docs](https://www.mongodb.com/docs/atlas/reference/glossary/#std-term-dead-letter-queue) for more information.",
 					},
+					"autoscaling": schema.SingleNestedAttribute{
```
AgustinBettati, line 96: q: Sibling `dlq` is still marked as required, is this accurate? Would it have to be defined every time `options` is defined?

internal/service/streamprocessor/resource_schema.go
```diff
@@ -90,6 +94,35 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+								Required:            true,
+								MarkdownDescription: "Flag that indicates whether autoscaling is enabled. Must be `true` when the `options.autoscaling` block is present. To disable autoscaling, remove the block rather than setting this to `false` (the backend does not persist a disabled config, so `enabled = false` is rejected during planning).",
+							},
+							"min_tier": schema.StringAttribute{
+								Optional:            true,
+								Computed:            true,
+								MarkdownDescription: "Tier floor for autoscaling (scale-down limit). Valid options are SP2, SP5, SP10, SP30, and SP50. When not set, it defaults to the workspace default tier.",
+								Validators: []validator.String{
+									stringvalidator.OneOf(streamProcessorTiers...),
+								},
```
AgustinBettati: Highly advise removing enum client side validations (general guidance we have in our resources), these become outdated quickly

internal/service/streamprocessor/resource_schema.go
```diff
@@ -90,6 +94,35 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+							},
+							"max_tier": schema.StringAttribute{
+								Optional:            true,
+								Computed:            true,
+								MarkdownDescription: "Tier ceiling for autoscaling (scale-up limit). Valid options are SP2, SP5, SP10, SP30, and SP50. When not set, it defaults to the workspace maximum tier.",
+								Validators: []validator.String{
```
AgustinBettati: same here

docs/data-sources/stream_processor.md
AgustinBettati: > enabled seems redundant as a Terraform argument

I agree here. I am curious why we didnt apply the same judgement at the API level, why does `autoScaling.enabled` actually exist?

internal/service/streamprocessor/resource_schema.go
```diff
@@ -90,6 +94,35 @@ func ResourceSchema(ctx context.Context) schema.Schema {
...
+						},
+						Attributes: map[string]schema.Attribute{
+							"enabled": schema.BoolAttribute{
+								Required:            true,
+								MarkdownDescription: "Flag that indicates whether autoscaling is enabled. Must be `true` when the `options.autoscaling` block is present. To disable autoscaling, remove the block rather than setting this to `false` (the backend does not persist a disabled config, so `enabled = false` is rejected during planning).",
+							},
+							"min_tier": schema.StringAttribute{
+								Optional:            true,
+								Computed:            true,
+								MarkdownDescription: "Tier floor for autoscaling (scale-down limit). Valid options are SP2, SP5, SP10, SP30, and SP50. When not set, it defaults to the workspace default tier.",
```
AgustinBettati: Unfortunate we did not follow https://mongodb.github.io/ipa/111#optional-fields-with-server-defaults on the API side and hence have to define O+C here.

AgustinBettati (commented):
Providing initial review

docs/data-sources/stream_processor.md
jwongmongodb: This is for the PATCH behavior, excluding the autoscaling config in a PATCH API does not mean that it will disable it since omission means to preserve the existing config for the PATCH contract. 

So we needed a way to disable the autoscaling field with the `enabled` field, for terraform we can simplify this with just the config.

internal/service/streamprocessor/resource_schema.go
jwongmongodb: DLQ and autoscaling are optional in the API, so when options is passed, it can optionally have autoscaling config.

docs/data-sources/stream_processor.md
jwongmongodb: are we okay with this terraform behavior to exclude enabled and have the existence of the Autoscaling config be the enabling factor for terraform @AgustinBettati. If yes, I can make the change now.

EX:
removing autoscaling config from an existing stream processor will disable autoscaling.

internal/service/streamprocessor/resource_schema.go
jwongmongodb: yes, I see that this was a miss. Can we leave this as O+C here for these fields? Maybe use something like `x-xgen-server-computed-when-client-omitted: true` to make it explicit?

EDIT: Looks like the this was added in June 22nd, and the API contract review was before that, so this is probably considered legacy.

internal/service/streamprocessor/resource_schema.go
jwongmongodb: I made DLQ optional now to match the API behavior, I believe this is not a breaking change since we are more relaxed in the provider.

internal/service/streamprocessor/resource_schema.go
jwongmongodb: Agreed. Removed the client-side enum validation in 4ca4f074e so the API remains the source of truth as supported tiers evolve.

internal/service/streamprocessor/resource_schema.go
jwongmongodb: done

docs/data-sources/stream_processor.md
jwongmongodb: I did this cleanup, let me know if this should be reverted to add back `enabled` to the autoscaling field. I believe this change to remove the `enabled` field is a better user experience.

internal/service/streamprocessor/resource_schema.go
jwongmongodb: I verified this in Cloud Dev with this terraform provider: when `min_tier` and `max_tier` are omitted, Atlas resolves them to the workspace defaults and returns them on read; Terraform then reaches a no-change plan. I would keep O+C here for now.

docs/data-sources/stream_processor.md
AgustinBettati: I am aligned, lets make sure to capture this decision in associated TD for future reference. 
In terms of the PATCH behaviour on the API side, one option we could have used (not sure if too late now) is that sending an explicit null (e.g. `autoScaling = null`) in the PATCH is the way of signalling we want to unset the property (https://mongodb.github.io/ipa/107#guidance point 5).

internal/service/streamprocessor/resource_schema.go
AgustinBettati: > Maybe use something like x-xgen-server-computed-when-client-omitted: true to make it explicit?

Yes that would be good, this way our API Spec flags explicitly the behaviour.

internal/service/streamprocessor/autoscaling.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	if stateAutoscaling != nil {
+		return &admin.StreamsAutoscaling{Enabled: new(false)}, nil
+	}
+	return nil, nil
+}
+
+// effectiveTierFromResp derives the read-only `effective_tier` from the API response.
+// Falls back to the baseline `tier` when the API does not return an explicit
+// effectiveTier (they are equal whenever autoscaling is disabled).
```
AgustinBettati: q: Isnt `effectiveTier` always returned?

internal/service/streamprocessor/autoscaling.go
```diff
@@ -0,0 +1,114 @@
...
+//
+// The autoscaling config (`options.autoscaling`) and the read-only top-level
+// `effectiveTier` field are still `@Hidden` in the Atlas API and are therefore
+// not part of the officially generated SDK yet. During development they are
+// provided by a hand-patched LOCAL copy of the versioned SDK wired in via the
+// `replace` directive in go.mod (see .sdkstub/, per the ASP "Terraform
+// Development" wiki's "Local validation in Cloud Dev" guidance). Once the fields
+// are un-hidden and land in the official SDK, remove the stub + replace directive;
+// no changes to this file should be required because it already targets the real
+// admin.StreamsAutoscaling type.
```
AgustinBettati: let make sure to remove this content once we adopt the released go sdk

internal/service/streamprocessor/model.go
```diff
@@ -28,6 +28,10 @@ func NewStreamProcessorReq(ctx context.Context, plan *TFStreamProcessorRSModel)
 	streamProcessor := &admin.StreamsProcessor{
 		Name:     plan.ProcessorName.ValueStringPointer(),
 		Pipeline: &pipeline,
+		// effectiveTier is read-only but the preview SDK currently models it as a
+		// required value type. Send explicit null rather than its invalid zero-value
+		// enum (empty string) until the generated SDK makes it optional/read-only.
+		NullFields: []string{"EffectiveTier"},
```
AgustinBettati: Lets make sure to clean this up once adopting stable SDK

docs/data-sources/stream_processor.md
jwongmongodb: I updated and tagged you in the updated TD reference to the terraform config: https://docs.google.com/document/d/1q5mmlV63bEk006CXpDEWM3nXDa41USzgvE_nvHkZGGs/edit?disco=AAACF73K4vg

jwongmongodb: I also made a post in the thread regarding the level of effort needed to make `autoscaling = null` before the release this Thursday [here](https://mongodb.slack.com/archives/C01T8F3BJFJ/p1787661782060209?thread_ts=1787563805.053499&cid=C01T8F3BJFJ).

internal/service/streamprocessor/autoscaling.go
jwongmongodb: yes, the service does this fallback already so I removed the unnecessary handling from the terraform provider

internal/service/streamprocessor/autoscaling.go
jwongmongodb: Done. The stale local-SDK/stub comment has been removed. The implementation now uses the versioned SDK already adopted on provider master.

internal/service/streamprocessor/model.go
jwongmongodb: Done, the stream processor implementation now uses the versioned SDK and `AtlasV2` client from provider master. The `effectiveTier` create-request workaround is no longer needed and has been removed as part of that update.

docs/data-sources/stream_processor.md
```diff
@@ -140,23 +146,34 @@ output "stream_processors_results" {
 
 ### Read-Only
 
+- `effective_tier` (String) Tier the stream processor is currently running on. When autoscaling is disabled this equals `tier`; when autoscaling is enabled it reflects the tier chosen by the autoscaler within the configured bounds.
 - `failover_enabled` (Boolean) Indicates whether this stream processor is eligible for failover. When `true`, an operator can trigger a failover event to migrate the stream processor to a secondary region configured in the workspace's `failover_regions`. Requires an Atlas-to-Atlas or Atlas-to-Kafka pipeline with `failover_regions` configured on the workspace.
 - `id` (String) Unique 24-hexadecimal character string that identifies the stream processor.
-- `options` (Attributes) Optional configuration for the stream processor. (see [below for nested schema](#nestedatt--options))
+- `options` (Attributes) Optional configuration for the stream processor. Empty `options` objects are not supported. (see [below for nested schema](#nestedatt--options))
```
jwongmongodb, line 152: @AgustinBettati 
I made `dlq` optional so it can be configured independently from autoscaling, which matches the MMS API contract.

This exposed a removal transition that the previous schema could not represent: historically, `dlq` was required whenever `options` was present, so Terraform could never plan a processor with `options` but no DLQ.

For PATCH, MMS/SPM treats an omitted DLQ as “preserve the existing DLQ” and an empty DLQ object (`dlq: {}`) as the explicit remove signal. The provider now translates removal of a configured Terraform DLQ into that empty-object PATCH payload.

I also added validation that rejects `options = {}`. If removing DLQ leaves no other option configured, users should omit `options` entirely. This keeps the Terraform state shape aligned with the API response, which omits `options` when neither DLQ nor autoscaling is configured.

docs/data-sources/stream_processor.md
jwongmongodb: As stated in this slack [thread](https://mongodb.slack.com/archives/C01T8F3BJFJ/p1787667359557549?thread_ts=1787563805.053499&cid=C01T8F3BJFJ), I updated the provider to use an explicit `autoscaling: null` PATCH when the autoscaling block is removed from an existing Terraform-managed processor.

I verified this in Cloud Dev with provider debug logging. The provider sent:

```json
{
  "name": "autoscaling-e2e-processor",
  "options": {
    "autoscaling": null
  },
  "pipeline": [
    {
      "$source": {
        "connectionName": "sample_stream_solar"
      }
    },
    {
      "$emit": {
        "connectionName": "__testLog"
      }
    }
  ],
  "tier": "SP10"
}
```

Cloud Dev accepted the update and cleared autoscaling.

This keeps the Terraform behavior declarative, presence of the autoscaling block enables it, while removing it disables it without exposing the API transport-level `enabled` field in Terraform configuration.

erabil-mdb (approved):
LGTM!

internal/service/streamprocessor/model.go
```diff
@@ -185,16 +205,36 @@ func NewTFStreamprocessorDSModel(ctx context.Context, projectID, instanceName, w
...
+		ConnectionName: dlqModel.ConnectionName.ValueStringPointer(),
+		Db:             dlqModel.DB.ValueStringPointer(),
+	}, nil
+}
+
 func ConvertOptionsToTF(ctx context.Context, options *admin.StreamsOptions) (*types.Object, diag.Diagnostics) {
-	if options == nil || !options.HasDlq() {
+	if options == nil || (!options.HasDlq() && options.Autoscaling == nil) {
 		return new(types.ObjectNull(OptionsObjectType.AttributeTypes())), nil
 	}
```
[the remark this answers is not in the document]
jwongmongodb: This compiles as written. The repository uses Go 1.26.4 (`go.mod`), where `new` accepts either a type or an expression. `types.ObjectNull(...)` returns a `types.Object`, so this produces the required `*types.Object`.

Verified locally with:

```text
go test ./internal/service/streamprocessor -run '^$' -count=1
make build

File: internal/service/streamprocessor/autoscaling.go [hunk 11]
```diff
@@ -0,0 +1,92 @@
...
+	if autoscaling.IsNull() || autoscaling.IsUnknown() {
+		return nil, nil
+	}
+	tfModel := &TFAutoscalingModel{}
+	if diags := autoscaling.As(ctx, tfModel, basetypes.ObjectAsOptions{}); diags.HasError() {
+		return nil, diags
+	}
+	// Block presence is Terraform's enablement signal. The API requires this
+	// transport field for create/PATCH semantics, but it is not exposed in state.
+	req := &admin.StreamsAutoscaling{Enabled: new(true)}
```
jwongmongodb, line 36: unfortunately, enabled=true needs to be sent to the API for autoscaling to work so this logic needs to stay.

internal/service/streamprocessor/resource.go
```diff
@@ -111,6 +111,13 @@ func (r *streamProcessorRS) Create(ctx context.Context, req resource.CreateReque
 		if plan.Tier.ValueString() != "" {
 			startWithOptions.SetTier(plan.Tier.ValueString())
 		}
+		// On the :startWith endpoint, `autoscaling` is TOP-LEVEL (no options wrapper).
```
jwongmongodb, line 114: this is from a convenience endpoint from the [adminAPI](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-startgroupstreamprocessorwith) where the terraform provider calls startsWithOptions if the `state=started` is used in the stream processor.

internal/service/streamprocessor/resource_test.go
```diff
@@ -124,7 +124,7 @@ func TestAccStreamProcessor_withTier(t *testing.T) {
 		}})
 }
 
-func TestAccStreamProcessor_withAutoscaling(t *testing.T) {
+func TestAccStreamProcessor_withOptionsDLQAutoscaling(t *testing.T) {
```
jwongmongodb, line 127: this now tests the options change with DLQ unsetting and setting with autoscaling setting and unsetting, also tests :startsWithOptions with autoscaling

docs/data-sources/stream_processor.md
AgustinBettati: Approach makes sense.

> an empty DLQ object (dlq: {}) as the explicit remove signal

Following the sibling `autoScaling` property, I would be consistent in allowing `dlq = null` as the removal signal in the API side. Also aligning to https://mongodb.github.io/ipa/107#guidance. (not urgent but for considering)

internal/service/streamprocessor/resource_test.go
AgustinBettati: Nice test 👍

AgustinBettati (approved):
LGTM, thanks for the follow ups here

docs/data-sources/stream_processor.md
jwongmongodb: Created this ticket for DLQ unsetting behavior https://jira.mongodb.org/browse/CLOUDP-439044

PR #4676, opened 2026-08-25
fix: Propagates API errors during delete, read, and wait operations of auto-generated resources

AgustinBettati:
## Description

Before this fix, auto-generated resources ignored every API error during delete, read, and wait operations and treated the failure as "resource not found". From the user's perspective:

| Operation | Before the fix | After the fix |
|---|---|---|
| Delete | `terraform destroy` reports a successful destroy even when the Atlas API rejects the delete (400, 401, 409, 500). The resource is removed from Terraform state but still exists in Atlas, with no signal to the operator. | The API error surfaces as a diagnostics error and the resource stays in Terraform state. A 404 is still tolerated as already deleted. |
| Read | An API error during `terraform plan` or refresh silently removes the resource from state, so the next apply tries to recreate it. Data sources report "Resource not found" instead of the real API error. | The API error surfaces and the resource stays in state; only a 404, or a hook signaling not-found, removes it.  |
| Wait (long-running operations) | While waiting for a delete to complete, a transient API error is taken as confirmation that the resource is gone, so the destroy finishes early and reports success. During create and update waits, the user sees a confusing "unexpected state 'DELETED'" error instead of the real one. | The polling error aborts the wait and the real API error surfaces. |

Root cause: `callAPI` returns a nil body on any failed call, `isEmptyJSON(nil)` is true, and `callDelete`, `handleReadCore`, and `refreshFunc` consulted `notFound` before `Err`, so every failure was classified as not-found.

Changes:

- `notFound` now classifies only genuine not-found: an HTTP 404, or a hook signaling it via the new `autogen.ErrNotFound` sentinel. All other errors propagate to diagnostics.
- The search deployment quirk of returning an ok status with an empty JSON body for missing resources moves out of the core into a `PostReadAPICall` hook on `searchdeploymentapi`, covering both read and delete-wait polling.
- The service account secret hooks wrap `autogen.ErrNotFound` when the secret is absent from the list response, so out-of-band deletion still removes the resource from state.
- New unit tests drive `HandleRead`, `HandleDataSourceRead`, `HandleDataSourceReadList`, `callDelete`, and `refreshFunc` through a real `config.MongoDBClient` against an `httptest` server across 400/401/404/409/500/204/200 responses.

Link to any related issue(s): CLOUDP-437881

## Type of change:

## Required Checklist:


internal/common/autogen/handle_operations.go
```diff
@@ -168,12 +167,11 @@ func HandleDataSourceReadList(ctx context.Context, req HandleReadReq) {
...
 		}
 		callResult := callReadWithHooksWithOptions(ctx, req.Client, paginatedParams, req, req.Hooks)
+		// Err covers every failure including not-found, as UntypedAPICall returns an error for any status >= 300.
+		// An ok response with an empty body is parsed below as a page without results.
 		if callResult.Err != nil {
 			return nil, callResult.Resp, callResult.Err
 		}
-		if notFound(callResult.Body, callResult.Resp) {
-			return nil, callResult.Resp, fmt.Errorf("resource not found")
-		}
```
AgustinBettati, line 177: The check for processing 404 was not reachable as it was being catched in if statement above (`callResult.Err != nil`). Only case this if captured was a 200 empty body `{}` with returned with a `resource not found`, now it will be hanlded as a regular successful response with no results.

AgustinBettati:
will wait on CI execution before merging

internal/common/autogen/handle_operations.go
```diff
@@ -423,13 +421,20 @@ func refreshFunc(ctx context.Context, wait *WaitReq, client *config.MongoDBClien
...
-func notFound(bodyResp []byte, apiResp *http.Response) bool {
-	return validate.StatusNotFound(apiResp) || isEmptyJSON(bodyResp)
+// ErrNotFound can be returned (wrapped) by custom hooks to signal that the resource does not exist
+// when the API cannot express it as an HTTP 404. Examples: reads implemented as a LIST call where
+// the hook searches for the element (service account secrets), or APIs that return an ok status code
+// with an empty JSON body for missing resources (search deployment).
+var ErrNotFound = errors.New("resource not found")
+
+// notFound returns if the API result indicates the resource is not found:
+// an HTTP 404 response, or a hook signaling it via ErrNotFound.
```
manupedrozo: Bit confusing :D

manupedrozo (approved):
Ty!

internal/common/autogen/handle_operations.go
AgustinBettati: Rephrased comment slightly. The sentinel gives hooks one consistent way to flag "resource not found" regardless of what the API actually returned, instead of each hook fabricating a fake 404 http.Response that misrepresents the real wire response.

erabil-mdb (approved):
LGTM!

internal/common/autogen/handle_operations_internal_test.go
```diff
@@ -0,0 +1,312 @@
+package autogen
```
EspenAlbert, line 1: is this a common pattern? Using tests in the same package?

EspenAlbert (approved):
nice. Would be great to have a real acceptance test for this also, but understand it might be tricky to trigger

internal/common/autogen/handle_operations_internal_test.go
AgustinBettati: Good callout, let me circle back here as this is diverging from common testing pattern in our repo, we have never used `*_internal_test.go`

AgustinBettati:
Merging so we have changes in upcoming release. Will follow up on unit testing structure that was used + assessing feasibility of capturing fixes in an acceptance test (verified fixes locally to be sure).

PR #4682, opened 2026-08-26
fix: Preserve streams resources when migrating workspace aliases

tcannon91:
## Summary
- Prevents stream connection replacement when an unchanged workspace is migrated from `instance_name` to `workspace_name`.
- Makes the alias migration one-way: reverting a resource that already uses `workspace_name` back to `instance_name` is rejected during planning.
- Avoids unnecessary stop/update/restart calls for alias-only stream processor changes and documents the migration.

## Testing
- `go test ./internal/service/streamconnection ./internal/service/streamprocessor`
- `git diff --check`

## Notes
- Acceptance coverage was added but not run locally because it requires configured Atlas acceptance-test credentials and infrastructure.


erabil-mdb (approved):
LGTM!

internal/service/streamconnection/workspace_name_plan_modifier.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+
+	if RequiresWorkspaceNameReplacement(stateWorkspaceName, stateInstanceName, planWorkspaceName, planInstanceName) {
+		resp.RequiresReplace = true
+	}
+}
+
+func IsWorkspaceNameAliasReversion(stateWorkspaceName, planWorkspaceName, planInstanceName types.String) bool {
+	return !stateWorkspaceName.IsNull() && !stateWorkspaceName.IsUnknown() &&
```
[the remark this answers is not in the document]
EspenAlbert: Confirmed. The prior connection importer populated both aliases, but the new reversion check uses a non-null state `workspace_name` alone to identify canonical state. An unchanged imported connection configured with `instance_name` will therefore fail planning after upgrade; the dual-alias legacy state needs to remain supported or be migrated safely.

internal/service/streamprocessor/model.go
```diff
@@ -20,6 +21,36 @@ func GetWorkspaceOrInstanceName(workspaceName, instanceName types.String) string
...
+	}
+
+	planWithoutAliases := *plan
+	planWithoutAliases.WorkspaceName = state.WorkspaceName
+	planWithoutAliases.InstanceName = state.InstanceName
+	// id and stats are computed API outputs. They can be unknown in an update
+	// plan even though no configured processor setting has changed.
+	planWithoutAliases.ProcessorID = state.ProcessorID
+	planWithoutAliases.Stats = state.Stats
+	return !plan.WorkspaceName.Equal(state.WorkspaceName) && reflect.DeepEqual(planWithoutAliases, *state)
```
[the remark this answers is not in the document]
EspenAlbert: Confirmed. The previous processor importer also wrote both aliases, and this validation consequently rejects an unchanged `instance_name` configuration for that historic state. The migration rule must distinguish a dual-alias legacy state from state produced by a completed workspace-name migration.

internal/service/streamprocessor/resource.go
```diff
@@ -44,6 +45,23 @@ func (r *streamProcessorRS) Schema(ctx context.Context, req resource.SchemaReque
...
+	}
+
+	if IsWorkspaceNameAliasReversion(state.WorkspaceName, plan.WorkspaceName, plan.InstanceName) {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("instance_name"),
+			"Cannot revert stream workspace alias",
+			"This resource already uses workspace_name in state. Use workspace_name instead of the deprecated instance_name attribute.",
+		)
+	}
+}
```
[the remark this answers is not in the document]
EspenAlbert: Verified: pointer model decoding is the established `ModifyPlan` pattern in this repository, and the nil guard handles null state during create. This does not introduce a diagnostic or validation gap.

internal/service/streamprocessor/model.go
```diff
@@ -20,6 +21,36 @@ func GetWorkspaceOrInstanceName(workspaceName, instanceName types.String) string
 	return instanceName.ValueString()
 }
 
+// IsAliasOnlyTransition reports whether a plan changes only the spelling of the
+// workspace identifier while preserving the same known workspace and every other
+// resource attribute.
+// IsWorkspaceNameAliasReversion reports whether a resource that already uses
+// workspace_name in state is being changed back to the deprecated instance_name.
+func IsWorkspaceNameAliasReversion(stateWorkspaceName, planWorkspaceName, planInstanceName types.String) bool {
```
EspenAlbert: This seems duplicated as the workspace_name_plan_modifier function

internal/service/streamprocessor/resource.go
```diff
@@ -44,6 +45,23 @@ func (r *streamProcessorRS) Schema(ctx context.Context, req resource.SchemaReque
 	conversion.UpdateSchemaDescription(&resp.Schema)
 }
 
+func (r *streamProcessorRS) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
EspenAlbert: Q: Why are we using a different mechanism (resource plan modifier here) than the stream connection?

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +54,43 @@ func TestAccStreamProcessor_basic(t *testing.T) {
...
 }
 
+func TestAccStreamProcessor_workspaceNameAliasMigration(t *testing.T) {
+	var (
+		projectID, instanceName = acc.ProjectIDExecutionWithStreamInstance(t)
+		randomSuffix            = acctest.RandString(5)
+		processorName           = "alias-migration-" + randomSuffix
+	)
+
+	resource.Test(t, resource.TestCase{
```
EspenAlbert: why is this not `resource.ParallelTest`? Add a comment if needed

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +54,43 @@ func TestAccStreamProcessor_basic(t *testing.T) {
...
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroyStreamProcessor,
+		Steps: []resource.TestStep{
+			{
+				Config: configMigration(t, projectID, instanceName, processorName, streamprocessor.CreatedState, randomSuffix, sampleSrcConfig, testLogDestConfig, "", nil),
```
EspenAlbert: nice test! But would be more readable if you store the step 1 and step 3 configs so it is clear they are the same

EspenAlbert (commented):
Looks like a few of the bot comments are valid here.
Would encourage to split out the shared logic and use the same method in both resources

internal/service/streamprocessor/resource.go
tcannon91: I think I have adjusted this in the direction you were looking for by bringing this up to the resource_schema's plan modifiers. Let me know if I missed what you were looking for.

EspenAlbert: Great job!
Yes. The processor now applies the shared `streamalias.ReversionPlanModifier` on `instance_name` in `resource_schema.go`, matching the connection's attribute-level mechanism. The resource-level `ModifyPlan` implementation was removed.

internal/service/streamalias/alias.go
```diff
@@ -0,0 +1,14 @@
+package streamalias
```
EspenAlbert: move to the common/customplanmodifier package

internal/service/streamprocessor/model_alias_transition_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/streamalias"
+	"github.com/mongodb/terraform-provider-mongodbatlas/internal/service/streamprocessor"
+)
+
+func TestIsWorkspaceNameAliasReversion(t *testing.T) {
```
EspenAlbert: should this live next to the streamalias file maybe?

internal/service/streamconnection/resource_stream_connection_test.go
```diff
@@ -170,19 +164,28 @@ func testCaseKafkaPlaintextMigration(t *testing.T) *resource.TestCase {
...
+				PlanOnly:    true,
+				ExpectError: regexp.MustCompile("Cannot revert stream workspace alias"),
+			},
 			{
 				ResourceName:            resourceName,
 				ImportStateIdFunc:       checkStreamConnectionImportStateIDFunc(resourceName),
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"authentication.password"},
+				ImportStateVerifyIgnore: []string{"authentication.password", "instance_name", "workspace_name"},
```
EspenAlbert: why do we have to add workspace_name to ignore?

internal/service/streamprocessor/resource_test.go
```diff
@@ -53,6 +54,45 @@ func TestAccStreamProcessor_basic(t *testing.T) {
 	resource.Test(t, *basicTestCase(t))
 }
 
+func TestAccStreamProcessor_workspaceNameAliasMigration(t *testing.T) {
```
EspenAlbert, line 57: nice!

EspenAlbert (commented):
Good job, a few nits and questions before approval 👍

internal/service/streamprocessor/model_alias_transition_test.go
tcannon91: Brought this test up to the common package with the other changes

internal/service/streamalias/alias.go
tcannon91: I did this and turned it into a little bit bigger change that makes the changes more "generic" and reusable in case others have the same use case as well.

EspenAlbert: Confirmed. `AliasReversion` now resides in `internal/common/customplanmodifier`, takes both attribute names as parameters, and is covered by focused unit tests.

internal/service/streamconnection/resource_stream_connection_test.go
EspenAlbert: The migration test no longer ignores either alias. Its prior successful step has already converted state to `workspace_name`, so import verification now confirms that import also writes the canonical `workspace_name` state.

EspenAlbert (approved):
Thank you for addressing all the comments!

internal/service/streamconnection/resource_schema.go
```diff
@@ -40,7 +41,8 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 			"instance_name": schema.StringAttribute{
 				Optional: true,
 				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.RequiresReplace(),
+					customplanmodifier.AliasReversion("workspace_name", "instance_name"),
```
manupedrozo, line 44: Is this considered a breaking change? I.e. you could previously revert to ~`workspace_name`~`instance_name` but not anymore.
cc @EspenAlbert (I have little context on these resources)

AgustinBettati: @tcannon91 Could you provide more context as to why we are now shifting towards a stricter deprecation mechanism (not allowing revert back to using deprecated attribute, import no longer allows leveraging deprecated attribute)? 
Both do seem like breaking changes so we have to be very cautious if this will be released in a minor version.

EspenAlbert: I guess you can consider it a breaking change for users with a pipeline that creates with workspace_name and then chooses instance_name, but IMO that set of action makes little sense in my mind.
I don't think this would force any existing configs to have to change so I consider it ok.

tcannon91: Yes, Espen is correct. We are just trying to prevent people from going towards `instance_name` _from_ `workspace_name`. 

I guess that is supported today but it is not a desirable direction we want customers to move. This should not break any existing customers on `instance_name` that are staying on `instance_name` for now. This is trying to nudge users harder towards workspace name so they have an easier transition when we do fully remove instance_name from our api. 

Also, streams customer base is pretty small and our product team works closely with our customers. If this is a pain point for customers our product team can coordinate with them directly.

.changelog/4682.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:new-guide
+Adds migration guide for Stream Instance to Stream Workspace
```
RafaelCenzano, line 2: [nit] should it be more specific and say for stream connection? the Stream instance to Stream workspace change was a separate change like this that allowed move from stream instance to stream workspace as it's own resource

tcannon91: We can -- this also applies to stream processors though so 🤷‍♂️

internal/service/streamconnection/resource_schema.go
tcannon91: Will wait for approval of this explanation before merging (or if apix wants to merge on my behalf go for it!)

manupedrozo: I see your points Taylor. We do hold a strict policy of no breaking changes for the provider, so even if breaking any existing config is highly unlikely, we tend to defer changes such as these to the next major version.

> This is trying to nudge users harder towards workspace name so they have an easier transition when we do fully remove instance_name from our api.

Seems like we have the deprecation message already for `instance_name` which is as much as we can do from a no-breaking-changes standpoint.
To clarify on the API removal, the provider won't be able migrate to the new API version until the next major provider version - moving to the new API version would be a breaking change (unless we get into custom mapping on the provider side, which we also tend to avoid).

tcannon91: I have stripped down the PR to be less strict around the instance name changes. I am not sure who from this list I should ping for reviews so I will wait until this thread is resolved to consider the PR good to go (if there are not other open threads)

manupedrozo: Thank you Taylor, good to resolve this thread on my end!

EspenAlbert: ok to resolve from me to 👍

PR #4683, opened 2026-08-26
fix: Avoid redundant autoscaling start request

jwongmongodb:
## Summary

Fixes a post-merge issue in the vertical autoscaling support added by #4665.

The provider persists autoscaling through the preceding Create or PATCH request. When a processor is started, it no longer resends the same autoscaling configuration through `:startWith`, which MMS/SPM treats as an additional autoscaling modification.

## Validation

- `go test ./internal/service/streamprocessor -count=1`
- `make verify`
- Cloud Dev direct Create → `STARTED` test using the local provider build
  - Create persisted autoscaling
  - `:startWith` sent only `tier`
  - processor reached `STARTED` successfully

## Follow-up

The underlying MMS/SPM `:startWith` autoscaling path remains a backend issue for direct API clients and is being tracked separately.

### DIFF
--- a/.changelog/4683.txt
+++ b/.changelog/4683.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/mongodbatlas_stream_processor: Avoids resending already persisted autoscaling configuration when starting a stream processor
+```
--- a/.changelog/4683.txt
+++ b/.changelog/4683.txt
@@ -1,3 +1,0 @@
-```release-note:bug
-resource/mongodbatlas_stream_processor: Avoids resending already persisted autoscaling configuration when starting a stream processor
-```
--- a/internal/service/streamprocessor/resource.go
+++ b/internal/service/streamprocessor/resource.go
@@ -111,8 +111,6 @@ func (r *streamProcessorRS) Create(ctx context.Context, req resource.CreateReque
 		if plan.Tier.ValueString() != "" {
 			startWithOptions.SetTier(plan.Tier.ValueString())
 		}
-		// Autoscaling is persisted through the preceding Create request. Do not repeat
-		// it in :startWith: that endpoint treats it as a new autoscaling modification.
 		_, err := connV2.StreamsAPI.StartStreamProcessorWith(ctx, projectID, workspaceOrInstanceName, processorName, startWithOptions).Execute()
 		if err != nil {
 			resp.Diagnostics.AddError(errorCreateStart, err.Error())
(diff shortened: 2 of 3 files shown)


internal/service/streamprocessor/resource_test.go
```diff
@@ -198,6 +196,38 @@ func TestAccStreamProcessor_withOptionsDLQAutoscaling(t *testing.T) {
...
 
+func TestAccStreamProcessor_withOptionsDLQAutoscalingCreateStarted(t *testing.T) {
+	var (
+		projectID, workspaceName = acc.ProjectIDExecutionWithStreamInstance(t)
+		_, clusterName           = acc.ClusterNameExecution(t, false)
+		randomSuffix             = acctest.RandString(5)
+		processorName            = "new-processor-autoscaling-started-" + randomSuffix
+	)
+
+	resource.Test(t, resource.TestCase{
```
jwongmongodb, line 207: this tests the STARTED flow from the CREATE call which is not covered in the TestAccStreamProcessor_withOptionsDLQAutoscaling which calls the STARTED flow in the UPDATE call, which is why this was missed. 

This edge case bug only affected the from scratch `STARTED` state generation, this will be fixed separately internally, but this change is still worth pushing to work in prod and also to clean up the redundant autoscaling config being set.

.changelog/4683.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
AgustinBettati: Is it safe to say this bug never existed for users (autoscaling has not yet been released)? In that case would avoid this changelog, we can use skip changelog entry for passing the CI check

internal/service/streamprocessor/resource.go
```diff
@@ -111,14 +111,6 @@ func (r *streamProcessorRS) Create(ctx context.Context, req resource.CreateReque
 		if plan.Tier.ValueString() != "" {
 			startWithOptions.SetTier(plan.Tier.ValueString())
 		}
-		// On the :startWith endpoint, `autoscaling` is TOP-LEVEL (no options wrapper).
-		autoscaling, diags := autoscalingFromOptions(ctx, plan.Options)
-		if diags.HasError() {
-			resp.Diagnostics.Append(diags...)
-			return
-		}
-		startWithOptions.Autoscaling = autoscaling
```
AgustinBettati, line 120: to understand, was this additional property simply redundant, or actually causing a failures? And what would be the change planned on the API side?

AgustinBettati (approved):
LGTM

.changelog/4683.txt
jwongmongodb: yea this never existed before this, it only affected the most recent autoscaling config. Granted I only tested in DEV so it may be a timing issue. But its still good to remove the redundant autoscaling config from being set

internal/service/streamprocessor/resource.go
jwongmongodb: this is redundant, but there appears to  be a race condition with the service that the api calls when the autoscaling payload was included. I believe that this 404 returned in the endpoint call is a bug that we will fix. But it's good to also remove the redundant config and fix this error path in the meantime. My understanding that any payload sent the startsWithOptions in this flow will trigger this issue: [here](https://jira.mongodb.org/browse/MHOUSE-17754) is the ticket filed

.changelog/4683.txt
AgustinBettati: In that case I would remove this changelog entry to avoid confusion on the user side

jwongmongodb: okay, this is now removed

lizo-mdb (approved):

PR #4687, opened 2026-08-27
feat: Support maintenance wave assignment without a maintenance window

marcabreracast:
## Description

Adapts `mongodbatlas_maintenance_window` to the upstream change that makes `dayOfWeek` optional on the maintenance window API, so customers can set a `wave_assignment` without defining a maintenance window.

- `day_of_week` and `hour_of_day` are now `Optional` (previously `Required`). This is backwards-compatible — existing configurations that set both continue to work unchanged.
- The provider only sends `day_of_week`/`hour_of_day` when they are set in the configuration, reusing the same technique already used for `wave_assignment` to distinguish an explicit `0` from an unset value.
- Range validation for these fields now relies on the Atlas API, which returns clear errors for out-of-range values.

Link to any related issue(s): CLOUDP-438200

## Type of change:

## Required Checklist:

## Further comments

Known limitation (tracked, to be resolved before master): removing `day_of_week`/`hour_of_day` from an existing full window does not clear the schedule in place — the API keeps the previous values (it ignores a `null` on those fields). Clearing a schedule requires resetting the resource. This has been raised with the upstream team.

The `atlas-sdk-go` preview pin remains temporary (see the `go.mod` TODO) and will be reverted to a released SDK before merging to master.


internal/service/maintenancewindow/resource_maintenance_window.go
```diff
@@ -39,25 +38,11 @@ func Resource() *schema.Resource {
...
-				Required: true,
-				ValidateFunc: func(val any, key string) (warns []string, errs []error) {
-					v := val.(int)
-					if v < 0 || v > 23 {
-						errs = append(errs, fmt.Errorf("%q value should be between 0 and 23, got: %d", key, v))
-					}
-					return
-				},
+				Optional: true,
 			},
```
marcabreracast, line 48: Removed the range `ValidateFunc`s for `day_of_week`/`hour_of_day` in favor of relying on the API's validation, consistent with the decision taken for `wave_assignment` on this resource ([context](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4510#discussion_r3427249043)). The API rejects out-of-range values with clear errors:
- `day_of_week` → `400 ATLAS_INVALID_MAINTENANCE_WINDOW_DAY_OF_WEEK`: "Invalid Day of Week. Value should be between 1 and 7 (Sunday = 1)."
- `hour_of_day` → `400 ATLAS_INVALID_MAINTENANCE_WINDOW_HOUR_OF_DAY`: "Invalid Hour of Day. Value should be between 0 and 23."

erabil-mdb (approved):
LGTM

internal/service/maintenancewindow/resource_maintenance_window.go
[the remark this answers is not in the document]
EspenAlbert: Q: Can we add a plan validator that enforce this documented all-or-none relationship? Otherwise a partial schedule passes Terraform validation and fails only after the provider sends the partial API request

docs/resources/maintenance_window.md
```diff
@@ -67,8 +67,8 @@ Use `defer` to defer the next scheduled maintenance event by one week. This only
 ## Argument Reference
 
 * `project_id` - The unique identifier of the project for the Maintenance Window, also known as `groupId` in the official documentation.
-* `day_of_week` - (Required) Day of the week when you would like the maintenance window to start as a 1-based integer: Su=1, M=2, T=3, W=4, T=5, F=6, Sa=7.
-* `hour_of_day` - (Required) Hour of the day when you would like the maintenance window to start. This parameter uses the 24-hour clock, where midnight is 0, noon is 12. Uses the project's configured timezone.
+* `day_of_week` - (Optional) Day of the week when you would like the maintenance window to start as a 1-based integer: Su=1, M=2, T=3, W=4, T=5, F=6, Sa=7. Must be set together with `hour_of_day`. Omit both to set a `wave_assignment` without a maintenance window.
+* `hour_of_day` - (Optional) Hour of the day when you would like the maintenance window to start. This parameter uses the 24-hour clock, where midnight is 0, noon is 12. Uses the project's configured timezone. Must be set together with `day_of_week`. Omit both to set a `wave_assignment` without a maintenance window.
```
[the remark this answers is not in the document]
EspenAlbert: This is more than a documentation gap: omitting the schedule from an existing resource leaves both fields out of the PATCH request, while `Read` writes Atlas's returned schedule back into state. Handle the removal transition and add an acceptance test from a scheduled resource to wave-only configuration so the resource converges

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/hashicorp/terraform-plugin-testing v1.16.0
 	github.com/jarcoal/httpmock v1.4.2
 	github.com/mongodb-forks/digest v1.1.0
-	github.com/mongodb/atlas-sdk-go v1.0.1-0.20260812085251-8093776e0ba0
+	github.com/mongodb/atlas-sdk-go v1.0.1-0.20260827071515-9bafc29a1ed5 // TODO: Temporary pin to the CLOUDP-417642 private-preview SDK branch for maintenance_window wave fields. Revert to the standard dev-latest/tagged release once CLOUDP-421608 ships private previews. Do not merge to master.
```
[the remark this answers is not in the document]
EspenAlbert: The explicit master-merge guard is intentional while the provider depends on the private-preview SDK branch; retain it until the preview SDK is available through the standard release path

internal/service/maintenancewindow/resource_test.go
```diff
@@ -206,6 +206,22 @@ func TestAccConfigRSMaintenanceWindow_waveAssignment(t *testing.T) {
...
+					checkExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "wave_assignment", "1"),
+					resource.TestCheckResourceAttr(dataSourceName, "wave_assignment", "1"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "effective_wave_assignment"),
+				),
+			},
+			{
+				// Omitting the schedule must produce an empty plan (guards against a phantom day_of_week/hour_of_day = 0 drift).
+				Config:   configWaveOnly(orgID, projectName, 1),
+				PlanOnly: true,
```
EspenAlbert, line 224: are we missing an assertion here?

EspenAlbert: are we also certain that this is really catching it? Do we need a maintenance configuration on the org that sets the `day_of_week` and `hour_of_day`?

docs/resources/maintenance_window.md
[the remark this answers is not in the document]
marcabreracast: This will be addressed by upstream team on the API side. Resolving

go.mod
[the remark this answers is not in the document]
marcabreracast: Agreed, retaining it while development is happening and prod SDK is available.

internal/service/maintenancewindow/resource_test.go
marcabreracast: > are we missing an assertion here?

Good catch. Added `day_of_week=0` and `hour_of_day=0` assertions to step 1 above, so the wave-only state values are now explicitly checked. This step stays assertion-free on purpose: `PlanOnly: true` is the assertion, as it fails if re-planning the identical config produces a non-empty plan. 

> are we also certain that this is really catching it? Do we need a maintenance configuration on the org that sets the day_of_week and hour_of_day?

No org setup needed. `day_of_week/hour_of_day` are per-project, not org-level.

internal/service/maintenancewindow/resource_maintenance_window.go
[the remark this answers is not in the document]
marcabreracast: We can yes, addressed in [28414ca](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4687/commits/28414ca2e84626aec6077e259572d97f8c753f44).
For context, the API also enforces this: a partial request returns `400 INVALID_PARAMETER: "dayOfWeek and hourOfDay must both be set"`, which is why I did not consider adding the plan validator and instead rely solely on the API.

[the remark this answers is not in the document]
EspenAlbert: Add a plan-only test for each one-field configuration. The new `RequiredWith` constraint is the provider behavior that moves this failure from the API to Terraform planning, but the current acceptance test covers only valid configurations

internal/service/maintenancewindow/resource_maintenance_window.go
```diff
@@ -259,10 +250,11 @@ func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	params := new(admin.GroupMaintenanceWindowPreviewUpdateRequest)
-	params.DayOfWeek = cast.ToInt(d.Get("day_of_week"))
-
-	if d.HasChange("hour_of_day") {
-		params.HourOfDay = new(cast.ToInt(d.Get("hour_of_day")))
+	if !d.GetRawConfig().GetAttr("day_of_week").IsNull() {
```
EspenAlbert, line 256: 🤖 related to the other comment:

Handle removal of an existing schedule and add coverage for that transition. Changing a scheduled resource to the documented wave-only configuration omits both schedule fields from the PATCH request, but `Read` continues to save Atlas's returned schedule values into state. The resource therefore cannot converge if Atlas retains omitted PATCH fields

internal/service/maintenancewindow/resource_test.go
EspenAlbert: I think you need to set `ExpectNonEmptyPlan: true` to ensure no plan changes.

Also, since this is SDK v2, can you double check these are not returned by the API?
That would still work for SDKv2 but when we upgrade to TPF we get inconsistent error

marcabreracast: But wouldn't `ExpectNonEmptyPlan: true` expect essentially that, a non-empty plan? 
What we're testing here is the opposite, an empty plan, which is exactly what `PlanOnly: true` does: it fails the step if the plan isn't empty. To my understanding `ExpectNonEmptyPlan: true` is for the reverse case, tests that expect a diff to remain, it actually fails with `"Expected a non-empty plan, but got an empty refresh plan"` when the plan is empty.

> Also, since this is SDK v2, can you double check these are not returned by the API?

Fields are not returned by the API so shouldn't be a problem when migrating to TPF

internal/service/maintenancewindow/resource_maintenance_window.go
marcabreracast: You're right. But we can't apply these changes yet because of how the API currently behaves: trying to perform a `PATCH ` with the attributes `dayOfWeek`/`hourOfDay` set to `null` gets ignored. I've flagged this to upstream team on  CLOUDP-439562 and I've created CLOUDP-440317 to follow up on the provider side once that's resolved.

internal/service/maintenancewindow/resource_maintenance_window.go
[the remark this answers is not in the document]
marcabreracast: Addressed in [6c2382d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4687/commits/6c2382d911bad827bf11068330f58bec6aa16393)

internal/service/maintenancewindow/resource_maintenance_window.go
EspenAlbert: Should we document this or add a TODO? Omitting both fields only works for a project that has no existing schedule; otherwise customers receive a non-converging configuration until the API supports clearing it

internal/service/maintenancewindow/resource_test.go
EspenAlbert: Ahh, great! Sorry seems like I got confused

EspenAlbert (approved):

internal/service/maintenancewindow/resource_maintenance_window.go
marcabreracast: Make sense, added a TODO in [e771b5a](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4687/commits/e771b5a5b199b7a7f6513dfbc5058cd2d583a084). Note that since this work is being done on a dev branch, we won't be releasing these changes until all the work has finished on the API side, so no need to add this to the docs.

PR #4693, opened 2026-08-31
feat: Adds copy policy items to cloud backup schedule

EspenAlbert:
Adds copy-policy items, last-N copies, and two request flags on the existing SDKv2 `mongodbatlas_cloud_backup_schedule` resource and data source. Users who need per-copy retention or a last-N window no longer have to PATCH the Admin API and then fight Terraform drift.

Jira: [CLOUDP-388562](https://jira.mongodb.org/browse/CLOUDP-388562)

Upstream: [CLOUDP-331665](https://jira.mongodb.org/browse/CLOUDP-331665) (Copy Snapshot Improvements)

## Type of change:

## Required Checklist:

## Further comments

Additive on the existing SDKv2 resource. No TPF rewrite. Nested `copy_settings.frequencies` is deprecated (warning only, no removal date). Existing `frequencies` configs still apply and stay empty-plan aside from the warning. Switching an entry to `copy_policy_items` or `last_number_of_snapshots` plans leftover frequencies away.
`copy_policy_items_enabled` cannot be reverted after the change, opened [\[CLOUDP-440472\] Document that copyPolicyItemsEnabled cannot be turned off once true](https://jira.mongodb.org/browse/CLOUDP-440472)

**Resource (and persistent fields on the data source):**

- `copy_policy_items_enabled`
- `copy_settings.copy_policy_items` (nested `frequency_type`, `retention_unit`, `retention_value`, computed `id`)
- `copy_settings.last_number_of_snapshots` (1-500)

**Resource only (request flags, not returned by GET):**

- `update_copy_snapshots`
- `delete_copy_snapshots`


internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
```diff
@@ -659,3 +714,100 @@ func getRequestPolicies(policiesItem []admin.DiskBackupApiPolicyItem, respPolici
...
+	}
+	if !computed {
+		s.Optional = true
+	}
+	return s
+}
+
+func resourceCustomizeDiff(_ context.Context, d *schema.ResourceDiff, _ any) error {
+	return ValidateCopySettingsModes(d)
+}
```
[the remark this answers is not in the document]
EspenAlbert: Implemented in CustomizeDiff with a focused SimpleDiff transition test

docs/resources/cloud_backup_schedule.md
```diff
@@ -226,9 +284,12 @@ resource "mongodbatlas_cloud_backup_schedule" "test" {
...
+* `update_snapshots` - (Optional) Specify true to apply the retention changes in the updated backup policy to snapshots that Atlas took previously.
+* `update_copy_snapshots` - (Optional) Specify true to apply the retention changes for updated copy policy items to snapshot copies that Atlas took previously.
+* `delete_copy_snapshots` - (Optional) Specify true to delete snapshot copies when their associated `copy_policy_items` are removed. Requires `copy_policy_items_enabled` to be true.
+
+  **Note** Atlas does not return `update_snapshots`, `update_copy_snapshots`, or `delete_copy_snapshots`. This is Atlas Admin API behavior, not Terraform. For more details about this resource see [Cloud Backup Schedule](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Cloud-Backups/operation/getBackupSchedule).
+* `copy_policy_items_enabled` - (Optional) Flag that selects copy-policy mode. Set to `true` to use `copy_policy_items` or `last_number_of_snapshots`. When `false` or omitted, use `frequencies`.
```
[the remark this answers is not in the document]
EspenAlbert: Documented the one-way transition alongside the copy-policy mode guidance

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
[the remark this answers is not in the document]
EspenAlbert: We don't do the true -> false transition since the API will raise an error but we don't know if that might change in the future.

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
```diff
@@ -659,3 +714,100 @@ func getRequestPolicies(policiesItem []admin.DiskBackupApiPolicyItem, respPolici
...
+
+		if hasItems && hasLastN {
+			return errors.New(errCopySettingsModes)
+		}
+		if (hasItems || hasLastN) && !enabled {
+			return errors.New(errCopyPolicyRequiresFlag)
+		}
+		// Leftover Optional+Computed frequencies are ignored when another mode is set so a frequencies config can switch in one apply.
+		if enabled && hasFreq && !hasItems && !hasLastN {
+			return errors.New(errFrequenciesWithFlag)
```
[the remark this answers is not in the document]
EspenAlbert: thanks, changed to support this

docs/resources/cloud_backup_schedule.md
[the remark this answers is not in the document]
EspenAlbert: updated the docs

internal/service/cloudbackupschedule/customize_diff.go
```diff
@@ -0,0 +1,167 @@
...
+	if copySettingsRawConfigEmpty(copySettingsRaw) {
+		return d.SetNew("copy_settings", []any{})
+	}
+	return clearFrequenciesWhenCopyPolicyEnabled(d)
+}
+
+// copySettingsRawConfigEmpty reports whether HCL omitted copy_settings. Also used at apply time by copySettingsForUpdate.
+func copySettingsRawConfigEmpty(raw cty.Value) bool {
+	// Omitted Optional+Computed nested blocks arrive as unknown (computed from state), not as null.
+	return raw.IsNull() || !raw.IsKnown() || raw.LengthInt() == 0
```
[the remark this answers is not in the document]
EspenAlbert: Fixed: `copySettingsRawConfigEmpty` now treats only a null or statically empty list as omitted, so an unknown dynamic-block collection keeps the planned value computed, and `copySettingsForUpdate` expands the apply-time config when the planned list is still computed; apply no longer deletes the configured entries

internal/service/cloudbackupschedule/customize_diff.go
```diff
@@ -0,0 +1,167 @@
...
+			copied["frequencies"] = schema.NewSet(schema.HashString, []any{})
+			changed = true
+		}
+		rewritten[i] = copied
+	}
+	return rewritten, changed
+}
+
+func validateCopySettingsModes(d *schema.ResourceDiff, copySettingsRaw cty.Value) error {
+	enabled, _ := d.Get("copy_policy_items_enabled").(bool)
```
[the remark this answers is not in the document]
EspenAlbert: Fixed: `validateCopySettingsModes` skips the flag-dependent checks when the raw config flag is unknown, so configurations that derive the flag from another resource plan normally and the API validates the resolved request at apply

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
```diff
@@ -100,6 +106,11 @@ func Resource() *schema.Resource {
 							Optional: true,
 							Computed: true,
 						},
+						"copy_policy_items": copyPolicyItemsSchema(false),
+						"last_number_of_snapshots": {
```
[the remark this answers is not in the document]
EspenAlbert: Intentional: we rely on API validation for value bounds instead of local schema validation; the docs state the 1 to 500 range and Atlas rejects out-of-range values at apply

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
```diff
@@ -659,3 +681,46 @@ func getRequestPolicies(policiesItem []admin.DiskBackupApiPolicyItem, respPolici
...
+	}
+	if !computed {
+		frequencyType.Required = true
+	}
+	retentionUnit := &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: computed,
+	}
+	if !computed {
+		retentionUnit.Optional = true
```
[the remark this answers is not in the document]
EspenAlbert: Intentional: retention requirements are enforced by the API, not local schema validation; the docs state `retention_unit` and `retention_value` are required except for `ondemand`, and Atlas rejects invalid items at apply

antellezr-mdb (approved):
LGTM

internal/service/cloudbackupschedule/customize_diff.go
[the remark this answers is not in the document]
EspenAlbert: No local fix: cty `IsKnown` is shallow, so a static `copy_settings` block that references computed values (for example `zone_id` from a cluster applied in the same run) is not treated as omitted; only a dynamic block whose `for_each` collection is itself computed arrives wholly unknown. That configuration is out of scope for this change; the apply path then uses prior state and a later apply converges

internal/service/cloudbackupschedule/customize_diff.go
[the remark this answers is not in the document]
EspenAlbert: No local fix: `copy_policy_items_enabled` is an Optional-only mode switch set statically in the configuration; deriving it from computed values is not a supported configuration and the plan error is the expected outcome

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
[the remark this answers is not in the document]
EspenAlbert: Intentional: we rely on API validation for value bounds instead of local schema validation; the docs state the 1 to 500 range and Atlas rejects out-of-range values at apply

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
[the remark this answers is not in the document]
EspenAlbert: Intentional: retention requirements are enforced by the API, not local schema validation; the docs state `retention_unit` and `retention_value` are required except for `ondemand`, and Atlas rejects invalid items at apply

docs/resources/cloud_backup_schedule.md
```diff
@@ -193,42 +193,55 @@ resource "mongodbatlas_cloud_backup_schedule" "test" {
...
+* `update_copy_snapshots` - (Optional) Specify true to apply the retention changes for updated copy policy items to snapshot copies that Atlas took previously. Requires `copy_policy_items_enabled` to be true.
+* `delete_copy_snapshots` - (Optional) Specify true to delete snapshot copies when their associated `copy_policy_items` are removed. Requires `copy_policy_items_enabled` to be true.
+
+  **Note** Atlas does not return `update_snapshots`, `update_copy_snapshots`, or `delete_copy_snapshots`. This is Atlas Admin API behavior, not Terraform. For more details about this resource see [Cloud Backup Schedule](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Cloud-Backups/operation/getBackupSchedule).
+* `copy_policy_items_enabled` - (Optional) Flag that selects copy-policy mode. Set to `true` to use `copy_policy_items` or `last_number_of_snapshots`. When `false` or omitted, use `frequencies`. This transition is one-way: after you apply with `copy_policy_items_enabled = true`, Atlas cannot disable copy-policy items. You can still switch each `copy_settings` entry between `copy_policy_items` and `last_number_of_snapshots`.
```
AgustinBettati, line 242: To confirm, we do not expect users to remove copy_policy_items_enabled once it has been set to true right?

internal/service/cloudbackupschedule/customize_diff.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func copySettingsFromRawConfig(rawConfig cty.Value) cty.Value {
+	if rawConfig.IsNull() || !rawConfig.IsKnown() {
+		return cty.NullVal(cty.DynamicPseudoType)
+	}
+	return rawConfig.GetAttr("copy_settings")
+}
+
+func clearFrequenciesWhenCopyPolicyEnabled(d *schema.ResourceDiff) error {
```
AgustinBettati, line 52: q: Doesnt the API stop returning frequencies once copy policy is enabled? Just trying to understand why we need the additional clearing logic considering a followup read would come with this value emtpy

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
```diff
@@ -70,6 +78,7 @@ func Resource() *schema.Resource {
 			"copy_settings": {
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true, // SetNew on the list requires Computed; omitting the block is still a delete via CustomizeDiff.
```
AgustinBettati, line 81: And we are using `SetNew` strictly because of new `resourceCustomizeDiff` logic right?

internal/service/cloudbackupschedule/resource_cloud_backup_schedule_test.go
```diff
@@ -369,6 +372,182 @@ func TestAccBackupRSCloudBackupSchedule_copySettings_zoneId(t *testing.T) {
...
+				{Region: "US_EAST_2"},
+			},
+			PitEnabled: true,
+		})
+		base                   = newCopyScheduleConfig(&clusterInfo)
+		configFrequencies      = base.withFrequencies()
+		configDailyAndOndemand = base.withDaily(7).withOndemand()
+		configUpdateCopies     = base.withDaily(3).withOndemand().withUpdateCopies()
+		configDeleteOndemand   = base.withDaily(3).withDeleteCopies()
+		configLastN            = base.withLastN(5)
```
AgustinBettati, line 510: really nice

internal/service/cloudbackupschedule/resource_cloud_backup_schedule_test.go
```diff
@@ -369,6 +372,182 @@ func TestAccBackupRSCloudBackupSchedule_copySettings_zoneId(t *testing.T) {
...
+			},
+			// update_copy_snapshots is config-only (GET omits it). Re-plan the same HCL so Read does not invent a false -> true diff.
+			{
+				Config: configUpdateCopies.HCL(),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+			},
```
AgustinBettati, line 538: nit: Non-empty plan would be caught in the previous step right?

examples/mongodbatlas_cloud_backup_schedule/main.tf
```diff
@@ -46,17 +46,22 @@ resource "mongodbatlas_cloud_backup_schedule" "test" {
   reference_minute_of_hour = 45
   restore_window_days      = 4
 
+  copy_policy_items_enabled = true
+
   copy_settings {
-    cloud_provider = "AWS"
-    frequencies = ["HOURLY",
```
AgustinBettati, line 51: nit: spotted some examples in examples/mongodbatlas_backup_compliance_policy that are still using legacy `frequencies`

docs/resources/cloud_backup_schedule.md
```diff
@@ -193,42 +193,55 @@ resource "mongodbatlas_cloud_backup_schedule" "test" {
...
+* `update_snapshots` - (Optional) Specify true to apply the retention changes in the updated backup policy to snapshots that Atlas took previously.
+* `update_copy_snapshots` - (Optional) Specify true to apply the retention changes for updated copy policy items to snapshot copies that Atlas took previously. Requires `copy_policy_items_enabled` to be true.
+* `delete_copy_snapshots` - (Optional) Specify true to delete snapshot copies when their associated `copy_policy_items` are removed. Requires `copy_policy_items_enabled` to be true.
+
+  **Note** Atlas does not return `update_snapshots`, `update_copy_snapshots`, or `delete_copy_snapshots`. This is Atlas Admin API behavior, not Terraform. For more details about this resource see [Cloud Backup Schedule](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Cloud-Backups/operation/getBackupSchedule).
```
AgustinBettati: Any terraform implication? Not certain what value this adds to the end user

AgustinBettati (commented):
Overall LGTM, just want to double check is we can reduce some of the complexity around new resourceCustomizeDiff or if it is fully needed

docs/resources/cloud_backup_schedule.md
EspenAlbert: Correct: once the flag has been applied `true`, the expectation is that it stays `true` for the life of the schedule. Removing it or setting it to `false` sends `copyPolicyItemsEnabled: false` in the update and Atlas rejects it, so the apply fails until the config sets it back; we let the API raise that error rather than blocking the transition at plan time

internal/service/cloudbackupschedule/customize_diff.go
EspenAlbert: Correct: once the flag is true, the GET omits frequencies and the read after the update clears state (`flattenFrequencies` returns an empty slice so `d.Set` wipes the Optional+Computed leftover). The CustomizeDiff clearing covers the switch apply itself: `frequencies` is Optional+Computed, so when the config drops it during the one-apply switch the SDK plans the state-carried set, the plan shows the old frequencies retained next to the new `copy_policy_items`, and `normalizeNullValues` restores the planned set into the written state, so the stale value only disappears after the follow-up read. The rewrite makes plan, apply, and read agree in one pass

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
EspenAlbert: Yes: SDKv2 `ResourceDiff.SetNew` only operates on Computed keys (`checkKey` in `resource_diff.go` rejects non-Computed), so rewriting the planned `copy_settings` block from `resourceCustomizeDiff` required `Computed: true` on the list. The tradeoff is that Optional+Computed keeps prior state when the block is omitted, which is why CustomizeDiff forces `[]` for a null config so delete-on-omit still works

internal/service/cloudbackupschedule/resource_cloud_backup_schedule_test.go
EspenAlbert: The previous step asserts post-apply state, not the next plan, so it cannot catch a false diff on re-plan. This step re-plans the same config and expects an empty plan because `update_copy_snapshots` is config-only (the GET never returns it); without it, a Read that invents a value would surface as a `false` to `true` diff only on a user's next apply

examples/mongodbatlas_cloud_backup_schedule/main.tf
EspenAlbert: Fixed: migrated both schedules in `examples/mongodbatlas_backup_compliance_policy` (`resource` and `module/modules/cluster_with_schedule`) to `copy_policy_items_enabled = true` with `copy_policy_items` (daily plus ondemand), matching the main example

docs/resources/cloud_backup_schedule.md
EspenAlbert: Agreed, removed the note entirely; the write-only behavior stays guarded by the acc test's re-plan ExpectEmptyPlan step

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
```diff
@@ -659,3 +685,46 @@ func getRequestPolicies(policiesItem []admin.DiskBackupApiPolicyItem, respPolici
 	}
 	return nil
 }
+
+func copyPolicyItemsSchema(computed bool) *schema.Schema {
+	frequencyType := &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: computed,
+	}
+	if !computed {
```
AgustinBettati: nit: These ifs seem like they could be fully inlined: 
```
func copyPolicyItemsSchema(computed bool) *schema.Schema {
	return &schema.Schema{
		Type:     schema.TypeList,
		Optional: !computed,
		Computed: computed,
		Elem: &schema.Resource{
				"id":              {Type: schema.TypeString, Computed: true},
				"frequency_type":  {Type: schema.TypeString, Required: !computed, Computed: computed},
				"retention_unit":  {Type: schema.TypeString, Optional: !computed, Computed: computed},
				"retention_value": {Type: schema.TypeInt, Optional: !computed, Computed: computed},
			},
		},
	}
}
```

AgustinBettati (approved):
LGTM, thanks for the follow ups

internal/service/cloudbackupschedule/resource_cloud_backup_schedule.go
EspenAlbert: thanks, cleaned in 566dfad4665fdd2aa64b54fe6eee6c7c7f945245

PR #4694, opened 2026-08-31
test: Moves autogen error handling tests to black box packages and adds a mocked lifecycle test

AgustinBettati:
## Description

Follow-up to #4676, addressing two review comments.

**Test package.** The `*_internal_test.go` files added in #4676 were the only same-package tests in the repo, and `contributing/testing-best-practices.md` requires a separate `_test` package. All four are now external test packages, and no export shim was needed.

In `handle_operations_test.go` the change that matters is that `TestNotFound` is gone. It asserted `notFound` against 8 hand-built `APICallResult` values, which is exactly where a synthesized combination can drift from what a real call produces. Those cases now arrive from actual HTTP responses through `HandleRead`, `HandleDelete` and `HandleDataSourceRead`: the 2xx-with-error case is a real undecodable body, and the no-HTTP-response case is a real transport failure against a closed server. The one case that cannot come from an HTTP response, the `ErrNotFound` sentinel, has its own test at all three `notFound` call sites.

The rest is mechanical. `TestCallDelete` and `TestRefreshFunc` keep their cases but run through `HandleDelete`, without and with a `Wait`. `TestHandleRead` grows from 4 cases to 8, since driving it from an httptest server makes the extra status codes cheap.

**Acceptance test.** Real Atlas cannot trigger the fix. Every auto-generated delete endpoint answers 404 for a missing resource, so it will not return a 4xx/5xx on the delete path on demand. `handle_operations_lifecycle_test.go` instead drives a full plan/apply/refresh/destroy cycle against canned responses using the existing `unit.TestAccProviderV6FactoriesWithMock`, on `mongodbatlas_ai_model_api_key`. It asserts both user visible symptoms: a 500 on refresh surfaces the error and leaves the resource in state, and a 409 on delete fails the destroy. No credentials needed, runs in `make test`.

Both are verified by reverting the fix: reverting `notFound` fails 13 subtests plus the lifecycle refresh step, and reverting only `callDelete` fails the lifecycle destroy step.

Link to any related issue(s):

## Type of change:

## Required Checklist:


internal/common/autogen/handle_operations_lifecycle_test.go
```diff
@@ -0,0 +1,165 @@
...
+			geography  = "ANY"
+			name       = "mocked-api-key"
+		}
+	`, lifecycleProjectID)
+}
+
+// mockAtlas answers the ai model API key endpoints, and lets a test step arm a failure for the next
+// read or delete. It is both the HTTP client modifier expected by unit.TestAccProviderV6FactoriesWithMock
+// and the round tripper it installs.
+type mockAtlas struct {
```
EspenAlbert, line 78: should this be in a helper file maybe?

EspenAlbert: keeping it local reads right to me. the recorded-mock infra (`unit.CaptureOrMockTestCaseAndRun`) matches requests against captured data files, so injecting synthetic failures mid-lifecycle, which is the whole point here since real Atlas never returns these on demand, would mean hand-editing captured responses. if more error-injection lifecycle tests show up, the reusable piece would be a small status-arming transport in `internal/testutil/unit` next to `HTTPClientModifier`; I'd extract at that point, not before

internal/common/autogen/handle_operations_test.go
```diff
@@ -0,0 +1,347 @@
...
+	srv.Close()
+	return newClient(t, srv.URL)
+}
+
+func newClient(t *testing.T, baseURL string) *config.MongoDBClient {
+	t.Helper()
+	client, err := config.NewClient(&config.Credentials{BaseURL: baseURL}, "test")
+	require.NoError(t, err)
+	return client
+}
```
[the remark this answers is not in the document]
EspenAlbert: 🤖 verified: `unreachableClient` closes the httptest server and keeps its URL, so the transport-error case depends on the released ephemeral port staying unbound; `go test ./...` runs packages in parallel, so a server from another package can rebind it and the case would get an HTTP response instead. a round tripper that always returns an error makes the failure deterministic. one more spot: `TestHandleRead` calls `unreachableClient(t)` in every subtest before overwriting it (`handle_operations_test.go:246`), so each subtest opens and closes a spare server; guarding with `if tc.unreachable` would remove that churn too

internal/common/autogen/handle_operations_internal_test.go
```diff
@@ -1,313 +0,0 @@
-package autogen
```
EspenAlbert, line 1: I suggest we discuss this in the tech sync

EspenAlbert (approved):
Really cool way of testing! I'm tempted to say we should already extract some helpers but it might be pre-mature. I would definitely raise an item in the tech sync!

internal/common/autogen/handle_operations_lifecycle_test.go
AgustinBettati: Kept it local for now since this is the only consumer

internal/common/autogen/handle_operations_test.go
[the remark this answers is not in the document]
AgustinBettati: Replaced the closed-server trick with a failingTransport round tripper that always returns an error, so the transport case no longer depends on an unbound port

internal/common/autogen/handle_operations_internal_test.go
AgustinBettati: Added a pointer to the black-box rule in contributing/testing-best-practices.md to the acceptance-test-patterns skill in this PR so it surfaces during development and review

AgustinBettati:
failing tests in `auto_gen` are unrelated timeout flaky errors associated to search index

PR #4697, opened 2026-09-01
feat: Add shard_size_limit_gb to advanced_cluster

lantoli:
## Description

Add `shard_size_limit_gb` under `auto_scaling.storage_config` for INFINITE advanced clusters.

- Support create, update, removal, and import for INFINITE replica sets, and expose shard limits through both data sources.
- Keep `shard_size_limit_gb` required inside the optional `storage_config` block because Atlas omits empty `storageConfig` from API responses. This omission behavior is settled API design (matches `autoScaling`), so removal is handled provider-side via a plan modifier and a forced `replicationSpecs` PATCH.
- Set `compute_enabled` explicitly in the INFINITE examples and acceptance tests because Atlas now requires an explicit `autoScaling.compute.enabled` on INFINITE cluster creates. The provider sends no default; omitting it returns an Atlas `MISSING_ATTRIBUTE` error at apply time.
- The `online-changed` documentation-link check is expected to fail until the [Atlas Infinite documentation page](https://www.mongodb.com/docs/atlas/infinite/atlas-infinite-landing/) is live. The link comes from the base branch and currently returns HTTP 404 ([failing check](https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/34126931731/job/101757692454)).

Link to any related issue(s): CLOUDP-430468.

## Type of change:

## Required Checklist:


docs/data-sources/advanced_cluster.md
```diff
@@ -260,6 +260,8 @@ Key-value pairs that categorize the cluster. Each key and value has a maximum le
 * `compute_scale_down_enabled` - Flag that indicates whether the instance size may scale down. 
 * `compute_min_instance_size` - Minimum instance size to which your cluster can automatically scale (such as M10). 
 * `compute_max_instance_size` - Maximum instance size to which your cluster can automatically scale (such as M40). 
+* `storage_config` - Settings that determine the configured per-shard data-size limit for an Atlas INFINITE cluster.
+  * `shard_size_limit_gb` - Maximum configured data size that MongoDB Cloud allows each shard to reach, expressed in gigabytes.
```
antellezr-mdb, line 264: In `docs/resources/advanced_cluster.md`, these settings are labeled as optional or required. Should we add those labels here too?

docs/data-sources/advanced_clusters.md
```diff
@@ -295,6 +295,8 @@ Key-value pairs that categorize the cluster. Each key and value has a maximum le
 * `compute_scale_down_enabled` - Flag that indicates whether the instance size may scale down.
 * `compute_min_instance_size` - Minimum instance size to which your cluster can automatically scale (such as M10).
 * `compute_max_instance_size` - Maximum instance size to which your cluster can automatically scale (such as M40).
+* `storage_config` - Settings that determine the configured per-shard data-size limit for an Atlas INFINITE cluster.
```
antellezr-mdb, line 298: In `docs/resources/advanced_cluster.md`, these settings are labeled as optional or required. Should we add those labels here too?

antellezr-mdb (approved):
LGTM with one question

internal/service/advancedcluster/plan_modifier.go
```diff
@@ -281,3 +285,31 @@ func minLen[T any](a, b []T) int {
...
+		return
+	}
+	if !req.PlanValue.IsNull() && !req.PlanValue.IsUnknown() {
+		attributes = req.PlanValue.Attributes()
+	}
+	attributes["storage_config"] = types.ObjectNull(storageConfigObjType.AttrTypes)
+	var diags diag.Diagnostics
+	resp.PlanValue, diags = types.ObjectValue(autoScalingWithStorageConfigObjType.AttrTypes, attributes)
+	resp.Diagnostics.Append(diags...)
+}
```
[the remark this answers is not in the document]
lantoli: `req.Plan` exists in the pinned Terraform Plugin Framework v1.19.0 (`ObjectRequest.Plan` is a `tfsdk.Plan`). The null check preserves destroy plans. This modifier is only attached to the schema containing `storage_config`, and null/unknown state is already guarded

internal/service/advancedcluster/schema.go
```diff
@@ -497,6 +500,23 @@ func autoScalingSchema() schema.SingleNestedAttribute {
...
+	result.PlanModifiers = []planmodifier.Object{clearRemovedStorageConfig{}}
+	result.Attributes["storage_config"] = schema.SingleNestedAttribute{
+		Optional:            true,
+		MarkdownDescription: descStorageConfig,
+		Attributes: map[string]schema.Attribute{
+			"shard_size_limit_gb": schema.Int64Attribute{
+				// TODO: CLOUDP-443184: Make Optional after Atlas preserves empty versus omitted storageConfig in API responses.
+				Required:            true,
+				MarkdownDescription: descShardSizeLimitGB,
+			},
```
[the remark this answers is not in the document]
lantoli: keeping range validation in Atlas to avoid duplicating its rules in Terraform. The acceptance tests cover Atlas rejecting zero and negative limits

docs/data-sources/advanced_cluster.md
lantoli: these are computed data-source outputs, so Optional/Required labels don't apply here. Those labels describe configurable resource inputs. I've updated the resource docs to show both inputs as Optional ahead of [CLOUDP-443184](https://jira.mongodb.org/browse/CLOUDP-443184), keeping the documentation review in this PR while the schema change follows in that ticket

docs/data-sources/advanced_clusters.md
lantoli: these are computed outputs under `results`, so Optional/Required labels would imply they can be configured. Keeping them without those labels matches the other fields in Attributes Reference

docs/resources/advanced_cluster.md
```diff
@@ -868,6 +868,8 @@ replication_specs = [
...
 * `compute_min_instance_size` - (Optional) Minimum instance size to which your cluster can automatically scale (such as M10). Atlas requires this parameter if `replication_specs[#].region_configs[#].auto_scaling.compute_scale_down_enabled` is true.
 * `compute_max_instance_size` - (Optional) Maximum instance size to which your cluster can automatically scale (such as M40). Atlas requires this parameter if `replication_specs[#].region_configs[#].auto_scaling.compute_enabled` is true.
+* `storage_config` - (Optional) Settings that determine the per-shard data-size limit for an Atlas INFINITE cluster.
+  * `shard_size_limit_gb` - (Optional) Maximum data size that MongoDB Cloud allows each shard to reach, expressed in gigabytes. The minimum value is `1`. Set the same value for every region configuration. Omit `shard_size_limit_gb` or `storage_config` to use the Atlas default limit.
```
AgustinBettati: nit: Could be worth leaving as `(Required)` reflecting current state, in followup ticket both schema and docs can be adjusted.

internal/service/advancedcluster/model_to_ClusterDescription20240805.go
```diff
@@ -51,6 +51,11 @@ func newAtlasReq(ctx context.Context, input *TFModel, diags *diag.Diagnostics) *
 		AdaptiveCapacity:                              conversion.NilForUnknown(input.AdaptiveCapacity, input.AdaptiveCapacity.ValueStringPointer()),
 		AdvancedConfiguration:                         newClusterAdvancedConfiguration(ctx, &input.AdvancedConfiguration, diags),
 	}
+	// Imported Infinite clusters can lack databaseEdition, so a shard limit also triggers cleanup.
+	if result.GetDatabaseEdition() == "INFINITE" || hasShardSizeLimit(result.GetReplicationSpecs()) {
```
AgustinBettati: Is there any way we could check the effective value to be fully sure it is an infinite cluster or not?

internal/service/advancedcluster/resource_autoscaling_test.go
```diff
@@ -0,0 +1,396 @@
...
+				"readOnlySpecs":{"instanceSize":"M20","nodeCount":0},
+				"analyticsSpecs":{"instanceSize":"M20","nodeCount":0}
+			}]
+		}]}`, string(encoded))
+	}).Return(admin.UpdateClusterApiRequest{ApiService: api}).Once()
+	apiError := errors.New("request inspected")
+	api.EXPECT().UpdateClusterExecute(mock.Anything).Return(nil, nil, apiError).Once()
+	prior, plan := model(true), model(false)
+	var resp resource.UpdateResponse
+	r.Update(ctx, resource.UpdateRequest{
```
AgustinBettati: Interesting approach for calling the top level resource TPF functions. Is it worth having a single test that calls `ModifyPlan` and then `Update`, considering in a regular tf apply the object Update receives is the plan modifier's output.

docs/resources/advanced_cluster.md
lantoli: changed to `(Required)` reflecting current state in a66ad29, `CLOUDP-443184` tracks the follow-up to make it optional

internal/service/advancedcluster/model_to_ClusterDescription20240805.go
lantoli: callers of `validateInfiniteClusterType` now pass `EffectiveDatabaseEdition` alongside `DatabaseEdition` at the server-response checkpoints, resolved via `resolveDatabaseEdition` (effective wins whenever known) in resource_database_edition.go, a66ad29. This spot in `newAtlasReq` only shapes the outgoing request from plan/config, there's no server response to check yet

internal/service/advancedcluster/resource_autoscaling_test.go
lantoli: removed `TestUpdateRemovesShardSizeLimitAcrossReplicationSpecs` (Update-only, redundant) in a66ad29 and kept `TestUpdateRemovesShardSizeLimitPreservesPlannedValues`, which does what you suggested: calls `PlanResourceChange` through the muxed server to exercise `ModifyPlan`, then feeds that output into `Update`

internal/service/advancedcluster/resource.go
```diff
@@ -530,6 +576,11 @@ func findClusterDiff(ctx context.Context, state, plan *TFModel, diags *diag.Diag
 	patchOptions := update.PatchOptions{
 		IgnoreInStatePrefix: []string{"replicationSpecs"}, // only use config values for replicationSpecs, state values might come from the UseStateForUnknown and shouldn't be used, `id` is added in updateLegacyReplicationSpecs
 	}
+	if shardSizeLimitRemoved(stateReq.ReplicationSpecs, planReq.ReplicationSpecs) {
+		// Atlas clears an omitted shardSizeLimitGB only when replicationSpecs is included in the PATCH.
+		patchOptions.ForceUpdateAttr = []string{"replicationSpecs"}
```
EspenAlbert, line 538: 🤖 the latest acceptance run on head (attempt 16242) is red in both tests that exercise the new removal path built here: removing `storage_config` while compute auto-scaling stays configured. `TestAccClusterAdvancedCluster_infiniteShardSizeLimit` fails at step 20/23 and `TestAccClusterAdvancedCluster_infiniteComputeAutoScaling` fails at step 11/12, each with `PATCH ... HTTP 500 (UNEXPECTED_ERROR)` from Atlas on the forced `replicationSpecs` removal request. removal of the whole `auto_scaling` block (`infiniteShardSizeLimit` step 13) and plain limit updates pass, so the failing payload shape is specifically the forced removal PATCH that keeps `autoScaling.compute`. the PR description acknowledges the docs-link failure but not this one. can you rerun the suite and confirm whether this is the known Atlas PATCH defect family from CLOUDP-443190 or something in the forced-removal payload, and record the outcome or gate before merge?

internal/service/advancedcluster/resource_database_edition.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var _ resource.ResourceWithValidateConfig = &rs{}
+
+func (r *rs) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var clusterType, databaseEdition types.String
+	resp.Diagnostics.Append(req.Config.GetAttribute(ctx, path.Root("cluster_type"), &clusterType)...)
+	resp.Diagnostics.Append(req.Config.GetAttribute(ctx, path.Root("database_edition"), &databaseEdition)...)
+	if !resp.Diagnostics.HasError() {
+		validateInfiniteClusterType(&resp.Diagnostics, clusterType.ValueString(), databaseEdition.ValueString(), "")
```
EspenAlbert: q: `ValidateConfig` rejects SHARDED/GEOSHARDED combined with `database_edition = "INFINITE"`, but nothing rejects `storage_config` when `database_edition` is `CORE` or omitted, even though the attribute description and the resource docs say these settings apply only to INFINITE clusters. a CORE configuration with `storage_config` surfaces only as an apply-time Atlas error (or as perpetual drift if Atlas silently ignores the field). is leaving this to Atlas deliberate, like the minimum-value decision tracked in CLOUDP-443184, or worth a plan-time check alongside the topology validation here?

internal/service/advancedcluster/resource.go
```diff
@@ -80,12 +80,22 @@ type rs struct {
...
+	var plan TFModel
 	diags := &resp.Diagnostics
 	diags.Append(req.Plan.Get(ctx, &plan)...)
+	validateInfiniteClusterType(diags, plan.ClusterType.ValueString(), plan.DatabaseEdition.ValueString(), "")
+	if diags.HasError() || req.State.Raw.IsNull() || req.Plan.Raw.IsFullyKnown() { // Return early unless it is an Update.
+		return
+	}
+	// Wait for configured expressions to resolve before copying state into unknown plan values.
+	// Otherwise an old value can conflict with the configuration resolved during apply.
+	if !req.Config.Raw.IsFullyKnown() {
```
EspenAlbert: q: Is this a bug fix?

internal/service/advancedcluster/resource.go
```diff
@@ -149,6 +160,23 @@ func (r *rs) Create(ctx context.Context, req resource.CreateRequest, resp *resou
 		return
 	}
 	clusterResp := createCluster(ctx, diags, r.Client, latestReq, waitParams)
+	if diags.HasError() {
+		return
+	}
+	var editionDiags diag.Diagnostics
+	validateInfiniteClusterType(&editionDiags, clusterResp.GetClusterType(), clusterResp.GetDatabaseEdition(), clusterResp.GetEffectiveDatabaseEdition())
```
EspenAlbert: why do we need this validation again if it is already done in the plan modifier?

internal/service/advancedcluster/resource.go
```diff
@@ -233,43 +261,52 @@ func (r *rs) Update(ctx context.Context, req resource.UpdateRequest, resp *resou
 	if diags.HasError() {
 		return
 	}
+	validateInfiniteClusterType(diags, plan.ClusterType.ValueString(), plan.DatabaseEdition.ValueString(), "")
```
EspenAlbert: why doing this twice?

internal/service/advancedcluster/resource_database_edition.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func validateInfiniteClusterType(diags *diag.Diagnostics, clusterType, databaseEdition, effectiveDatabaseEdition string) {
+	if isShardedClusterType(clusterType) && resolveDatabaseEdition(databaseEdition, effectiveDatabaseEdition) == "INFINITE" {
+		diags.AddAttributeError(path.Root("cluster_type"), "Unsupported INFINITE cluster type",
+			fmt.Sprintf("INFINITE clusters with cluster_type = %q are not supported by this version of the MongoDB Atlas Terraform provider. Support for SHARDED and GEOSHARDED requires a newer provider version that explicitly enables these topologies once available.", clusterType))
+	}
+}
+
+func (r *rs) prepareUpdateDatabaseEdition(ctx context.Context, diags *diag.Diagnostics, state, plan *TFModel, patch *admin.ClusterDescription20240805) {
```
EspenAlbert, line 20: maybe worth a docstring why this is needed?

internal/service/advancedcluster/resource_database_edition_test.go
```diff
@@ -0,0 +1,988 @@
+package advancedcluster_test
```
EspenAlbert, line 1: nice to have these tests split out!

EspenAlbert (commented):
Feels like a lot of changes for a few fields. Not sure I understand why we need so many calls to the validateInfiniteClustertype everywhere.

internal/service/advancedcluster/plan_modifier_test.go
```diff
@@ -80,3 +95,145 @@ func TestPlanChecksClusterTwoRepSpecsWithAutoScalingAndSpecs(t *testing.T) {
...
+			require.NoError(t, err)
+			result, err := server.PlanResourceChange(ctx, &tfprotov6.PlanResourceChangeRequest{
+				TypeName: "mongodbatlas_advanced_cluster", PriorState: dynamic(prior),
+				ProposedNewState: dynamic(proposed), Config: dynamic(model("M20", majorVersion)),
+			})
+			require.NoError(t, err)
+			require.Empty(t, result.Diagnostics)
+			plan, err := result.PlannedState.Unmarshal(typ)
+			require.NoError(t, err)
+			for _, attribute := range computedAttributes {
```
EspenAlbert: the `unknown` subtest asserts the computed attributes keep their state values but never asserts that `mongo_db_major_version` itself stays unknown in the planned state. I verified by making `unknownInConfig` return nil: this test still passes, so the top-level `keepUnknown` wiring has no coverage; only the nested `replication_specs` case is pinned, by `TestPlanAutoScalingStorageConfigUnknown`. a silent regression would copy state into unresolved top-level config values, the exact inconsistent-plan bug this commit fixes. adding a `require.False(...IsKnown())` check on `mongo_db_major_version` in the `unknown` subtest pins it; I confirmed that assertion passes on head and fails with `unknownInConfig` disabled

internal/service/advancedcluster/resource.go
lantoli: yes, and it's pre-existing on master so moving fix to #4718

internal/service/advancedcluster/resource.go
lantoli: i've made multiple changes to the tests since the code this review refers to. I'll make sure the full acceptance suite passes before merging

internal/service/advancedcluster/resource_database_edition.go
lantoli: it's deliberate: we don't normally duplicate Atlas's validation in the provider. Atlas rejects `storageConfig` on a CORE cluster with a clear apply-time error, and the edition isn't reliably known at plan time after an import, so a plan-time check could false-reject. The INFINITE-only scope is documented on the `storage_config` attribute in the resource docs.

The SHARDED/GEOSHARDED rejection is a different case: it isn't there to catch a user error early, it's to protect the implementation from a topology Atlas doesn't support yet. Once Atlas adds SHARDED/GEOSHARDED, that guard comes out. `storage_config` on CORE is already a defined Atlas error, not unsupported behavior we'd have to unscramble later.

internal/service/advancedcluster/resource.go
lantoli: that validation is gone now. The code was simplified since this review: resource_database_edition.go was removed and the edition resolution moved inline, using database_edition. oreffective_database_edition when appropriate.

internal/service/advancedcluster/resource_database_edition.go
lantoli: code removed and logic simplified

internal/service/advancedcluster/resource.go
lantoli: That double call is gone. validateInfiniteClusterType was removed entirely when the INFINITE topology rejection moved out of this PR, so Update no longer calls it.

internal/service/advancedcluster/resource.go
EspenAlbert: 🤖 consider skipping, looks like overall quite good coverage:

The current head removes both acceptance steps that cleared `storage_config` while compute remained configured. `infiniteShardSizeLimit` now clears only after compute is disabled and the whole `auto_scaling` block is omitted; `infiniteComputeAutoScaling` adds storage but never clears it. the mock request test still checks the intended payload, but it cannot establish that Atlas accepts the forced PATCH. please retain or add a focused create-with-compute-and-storage then clear-storage lifecycle with sufficient settling, so we can verify the prior 500s were timing rather than this payload

internal/service/advancedcluster/resource_storage_test.go
```diff
@@ -0,0 +1,512 @@
...
+			},
+			"auto scaling": {
+				config: tftypes.UnknownValue, unknownPath: autoScalingPath,
+			},
+		} {
+			t.Run(fmt.Sprintf("prior_storage=%t/%s", priorStorage, name), func(t *testing.T) {
+				server, err := acc.TestAccProviderV6Factories["mongodbatlas"]()
+				require.NoError(t, err)
+				planValue := func(config *tfprotov6.DynamicValue, path *tftypes.AttributePath) tftypes.Value {
+					result, err := server.PlanResourceChange(ctx, &tfprotov6.PlanResourceChangeRequest{
```
EspenAlbert, line 98: this looks like quite a new style? Would appreciate some more context or contributing guidance on when this is useful.
On a first look it looks very intimidating, with a double for loop and many different function calls, also tfprotov6 calls.
Would consider trying to stick to more config friendly options

internal/service/advancedcluster/resource_storage_test.go
```diff
@@ -0,0 +1,512 @@
...
+				initial := planValue(model(tc.config), tc.unknownPath)
+				require.False(t, initial.IsKnown(), "configured unknown storage values must not be replaced by prior state")
+				resolved := planValue(model(settings(map[string]any{"shard_size_limit_gb": int64(2048)})), limitPath)
+				require.Equal(t, tftypes.NewValue(tftypes.Number, int64(2048)), resolved, "the plan must accept the value resolved during apply")
+			})
+		}
+	}
+}
+
+func TestAutoScalingStorageConfigUnknownLifecycle(t *testing.T) {
```
EspenAlbert, line 118: nice, this looks more readable to me

internal/service/advancedcluster/plan_modifier.go
```diff
@@ -287,3 +291,31 @@ func minLen[T any](a, b []T) int {
 	}
 	return lb
 }
+
+// clearRemovedStorageConfig prevents the computed parent from retaining a configured limit after removal.
+type clearRemovedStorageConfig struct{}
```
EspenAlbert, line 296: nice!

internal/service/advancedcluster/model_to_ClusterDescription20240805.go
```diff
@@ -194,20 +200,42 @@ func newRegionConfig(ctx context.Context, input types.List, diags *diag.Diagnost
...
 
 func newAdvancedAutoScalingSettings(ctx context.Context, input types.Object, diags *diag.Diagnostics) *admin.AdvancedAutoScalingSettings {
-	var resp *admin.AdvancedAutoScalingSettings
 	if input.IsUnknown() || input.IsNull() {
-		return resp
+		return nil
+	}
+	result := new(admin.AdvancedAutoScalingSettings)
+	attributes := input.Attributes()
+	if storageConfig, ok := attributes["storage_config"]; ok {
```
EspenAlbert: [nit] might be worth a comment on why we have to remove the storage config before converting to TFAutoScalingModel?

EspenAlbert (approved):
Good job on the refactoring! A lot of work went into this. Left a few comments, but nothing blocking :clap:

internal/service/advancedcluster/resource_storage_test.go
EspenAlbert: the protocol test deliberately isolates unknown child, `storage_config` object, and `auto_scaling` object shapes against prior state. `TestAutoScalingStorageConfigUnknownLifecycle` immediately below separately verifies a config-facing `terraform_data` expression through Terraform plan and apply. keeping the focused protocol test is appropriate because expressing each parent-object unknown independently in HCL would require several heavier fixtures; a short comment stating that split would make the intent clearer

internal/service/advancedcluster/resource.go
EspenAlbert: 402457552 adds exactly this: `infiniteShardSizeLimit` now configures compute plus storage right after create, clears storage while compute stays configured, and re-adds storage, all early in the chain where no rejected-PATCH settle race applies. that closes the coverage request; the remaining condition is the green full-suite acceptance run you already committed to, since this step now fails deterministically if Atlas rejects the forced-removal payload

EspenAlbert: f3028bd85 reverts the focused lifecycle in 402457552, so this request is active again: the current suite clears storage only when the whole `auto_scaling` block is absent, while the compute lifecycle never clears storage. The mock request test cannot establish whether Atlas accepts the forced PATCH retaining `autoScaling.compute`, which previously failed with HTTP 500. Please restore the focused lifecycle or document and test an alternative Atlas-backed path

internal/service/advancedcluster/resource_storage_test.go
EspenAlbert: the protocol test deliberately isolates unknown child, `storage_config` object, and `auto_scaling` object shapes against prior state. `TestAutoScalingStorageConfigUnknownLifecycle` immediately below separately verifies a config-facing `terraform_data` expression through Terraform plan and apply. keeping the focused protocol test is appropriate because expressing each parent-object unknown independently in HCL would require several heavier fixtures; a short comment stating that split would make the intent clearer

internal/service/advancedcluster/resource.go
lantoli: will do in CLOUDP-447210 after API is fixed

internal/service/advancedcluster/resource_storage_test.go
lantoli: fair, it's dense. it drives the real `PlanResourceChange` through the muxed server because the behavior under test is plan-time unknown handling; the value has to be genuinely unknown at plan time, which config-based tests can't express. 

the `clusterSchema`/`clusterDynamic`/`planTestValue` helpers keep the fixtures readable, same approach as `plan_modifier_test.go`.

internal/service/advancedcluster/model_to_ClusterDescription20240805.go
lantoli: there's a comment now on `newAutoScalingWithStorageConfig`: the object is narrowed to the base `autoScalingObjType` before decoding because `TFAutoScalingModel` has no storage field and the framework's `As` rejects unknown attributes.

PR #4699, opened 2026-09-02
feat: Add include_system_managed attribute in service account plural datasources

manupedrozo:
## Description

Updated autogen to set optional+computed query params to optional only, these attributes are never returned by the API so the spec defining a default should not promote them to computed.
Regenerated service account resources to add `include_system_managed`.
No other autogen resource is impacted by the change.

Keeping this change separated from the MCP Config dev branch to avoid mixing codegen changes with new MCP resources.

Testing: 
* Will add/update acc tests in the MCP Config dev branch - We need an MCP config to verify its service account is returned by the SA plural DS when `include_system_managed = true`.
* Manually verified that the query param works as expected for both datasources.

Link to any related issue(s): CLOUDP-433211

## Type of change:

## Required Checklist:

### DIFF
--- a/internal/serviceapi/serviceaccount/plural_data_source_schema.go
+++ b/internal/serviceapi/serviceaccount/plural_data_source_schema.go
@@ -14,6 +14,11 @@ import (
 func PluralDataSourceSchema(ctx context.Context) dsschema.Schema {
 	return dsschema.Schema{
 		Attributes: map[string]dsschema.Attribute{
+			"include_system_managed": dsschema.BoolAttribute{
+				Computed:            true,
+				Optional:            true,
+				MarkdownDescription: "Flag that indicates whether system-managed Service Accounts (such as those used for MCP ingress/egress integrations) are included in the response. When false, only user-managed Service Accounts are returned.",
+			},
 			"org_id": dsschema.StringAttribute{
 				Required:            true,
 				MarkdownDescription: "Unique 24-hexadecimal digit string that identifies the organization that contains your projects.",
@@ -83,8 +88,9 @@ func PluralDataSourceSchema(ctx context.Context) dsschema.Schema {
 }
 
 type TFPluralDSModel struct {
-	OrgId   types.String                                        `tfsdk:"org_id" autogen:"omitjson"`
-	Results customtypes.NestedListValue[TFPluralDSResultsModel] `tfsdk:"results" autogen:"omitjson"`
+	IncludeSystemManaged types.Bool                                          `tfsdk:"include_system_managed" autogen:"omitjson"`
+	OrgId                types.String                                        `tfsdk:"org_id" autogen:"omitjson"`
+	Results              customtypes.NestedListValue[TFPluralDSResultsModel] `tfsdk:"results" autogen:"omitjson"`
 }
 type TFPluralDSResultsModel struct {
 	ClientId    types.String                                               `tfsdk:"client_id" autogen:"omitjson"`
--- a/internal/serviceapi/serviceaccount/plural_data_source_schema.go
+++ b/internal/serviceapi/serviceaccount/plural_data_source_schema.go
@@ -15,7 +15,6 @@ func PluralDataSourceSchema(ctx context.Context) dsschema.Schema {
 	return dsschema.Schema{
 		Attributes: map[string]dsschema.Attribute{
 			"include_system_managed": dsschema.BoolAttribute{
-				Computed:            true,
 				Optional:            true,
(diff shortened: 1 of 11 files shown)


tools/codegen/gofilegen/datasource/data_source_file.go
```diff
@@ -79,16 +79,15 @@ func pluralizeName(name string) string {
...
 func getQueryParams(attributes codespec.Attributes) []codetemplate.Param {
 	var queryParams []codetemplate.Param
 
 	for i := range attributes {
-		// Only consider optional attributes as query parameters
-		if attributes[i].ComputedOptionalRequired == codespec.Optional {
+		if attributes[i].ComputedOptionalRequired == codespec.Optional || attributes[i].ComputedOptionalRequired == codespec.ComputedOptional {
 			param := codetemplate.Param{
 				PascalCaseName: stringcase.Capitalize(attributes[i].TFModelName),
-				CamelCaseName:  stringcase.Uncapitalize(attributes[i].APIName),
```
EspenAlbert, line 91: why remove this?

EspenAlbert (approved):

manupedrozo:
I think we can actually make this new attr computed only. Will verify before merging.

tools/codegen/gofilegen/datasource/data_source_file.go
manupedrozo: The query param must match the API name for the attribute, we shouldn't modify it here. 
Was not breaking anything since the norm is for API names to start with a lowercase, just preventing a potential future issue.

internal/serviceapi/serviceaccount/plural_data_source_schema.go
```diff
@@ -14,6 +14,11 @@ import (
 func PluralDataSourceSchema(ctx context.Context) dsschema.Schema {
 	return dsschema.Schema{
 		Attributes: map[string]dsschema.Attribute{
+			"include_system_managed": dsschema.BoolAttribute{
+				Computed:            true,
+				Optional:            true,
+				MarkdownDescription: "Flag that indicates whether system-managed Service Accounts (such as those used for MCP ingress/egress integrations) are included in the response. When false, only user-managed Service Accounts are returned.",
+			},
```
AgustinBettati, line 20: Would expect this attribute to be optional only right? considering it is not present in the API response.

manupedrozo: Yup, same https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4699#issuecomment-5508033180

Taking a different approach for codegen to adjust these to optional so we don't need to override manually moving forward.

tools/codegen/models/service_account.yaml
```diff
@@ -337,8 +337,20 @@ data_sources:
               present_in_any_response: false
               request_only_required_on_create: false
     plural:
-        description: Returns all Service Accounts for the specified Organization.
+        description: Returns all Service Accounts for the specified Organization. By default, system-managed Service Accounts are excluded. Set `includeSystemManaged=true` to include them.
```
AgustinBettati, line 340: q: How does a user who sets include_system_managed = true know which entries are system-managed? I had the vague memory each result had a boolean flag `is_system_managed` but that does not seem to be the case.

manupedrozo: Yes! We touched on this with upstream, there is no public field on the SA yet. Lets follow up offline.

internal/serviceapi/serviceaccount/plural_data_source_schema.go
manupedrozo: See [5b3d8e3](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4699/commits/5b3d8e32473960606b1b5df319539f1ab5dc3e76)

tools/codegen/codespec/api_to_provider_spec_mapper.go
```diff
@@ -323,6 +323,14 @@ func queryParamsToAttributes(op *high.Operation) (Attributes, error) {
 			log.Printf("[WARN] Query param %s could not be mapped: %s", paramName, err)
 			continue
 		}
+
+		// Query params are never returned in API responses.
+		// When the spec defines a default, these are promoted to ComputedOptional.
+		// Roll them back to Optional only.
+		if parameterAttribute.ComputedOptionalRequired == ComputedOptional {
```
AgustinBettati, line 330: nice, agree this should be during the model generation phase

AgustinBettati (approved):
LGTM

antellezr-mdb (approved):

PR #4705, opened 2026-09-06
Treat missing Atlas database user as successful delete

### LINKED ISSUE #4656: [Bug]: mongodbatlas_database_user Delete fails on USER_NOT_FOUND

- [x] I have searched the existing issues

### Provider Version

v1.30.0

### Terraform Version

pulumi-mongodbatlas 4.13.0, which vendors provider v1.30.0

### Terraform Edition

Terraform Open Source (OSS)

### Current Behavior

This is the Delete-side counterpart of #3069, which fixed `Read()` for users that have already been deleted on Atlas's side but remain present in Terraform state. `Delete()` hard-fails in that same situation.

When state records a `mongodbatlas_database_user` that no longer exists in Atlas (e.g., vanishing after a CI deployment was stopped halfway through), the next destroy or replace calls the delete API, receives `404 USER_NOT_FOUND`, and the provider aborts the entire run. Since the purpose of a delete is to ensure the user does not exist, a 404 in this path indicates that the desired outcome has been reached rather than a failure that should stop the apply.

`Delete()` currently discards the HTTP response, so a 404 is indistinguishable from any other error:

```go
// internal/service/databaseuser/resource_database_user.go
_, err := connV2.DatabaseUsersAPI.DeleteDatabaseUser(
	ctx,
	state.ProjectID.ValueString(),
	state.AuthDatabaseName.ValueString(),
	state.Username.ValueString()).Execute()
if err != nil {
	resp.Diagnostics.AddError("error when destroying the database user resource", err.Error())
	return
}
```

Capturing the response and treating `validate.StatusNotFound` as a successful delete, as `teamprojectassignment` does in its `Delete`, would address this.

I believe this aligns with [HashiCorp's framework guidance](https://developer.hashicorp.com/terraform/plugin/framework/resources/delete#recommendations) and the behavior several sibling resources already implement in `Delete` (e.g., `teamprojectassignment` and `projectipaccesslist`).

```hcl
resource "mongodbatlas_database_user" "oidc_user" {
  project_id         = var.project_id
  username           = "${var.tenant_id}/my-workload-identity"
  auth_database_name = "$external"
  oidc_auth_type     = "USER"

  roles {
    role_name     = "readWrite"
    database_name = var.database_name
  }
}
```

### Steps To Reproduce

1. Apply the config above so the user exists in both Atlas and Terraform state.
2. Delete the user directly in the Atlas UI (or interrupt an apply after it deleted the user but before it saved state).
3. Run terraform destroy, or any apply that plans a delete/replace of that user, without refreshing first.
4. The run fails immediately with USER_NOT_FOUND instead of completing.

### Logs

```txt

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Hashim1999164:
## Summary
When destroy calls Delete for mongodbatlas_database_user and Atlas returns 404 USER_NOT_FOUND, treat that as success so apply can finish. This matches Read behavior and sibling resources such as teamprojectassignment.

Closes #4656

## Test plan
1. Unit tests for the databaseuser package passed locally
2. Reviewers can confirm Delete ignores StatusNotFound the same way as teamprojectassignment

### DIFF
--- a/internal/service/databaseuser/resource_database_user.go
+++ b/internal/service/databaseuser/resource_database_user.go
@@ -364,12 +364,16 @@ func (r *databaseUserRS) Delete(ctx context.Context, req resource.DeleteRequest,
 	}
 
 	connV2 := r.Client.AtlasV2
-	_, err := connV2.DatabaseUsersAPI.DeleteDatabaseUser(
+	httpResponse, err := connV2.DatabaseUsersAPI.DeleteDatabaseUser(
 		ctx,
 		state.ProjectID.ValueString(),
 		state.AuthDatabaseName.ValueString(),
 		state.Username.ValueString()).Execute()
 	if err != nil {
+		// Already deleted on the Atlas side: desired end state is reached.
+		if validate.StatusNotFound(httpResponse) {
+			return
+		}
 		resp.Diagnostics.AddError("error when destroying the database user resource", err.Error())
 		return
 	}


marcabreracast:
Hi @Hashim1999164,
Thank you for reaching out with this PR! As my colleague @AgustinBettati mentioned in the [original issue](https://github.com/mongodb/terraform-provider-mongodbatlas/issues/4656#issuecomment-5369938259), we're currently discussing a provider-wide convention approach as this inconsistency behaviour affects several resources. 

I f we go with your proposed approach, following our Contributing Guidelines in regards to [Merging a Pull Request](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/development-setup.md#merging-a-pull-request), we will create a branch from this PR and cherry pick your contribution. 

I will keep this PR open for now so I can inform you when there's any relevant updates regarding this topic. Thank you for your patience!

Hashim1999164:
@marcabreracast yeah that makes sense if a bunch of resources do the same thing. no rush, cherry pick is fine whenever you land on the convention.

PR #4709, opened 2026-09-07
feat: Add `mongodbatlas_org_delegation_settings` resource

oarbusi:
## Description

Adds the autogenerated `mongodbatlas_org_delegation_settings` resource and data source, a singleton organization-level resource that manages delegation settings (GET/PATCH on `/api/atlas/v2/orgs/{orgId}/delegationSettings`, API version 2025-03-12). It exposes 4 attributes: `delegated_mcp_access`, `delegated_partner_access`, `idle_refresh_token_lifetime`, and `maximum_refresh_token_lifetime`.

The access policies are optional+computed since Atlas returns their default (`DISALLOWED`) when omitted. The token lifetimes are optional-only and use `send_null_as_null_on_update`, so removing them from the configuration sends null and resets them to the system default, as documented by the API.

The API has no DELETE endpoint (settings always exist), so the delete operation is a no-op that only emits a warning and performs no API calls; Terraform only removes the resource from the state. This matches the `project_settings_api` singleton pattern.

The data source is backed by the same GET endpoint and mirrors the resource attributes.

Link to any related issue(s): CLOUDP-442139

## Type of change:

## Required Checklist:

## Further comments

- The optional+computed overrides on the two access policies are a stopgap: ideally the spec annotates them as server-computed-when-client-omitted so the overrides can be dropped on a future regeneration.


tools/codegen/config.yml
```diff
@@ -1836,3 +1836,32 @@ resources:
         overrides:
           project_id:
             description: *project_id_description
+
+  org_delegation_settings:
```
manupedrozo, line 1840: Any concerns adding the data source (now or in a follow-up PR)? 
Saw the PR description, I believe `project_settings_api` was implemented before DSs where supported in autogen so wouldn't necessarily follow that one as the example. `ai_model_rate_limit` is an example autogen singleton with a datasource.

tools/codegen/atlasapispec/multi-version-api-spec.flattened.yml
manupedrozo, line 1: Can we skip the api spec changes? I think there are no necessary additions (and models are enough for the CI checks).

examples/mongodbatlas_org_delegation_settings/main.tf
```diff
@@ -0,0 +1,27 @@
+resource "mongodbatlas_org_delegation_settings" "this" {
+  org_id                         = var.atlas_org_id
+  delegated_mcp_access           = "READ_ONLY"
+  delegated_partner_access       = "DISALLOWED"
+  idle_refresh_token_lifetime    = 3600
+  maximum_refresh_token_lifetime = 86400
+}
+
+output "org_delegation_settings_delegated_mcp_access" {
```
manupedrozo, line 13: Not sure we need the outputs, no strong opinion though.

internal/serviceapi/orgdelegationsettings/resource_test.go
```diff
@@ -0,0 +1,205 @@
...
+	return func(s *terraform.State) (string, error) {
+		rs, ok := s.RootModule().Resources[resourceName]
+		if !ok {
+			return "", fmt.Errorf("not found: %s", resourceName)
+		}
+		return rs.Primary.Attributes["org_id"], nil
+	}
+}
+
+type delegationSettings struct {
```
manupedrozo, line 177: Untyped client really makes these more verbose. 
No blocker for this one, may need to revisit our decision to use untyped client in tests.

internal/serviceapi/orgdelegationsettings/resource_test.go
```diff
@@ -0,0 +1,205 @@
...
+	// captured before the test and restored afterwards to leave the org unchanged.
+	baseline, err := getDelegationSettings(orgID)
+	if err != nil {
+		t.Fatalf("failed to capture current delegation settings for org %s: %v", orgID, err)
+	}
+	t.Cleanup(func() {
+		if err := updateDelegationSettings(orgID, baseline); err != nil {
+			t.Errorf("failed to restore delegation settings for org %s: %v", orgID, err)
+		}
+	})
```
manupedrozo, line 39: Nice! I think we'll need to figure out another approach for these org tests moving forward (e.g. create a new org).

manupedrozo (approved):
Thanks for the quick implementation! Left some comments, nothing blocking.

internal/serviceapi/orgdelegationsettings/resource_schema.go
```diff
@@ -0,0 +1,50 @@
...
+				MarkdownDescription: "Policy that controls whether partner delegated access is permitted within this organization. Possible values are `DISALLOWED` and `READ_WRITE`. Defaults to `DISALLOWED`.",
+			},
+			"idle_refresh_token_lifetime": schema.Int64Attribute{
+				Optional:            true,
+				MarkdownDescription: "Maximum number of seconds a refresh token may be idle before it expires. Omit to leave unchanged; set to null to reset to the system default. Must be between 1 and 31536000 (1 year) when provided.",
+			},
+			"maximum_refresh_token_lifetime": schema.Int64Attribute{
+				Optional:            true,
+				MarkdownDescription: "Maximum lifetime of a refresh token in seconds, regardless of activity. Omit to leave unchanged; set to null to reset to the system default. Must be between 1 and 31536000 (1 year) when provided.",
+			},
```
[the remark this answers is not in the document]
manupedrozo: Worth taking a look at this one.

internal/serviceapi/orgdelegationsettings/resource_test.go
oarbusi: Agreed, it is verbose. Followed the current autogen test convention here; happy to switch to a typed SDK helper or a shared acc helper if we revisit that decision.

tools/codegen/atlasapispec/multi-version-api-spec.flattened.yml
oarbusi: Done, reverted both spec files to master. You were right: the delegationSettings path and both schemas were already in the spec, so the refresh added no necessary changes.

internal/serviceapi/orgdelegationsettings/resource_test.go
oarbusi: Thanks! Agreed, creating a fresh org per test would remove the baseline capture/restore dance and let these run in parallel.

tools/codegen/config.yml
oarbusi: Good point about `project_settings_api` predating data source support in autogen. No concerns, added the data source in this PR (same GET endpoint): `datasources.read` in config.yml, provider registration, data source checks in the existing acceptance test, docs template, example usage, and a `new-datasource` changelog block.

examples/mongodbatlas_org_delegation_settings/main.tf
oarbusi: Keeping them: they surface in the generated docs page and show what the resource exposes, same pattern as the `ai_model_rate_limit` example. Also added one output sourced from the new data source.

internal/serviceapi/orgdelegationsettings/resource_schema.go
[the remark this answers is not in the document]
oarbusi: Looked into this one. The suggested alternative (Optional+Computed, with users setting the attribute explicitly to null to reset) is not implementable in Terraform: `attr = null` is identical to omitting the attribute, and with Computed a null config falls back to the prior state value, so null would never reach the API. Optional-only + `send_null_as_null_on_update` is the only way to offer the documented reset-to-default behavior.

The perpetual-diff concern would only materialize if Atlas returned a concrete default for unset lifetimes. The acceptance test verifies the opposite: the reset step asserts both attributes disappear from state after sending null, and the step requires an empty post-apply plan (green in CI). If Atlas ever changes that behavior, the daily test suite will fail and we can revisit.

examples/mongodbatlas_org_delegation_settings/variables.tf
```diff
@@ -0,0 +1,17 @@
...
+  type        = string
+  default     = ""
+}
+
+variable "atlas_client_secret" {
+  description = "MongoDB Atlas Service Account Client Secret"
+  type        = string
+  sensitive   = true
+  default     = ""
+}
```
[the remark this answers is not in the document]
oarbusi: This matches the current repo convention: 43 example provider configurations feed client_id/client_secret via variables (including the most recently added examples, e.g. `mongodbatlas_metric_integration`), while only 2 use an empty provider block. Keeping as is.

examples/mongodbatlas_org_delegation_settings/README.md
```diff
@@ -0,0 +1,18 @@
...
+This example shows how to configure the delegation settings of a MongoDB Atlas organization with the `mongodbatlas_org_delegation_settings` resource. Delegation settings control how MCP (Model Context Protocol) and partner delegated access are permitted within the organization, as well as the refresh token lifetimes.
+
+## Important Notes
+
+Delegation settings are a singleton at the organization level: the settings always exist, so creating the resource updates the existing settings. The delete operation is a no-op that only emits a warning and performs no API calls, so the settings in Atlas are left unchanged; Terraform only removes the resource from the state.
+
+## Variables Required to be set:
+- `atlas_client_id`: MongoDB Atlas Service Account Client ID
+- `atlas_client_secret`: MongoDB Atlas Service Account Client Secret
+- `atlas_org_id`: Organization ID where the Delegation Settings will be configured
```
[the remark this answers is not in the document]
oarbusi: This matches the current repo convention: 43 example provider configurations feed client_id/client_secret via variables (including the most recently added examples, e.g. `mongodbatlas_metric_integration`), while only 2 use an empty provider block. Keeping as is.

internal/serviceapi/orgdelegationsettings/resource_test.go
```diff
@@ -0,0 +1,205 @@
...
+func TestAccOrgDelegationSettings_basic(t *testing.T) {
+	acc.SkipInUnitTest(t) // baseline capture below runs before resource.Test skips
+	orgID := os.Getenv("MONGODB_ATLAS_ORG_ID")
+
+	// Delete is a no-op for this singleton resource, so the current settings are
+	// captured before the test and restored afterwards to leave the org unchanged.
+	baseline, err := getDelegationSettings(orgID)
+	if err != nil {
+		t.Fatalf("failed to capture current delegation settings for org %s: %v", orgID, err)
+	}
```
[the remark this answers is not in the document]
oarbusi: Good catch, fixed: `acc.PreCheckBasic(t)` now runs right after the unit-test skip, before the baseline capture, so missing credentials or org id fail fast with the intended message.

examples/mongodbatlas_org_delegation_settings/provider.tf
```diff
@@ -0,0 +1,4 @@
+provider "mongodbatlas" {
+  client_id     = var.atlas_client_id
+  client_secret = var.atlas_client_secret
+}
```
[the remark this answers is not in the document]
oarbusi: This matches the current repo convention: 43 example provider configurations feed client_id/client_secret via variables (including the most recently added examples, e.g. `mongodbatlas_metric_integration`), while only 2 use an empty provider block. Keeping as is.

docs/resources/org_delegation_settings.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Optional
+
+- `delegated_mcp_access` (String) Policy that controls how MCP (Model Context Protocol) delegated access is permitted within this organization. Possible values are `DISALLOWED`, `READ_ONLY`, and `READ_WRITE`. Defaults to `DISALLOWED`.
+- `delegated_partner_access` (String) Policy that controls whether partner delegated access is permitted within this organization. Possible values are `DISALLOWED` and `READ_WRITE`. Defaults to `DISALLOWED`.
+- `idle_refresh_token_lifetime` (Number) Maximum number of seconds a refresh token may be idle before it expires. Omit to leave unchanged; set to null to reset to the system default. Must be between 1 and 31536000 (1 year) when provided.
+- `maximum_refresh_token_lifetime` (Number) Maximum lifetime of a refresh token in seconds, regardless of activity. Omit to leave unchanged; set to null to reset to the system default. Must be between 1 and 31536000 (1 year) when provided.
+
+## Import
+Import the Delegation Settings resource by using the Organization ID, e.g.
```
antellezr-mdb: Consider replacing "e.g." with "for example". MongoDB Style guide suggests to use  English alternatives.
https://www.mongodb.com/docs/meta/style-guide/terminology/general-term-guidelines/avoid-obscure-words/#std-label-avoid-obscure-words

antellezr-mdb (approved):

docs/resources/org_delegation_settings.md
oarbusi: Done, replaced with "For example:" in the docs template and regenerated the doc.

internal/serviceapi/orgdelegationsettings/resource_test.go
AgustinBettati: Would keep in mind it is not common we call update operations with SDK during our acceptance test logic, a simple GET for checkDestroy is more simple

PR #4710, opened 2026-09-07
refactor: Adds SDKv2 raw config helpers

EspenAlbert:
## Description

Adds `internal/common/sdkv2config` so SDKv2 resources can read HCL from `GetRawConfig` as TPF-like values (null, unknown, or set). `Get` / `GetOk` / `GetOkExists` cannot tell an unset attribute from `""`, `false`, or `0`, and Optional+Computed `Get()` still returns leftover state after the user drops the attribute from HCL.

Call sites no longer call `GetRawConfig` themselves. Helpers return typed wrappers with `IsNull()`, `IsUnknown()`, and `HasChange()`. Callers compose those; there is no `IsNullUnchanged`.

Stacked on [#4693](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4693). Switches that PR's backup-schedule raw-config reads, plus `operations_contact` on `mongodbatlas_organization` from [#4700](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4700) after merging `master`.

Jira: [CLOUDP-388562](https://jira.mongodb.org/browse/CLOUDP-388562)

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue). Please, add the "bug" label to the PR.
- [ ] New feature (non-breaking change which adds functionality). Please, add the "enhancement" label to the PR. A migration guide must be created or updated if the new feature will go in a major version.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected). Please, add the "breaking change" label to the PR. A migration guide must be created or updated.
- [x] This change requires a documentation update
- [ ] Documentation fix/enhancement

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have read the [contributing guides](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/README.md)
- [x] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
- [x] I have added tests that prove my fix is effective or that my feature works per HashiCorp requirements
- [x] I have added any necessary documentation (if appropriate)
- [ ] I have run make fix and verified my code
- [ ] If changes include deprecations or removals I have added appropriate changelog entries.
- [ ] If changes include removal or addition of 3rd party GitHub actions, I updated our internal document. Reach out to the APIx Integration slack channel to get access to the internal document.

## Further comments

[...]

PR #4713, opened 2026-09-07
feat: Add custom session timeouts to mongodbatlas_organization

marcabreracast:
## Description

Adds `custom_session_timeouts` to the `mongodbatlas_organization` resource and its data sources, allowing organization owners to manage absolute and idle session timeouts for the Atlas UI at the organization level, including clearing them back to the server defaults.

Related issue: https://jira.mongodb.org/browse/CLOUDP-294579

## Type of change:

## Required Checklist:

## Local verification

Resource acceptance tests run locally against cloud-dev with a paying org:

```
=== RUN   TestAccConfigRSOrganization_Settings
--- PASS: TestAccConfigRSOrganization_Settings (132.09s)
=== RUN   TestAccConfigRSOrganization_Settings_OperationsContact
--- PASS: TestAccConfigRSOrganization_Settings_OperationsContact (197.46s)
=== RUN   TestAccConfigRSOrganization_Settings_Timeouts
--- PASS: TestAccConfigRSOrganization_Settings_Timeouts (388.48s)
PASS
ok  	github.com/mongodb/terraform-provider-mongodbatlas/internal/service/organization
```


antellezr-mdb (approved):

examples/mongodbatlas_organization/organization-step-1/variables.tf
```diff
@@ -22,6 +22,14 @@ variable "operations_contact" {
   description = "Email address, typically a distribution list, for the organization to receive proactive notifications about its infrastructure"
   default     = null
 }
+variable "custom_session_timeouts" {
+  type = list(object({
+    absolute_session_timeout_in_seconds = number
+    idle_session_timeout_in_seconds     = number
+  }))
```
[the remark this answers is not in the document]
marcabreracast: Addressed in [807a5fe](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/807a5fe07bcb8703a63fd7c4e59e2e6559fa2240)

marcabreracast:
FYI - Coordination with [#4710](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4710): that PR refactors the same `newOrganizationSettings` raw-config reads into shared `sdkv2config` helpers, and its `NestedCollectionLen` covers exactly the pattern `custom_session_timeouts uses`. We can wait until the PR is merged in order to rebase and benefit from those changes

internal/service/organization/resource_organization.go
```diff
@@ -392,9 +416,69 @@ func newOrganizationSettings(d *schema.ResourceData) *admin.OrganizationSettings
...
+		obj["absolute_session_timeout_in_seconds"] = v
+	}
+	if idleOK {
+		obj["idle_session_timeout_in_seconds"] = w
+	}
+	return []any{obj}
+}
+
+// sessionTimeoutDiffSuppress treats a state value of 0 as equal to an unset (null) config value for the
+// nested custom_session_timeouts children. SDKv2 materializes an omitted non-Computed TypeInt child as 0,
```
AgustinBettati: This the default behaviour of SDK v2 (not defined in config -> zero value in state). 
Curious why a custom diff suppress is needed here? I would expect we have other SDK v2 resources with optional-only attributes we can use as reference.

docs/resources/organization.md
```diff
@@ -73,12 +73,18 @@ output "service_account_secret" {
...
 
+### Custom Session Timeouts
+
+* `absolute_session_timeout_in_seconds` - (Optional) Specifies the absolute session timeout duration in seconds. When set to `null`, the field's value is unset, and the default value of 43,200 seconds (12 hours) is applied. Accepted values range between a minimum of 3,600 seconds (1 hour) and a maximum of 43,200 seconds (12 hours).
+* `idle_session_timeout_in_seconds` - (Optional) Specifies the idle session timeout duration in seconds. When set to `null`, the field's value is unset, and the default behavior depends on the context: no timeout for Atlas Commercial, and 600 seconds (10 minutes) for Atlas for Government. Accepted values start at a minimum of 300 seconds (5 minutes). For Atlas Commercial, the maximum value cannot exceed the configured absolute session timeout. For Atlas for Government, the maximum value is capped at 600 seconds (10 minutes).
```
AgustinBettati, line 86: `600 seconds (10 minutes) for Atlas for Government`
Would this imply that for gov we would face a non-empty plan if the attribute is left emtpy in the config? (considering the attribute is currently optional-only)

internal/service/organization/resource_organization_test.go
```diff
@@ -132,12 +140,18 @@ func TestAccConfigRSOrganization_Settings(t *testing.T) {
...
+				Check: checkAggr(orgOwnerID, name, description, settingsConfig,
+					resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.0.absolute_session_timeout_in_seconds", "3600"),
+					resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.0.idle_session_timeout_in_seconds", "300")),
 			},
 			{
 				PreConfig: sleepForSettingsRateLimit,
-				Config:    configWithSettings(orgOwnerID, name, description, roleName, withOperationsContact(settingsConfig, "test-updated@mongodb.com")),
-				Check:     checkAggr(orgOwnerID, name, description, withOperationsContact(settingsConfig, "test-updated@mongodb.com")),
+				Config:    configWithSettings(orgOwnerID, name, description, roleName, withOperationsContact(settingsConfigUpdated, "test-updated@mongodb.com")),
```
AgustinBettati: Can be worth testing the use-case of removing only one child attribute while leaving the other one present

internal/service/organization/resource_organization.go
marcabreracast: Yep you're right. Suppressed this in [b8585a8](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/b8585a8d41875b7109296bec31ac6ccd3463bd04)

internal/service/organization/resource_organization_test.go
marcabreracast: Sure! Addressed in [7fd3b82](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/7fd3b826d9039213d9b793793f31313a8ed88005)

docs/resources/organization.md
marcabreracast: That's a good point, let me chat to the upstream team about this behaviour, due to our testing limitations here.

marcabreracast: Confirmed by upstream team that this does not affect the API: There is no difference for Gov vs. Commercial of the API response in this case, so if `idleSessionTimeoutInSeconds` is unset, it'll be `null` instead of 600. And if neither absolute nor idle timeout is set, the whole `customSessionTimeouts` will be `null`, so we won't be facing a non-empty plan.

marcabreracast:
Follow-up from [previous comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713#issuecomment-5581432730). Improvement on the work that https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4710 implements does not include nested blocks. Agreement to merge this work first and then create a follow-up PR that includes this improvement so this work is not blocked.

internal/service/organization/resource_organization.go
```diff
@@ -392,9 +416,69 @@ func newOrganizationSettings(d *schema.ResourceData) *admin.OrganizationSettings
...
+}
+
+func flattenCustomSessionTimeouts(cst *admin.CustomSessionTimeouts) []any {
+	if cst == nil {
+		return nil
+	}
+	obj := map[string]any{}
+	v, absOK := cst.GetAbsoluteSessionTimeoutInSecondsOk()
+	w, idleOK := cst.GetIdleSessionTimeoutInSecondsOk()
+	if !absOK && !idleOK {
```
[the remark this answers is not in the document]
EspenAlbert: 🤖 verified against head: `flattenCustomSessionTimeouts` returns nil when both children are unset (`internal/service/organization/resource_organization.go:460`), so state holds zero items while the config keeps one, and the count diff re-proposes the reset on every plan

two additional entry points into this form:
- `examples/mongodbatlas_organization/organization-step-1/main.tf:11` renders the block unconditionally and both variables default to `null`, so the example as shipped produces exactly the one-item all-null block
- the per-child `When set to null` reset wording in the docs steers a user resetting both fields into the same form, and there is no guidance for clearing the block as a whole

block removal, `= null`, and `= []` all converge; only the one-item all-null block does not, so the fix needs to either keep one state item for an all-null response or suppress the count diff for that shape

docs/data-sources/organization.md
```diff
@@ -34,9 +34,16 @@ In addition to all arguments above, the following attributes are exported:
 * `gen_ai_features_enabled` - (Optional) Flag that indicates whether this organization has access to generative AI features. This setting only applies to Atlas Commercial and defaults to `true`. With this setting on, Project Owners may be able to enable or disable individual AI features at the project level. To learn more, see https://www.mongodb.com/docs/generative-ai-faq/.
 * `security_contact` - (Optional) String that specifies a single email address for the specified organization to receive security-related notifications. Specifying a security contact does not grant them authorization or access to Atlas for security decisions or approvals.
 * `operations_contact` - (Optional) String that specifies a distribution list email address for the specified organization to receive proactive notifications about its infrastructure. The operations contact is used for notifications only and is not authorized to make decisions or approvals.
+* `custom_session_timeouts` - (Optional) Block that specifies the custom session timeout settings for the organization. See [Custom Session Timeouts](#custom-session-timeouts).
```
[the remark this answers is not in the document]
EspenAlbert: the `(Optional)` marker itself matches the sibling convention here: `gen_ai_features_enabled`, `security_contact`, and `operations_contact` are also Computed-only in the data-source schemas yet labeled `(Optional)` in these docs, so changing only this line would read as inconsistent

the part worth trimming is the copied field prose: `When set to null, the field's value is unset...` and `Accepted values range...` describe configuration semantics for a computed-only output, and that copied text is the actual input/output confusion in the data-source docs (same copy in `docs/data-sources/organizations.md`)

internal/service/organization/resource_organization_test.go
```diff
@@ -132,12 +140,26 @@ func TestAccConfigRSOrganization_Settings(t *testing.T) {
...
 				Config: configWithSettings(orgOwnerID, name, description, roleName, settingsConfig),
-				Check:  checkAggr(orgOwnerID, name, description, settingsConfig),
+				Check: checkAggr(orgOwnerID, name, description, settingsConfig,
+					resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.0.absolute_session_timeout_in_seconds", "3600"),
+					resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.0.idle_session_timeout_in_seconds", "300")),
+			},
+			{
+				PreConfig: sleepForSettingsRateLimit,
+				Config:    configWithSettings(orgOwnerID, name, description, roleName, &admin.OrganizationSettings{CustomSessionTimeouts: &admin.CustomSessionTimeouts{AbsoluteSessionTimeoutInSeconds: conversion.IntPtr(3600)}}),
```
EspenAlbert: [nit] Struggling a bit to read this test steps. What is changing and what are the checks? Some uses intermediate variables others creates new values inline

EspenAlbert (commented):
Struggled a bit to understand what is the API behavior here. Do we need to reset on the root level field or using an explicit `null` in both child attributes?

docs/data-sources/organization.md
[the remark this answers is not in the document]
marcabreracast: Addressed in [2cf3a57](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/2cf3a5777fe1e634a95b1754435a0ad3c41c36a6)

internal/service/organization/resource_organization.go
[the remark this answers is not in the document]
EspenAlbert: TY!

internal/service/organization/resource_organization_test.go
```diff
@@ -504,9 +556,24 @@ func configWithSettings(orgOwnerID, name, description, roleNames string, setting
...
+// conversion.StringPtr, because StringPtr("") returns nil and would drop the attribute from the
+// emitted HCL, silently turning an intended empty string into "field absent".
 func withOperationsContact(settings *admin.OrganizationSettings, operationsContact string) *admin.OrganizationSettings {
 	updated := *settings
-	updated.OperationsContact = conversion.StringPtr(operationsContact)
+	updated.OperationsContact = &operationsContact
+	return &updated
+}
+
+// withTimeouts returns settings with the custom session timeouts block replaced, so a test step
```
EspenAlbert, line 596: nit: the `withTimeouts` doc comment claims a test step `only changes those two attributes`, but all three call sites pass an empty `&admin.OrganizationSettings{}` base, so the emitted config also drops `gen_ai_features_enabled` (schema default `true`) and the other settings, and steps following a `gen_ai_features_enabled = false` config flip it back to true; reword the comment or pass the previous step's settings as the base the way `withOperationsContact` does

internal/service/organization/resource_organization_test.go
```diff
@@ -131,37 +139,81 @@ func TestAccConfigRSOrganization_Settings(t *testing.T) {
...
+				Config:    configWithSettings(orgOwnerID, nameUpdated, description, roleName, withTimeouts(&admin.OrganizationSettings{}, nil, nil)),
+				Check:     resource.TestCheckResourceAttr(resourceName, "custom_session_timeouts.#", "0"),
+			},
+			{
+				// Regression guard: the all-null block must not re-propose the 1 -> 0 count diff.
+				// customSessionTimeoutsEmptyBlockSuppress hides it when state is empty, so this plan
+				// is empty (PlanOnly with ExpectNonEmptyPlan=false fails otherwise).
+				PreConfig: sleepForSettingsRateLimit,
+				Config:    configWithSettings(orgOwnerID, nameUpdated, description, roleName, withTimeouts(&admin.OrganizationSettings{}, nil, nil)),
+				PlanOnly:  true,
```
EspenAlbert, line 242: q: the local run pasted in the PR description predates the DiffSuppressFunc fix and the new empty-block and PlanOnly steps, and this test skips CI via `acc.SkipTestForCI`, so the regression guard has no evidenced run; can you re-run `TestAccConfigRSOrganization_Settings` at the new head and refresh the log

internal/service/organization/resource_organization.go
```diff
@@ -107,6 +108,24 @@ func Resource() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"custom_session_timeouts": {
+				Type:             schema.TypeList,
+				Optional:         true,
+				MaxItems:         1,
+				DiffSuppressFunc: customSessionTimeoutsEmptyBlockSuppress,
```
marcabreracast, line 115: FYI @EspenAlbert: Regarding [your review comment](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713#discussion_r3969259947) and the API behaviour, there's two ways a customer can reset both timeouts, and the API treats them identically:

1. Set the whole block to `null`, remove the block, or set it to `[]`.
2. Declare the block with both children `null`.

I verified both against the API: each PATCH returns 200, and afterwards the GET response omits `customSessionTimeouts`. That detail would rule out the "keep one state item", so went with the second option you named, suppressing the shape by using `DiffSuppressFunc` on the block. It only suppresses when state is already empty and both children are null, so a real reset still applies and the following plan converges.
 I've also added additional steps to check this behaviour in the acceptance tests, and execution was successful: 

```
=== RUN   TestAccConfigRSOrganization_Settings
--- PASS: TestAccConfigRSOrganization_Settings (587.34s)
PASS
ok  	github.com/mongodb/terraform-provider-mongodbatlas/internal/service/organization	588.460s
```

internal/service/organization/resource_organization_test.go
marcabreracast: Did a refactor of them in c2ea537. Let me know if it's a bit better

internal/service/organization/resource_organization.go
```diff
@@ -393,11 +416,84 @@ func newOrganizationSettings(d *schema.ResourceData) *admin.OrganizationSettings
...
+	if old != "0" && old != "" {
+		return false // state holds a value; keep the diff so the reset is applied
+	}
+	return customSessionTimeoutsAllNullConfig(d)
+}
+
+// customSessionTimeoutsAllNullConfig reports whether raw config declares the
+// custom_session_timeouts block with both children unset.
+func customSessionTimeoutsAllNullConfig(d *schema.ResourceData) bool {
+	raw := d.GetRawConfig().GetAttr("custom_session_timeouts")
```
EspenAlbert, line 483: can this be replaced with the sdkv2config.CollectionEmpty ?

internal/service/organization/resource_organization.go
```diff
@@ -392,9 +416,69 @@ func newOrganizationSettings(d *schema.ResourceData) *admin.OrganizationSettings
...
+	}
 	return settings
 }
 
+// expandCustomSessionTimeouts maps the block into the SDK object using raw config so an
+// unset child is sent as an explicit null.
+func expandCustomSessionTimeouts(d *schema.ResourceData) *admin.CustomSessionTimeouts {
+	raw := d.GetRawConfig().GetAttr("custom_session_timeouts")
+	cst := &admin.CustomSessionTimeouts{}
+	if raw.LengthInt() == 0 { // block present but empty -> reset both to defaults
```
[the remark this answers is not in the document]
EspenAlbert: based on the other comment since `custom_session_timeouts = []` also reset, we can just return `cst` directly?

internal/service/organization/resource_organization_test.go
EspenAlbert: Thanks! That helped a lot. However, now there is 10 steps in this test.
I think we should split it to more specific tests.

internal/service/organization/resource_organization.go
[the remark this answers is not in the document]
EspenAlbert: verified against the vendored SDK: not quite — `SetAbsoluteSessionTimeoutInSecondsNil` does more than nil the pointer, it also appends to `NullFields`, and `CustomSessionTimeouts.MarshalJSON` honors `NullFields`, so the current branch sends `"absoluteSessionTimeoutInSeconds": null, "idleSessionTimeoutInSeconds": null` while a fresh `cst` would marshal `{}` with both children omitted; that is a different request body, and `{}` may read as no-change instead of a reset

the zero-length guard also has to stay on its own: `raw.Index(cty.NumberIntVal(0))` on an empty tuple panics (go-cty indexes the backing slice directly), so `custom_session_timeouts = []` cannot fall through to the element path

internal/service/organization/resource_organization.go
EspenAlbert: yes — `sdkv2config.CollectionEmpty(d, "custom_session_timeouts")` returns true for exactly the cases the inline condition guards (attr null, unknown, or known length 0) and false when the whole raw-config object is null or unknown, which matches the current `GetAttr`-on-null-object behavior, so `if sdkv2config.CollectionEmpty(d, "custom_session_timeouts") { return false }` is a drop-in replacement; the package is already imported in this file for `operations_contact`

marcabreracast: I believe we agreed on doing those changes on a follow-up PR so the change is contained and easier to track

internal/service/organization/resource_organization.go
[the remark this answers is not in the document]
marcabreracast: Apologies as I made a mistake as discussed offline, the API treats `custom_session_timeouts = []` as a no-op, so in [285e65e](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/285e65ea6879237b0aa99bb5ea8336653def6527) I addressed this.

internal/service/organization/resource_organization_test.go
marcabreracast: Sure, since I split the tests per your recommendation in [5febe3b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/5febe3bbb8c994e325e5aa06c2e15fc7eea1c666), added the log with the tests passing in the PR description

internal/service/organization/resource_organization_test.go
marcabreracast: Addressed in [ff92b92](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/ff92b92b464b02707b761b20eb489968a769a2c4)

internal/service/organization/resource_organization.go
marcabreracast: I believe we agreed in doing these changes in a follow-up PR for tracking purposes and so we can test that the main behaviour does not change

internal/service/organization/resource_organization.go
AgustinBettati, line 115: So `customSessionTimeoutsEmptyBlockSuppress` is additional complexity at the benefit of supporting `custom_session_timeouts {}` for unsetting attributes right? A small comment with this clarification can help.

AgustinBettati (approved):
LGTM

internal/service/organization/resource_organization.go
EspenAlbert: verified the wire shapes with a throwaway marshal test: `custom_session_timeouts {}` PATCHes `customSessionTimeouts: {"absoluteSessionTimeoutInSeconds": null, "idleSessionTimeoutInSeconds": null}` while `= []`, `= null`, and block removal PATCH top-level `customSessionTimeouts: null`

so the premise depends on the open question from the `= []` thread: if the API no-ops the all-null object as reported there, the suppress is not supporting unsetting via `{}` at all, it only keeps the empty block a clean no-op when state is already empty, and unsetting works via `= []`, `= null`, or removing the block; the comment requested here should capture which of the two it is (raised separately on the switch at `resource_organization.go:421`)

internal/service/organization/resource_organization.go
```diff
@@ -393,11 +416,79 @@ func newOrganizationSettings(d *schema.ResourceData) *admin.OrganizationSettings
 	case contact.Set():
 		settings.SetOperationsContact(contact.ValueString())
 	}
+	raw := d.GetRawConfig().GetAttr("custom_session_timeouts")
+	switch {
+	case raw.IsKnown() && !raw.IsNull() && raw.LengthInt() > 0:
```
EspenAlbert: after `285e65e` the reset forms diverge at the wire level, and the empty-block form sends exactly the request reported as an API no-op: a block whose children are both null still goes through `expandCustomSessionTimeouts` and PATCHes `customSessionTimeouts: {"absoluteSessionTimeoutInSeconds": null, "idleSessionTimeoutInSeconds": null}`, while `= []`, `= null`, and block removal now send top-level `customSessionTimeouts: null`

if the no-op statement on the `= []` thread holds, `custom_session_timeouts {}` (or both children set to `null`) against a state with values is a perpetual plan: the apply no-ops, the read restores the values, and the diff re-proposes every plan (`customSessionTimeoutsEmptyBlockSuppress` only fires when state is already empty), and the new test steps expecting `# = 0` plus a clean `PlanOnly` cannot pass; if instead the pasted passing log was captured at this head, the all-null object does reset and the no-op statement is a misdiagnosis

either way the safe fix is to route the all-null block to `SetCustomSessionTimeoutsNil()` so every reset form sends top-level null; that converges under both readings and matches the test expectations
can you confirm with upstream which behavior the all-null object actually has?

docs/resources/organization.md
```diff
@@ -73,12 +73,18 @@ output "service_account_secret" {
...
+* `skip_default_alerts_settings` - (Optional) Flag that indicates whether to prevent Atlas from automatically creating organization-level alerts not explicitly managed through Terraform. Defaults to `true`.
 
 ~> **NOTE:** - If you create an organization with our Terraform provider version >=1.30.0, this field is set to `true` by default.<br> - If you have an existing organization created with our Terraform provider version <1.30.0, this field might be `false`, which is the [API default value](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-createorganization). To prevent the creation of future default alerts, set this explicitly to `true`.
 
 * `service_account` - (Optional) Block to create a Service Account instead of a Programmatic API Key when creating the organization. The API does not allow creating both in the same request. Mutually exclusive with `description` and `role_names`. This block can't be updated after creation. See [Service Account](#service-account).
 
+### Custom Session Timeouts
```
EspenAlbert, line 83: [nit] the docs describe per-child `null` semantics but never explain how to clear the block as a whole, maybe worth adding whole-block clear guidance like `operations_contact` has ("Remove the attribute from your configuration, or set it to `null`, to clear") once the empty-block behavior is confirmed

EspenAlbert (approved):
Great job! Thank you for addressing all the comments. Looking forward to the sdkv2config reuse PR :D

docs/resources/organization.md
marcabreracast: Addressed in [1dca0e2](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/1dca0e235a179bf63867ac3c89cd88d2beee3f58)

internal/service/organization/resource_organization.go
marcabreracast: Addressed in [1dca0e2](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4713/commits/1dca0e235a179bf63867ac3c89cd88d2beee3f58)

marcabreracast:
The acceptance `config` group failures are unrelated to this PR. They are all in `internal/service/team` (and `atlasuser` in one run) and also fail in `master` https://github.com/mongodb/terraform-provider-mongodbatlas/actions/runs/34505005382/job/102964948790

PR #4716, opened 2026-09-08
chore: Clear maintenance schedule on wave-only transition

marcabreracast:
## Description

Allows `mongodbatlas_maintenance_window` to clear an existing maintenance schedule when transitioning to a wave-only configuration (schedule omitted, `wave_assignment` retained). The update path now sends `day_of_week` and `hour_of_day` as explicit null via the SDK `NullFields` mechanism when they are removed from config, so Atlas clears the schedule instead of retaining the old values. Also adds a follow-up `Read` after `Update` to refresh computed attributes that may change server-side during the transition.

Link to any related issue(s): CLOUDP-440317, blocked by CLOUDP-439562 (now resolved)

## Type of change:

Internal work on an unreleased feature (maintenance waves). None of the above apply.

## Required Checklist:

## Further comments

The `Read` after `Update` fixes a pre-existing issue where computed attributes (such as `time_zone_id`) went stale after an update until the next plan/refresh. This was latent until the wave-only transition introduced a path where a computed attribute changes value mid-lifecycle. No changelog entry is needed as this is internal work on an unreleased feature; the user-facing changelog is covered by the overall maintenance waves enhancement.

### DIFF
--- a/internal/service/maintenancewindow/resource_maintenance_window.go
+++ b/internal/service/maintenancewindow/resource_maintenance_window.go
@@ -285,7 +285,7 @@ func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		}
 	}
 
-	return nil
+	return resourceRead(ctx, d, meta)
 }
 
 func resourceDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
--- a/internal/service/maintenancewindow/resource_maintenance_window.go
+++ b/internal/service/maintenancewindow/resource_maintenance_window.go
@@ -233,14 +233,16 @@ func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	params := new(admin.GroupMaintenanceWindowPreviewUpdateRequest)
-	// TODO(CLOUDP-440317): omitting day_of_week/hour_of_day leaves them out of the PATCH rather than
-	// clearing them, so removing an existing schedule to go wave-only does not converge. Send them as
-	// explicit null (SetDayOfWeekNil/SetHourOfDayNil) once the API honors null unset (CLOUDP-439562).
+
 	if !d.GetRawConfig().GetAttr("day_of_week").IsNull() {
-		params.DayOfWeek = new(d.Get("day_of_week").(int))
+		params.SetDayOfWeek(d.Get("day_of_week").(int))
+	} else if d.HasChange("day_of_week") {
+		params.SetDayOfWeekNil()
 	}
 	if !d.GetRawConfig().GetAttr("hour_of_day").IsNull() {
-		params.HourOfDay = new(d.Get("hour_of_day").(int))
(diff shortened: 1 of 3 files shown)


internal/service/maintenancewindow/resource_maintenance_window.go
```diff
@@ -283,7 +285,7 @@ func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		}
 	}
 
-	return nil
+	return resourceRead(ctx, d, meta)
```
marcabreracast, line 289: Added this `Read` after `Update` to refresh computed attributes that the API may omit or change server-side. Without it, values like `time_zone_id` (which is absent from the GET response for a wave-only window) go stale in state until the next plan. This was a pre-existing issue that surfaced when the wave-only transition introduced a path where a computed attribute changes value mid-lifecycle.

internal/service/maintenancewindow/resource_maintenance_window.go
```diff
@@ -233,14 +233,16 @@ func resourceUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
-		params.DayOfWeek = new(d.Get("day_of_week").(int))
+		params.SetDayOfWeek(d.Get("day_of_week").(int))
+	} else if d.HasChange("day_of_week") {
+		params.SetDayOfWeekNil()
 	}
 	if !d.GetRawConfig().GetAttr("hour_of_day").IsNull() {
-		params.HourOfDay = new(d.Get("hour_of_day").(int))
+		params.SetHourOfDay(d.Get("hour_of_day").(int))
+	} else if d.HasChange("hour_of_day") {
+		params.SetHourOfDayNil()
```
AgustinBettati, line 246: The `else if d.HasChange("hour_of_day")` gate misses the case where the existing schedule is at midnight. `hour_of_day = 0` is both a valid value and the SDKv2 TypeInt zero value, so removing the schedule from config produces no diff entry for it and the PATCH sends `{"dayOfWeek": null}` with `hourOfDay` omitted. `day_of_week` is unaffected because it is 1-based. Not certain how API behaves in this case.

One option: Suggest gating both branches on `d.HasChange("day_of_week") || d.HasChange("hour_of_day")`, since the two are mutually `RequiredWith`.

Lets make sure to capture this case in an acceptance test

marcabreracast: Good catch! Addressed this in [32978eb](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4716/commits/32978eb20fc70d94dfbe145d4a7c2ccee7099e47) and added a separate acceptance test since I want to make sure to specifically cover this case (`TestAccConfigRSMaintenanceWindow_waveAssignmentMidnight`), but let me know if you prefer to fold it into another test.

AgustinBettati (approved):
LGTM, thanks for the additional test coverage

PR #4718, opened 2026-09-09
fix: Fixes plan and apply errors when attributes are not known until apply

lantoli:
## Description

Planning an update to `mongodbatlas_advanced_cluster` fails when an optional attribute is set from an expression that is not resolved yet, such as a module output or `merge()`. 
Found while doing #4697. It is an independent pre-existing defect on `master`, so it is fixed here rather than in that branch.

The value is unknown at plan time, and the provider mishandles it in three independent places:
- `handleModifyPlan` calls `CopyUnknowns`, which replaces the unknown plan value with the state value. The planned value then contradicts the configuration, so Terraform reports `Provider produced invalid plan`.
- `newResourceTag` and `newComponentLabel` convert `tags` and `labels` without an unknown guard, so building the Atlas request fails with `Value Conversion Error` during plan, before any Atlas call.
- `accept_data_risks_and_force_replica_set_reconfig` reaches the time parser as an empty string, failing with `error converting AcceptDataRisksAndForceReplicaSetReconfig`.

This affects every attribute that is `Optional` without `Computed`, currently: `tags`, `labels`, `adaptive_capacity`, `retain_backups_enabled`, `use_effective_fields`, `pinned_fcv`, `backing_provider_name`, `accept_data_risks_and_force_replica_set_reconfig` and `timeouts`. Their value can only come from the configuration, so keeping them unknown is always correct. `Optional+Computed` attributes need a second fix. Terraform accepts a provider planning a value it owns for them at plan time, but the plan re-run during apply sees the resolved configuration, finds `replication_specs` changed and turns attributes such as `replication_specs.0.container_id` unknown, and known to unknown fails the apply with `Provider produced inconsistent final plan`. This is reproducible on `master` with `auto_scaling` set from another resource on a plain `REPLICASET`.

`schemafunc.UnknownInConfig` reports the top-level attributes the configuration leaves unknown and adds them to `keepUnknown`, so they stay unknown until the expression resolves. The name is top-level because an unknown nested in `replication_specs` also hides whether that list changes. `adjustRegionConfigsChildren` is skipped for unresolved `replication_specs` for the same reason: the values it fills from state, such as `read_only_specs`, depend on specs that are not resolved yet, and the plan re-run during apply fills them once they are.

### Reproduction

Create the cluster without the tag, then add it and apply again:

```hcl
resource "terraform_data" "tags" {
  input = { owner = "team-atlas" }
}

resource "mongodbatlas_advanced_cluster" "test" {
  project_id   = var.project_id
  name         = "example"
  cluster_type = "REPLICASET"

  tags = terraform_data.tags.output // Attribute updated

  replication_specs = [{
    region_configs = [{
      provider_name   = "AWS"
      region_name     = "US_EAST_1"
      priority        = 7
      electable_specs = { instance_size = "M10", node_count = 3 }
    }]
  }]
}
```

Before this change `terraform plan` fails, without reaching Atlas:

```
Error: Value Conversion Error

Received unknown value, however the target type cannot handle unknown values.
Target Type: map[string]basetypes.StringValue
```

Link to any related issue(s): CLOUDP-430468

## Type of change:

## Required Checklist:


lizo-mdb (approved):
LGTM

internal/service/advancedcluster/plan_modifier_test.go
```diff
@@ -46,6 +46,21 @@ func TestPlanChecksClusterTwoRepSpecsWithAutoScalingAndSpecs(t *testing.T) {
 					plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionNoop),
 				},
 			},
+			{
+				// These attributes are Optional without Computed, so an unresolved expression must
+				// stay unknown instead of being replaced by the state value.
+				ConfigFilename: "main_unresolved_optional_attributes.tf",
```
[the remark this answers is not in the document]
lantoli: the fixture is intentionally partial: it covers one attribute per affected kind (map, bool, string, date, nested), not all nine. We are keeping `configOnlyAttributes` as a plain literal list rather than adding a schema-driven guard test; a reviewer will catch a missing entry when a new Optional-only attribute is added, same as the existing `volatileAttributes` list right above it, which has no such guard either.

internal/service/advancedcluster/plan_modifier.go
```diff
@@ -58,7 +58,9 @@ func handleModifyPlan(ctx context.Context, diags *diag.Diagnostics, state, plan
 		return
 	}
 	attributeChanges := schemafunc.NewAttributeChanges(ctx, state, plan)
-	keepUnknown := []string{"connection_strings", "state_name", "mongo_db_version", "config_server_type"} // Volatile attributes, should not be copied from state
+	// Optional attributes without Computed, an unknown value is an unresolved config expression, not a state value
+	keepUnknown := []string{"accept_data_risks_and_force_replica_set_reconfig", "adaptive_capacity", "backing_provider_name", "labels", "pinned_fcv", "retain_backups_enabled", "tags", "timeouts", "use_effective_fields"}
```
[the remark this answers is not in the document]
lantoli: we decided against adding a schema-driven guard test for `configOnlyAttributes`. It stays a plain literal list, matching `volatileAttributes` right above it, which has never had one either. A missing entry surfaces at review time when a new Optional-only attribute is added, not through a dedicated test.

internal/service/advancedcluster/plan_modifier_test.go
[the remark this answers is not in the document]
lantoli: same reasoning as the reply above: this fixture is intentionally partial (one attribute per affected kind), and we are not adding a schema-driven guard for `configOnlyAttributes`. It is a plain literal list like `volatileAttributes`, and review will catch a missing entry when a new Optional-only attribute is added.

lantoli:
TestAccMockableAdvancedCluster_replicasetAdvConfigUpdate  failing test in PR checks is not related to this PR and is also failing in master in TestSuite

internal/service/advancedcluster/plan_modifier.go
EspenAlbert: Why do we need to add this to never copied from state? Can the CopyUnknowns instead handle this?

EspenAlbert: CopyUnknowns cannot derive the set itself: it reflects over the TFModel fields and only sees tfsdk names and attr values, with no access to the schema, so it cannot tell Optional-without-Computed from Optional+Computed. da7afc874 tried the schema-walk derivation (with a drift test) and b12f5e1e0 replaced it with this explicit list so both name sets sit next to the use site. I verified the list against the schema: the seven root Optional-only attributes plus pinned_fcv, timeouts, and the nested backing_provider_name are exactly the Optional-without-Computed set, so nothing is missing today

internal/service/advancedcluster/plan_modifier_test.go
EspenAlbert: q: the fixture exercises 7 of the 9 `keepUnknown` names, but `pinned_fcv` and `timeouts` are left out. any reason? `pinned_fcv` is the only object-typed entry, and its nested `expiration_date` going unknown is the distinct repro shape for a SingleNestedAttribute (`pinned_fcv = { expiration_date = terraform_data.unresolved_date.output }` plus `plancheck.ExpectUnknownValue(resourceName, tfjsonpath.New("pinned_fcv").AtMapKey("expiration_date"))`), so covering it would also guard the one entry whose removal no current test would catch

internal/service/advancedcluster/plan_modifier.go
lantoli: thanks, you're right, updated `CopyUnknowns` to handle this in d4d70a6. The implementation is a bit more involved, but it avoids maintaining a separated list. I also simplified the func by removing the unused `keepUnkownCall` paramater

internal/service/advancedcluster/plan_modifier.go
```diff
@@ -28,8 +30,28 @@ var (
 	}
 )
 
+// unknownInConfig lists the top-level attributes the configuration still leaves unknown, using the naming of
+// keepUnknown. Planning their state value hides a change that the plan re-run during apply then reveals, so the
+// apply fails with "Provider produced inconsistent final plan". The Optional-without-Computed rule in
+// CopyUnknowns does not cover this: cluster_type and auto_scaling are also Computed. The name is top-level
+// because an unknown nested in replication_specs hides whether that list changes.
+func unknownInConfig(config tftypes.Value) []string {
```
EspenAlbert: consider refactoringr
does this belong in a helper functoin? Doesn't seem related to advanced_cluster?
Might be better in the schemafunc as part of copyUnknown?

internal/service/advancedcluster/plan_modifier_test.go
```diff
@@ -46,6 +46,38 @@ func TestPlanChecksClusterTwoRepSpecsWithAutoScalingAndSpecs(t *testing.T) {
 					plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionNoop),
 				},
 			},
+			{
```
EspenAlbert, line 49: nice with plan tests for this!

EspenAlbert (approved):
Thank you for following up on the comments. Ok to merge current style 👍

internal/service/advancedcluster/plan_modifier.go
lantoli: you're right, changed in 4ab3744

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -35,58 +37,88 @@ func HasUnknowns(obj any) bool {
...
+func copyUnknowns(ctx context.Context, src, dest any, attributes map[string]schema.Attribute, keepUnknown []string) {
 	valSrc, valDest := validateStructPointers(src, dest)
 	typeSrc := valSrc.Type()
 	typeDest := valDest.Type()
 	for i := range typeDest.NumField() {
 		fieldDest := typeDest.Field(i)
 		name, tfName := fieldNameTFName(&fieldDest)
 		srcValue := valSrc.FieldByName(name).Interface()
-		if keepUnknownCall(tfName, srcValue.(attr.Value)) {
+		if shouldKeepUnknown(attributes, keepUnknown, tfName) {
```
AgustinBettati: help from 🤖 (worth double checking)

This skips the attribute before recursing, so an Optional-only object with a Computed child never gets that child from state. pinned_fcv is that case: version is Computed.
Suggestion: apply the Optional-only rule only when the dest value itself is unknown, here and at line 202, so a known object still recurses for its Computed children

internal/common/schemafunc/plan_modifier_state_for_unknown.go
```diff
@@ -35,58 +37,88 @@ func HasUnknowns(obj any) bool {
...
 }
 
-type KeepUnknownFunc func(string, attr.Value) bool
+// UnknownInConfig lists the top-level attributes that the configuration leaves unknown, to be passed as
+// keepUnknown to CopyUnknowns. Planning their state value hides a change that the plan Terraform re-runs
+// during apply then reveals, so the apply fails with "Provider produced inconsistent final plan". This is
+// not covered by CopyUnknowns keeping the Optional-without-Computed attributes unknown, which leaves out the
+// Optional+Computed ones. The name is top-level because an unknown nested in a list also hides whether that
+// list changes.
+func UnknownInConfig(config tftypes.Value) []string {
```
AgustinBettati, line 46: q: This function focuses in finding top-level. Why is nested information not needed or extracted at this point?

AgustinBettati (commented):
Was not able to dive too deep here, hard to grasp the end to end. Leaving comments that do not block the merge.

internal/common/schemafunc/plan_modifier_state_for_unknown.go
lantoli: nice catch, fixed in 97cb879. `pinned_fcv.version` was planned as unknown instead of taking its state value, as the parent is skipped before recursing

it shows up on every update of a cluster with `pinned_fcv` set: the framework marks Computed attributes that are null in the configuration as unknown, and `CopyUnknowns` was no longer filling it back. it is only extra noise in the plan, nothing would break if we left it, but the fix is small so let's take it

one addition while doing it: traversing a known object also exposes its `Required` children, so the rule is now "not `Computed`" instead of "`Optional` and not `Computed`", otherwise an unresolved `expiration_date = module.x.date` would take the state value. tests added for both directions

internal/common/schemafunc/plan_modifier_state_for_unknown.go
lantoli: `keepUnknown` matches attribute names, not paths, so nested granularity cannot be expressed there. It is not needed either: an unknown nested in `replication_specs` hides whether that list changes, so the whole attribute has to stay unknown until it resolves. The per-level precision comes from `CopyUnknowns`, which walks the nested attributes itself

PR #4726, opened 2026-09-14
fix: Avoids inconsistent result error when Atlas omits specs of nodeless region configs

### LINKED ISSUE #4238: [Bug]: produced an unexpected new value

- [x] I have searched the existing issues

### Provider Version

2.7.0

### Terraform Version

1.14.6

### Terraform Edition

Terraform Open Source (OSS)

### Current Behavior

I added `use_effective_fields = true` on my imported cluster and then added `electable_specs` as well since I was confronted with `REPLICATION_SPECS_REQUIRES_ELECTABLE_SPECS`. The apply seems to succeed within mongo but then fails on the state/output side of things:

> Error: Provider produced inconsistent result after apply
> │ 
> │ When applying changes to mongodbatlas_advanced_cluster.this, provider "provider[\"registry.terraform.io/mongodb/mongodbatlas\"]" produced an unexpected new value: .replication_specs[0].region_configs[0].analytics_specs.disk_size_gb: was cty.NumberIntVal(2048),
> │ but now null.
> │ 
> │ This is a bug in the provider, which should be reported in the provider's own issue tracker.

```hcl
resource "mongodbatlas_advanced_cluster" "this" {
  project_id           = local.project_id
  name                 = local.cluster_name
  cluster_type         = "REPLICASET"
  use_effective_fields = true
  replication_specs = [
    {
      region_configs = [
        {
          provider_name = "AZURE"
          priority      = 7
          region_name   = "GERMANY_WEST_CENTRAL"
          analytics_specs = {
            disk_iops     = 7500
            disk_size_gb  = 2048
            instance_size = "M60"
            node_count    = 0
          }
          auto_scaling = {
            compute_enabled            = true
            compute_max_instance_size  = "M60"
            compute_min_instance_size  = "M50"
            compute_scale_down_enabled = true
            disk_gb_enabled            = true
          }
          electable_specs = {
            instance_size = "M50"
            node_count    = 3
          }
        }
      ]
    }
  ]
}
```

### Steps To Reproduce

not sure, create above without effective fields and then add it, perhaps

### Logs

```txt

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ZayanKhan-12:
## Description

Fixes the `Provider produced inconsistent result after apply` error reported when a region config has an `analytics_specs` block with `node_count = 0` and `use_effective_fields = true`.

Link to any related issue(s): Fixes #4238

### Root cause

Atlas omits hardware spec attributes for a spec that describes no nodes — with nothing provisioned there is no hardware to report. It is most visible with `use_effective_fields` enabled, where the specs returned describe the effective (actually provisioned) hardware rather than what the client sent.

Those attributes are `Optional + Computed`, so Terraform requires the value in state after apply to equal the config value whenever the practitioner set one. `newSpecsObjType` maps an omitted field straight to null via `types.Float64PointerValue(nil)`, which contradicts the config and aborts the apply:

```
Error: Provider produced inconsistent result after apply
... .replication_specs[0].region_configs[0].analytics_specs.disk_size_gb:
was cty.NumberIntVal(2048), but now null.
```

That also explains the reporter's workaround — removing `analytics_specs.disk_size_gb` from the config leaves nothing for the null to contradict, and the plan settles.

### Fix

Keep the value from the previous model for attributes of a nodeless spec that Atlas did not report.

The override runs inside `OverrideAttributesWithPrevStateValue`, already the single reconciliation point called from `getBasicClusterModel` for Create, Read and Update — so `modelIn` is the plan in Create/Update and the prior state in Read, and in every case holds the value to fall back to. Handling it in Read as well is what keeps subsequent plans empty rather than showing a permanent diff.

Two guards keep the change narrow:

- **Specs that have nodes are left untouched.** There Atlas reports the hardware actually backing them, and that value must win — this must not mask auto-scaling adjustments. Only a spec whose `node_count` is 0 is considered, taken from Atlas' response when it reported one and from the previous model otherwise.
- **Only known values are copied.** `modelIn` is the plan in Create/Update, where computed attributes can be unknown, and unknown must never reach state. This is why the existing `copyAttrIfDestNotKnown` is not reused: it treats null and unknown alike and is only safe where the source is state.

The same omission is possible for `electable_specs` and `read_only_specs`, so all three are handled uniformly.

`OverrideAttributesWithPrevStateValue` gains a `ctx` parameter, needed to walk the nested list/object values. Both call sites (`resource.go`, `model_flex.go`) already had one in scope.

### A note on scope

This is the provider-side half. @AgustinBettati mentioned in the issue that the inconsistency originates upstream in the API — that remains worth fixing there, and this change does not depend on it either way: it makes the provider tolerant of an omitted attribute for a spec that has no nodes, which is reasonable behaviour regardless of whether the API is later changed to echo the value back.

I also deliberately did **not** handle the case where Atlas omits the entire specs *object*, only individual attributes within it. The reported error is anchored at `.analytics_specs.disk_size_gb`, so the object is present; resurrecting a whole absent block felt like speculative scope for a bug report that doesn't show it. Happy to add it if you've seen that variant.

## Type of change:

## Required Checklist:

### Tests

`TestAdvancedCluster_overrideUnreportedInactiveSpecs` encodes the issue directly, plus the cases that keep the fix honest:

| Case | Asserts |
|---|---|
| issue 4238 scenario | `disk_size_gb` 2048 survives an omission for a 0-node analytics spec |
| every omitted attribute of an inactive spec | `disk_iops`, `ebs_volume_type`, `instance_size`, `node_count` too |
| Atlas values win when reported | the override never overwrites a value Atlas actually sent |
| a spec with nodes is left untouched | auto-scaling adjustments are not masked |
| unknown plan values never reach state | an unknown source is skipped, leaving null |
| `node_count` omitted by Atlas | falls back to the previous model |
| spec absent from the response / from the previous model | neither is resurrected nor disturbed |

Plus `...AllSpecTypes` (analytics, electable and read-only) and `...NoReplicationSpecs` (null/unknown/empty lists, guarding the flex path).

**Verified the tests actually catch the bug**: with the `overrideUnreportedInactiveSpecs` call commented out, the issue-4238 case fails with exactly the reported shape — `expected: Float64Value{state:0x2 (known), 2048}`, `actual: Float64Value{state:0x0 (null)}`.

### Verification run

- `go build ./...` — clean
- `go test ./...` — 0 failures repo-wide
- `go test ./internal/service/advancedcluster/... -race -count=1` — pass
- `golangci-lint run` (v2.12.2, the pinned version) — 0 issues repo-wide
- `gofmt -s -l`, `goimports -l`, `go vet` on the changed files — clean

Note that the plan-check tests in this package require the `terraform` binary on `PATH`; without it they panic with `terraform not found in PATH`, which reads like a code failure but isn't.

### What I could not verify

I have no Atlas org, so **acceptance tests were not run** and no MacT capture is included — a recorded capture of an apply that enables `use_effective_fields` on a cluster with a 0-node `analytics_specs` would be the natural end-to-end regression test, and I'd rather a maintainer record it against real Atlas than hand-forge the YAML. The root cause above is inferred from the reported error, the reporter's workaround, and the response-mapping code path; if your reproduction shows Atlas omitting something different, the guard in `overrideUnreportedSpecAttrs` is where to adjust.

The changelog entry (`.changelog/<PR>.txt`) is added in a follow-up commit now that the PR number exists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Amg4ZK9JA4ePZxHUhZwnjv


AgustinBettati:
Hi @ZayanKhan-12. Thank you for the contribution.

As mentioned in https://github.com/mongodb/terraform-provider-mongodbatlas/issues/4238 the root cause of this issue is due to inconsistent API behaviour. Specifically, when leveraging `use_effective_fields` and defining different `disk_size_gb` values in `electableSpecs` vs `analyticsSpecs`/`readOnlySpecs`, `electableSpecs` is used as the single source for `disk_size_gb` value across all specs.

Our current posture remains on fixing this issue on the API side and hence avoiding additional complexity on a specific client (in this case the Terraform provider). This decision is also considering there is a straight forward path for mitigating the non-empty plan (aligning disk_size_gb value across the specs). Any further updates will be left in https://github.com/mongodb/terraform-provider-mongodbatlas/issues/4238.

PR #4727, opened 2026-09-15
feat: Adds `without_initial_secret` to mongodbatlas_service_account

EspenAlbert:
## Description

Adds a `without_initial_secret` attribute to `mongodbatlas_service_account`. When set to `true`, the provider sends `withoutInitialSecret: true` to the Atlas API, so Atlas creates the Service Account without generating a bootstrap secret. The resource then has an empty `secrets` list, and the configuration creates its own secrets with `mongodbatlas_service_account_secret`.

`secret_expires_after_hours` is no longer required on create. It stays `optional` and create-only. The API contract is:

- `without_initial_secret = true` requires `secret_expires_after_hours` to be unset.
- Setting neither is rejected by the API (`ARGUMENT_NOT_PROVIDED`).
- Setting both is rejected by the API (`MUTUALLY_EXCLUSIVE_ARGUMENTS`).

The provider relies on the API error for these two validation cases rather than duplicating the check at plan time.

Link to any related issue(s): [CLOUDP-444252](https://jira.mongodb.org/browse/CLOUDP-444252), [CLOUDP-375930](https://jira.mongodb.org/browse/CLOUDP-375930)

## Type of change:

## Required Checklist:

### Schema: `without_initial_secret` is Optional + Computed

`without_initial_secret` is `Optional` and `Computed`, and uses `customplanmodifier.CreateOnlyBoolWithDefault(false)`. The plan modifier injects the `false` default into the plan on create so the resource Create handler needs no custom logic, and it rejects a change on update. Terraform Plugin Framework requires a plan-time default to sit on an Optional + Computed attribute, so the attribute is Computed as well.

`true` is only meaningful at create time. On update, the API omits `withoutInitialSecret` from responses, so the state value stays whatever the user set at create (default `false`), and the attribute is not re-read.

### Codegen: create-only plan modifier for computed optional bools with a default

`setCreateOnlyValue` in `tools/codegen/codespec/config.go` skipped every computed attribute when deciding whether to emit a create-only plan modifier. That guard exists because a plain `CreateOnly()` modifier can see an unknown value at plan time on a computed attribute and fail the update validation.

This PR narrows the guard: a called-out `ComputedOptional` bool that has a default and that the API never returns keeps the create-only behavior, because `CreateOnlyBoolWithDefault` handles the unknown value and the default. All other computed attributes keep the previous behavior. `TestApplyTransformationsToResource_CreateOnlyTransformation` covers the new cases.

Regenerating all resources with `make autogen-generate-resources step=code-gen` changes only `without_initial_secret`.

### Generated files

The tracked OpenAPI spec in `tools/codegen/atlasapispec/` is stale for both `withoutInitialSecret` and `systemManaged`. I did not commit a refreshed spec: the upstream `dev` spec is used as a local input only. Both the raw and flattened spec files are left at `HEAD`, and no spec diff is part of this PR.

The flow used to update the model:

1. `make autogen-update-api-spec spec_source=/tmp/upstream-v2-dev.yaml` writes the raw and flattened spec locally.
2. `make autogen-generate-resources resource_name=service_account step=model-gen` regenerates `tools/codegen/models/service_account.yaml` from that local spec.
3. `make autogen-generate-resources resource_name=service_account step=code-gen` regenerates `internal/serviceapi/serviceaccount/resource_schema.go`.
4. `git checkout -- tools/codegen/atlasapispec/` reverts the spec files so they stay out of the PR.

The upstream `dev` spec now carries the `systemManaged` field on `OrgServiceAccount` (this is what the hand-vendored `system_managed` attribute came from) and `withoutInitialSecret` on `OrgServiceAccountRequest`, and it drops `secretExpiresAfterHours` from the request's required list. Regeneration therefore reproduces `system_managed` instead of dropping it, so the model no longer needs a hand edit.

CI's `check-autogen-resources` workflow runs `autogen-generate-resources step=code-gen` against the committed models, so the committed model is the source of truth for generated code and the local spec is not needed to reproduce it.


internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +83,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+				// without_initial_secret defaults to false on create, so setting it to true on update must be rejected.
+				Config:      configBasic(orgID, name, "description", []string{"ORG_READ_ONLY"}, nil, "without_initial_secret = true"),
+				PlanOnly:    true,
+				ExpectError: regexp.MustCompile("without_initial_secret cannot be updated"),
+			},
+		},
+	})
+}
+
+func TestAccServiceAccount_withoutInitialSecret(t *testing.T) {
```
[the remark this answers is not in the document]
EspenAlbert: Wrong, also valid in qa

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +83,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet(dataSourceName, "client_id"),
+				),
+			},
+			{
+				ResourceName:                         resourceName,
+				ImportStateIdFunc:                    importStateIDFunc(resourceName),
+				ImportStateVerifyIdentifierAttribute: "client_id",
+				ImportState:                          true,
+				ImportStateVerify:                    true,
+				ImportStateVerifyIgnore:              []string{"without_initial_secret"},
```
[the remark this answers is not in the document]
EspenAlbert: Fixed, `without_initial_secret` is now in the basic test's `ImportStateVerifyIgnore` next to `secret_expires_after_hours`. Exactly the failure mode you describe: the create path stores the injected `false` default while an import read leaves the attribute null because the API never returns it.

templates/data-sources/service_account.md.tmpl
```diff
@@ -9,7 +9,7 @@ subcategory: "Service Accounts"
 ~> **IMPORTANT WARNING:** Managing Service Accounts with Terraform **exposes sensitive organizational secrets** in Terraform's state. We suggest following [Terraform's best practices](https://developer.hashicorp.com/terraform/language/state/sensitive-data).
 
 ## Example Usages
-{{ tffile (printf "examples/%s/main.tf" .Name )}}
+{{ tffile "examples/mongodbatlas_service_account/bootstrap_secret/main.tf" }}
```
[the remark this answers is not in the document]
EspenAlbert: I checked this by running the generation locally: `tfplugindocs generate` from the current templates produces byte-identical output for `docs/data-sources/service_account.md` (diff against the tracked file is empty). The template now points at `examples/mongodbatlas_service_account/bootstrap_secret/main.tf`, which is the unchanged rename of the old root `main.tf` the tracked docs were generated from, so regeneration is a no-op. The latest CI run confirms `generate-doc-check: SUCCESS`; no doc change is needed.

templates/data-sources/service_accounts.md.tmpl
```diff
@@ -9,7 +9,7 @@ subcategory: "Service Accounts"
 ~> **IMPORTANT WARNING:** Managing Service Accounts with Terraform **exposes sensitive organizational secrets** in Terraform's state. We suggest following [Terraform's best practices](https://developer.hashicorp.com/terraform/language/state/sensitive-data).
 
 ## Example Usages
-{{ tffile (printf "examples/mongodbatlas_service_account/main.tf" )}}
+{{ tffile "examples/mongodbatlas_service_account/bootstrap_secret/main.tf" }}
```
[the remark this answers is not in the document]
EspenAlbert: Same result as the singular data source: I ran the doc generation locally and the output for `docs/data-sources/service_accounts.md` is byte-identical to the tracked file, because both templates now render `bootstrap_secret/main.tf`, the unchanged rename of the old root `main.tf`. `generate-doc-check` passes in CI; no doc change is needed.

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +83,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+		orgID = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		name  = acc.RandomName()
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		Steps: []resource.TestStep{
+			{
+				Config:      configBasic(orgID, name, "Without initial secret", []string{"ORG_READ_ONLY"}, new(24), "without_initial_secret = true"),
+				ExpectError: regexp.MustCompile("(?i)mutually|withoutInitialSecret|exclusive"),
```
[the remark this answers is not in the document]
EspenAlbert: Don't want another test for API validation, one is enough

.changelog/4727.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
```
EspenAlbert, line 1: the changelog entry file is `.changelog/4721.txt`, but this PR is 4727, so `check-changelog-entry-file` fails with "Changelog entry file .changelog/4727.txt not found" and the go-changelog release tooling, which reads `.changelog/{PR_NUMBER}.txt`, never picks up the entry. rename the file to `.changelog/4727.txt` per the naming convention in `contributing/changelog-process.md`

templates/resources/service_account.md.tmpl
```diff
@@ -10,7 +10,13 @@ subcategory: "Service Accounts"
 
 ## Example Usages
 
-{{ tffile (printf "examples/%s/main.tf" .Name )}}
+The following example creates a Service Account and lets Atlas generate its first secret.
+
+{{ tffile "examples/mongodbatlas_service_account/bootstrap_secret/main.tf" }}
+
+### Further Examples
+- [Bootstrap secret](https://github.com/mongodb/terraform-provider-mongodbatlas/tree/v2.17.0/examples/mongodbatlas_service_account/bootstrap_secret)
```
EspenAlbert: the new "Further Examples" links pinned `tree/v2.17.0/`, but `examples/mongodbatlas_service_account/bootstrap_secret/` and `without_initial_secret/` are new in this PR and do not exist at that tag, so both links 404 and `Docs Link Check / online-changed` fails. switched them to `tree/master/`, the convention PR #4670 used for brand-new examples; the release bot pins these links to the release tag at release time (commit d8a0cb35c). note the online link check still fails on the PR itself until the branch merges, the same transient failure #4670 shipped with; q: fine to keep master links, or prefer dropping the links until the next release?

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +84,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+				// without_initial_secret defaults to false on create, so setting it to true on update must be rejected.
+				Config:      configBasic(orgID, name, "description", []string{"ORG_READ_ONLY"}, nil, "without_initial_secret = true"),
+				PlanOnly:    true,
+				ExpectError: regexp.MustCompile("without_initial_secret cannot be updated"),
+			},
+		},
+	})
+}
+
+func TestAccServiceAccount_withoutInitialSecret(t *testing.T) {
```
EspenAlbert: `TestAccServiceAccount_withoutInitialSecret` and `TestAccServiceAccount_withoutInitialSecretWithExpiration` still run unconditionally, but `withoutInitialSecret` is only available in cloud-dev and the scheduled suite routes Sunday runs to `atlas_cloud_env=qa`, so the service_account group goes red in QA every Sunday until rollout. `SkipUnlessCloudDev`/`InCloudDev` are committed in `internal/testutil/acc/skip.go` with no callers, so the gating calls appear to have been dropped from commit 1e6623c77. re-applying the `acc.SkipUnlessCloudDev` calls so default dev CI keeps running these tests and QA/prod runs skip

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +84,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+				ImportStateVerify:                    true,
+				ImportStateVerifyIgnore:              []string{"without_initial_secret"},
+			},
+		},
+	})
+}
+
+// TestAccServiceAccount_withoutInitialSecretWithExpiration verifies the API mutual exclusion contract:
+// sending withoutInitialSecret together with secretExpiresAfterHours fails the apply with the API error.
+func TestAccServiceAccount_withoutInitialSecretWithExpiration(t *testing.T) {
```
EspenAlbert: the neither-set validation path is still uncovered: the mutual-exclusion case has a test, but no test omits both `without_initial_secret` and `secret_expires_after_hours`, which the API rejects with `ARGUMENT_NOT_PROVIDED` now that plan-time validation is deliberately absent. adding `TestAccServiceAccount_neitherSet` (cloud-dev gated, `CheckDestroy` kept, expects the apply-time API error)

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +84,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet(dataSourceName, "client_id"),
+				),
+			},
+			{
+				ResourceName:                         resourceName,
+				ImportStateIdFunc:                    importStateIDFunc(resourceName),
+				ImportStateVerifyIdentifierAttribute: "client_id",
+				ImportState:                          true,
+				ImportStateVerify:                    true,
+				ImportStateVerifyIgnore:              []string{"without_initial_secret"},
```
EspenAlbert, line 141: verified manually both migration and upgrade experience.
Similar to `delete_on_create_timeout`:


Upgrading is invisible. Importing a `true`-created account into a config that keeps `true`
always errors at plan time, and the attribute cannot be reconstructed from the API.

The generated doc's note says `without_initial_secret` "should be omitted when importing",
which covers the removal. The error text points at "use the state value", which does not
exist on this path because the state value is null.

docs/resources/service_account.md
```diff
@@ -10,6 +10,8 @@ subcategory: "Service Accounts"
 
 ## Example Usages
 
+The following example creates a Service Account and lets Atlas generate its first secret.
```
manupedrozo: What do you think of changing docs to suggest `without_initial_secret = true` by default? 
It provides the best terraform experience by avoiding the need to import the generated secret so I'd lead customers in that direction.
We could also have both approaches here.

I think the recommended approach should be to define an SA with no secret + 1 secret resource.

examples/mongodbatlas_service_account/bootstrap_secret/providers.tf
```diff
@@ -0,0 +1,2 @@
+provider "mongodbatlas" {
+}
```
manupedrozo: no credential vars?

examples/mongodbatlas_service_account/without_initial_secret/providers.tf
```diff
@@ -0,0 +1,2 @@
+provider "mongodbatlas" {
```
manupedrozo: Same here

examples/mongodbatlas_service_account/without_initial_secret/main.tf
```diff
@@ -0,0 +1,31 @@
+# Create a Service Account without a bootstrap secret, then create the first managed secret explicitly.
```
manupedrozo: Nice, I think we should present this more strongly as the default, recommended approach.

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +84,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+		orgID = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		name  = acc.RandomName()
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		CheckDestroy:             checkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: configBasic(orgID, name, "Without initial secret", []string{"ORG_READ_ONLY"}, nil, "without_initial_secret = true"),
```
manupedrozo: Can we pass a bool to `configBasic` and `checkBasic` for `without_initial_secret` instead of sending in hcl only to `configBasic`?

tools/codegen/codespec/config.go
```diff
@@ -449,8 +449,14 @@ func ApplyDeleteOnCreateTimeoutTransformation(resource *Resource) {
...
-	// CreateOnly plan modifier will not be applied for computed attributes
-	if attr.ComputedOptionalRequired == Computed || attr.ComputedOptionalRequired == ComputedOptional {
+	// A plain CreateOnly plan modifier is not applied to computed attributes: their value may be
+	// unknown at plan time, which would falsely fail the update validation. The exception is a
+	// computed optional bool with a default that the API never returns, where
+	// CreateOnlyBoolWithDefault sets the default and keeps the create-only behavior.
+	if attr.ComputedOptionalRequired == Computed {
+		return
+	}
+	if attr.ComputedOptionalRequired == ComputedOptional && !isCreateOnlyBoolWithDefault(attr) {
```
manupedrozo: If we leave the attribute optional only, could we avoid the override + this change?
Seems like we are forcing computed to an attribute that given the api spec is not and then we are fixing it back here, I might be missing something!

If we do want this behavior, I think we should add this particular createOnly plan modifier through config/custom hook instead of changing the codegen for a case we are forcing.

manupedrozo (commented):
Nice, left comments.

* Planning to cover `project_service_account` changes and the [rotation guide](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/service-account-secret-rotation) in a separate PR?
* Ideally we'd also update other examples/guides to use the `without_initial_secret` attribute. A relevant one is the [PAK to SA migration guide](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/migrate-pak-to-service-account).

docs/resources/service_account.md
EspenAlbert: Agreed, done. The resource docs now lead with the `without_initial_secret` flow: the Example Usages section inlines `without_initial_secret/main.tf` (Service Account plus a `mongodbatlas_service_account_secret` resource) framed as the recommended approach, and the bootstrap flow moved to Further Examples as the alternative. Both approaches stay documented, and the examples README now presents the same ordering.

examples/mongodbatlas_service_account/bootstrap_secret/providers.tf
EspenAlbert: Restored. `bootstrap_secret/` now has a `provider.tf` wiring `client_id`/`client_secret` to `var.atlas_client_id`/`var.atlas_client_secret`, the two variables in `variables.tf`, and the README lists them under Variables, matching `flex_cluster` and `resource_policy`.

examples/mongodbatlas_service_account/without_initial_secret/providers.tf
EspenAlbert: Same fix here: `provider.tf` with the credential vars, `variables.tf` entries, and the README variable list.

examples/mongodbatlas_service_account/without_initial_secret/main.tf
EspenAlbert: Done. The example README now opens by stating this is the recommended way to create a Service Account, with the why: every secret is declared in the configuration, so there is no bootstrap secret to import and no targeted apply. The resource docs present it as the primary example too (see the resource-doc thread above).

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Done. `configBasic` now takes `withoutInitialSecret bool` and emits the attribute line from it (the `extraArgs` variadic is gone), and `checkBasic` takes the bool and asserts the stored value, so the basic test also verifies the injected `false` default end to end.

tools/codegen/codespec/config.go
EspenAlbert: You're not missing anything on the spec side: `withoutInitialSecret` is request-only and never returned, so the natural model is optional. The computed forcing exists to carry the default, and the default exists for the guardrail:

- optional-only without a default emits plain `CreateOnly()`, whose validation only fires when state is non-null. Created with the attribute omitted (the common path now that it's optional), state stays null, so a later `without_initial_secret = true` in config is silently accepted: no plan error, and the PATCH body omits the field so the API is never told. The update-rejection step in `TestAccServiceAccount_createOnlyAttributes` would also stop failing.
- optional-only with the default doesn't work either: `CreateOnlyBoolWithDefault` writes the default into the plan on create, and the framework rejects plan values that diverge from the config on non-computed attributes. That's why the modifier's doc says to always use Optional+Computed with a default.
- so ComputedOptional it is (the config.yml override), and then the codegen guard needed narrowing because model-gen serializes `create_only` through `setCreateOnlyValue`: with the old guard, regenerating the model drops `create_only` for this attribute, and hand-setting it in the yaml would pass code-gen CI but silently revert on the next model-gen.

The narrowed case also mirrors the existing `delete_on_create_timeout` shape (ComputedOptional + default + create-only + never in response), which today is only reachable through the hand-injected timeout transformation; the change makes that combination reachable for spec-derived attributes. If you'd rather not touch codegen, the alternative is keeping the model natural and wiring `CreateOnlyBoolWithDefault` through a resource-specific schema hook; happy to switch, but it special-cases what the guard change generalizes.

internal/serviceapi/serviceaccount/resource_test.go
```diff
@@ -83,19 +84,75 @@ func TestAccServiceAccount_createOnlyAttributes(t *testing.T) {
...
+		orgID = os.Getenv("MONGODB_ATLAS_ORG_ID")
+		name  = acc.RandomName()
+	)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { acc.PreCheckBasic(t) },
+		ProtoV6ProviderFactories: acc.TestAccProviderV6Factories,
+		Steps: []resource.TestStep{
+			{
+				Config:      configBasic(orgID, name, "Without initial secret", []string{"ORG_READ_ONLY"}, new(24), "without_initial_secret = true"),
+				ExpectError: regexp.MustCompile("(?i)mutually|withoutInitialSecret|exclusive"),
```
manupedrozo: I would avoid introducing API error tests, this is the type of test that should be on the API side (which I would assume it already is).

docs/resources/service_account.md
EspenAlbert: Implemented. The resource docs now lead with `without_initial_secret` and an explicit `mongodbatlas_service_account_secret` as the recommended workflow; the bootstrap-secret flow remains a Further Example. The example README uses the same ordering.

examples/mongodbatlas_service_account/bootstrap_secret/providers.tf
EspenAlbert: Implemented. `bootstrap_secret/` now has `provider.tf`, the credential variables in `variables.tf`, and the README variable list, matching `flex_cluster` and `resource_policy`.

examples/mongodbatlas_service_account/without_initial_secret/providers.tf
EspenAlbert: Implemented here too: `provider.tf` uses the credential variables, `variables.tf` declares them, and the README lists them.

examples/mongodbatlas_service_account/without_initial_secret/main.tf
EspenAlbert: Implemented. The example README and primary resource-doc example now identify this as the recommended path because every secret stays in configuration, with no bootstrap-secret import or targeted apply.

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Implemented. `configBasic` and `checkBasic` now both take `withoutInitialSecret bool`; the former emits the configuration and the latter asserts the stored value.

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Agreed, removed `TestAccServiceAccount_withoutInitialSecretWithExpiration`. The provider delegates both invalid-request combinations to Atlas, so API-side tests own those error contracts; provider coverage stays focused on the successful no-bootstrap create and the local create-only plan modifier.

docs/resources/service_account.md
```diff
@@ -59,7 +65,8 @@ output "service_accounts_results" {
 
 ### Optional
 
-- `secret_expires_after_hours` (Number) The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. This attribute is required when creating the Service Account and you cannot update it later.
+- `secret_expires_after_hours` (Number) The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. Required when `without_initial_secret` is false or omitted. Must not be set when `without_initial_secret` is true. Cannot be updated after creation.
```
JuliaMongo: Do not set this field if you set `without_initial-secret` to `true`. You can't update this field after you create the secret.

docs/resources/service_account.md
```diff
@@ -59,7 +65,8 @@ output "service_accounts_results" {
...
+- `without_initial_secret` (Boolean) When true, creates the Service Account without generating an initial secret. `secret_expires_after_hours` must not be set when this is true. Defaults to false.
```
JuliaMongo: do not set .... (instead of must not be set) -- active voice. So:
If this field is set to `true`, do not set `secret_expires....

docs/resources/service_account.md
```diff
@@ -86,6 +93,6 @@ Import the Service Account resource by using the Organization ID and Client ID i
 $ terraform import mongodbatlas_service_account.test 6117ac2fe2a3d04ed27a987v/mdb_sa_id_1234567890abcdef12345678
 ```
 
--> **NOTE:** `secret_expires_after_hours` is not populated during import and should be omitted in the resource definition when importing the resource.
+-> **NOTE:** `secret_expires_after_hours` and `without_initial_secret` are not populated during import and should be omitted in the resource definition when importing the resource.
```
JuliaMongo: nit: suggest: omit them in the resource definition -- change to active voice from "should be omitted"

examples/mongodbatlas_service_account/bootstrap_secret/README.md
```diff
@@ -0,0 +1,50 @@
+# Service Account with a bootstrap secret
+
+Create a Service Account and let Atlas generate its first secret by setting `secret_expires_after_hours`.
```
JuliaMongo: Does "first" imply that there will be subsequent secrets created after this first one expires?

Note: as written this sounds like setting the "after hours" flag creates a service account and causes Atlas to generate its first secret. I would separate the mention of "secret expires " into a new sentence. 
Create a service Account in which Atlas: - generates a secret without you having to provide it, and 
- sets `secret expired after hours`

examples/mongodbatlas_service_account/bootstrap_secret/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+The secret value is returned only once, at creation time. The example exposes it through the sensitive output `service_account_first_secret`.
+
+## Prerequisites
+
+- Service Account with Organization Owner permissions used for provider authentication.
+- Credentials come from the environment (`MONGODB_ATLAS_CLIENT_ID` and `MONGODB_ATLAS_CLIENT_SECRET`).
```
JuliaMongo: Credentials come from the following environment variables:

examples/mongodbatlas_service_account/bootstrap_secret/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Variables
+
+- `org_id`: Atlas Organization ID where the Service Account is created.
```
JuliaMongo: where you create the Service Account 

active voice

examples/mongodbatlas_service_account/without_initial_secret/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+**3. Plan and apply.**
+
+```bash
+terraform plan
+terraform apply
+```
+
+`without_initial_secret` and `secret_expires_after_hours` are mutually exclusive. Set `without_initial_secret = true` and omit `secret_expires_after_hours`, or set `secret_expires_after_hours` and omit `without_initial_secret`. Atlas rejects a create request that sets both.
```
JuliaMongo: This bit would be useful to repeat in the relevant other places, I think.

templates/resources/service_account.md.tmpl
```diff
@@ -20,6 +26,6 @@ Import the Service Account resource by using the Organization ID and Client ID i
 $ terraform import mongodbatlas_service_account.test 6117ac2fe2a3d04ed27a987v/mdb_sa_id_1234567890abcdef12345678
 ```
 
--> **NOTE:** `secret_expires_after_hours` is not populated during import and should be omitted in the resource definition when importing the resource.
+-> **NOTE:** `secret_expires_after_hours` and `without_initial_secret` are not populated during import and should be omitted in the resource definition when importing the resource.
```
JuliaMongo: change to active voice: omit .....

tools/codegen/models/service_account.yaml
```diff
@@ -76,7 +76,7 @@ schema:
           present_in_any_response: true
           request_only_required_on_create: false
         - int64: {}
-          description: The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. This attribute is required when creating the Service Account and you cannot update it later.
+          description: The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. Required when `without_initial_secret` is false or omitted. Must not be set when `without_initial_secret` is true. Cannot be updated after creation.
```
JuliaMongo: Do not set (as commented before). Active voice.

tools/codegen/models/service_account.yaml
```diff
@@ -183,6 +183,18 @@ schema:
           create_only: false
           present_in_any_response: true
           request_only_required_on_create: false
+        - bool:
+            default: false
+          description: When true, creates the Service Account without generating an initial secret. `secret_expires_after_hours` must not be set when this is true. Defaults to false.
```
JuliaMongo: do not set ... active voice

tools/codegen/config.yml
```diff
@@ -1257,8 +1257,14 @@ resources:
         created_at:
           immutable_computed: true
         secret_expires_after_hours:
-          # Adding clarification that the attribute is required for creation and can't be updated later.
-          description: "The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. This attribute is required when creating the Service Account and you cannot update it later."
+          # Clarifying that the attribute is required unless the account is created without an initial secret, and can't be updated later.
+          description: "The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. Required when `without_initial_secret` is false or omitted. Must not be set when `without_initial_secret` is true. Cannot be updated after creation."
```
JuliaMongo: when you set .... to `false` or when you omit the `...`.

tools/codegen/config.yml
```diff
@@ -1257,8 +1257,14 @@ resources:
...
+        without_initial_secret:
+          # Optional+Computed because the create-only plan modifier injects a default in the plan on create.
+          computability:
+            optional: true
+            computed: true
+          description: "When true, creates the Service Account without generating an initial secret. `secret_expires_after_hours` must not be set when this is true. Defaults to false."
```
JuliaMongo: When this flag is set to `true`, do not set `...`

JuliaMongo (approved):
LGTM for copy review with a few suggestions to use active voice in a few places (all optional).

tools/codegen/config.yml
```diff
@@ -1257,8 +1257,14 @@ resources:
...
         secret_expires_after_hours:
-          # Adding clarification that the attribute is required for creation and can't be updated later.
-          description: "The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. This attribute is required when creating the Service Account and you cannot update it later."
+          # Clarifying that the attribute is required unless the account is created without an initial secret, and can't be updated later.
+          description: "The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. Required when `without_initial_secret` is false or omitted. Must not be set when `without_initial_secret` is true. Cannot be updated after creation."
+        without_initial_secret:
+          # Optional+Computed because the create-only plan modifier injects a default in the plan on create.
+          computability:
+            optional: true
+            computed: true
```
AgustinBettati: Not fully clear here (comment justifying why O+C). 
Isnt the why related to `without_initial_secret` being returned in the response?

tools/codegen/codespec/config.go
AgustinBettati: Got the same doubt, would be clear on API behaviour and work backwards from there.

EspenAlbert:
> Nice, left comments.
> 
> * Planning to cover `project_service_account` changes and the [rotation guide](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/service-account-secret-rotation) in a separate PR?
> * Ideally we'd also update other examples/guides to use the `without_initial_secret` attribute. A relevant one is the [PAK to SA migration guide](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/migrate-pak-to-service-account).

Thanks for the comments, will do a tech sync item to make the decision, seems tempting to simplify this for users and use without_initial_secret when the secret_expires is not set.

docs/resources/service_account.md
EspenAlbert: Implemented. The primary resource-doc example now uses `without_initial_secret` with an explicit `mongodbatlas_service_account_secret`; the bootstrap-secret flow remains the alternative.

examples/mongodbatlas_service_account/bootstrap_secret/providers.tf
EspenAlbert: Keeping the provider block empty so the example uses the provider's standard environment configuration. The README now explicitly lists `MONGODB_ATLAS_CLIENT_ID` and `MONGODB_ATLAS_CLIENT_SECRET`.

examples/mongodbatlas_service_account/without_initial_secret/providers.tf
EspenAlbert: Same approach here: empty provider block, with the README explicitly listing the standard environment variables.

examples/mongodbatlas_service_account/without_initial_secret/main.tf
EspenAlbert: Implemented. The example README and primary resource docs identify this as the recommended path because every secret stays in configuration, with no bootstrap-secret import or targeted apply.

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Implemented. `configBasic` and `checkBasic` now take `withoutInitialSecret bool`; the configuration helper emits the field and the check helper asserts its stored value.

tools/codegen/codespec/config.go
EspenAlbert: The API contract is an optional request-only boolean with default `false`; Atlas does not return it on reads. We need the provider to retain an explicit `false` in state after an omitted create, otherwise a later `without_initial_secret = true` is a no-op PATCH that `CreateOnly()` cannot reject from null state. Optional-only cannot receive the plan-time default because the framework rejects a defaulted plan value that differs from a non-computed configuration. The ComputedOptional override and `CreateOnlyBoolWithDefault` therefore preserve the create-only guardrail; the generator change makes model regeneration retain that modifier instead of requiring a fragile hand edit. A resource schema hook can implement the same special case, but would not remove the required Optional+Computed model behavior.

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Agreed, removed `TestAccServiceAccount_withoutInitialSecretWithExpiration`. Atlas owns invalid-request contract coverage; the provider suite retains successful no-bootstrap creation and local create-only behavior.

docs/resources/service_account.md
EspenAlbert: Updated the schema descriptions to state both valid choices directly: set this field when you set `without_initial_secret` to false or omit it; do not set it when you set `without_initial_secret` to true; you cannot update it after you create the Service Account.

docs/resources/service_account.md
EspenAlbert: Updated to imperative wording: if you set `without_initial_secret` to true, do not set `secret_expires_after_hours`.

docs/resources/service_account.md
EspenAlbert: Updated the import note to instruct users to omit both request-only inputs in the resource definition.

examples/mongodbatlas_service_account/bootstrap_secret/README.md
EspenAlbert: Reworded the bootstrap example to separate account creation, Atlas-generated secret creation, and `secret_expires_after_hours`.

examples/mongodbatlas_service_account/bootstrap_secret/README.md
EspenAlbert: Updated both example READMEs to list `MONGODB_ATLAS_CLIENT_ID` and `MONGODB_ATLAS_CLIENT_SECRET` explicitly.

examples/mongodbatlas_service_account/bootstrap_secret/README.md
EspenAlbert: Updated `org_id` to describe the organization where the configuration creates the Service Account.

examples/mongodbatlas_service_account/without_initial_secret/README.md
EspenAlbert: Repeated the mutually exclusive input guidance in the resource docs, the examples overview, and the bootstrap-secret example.

templates/resources/service_account.md.tmpl
EspenAlbert: Updated the template import instruction to say "omit".

tools/codegen/models/service_account.yaml
EspenAlbert: Updated the source model description to use the active imperative wording.

tools/codegen/models/service_account.yaml
EspenAlbert: Updated the source model description to instruct users not to set the expiration field with this flag.

tools/codegen/config.yml
EspenAlbert: Updated the override description to distinguish setting `without_initial_secret` to false from omitting it.

tools/codegen/config.yml
EspenAlbert: Updated the override description to say that when you set this field to true, do not set `secret_expires_after_hours`.

tools/codegen/config.yml
EspenAlbert: Clarified the comment: Optional+Computed permits the modifier's create-time default, and Atlas omits this create-only field from responses, so create/read retain the planned value in state.

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Implemented. `configBasic` now takes `withoutInitialSecret bool`; `checkBasic` checks for the configured `true` value or an omitted attribute, matching the optional-only schema.

tools/codegen/codespec/config.go
EspenAlbert: Implemented the optional-only design: the provider clears the API default and uses the normal create-only modifier, so it does not write an unconfigured value to state. This removes the computed override and prior specialized generator path. The PR description still needs the corresponding Optional-only/ClearDefault update.

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: Agreed, removed `TestAccServiceAccount_withoutInitialSecretWithExpiration`. Atlas owns invalid-request contract coverage; provider tests retain successful no-bootstrap creation and local plan behavior.

docs/resources/service_account.md
EspenAlbert: Thank you, changed in 0f737c32d53e386e4d8be3d0c3c9f275003aff0b

docs/resources/service_account.md
EspenAlbert: Updated the authored import note to instruct users to omit both request-only fields when importing a Service Account.

templates/resources/service_account.md.tmpl
EspenAlbert: Updated with the same imperative import instruction.

tools/codegen/models/service_account.yaml
EspenAlbert: Thank you, changed in 0f737c3

tools/codegen/config.yml
EspenAlbert: Thank you, changed in 0f737c3

examples/mongodbatlas_service_account/README.md
```diff
@@ -1,27 +1,61 @@
...
-terraform output -raw service_account_first_secret
+terraform output -raw secret
 ```
 
-For secret rotation, see [Guide: Service Account Secret Rotation](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/service-account-secret-rotation).
+For managing and rotating secrets, see [Guide: Service Account Secret Rotation](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/guides/service-account-secret-rotation).
+
+## Prerequisites
+
+- Service Account with Organization Owner permissions used for provider authentication.
```
EspenAlbert, line 21: the root example config assigns provider credentials from `var.atlas_client_id` and `var.atlas_client_secret`, but this README only tells users to set `MONGODB_ATLAS_CLIENT_ID` and `MONGODB_ATLAS_CLIENT_SECRET` and its `terraform.tfvars` snippet contains only `org_id`. copying it leaves both configured provider inputs at their empty defaults. document the credential variables and include them in the `terraform.tfvars` example

internal/serviceapi/serviceaccount/resource_test.go
EspenAlbert: no longer valid

docs/resources/service_account.md
EspenAlbert: Thank you, changed in https://github.com/mongodb/terraform-provider-mongodbatlas/commit/0f737c32d53e386e4d8be3d0c3c9f275003aff0b

examples/mongodbatlas_service_account/bootstrap_secret/README.md
EspenAlbert: example removed

examples/mongodbatlas_service_account/bootstrap_secret/README.md
EspenAlbert: thank you, example removed

examples/mongodbatlas_service_account/bootstrap_secret/README.md
EspenAlbert: thank you, example removed

examples/mongodbatlas_service_account/without_initial_secret/README.md
EspenAlbert: Thanks, updated in the description

tools/codegen/config.yml
EspenAlbert: Thank you, changed in 0f737c3

tools/codegen/config.yml
EspenAlbert: discussed offline, now optional only

examples/mongodbatlas_service_account/bootstrap_secret/providers.tf
EspenAlbert: added

examples/mongodbatlas_service_account/without_initial_secret/providers.tf
EspenAlbert: added

docs/resources/service_account.md
EspenAlbert: good idea, done

.changelog/4727.txt
```diff
@@ -0,0 +1,3 @@
...
+resource/mongodbatlas_service_account: Adds `without_initial_secret` attribute to create a Service Account without generating an initial secret. `secret_expires_after_hours` is now optional when `without_initial_secret` is `true`
```
AgustinBettati, line 2: Shouldn't `project_service_account` have the same change? Or was this not added on the API side?

tools/codegen/codespec/config.go
```diff
@@ -389,6 +392,23 @@ func getComputabilityFromConfig(computability config.Computability) ComputedOpti
 	return Required
 }
 
+// clearStaticDefault drops a static default copied from the API spec. Code generation otherwise emits
+// a schema Default, which forces computed_optional and is rejected by terraform-plugin-framework on a
```
AgustinBettati, line 396: `generation otherwise emits a schema Default`
Is this statement accurate? From what I recall we used API Spec default to determine if an attribute was Optional + Computed, or only Optional (no default value actually reflected in schema). With this in mind, curious of config override of defining it is optional only is sufficient.

AgustinBettati (commented):
Thanks for followups! 2 smaller questions on my side

internal/serviceapi/serviceaccount/resource_schema.go
```diff
@@ -88,6 +88,11 @@ func ResourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "Indicates whether the Service Account is system managed.",
 			},
+			"without_initial_secret": schema.BoolAttribute{
+				Optional:            true,
+				MarkdownDescription: "When true, creates the Service Account without generating an initial secret. If you set this field to true, do not set `secret_expires_after_hours`.",
+				PlanModifiers:       []planmodifier.Bool{customplanmodifier.CreateOnly()},
```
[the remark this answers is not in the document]
EspenAlbert: Valid. Keeping `without_initial_secret` Optional-only, but adding a resource-specific schema modifier that rejects a non-null value on update when the current state is null. This prevents the no-op PATCH and the incorrect plan-to-state value while preserving the cleared API default.

EspenAlbert (commented):
> `generation otherwise emits a schema Default`
> Is this statement accurate? From what I recall we used API Spec default to determine if an attribute was Optional + Computed, or only Optional (no default value actually reflected in schema). With this in mind, curious of config override of defining it is optional only is sufficient. 

Good catch, you're right -- the modifier is chosen regardless of computability, and that is the actual bug. Fixed it there instead of working around it.

`planModifierStatement` selected `CreateOnlyBoolWithDefault` for any create-only bool with a default, ignoring `ComputedOptionalRequired`. But a default on a non-computed attribute is rejected by terraform-plugin-framework, so that modifier is only valid for `computed_optional`. The generator now checks for `ComputedOptional` before choosing it and otherwise falls back to the plain `CreateOnly()`.

With that, the `computability: optional` override is sufficient by itself and `clear_default` is gone, along with the `ClearDefault` override field, `clearStaticDefault`, and their tests.

Why it matters for this attribute: the `dev` spec declares `default: false` on `withoutInitialSecret`. `CreateOnlyBoolWithDefault(false)` treats a null plan value as not-known and injects `false` on create, so an omitted `without_initial_secret` would be written to state as `false` even though the user never configured it. After import a later `without_initial_secret = true` would then be a `false -> true` transition and get rejected even though Atlas never applied the flag. Plain `CreateOnly()` keeps the omitted value null.

This makes the field behave like `secret_expires_after_hours`, which has the identical model shape (`optional`, `create_only`, `omit_in_update_body`, not present in any response): configured value is sent on create, omitted value stays null, and the provider never invents a value.

Blast radius is zero: regenerating every resource changes only `without_initial_secret`. All nine existing `CreateOnlyBoolWithDefault` uses are `delete_on_create_timeout`, which is `computed_optional`, so they keep the default-injecting modifier. `schema_attribute_test.go` covers both branches.

We are not adding a resource-local modifier for the null -> true update case. After an import both attributes are null, and setting either in config produces a non-empty plan whose PATCH omits the field -- a harmless no-op with no server-side change. That simple, consistent UX is preferable to failing hard on an attribute that is merely create-only.

EspenAlbert (commented):
> Shouldn't `project_service_account` have the same change? Or was this not added on the API side? 

Follow-up work

docs/data-sources/service_account.md
oarbusi, line 1: the data source is not being used in this docs example? only resource. should we include the data source in the example?

docs/resources/service_account.md
```diff
@@ -59,7 +59,8 @@ output "service_accounts_results" {
 
 ### Optional
 
-- `secret_expires_after_hours` (Number) The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. This attribute is required when creating the Service Account and you cannot update it later.
+- `secret_expires_after_hours` (Number) The expiration time of the new Service Account secret, provided in hours. The minimum and maximum allowed expiration times are subject to change and are controlled by the organization's settings. Set this field when you set `without_initial_secret` to false or omit `without_initial_secret`. Do not set this field when you set `without_initial_secret` to true. You cannot update this field after you create the Service Account.
+- `without_initial_secret` (Boolean) When true, creates the Service Account without generating an initial secret. If you set this field to true, do not set `secret_expires_after_hours`.
```
oarbusi, line 63: nit: i would be more explicit on the fact that this flag only has an effect at create time

oarbusi (approved):

tools/codegen/codespec/config.go
EspenAlbert: Update: the generator logic changed again after this reply, and the PR body now describes the shipped design. `without_initial_secret` stays Optional-only with the API default retained in the model; the schema generator emits `CreateOnlyBoolWithDefault` only for ComputedOptional attributes and plain `CreateOnly()` for Optional-only ones. No clear_default mechanism.

internal/serviceapi/serviceaccount/resource_schema.go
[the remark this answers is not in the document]
EspenAlbert: Valid observation, and the behavior is a deliberate trade-off in this PR. `without_initial_secret` is a create-time request input that Atlas never returns, so the provider retains configured create-only values in state the same way it does for `secret_expires_after_hours`. A user who adds it to an existing configuration gets a no-op apply; the `secrets` list in state continues to truthfully show the bootstrap secret Atlas generated. The alternative is a resource-local modifier that rejects a non-null value on a null state (including after import), which we can add if strictness is preferred over config-mirroring; happy to switch on maintainer preference.

AgustinBettati (approved):
LGTM, thanks for the details, I'm aligned with the approach

PR #4735, opened 2026-09-16
feat: Creates Service Account without an initial secret when secret_expires_after_hours is omitted

EspenAlbert:
## Description

`mongodbatlas_service_account` could not create a Service Account without a bootstrap secret. PR [#4727](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4727) exposed a new `without_initial_secret` attribute to do it. Review asked whether the flag is needed at all.

This PR takes the alternative approach: no new attribute. Omitting `secret_expires_after_hours` is the signal for "create without a bootstrap secret", and the provider sets `withoutInitialSecret: true` on the POST body for you.

## Approach

- `tools/codegen/config.yml` ignores `without_initial_secret` in the resource schema. Users never see it and it never lands in state.
- `internal/serviceapi/serviceaccount/resource_custom_hooks.go` drops the `RequestOnlyRequiredOnCreate` plan modifier from `secret_expires_after_hours`, so omission is valid on create. `CreateOnly` stays, so changing the hours after creation still fails.
- The same file adds `PreCreateAPICall`, which sets `withoutInitialSecret: true` when `secretExpiresAfterHours` is absent from the body. A decode or marshal error returns the original body so the API fails on it instead of silently creating an account with a bootstrap secret.

Empty `secrets` is the only signal that an account was created without a secret. Import and re-plan show `No changes`, matching the existing behavior for `secret_expires_after_hours`.

## Trade-off

Removing the guardrail means a forgotten `secret_expires_after_hours` now creates an account with no bootstrap secret instead of failing the plan. That is the accepted decision: omission is documented as the recommended flow, which leads customers away from managing an org-owned bootstrap secret in Terraform state.

## API spec

The tracked OpenAPI spec in `tools/codegen/atlasapispec/` is stale for both `withoutInitialSecret` and `systemManaged`. No refreshed spec is part of this PR: the upstream dev spec is a local input only, and both the raw and flattened spec files stay at HEAD. The flow used:

```sh
make autogen-update-api-spec spec_source=/tmp/upstream-v2-dev.yaml   # local spec only
make autogen-generate-resources resource_name=service_account step=model-gen
make autogen-generate-resources resource_name=service_account step=code-gen
git checkout -- tools/codegen/atlasapispec/                          # keep the spec out of the PR
```

Regenerating against the committed spec instead would drop `system_managed`, so `step=model-gen` is only valid with the refreshed local spec. Both `model-gen` and `code-gen` are idempotent under the flow above.

## Notes for reviewers

- PR [#4727](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4727) is left open as the explicit-flag alternative. The two cannot both expose a `without_initial_secret` attribute. This PR does not touch it.
- Follow-up work, not in this PR: applying the same behavior to `mongodbatlas_project_service_account` for project-level parity, and updating the [Service Account Secret Rotation guide](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/docs/guides/service-account-secret-rotation.md) to cover the no-initial-secret flow.
- The `resource_custom_hooks.go` schema hook compares plan modifier types with `reflect.TypeOf` rather than an interface assertion. `customplanmodifier.RequestOnlyRequiredOnCreateModifier` and `CreateOnlyModifier` have identical method sets, so an assertion matches both and silently drops `CreateOnly` as well.

## Type of change:

## Required Checklist:


EspenAlbert:
decided to stick with the optional-only approach

PR #4737, opened 2026-09-16
feat: Add OAuth 2.0 support to mongodbatlas_metric_integration resource

marcabreracast:
## Description

Adds OAuth 2.0 support to the `mongodbatlas_metric_integration` resource and its data sources, alongside the existing header-based authentication. With `auth_type = "OAUTH2"`, the integration authenticates to an OTel destination through an OAuth 2.0 token endpoint, using either a shared client secret (`client_auth_method = "CLIENT_SECRET"`) or an Atlas-managed signing key (`client_auth_method = "PRIVATE_KEY_JWT"`).

Adds an `oauth` block with:
- `client_auth_method`, `token_endpoint`, `client_id`
- `scopes`, `token_request_params` (optional)
- `client_secret` (write-only, `CLIENT_SECRET` only)
- `signing_key_info` (read-only, `PRIVATE_KEY_JWT` only): `kid`, `algorithm`, `jwks_uri`, `created_at`

The `client_secret` is write-only and never returned by the API. For `PRIVATE_KEY_JWT`, Atlas generates and manages the signing key server-side.

Link to any related issue(s): CLOUDP-412204

## Type of change:

## Required Checklist:

## Further comments

Validated locally against cloud-dev on a project with the OAuth feature flag enabled:

```
--- PASS: TestAccMetricIntegration_oauthClientSecret (10.45s)
--- PASS: TestAccMetricIntegration_oauthPrivateKeyJWT (8.08s)
--- PASS: TestAccMetricIntegration_oauthPrivateKeyJWTRejectsClientSecret (0.95s)
--- PASS: TestAccMetricIntegration_oauthToHeader (7.98s)
PASS
ok  github.com/mongodb/terraform-provider-mongodbatlas/internal/serviceapi/metricintegration
```

The OAuth acceptance tests are temporarily skipped in CI because the OAuth metric integration fields are still behind a feature flag and not available on the CI test project. They run locally against a cloud-dev project with the flag enabled and will be un-gated once the fields are available in prod.

Examples for the new `oauth` block are deferred to CLOUDP-445734 and will be added there.

The OpenAPI spec changes in this PR are committed from the `dev` branch of `mongodb/openapi`, since the OAuth metric integration fields are only available there and not yet in the prod spec. This will be re-synced to prod once the fields are deployed.


marcabreracast:
Test failures are unrelated to this implementation. Note that currently new added tests are being skipped in CI, see **Further Notes** section of the PR description for further details

docs/resources/metric_integration.md
```diff
@@ -64,13 +64,17 @@ output "metric_integration_ids" {
...
+- `auth_type` (String) Authentication method the integration uses when exporting metrics to the endpoint. `HEADER` authenticates with the static HTTP headers provided in the `headers` field, which must be set when this value is used. `OAUTH2` acquires a bearer token from an OAuth 2.0 token endpoint using the `oauth` field.
 - `endpoint` (String) OpenTelemetry collector endpoint URL. Must use HTTPS.
 - `integration_type` (String) Type of metric integration. Identifies which protocol will be used for the integration. This value cannot be modified after the integration is created.
 - `metric_selection` (Set of String) Array of metric categories to export. Determines which types of metrics are sent to the integration.
 - `project_id` (String) Unique 24-hexadecimal digit string that identifies your project, also known as `groupId` in the official documentation.
 - `provider_type` (String) The provider type for the metric integration. Identifies the third-party service provider.
 
+### Optional
```
AgustinBettati, line 74: q: Any examples worth adjusting to cover OAuth?

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -204,3 +216,259 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+		extraAttrs = append(extraAttrs, "scopes = "+hcl.StringSliceToHCL(scopes))
+	}
+	extraAttrs = append(extraAttrs, tokenRequestParams)
+	return configOauth(projectID, endpoint, tokenEndpoint, clientID, "PRIVATE_KEY_JWT", withDS, oauthExtrasHCL(extraAttrs...))
+}
+
+// TestAccMetricIntegration_oauthClientSecret covers the CLIENT_SECRET OAuth path
+func TestAccMetricIntegration_oauthClientSecret(t *testing.T) {
+	// TODO: remove once the OAuth metric integration fields are available in prod and before merging to master
+	acc.SkipTestForCI(t)
```
AgustinBettati: Not sure why "not available in prod" would limit us running these tests in CI against cloud-dev, can you clarify?

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -204,3 +216,259 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+		resource "mongodbatlas_metric_integration" "test" {
+			project_id              = %[1]q
+			integration_type        = "OTEL"
+			provider_type           = "CUSTOM"
+			auth_type               = "OAUTH2"
+			aggregation_temporality = "DELTA"
+			endpoint                = %[2]q
+			metric_selection        = ["ATLAS_STREAM_PROCESSING"]
+
+			oauth = {
```
AgustinBettati, line 413: Do we have a test checking full removal of oauth attribute? Just want confirm the behaviour, I imagine this will unset the configured value.

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -30,8 +34,17 @@ const (
 			depends_on = [mongodbatlas_metric_integration.test]
 		}
 	`
+	// Dummy endpoints for OAuth integrations.
+	oauthTokenEndpoint = "https://192.0.2.2/oauth2/token" //nolint:gosec // Test data
+	oauthClientID      = "atlas-otel-test"                //nolint:gosec // Test data
+	oauthEndpoint      = "https://192.0.2.1/v1/metrics"
+	// Pre-rendered token_request_params HCL fragments, passed to the config builders.
+	trParamsClientSecret  = `token_request_params = { resource = "atlas:otel:test" }`
+	trParamsPrivateKeyJWT = `token_request_params = { resource = "JPMC:URI:OTel" }`
```
AgustinBettati:
```suggestion
	trParamsPrivateKeyJWT = `token_request_params = { resource = "TEST:URI:OTel" }`
```

.changelog/4737.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:enhancement
+resource/mongodbatlas_metric_integration: Adds OAuth 2.0 support
```
JuliaMongo: Can this be made longer, or are we keeping it short because that's the style? 

such as: (taking from the PR description)

Adds OAuth 2.0 support to the `mongodbatlas_metric_integration` resource and its data sources, alongside the existing header-based authentication. With `auth_type = "OAUTH2"`, the integration authenticates to an OTel destination through an OAuth 2.0 token endpoint, using either a shared client secret (`client_auth_method = "CLIENT_SECRET"`) or an Atlas-managed signing key (`client_auth_method = "PRIVATE_KEY_JWT"`).

docs/data-sources/metric_integration.md
```diff
@@ -72,6 +72,7 @@ output "metric_integration_ids" {
 - `headers_redacted` (Attributes List) HTTP headers for authentication and configuration. Values are redacted and never returned in plaintext. (see [below for nested schema](#nestedatt--headers_redacted))
 - `integration_type` (String) Type of metric integration. Identifies which protocol will be used for the integration.
 - `metric_selection` (Set of String) Array of metric categories to export. Determines which types of metrics are sent to the integration.
+- `oauth` (Attributes) OAuth 2.0 configuration returned for a metric integration. Secrets are never returned. (see [below for nested schema](#nestedatt--oauth))
```
JuliaMongo, line 75: should this say: Attributes list, like line 72 does?

Replace "see below for" with "To learn more, see [nested schema] ((#nestedatt--oauth))

docs/data-sources/metric_integration.md
```diff
@@ -82,4 +83,27 @@ Read-Only:
...
+
+<a id="nestedatt--oauth"></a>
+### Nested Schema for `oauth`
+
+Read-Only:
+
+- `client_auth_method` (String) How the client authenticates to the token endpoint.
+- `client_id` (String) OAuth 2.0 client identifier registered with the token endpoint.
+- `scopes` (Set of String) OAuth 2.0 scopes requested on the token.
+- `signing_key_info` (Attributes) Read-only metadata for the Atlas-managed signing key used by `PRIVATE_KEY_JWT`. Present only for that method. Register the jwks_uri with your identity provider. Atlas rotates the underlying key without changing this URL. (see [below for nested schema](#nestedatt--oauth--signing_key_info))
```
JuliaMongo, line 95: should this be Attributes list? 

same comment re: learn more. 
Question re: 
Register the jwks_uri with your identity provider. -- I don't have context as a reader and would be confused. Might need a bit more explanation in words.

docs/data-sources/metric_integration.md
```diff
@@ -82,4 +83,27 @@ Read-Only:
...
+- `token_endpoint` (String) OAuth 2.0 token endpoint URL.
+- `token_request_params` (Map of String) Provider-specific parameters added to the token request.
+
+<a id="nestedatt--oauth--signing_key_info"></a>
+### Nested Schema for `oauth.signing_key_info`
+
+Read-Only:
+
+- `algorithm` (String) Signing algorithm of the Atlas-managed key.
+- `created_at` (String) When the currently active signing key was created. This parameter expresses its value in the ISO 8601 timestamp format in UTC.
```
JuliaMongo, line 105: The timestamp for when Atlas created the ....

docs/data-sources/metric_integration.md
```diff
@@ -82,4 +83,27 @@ Read-Only:
...
+- `jwks_uri` (String) Public JWKS URL serving this integration's signing keys. Fixed for the lifetime of the integration.
```
JuliaMongo, line 106: what does "Fixed for the lifetime of the integration." mean?

might be worth rewording this. 
Also this line 106 contains the definition in words for what is jwks uri -- consider using these same words Public JWKS URL in all other places too where this is discussed.

docs/data-sources/metric_integration.md
```diff
@@ -82,4 +83,27 @@ Read-Only:
...
+- `kid` (String) Key ID stamped on client assertions, the `SHA-1` thumbprint of the key certificate in uppercase hexadecimal. Changes when Atlas rotates the key.
```
JuliaMongo, line 107: shouldn't ID in kid be capitalized? kID? Ignore if this is not possible or if this is already an established convention.
Reads strangely otherwise. 

Also: Key ID stamped on client assertions, the `SHA-1` thumbprint of the key certificate in uppercase hexadecimal. -- this is a run in sentence. Break into two. Start with the fact that this is the key ID that represents the SHA . then say that Atlas stamps it on client assertions. Then talk about its format.

docs/data-sources/metric_integrations.md
```diff
@@ -86,6 +86,7 @@ Read-Only:
 - `integration_type` (String) Type of metric integration. Identifies which protocol will be used for the integration.
 - `metric_integration_id` (String) Unique identifier of the metric integration configuration.
 - `metric_selection` (Set of String) Array of metric categories to export. Determines which types of metrics are sent to the integration.
+- `oauth` (Attributes) OAuth 2.0 configuration returned for a metric integration. Secrets are never returned. (see [below for nested schema](#nestedatt--results--oauth))
```
JuliaMongo, line 89: same comment for To learn more, see

docs/data-sources/metric_integrations.md
```diff
@@ -96,4 +97,27 @@ Read-Only:
...
+
+<a id="nestedatt--results--oauth"></a>
+### Nested Schema for `results.oauth`
+
+Read-Only:
+
+- `client_auth_method` (String) How the client authenticates to the token endpoint.
+- `client_id` (String) OAuth 2.0 client identifier registered with the token endpoint.
+- `scopes` (Set of String) OAuth 2.0 scopes requested on the token.
+- `signing_key_info` (Attributes) Read-only metadata for the Atlas-managed signing key used by `PRIVATE_KEY_JWT`. Present only for that method. Register the jwks_uri with your identity provider. Atlas rotates the underlying key without changing this URL. (see [below for nested schema](#nestedatt--results--oauth--signing_key_info))
```
JuliaMongo, line 109: same comments as above

docs/resources/metric_integration.md
```diff
@@ -64,13 +64,17 @@ output "metric_integration_ids" {
 ### Required
 
 - `aggregation_temporality` (String) The temporality to send to the metric integration.
-- `auth_type` (String) Authentication method the integration uses when exporting metrics to the endpoint. `HEADER` authenticates with the static HTTP headers provided in the `headers` field, which must be set when this value is used.
+- `auth_type` (String) Authentication method the integration uses when exporting metrics to the endpoint. `HEADER` authenticates with the static HTTP headers provided in the `headers` field, which must be set when this value is used. `OAUTH2` acquires a bearer token from an OAuth 2.0 token endpoint using the `oauth` field.
```
JuliaMongo, line 67: this is a run in sentence: `HEADER` authenticates with the static HTTP headers provided in the `headers` field, which must be set when this value is used. Break it into two, and name what you mean when you say "which" -- start a new sentence. Then see if the order of these must be changed, so it is logical. 

Re last sentence about OATH2, -- does this mean that we have some other oath, and that behaves differently but we only mention the new one but don't qualify the existing one? This will look like it is missing context to the reader who may need to compare. Please consider adding parallel structures: 
For this type, ... this behavior happens
For this type (oath2,) this behavior happens

docs/resources/metric_integration.md
```diff
@@ -64,13 +64,17 @@ output "metric_integration_ids" {
...
 - `endpoint` (String) OpenTelemetry collector endpoint URL. Must use HTTPS.
 - `integration_type` (String) Type of metric integration. Identifies which protocol will be used for the integration. This value cannot be modified after the integration is created.
 - `metric_selection` (Set of String) Array of metric categories to export. Determines which types of metrics are sent to the integration.
 - `project_id` (String) Unique 24-hexadecimal digit string that identifies your project, also known as `groupId` in the official documentation.
 - `provider_type` (String) The provider type for the metric integration. Identifies the third-party service provider.
 
+### Optional
+
+- `oauth` (Attributes) OAuth 2.0 client credentials configuration. Required when auth_type is `OAUTH2`. Secrets are never returned. (see [below for nested schema](#nestedatt--oauth))
```
JuliaMongo, line 76: Required -- active voice -- we are asking to specify them. 
Secrets are never returned -- active voice -- who never returns them? Atlas never returns secrets.

docs/resources/metric_integration.md
```diff
@@ -93,6 +97,37 @@ Required:
 - `value` (String, Sensitive) Header value.
 
 
+<a id="nestedatt--oauth"></a>
+### Nested Schema for `oauth`
+
+Required:
+
+- `client_auth_method` (String) How the client authenticates to the token endpoint. `CLIENT_SECRET` sends a shared secret. `PRIVATE_KEY_JWT` signs a client assertion with an Atlas-generated, Atlas-managed key. Register the returned JWKS URL with your identity provider.
```
JuliaMongo, line 105: should registering happen first? I am not sure about the order here.

docs/resources/metric_integration.md
```diff
@@ -93,6 +97,37 @@ Required:
...
+- `client_id` (String) OAuth 2.0 client identifier registered with the token endpoint.
+- `token_endpoint` (String) OAuth 2.0 token endpoint URL. Must use HTTPS.
+
+Optional:
+
+- `client_secret` (String, Sensitive) Shared client secret. Required when client_auth_method is `CLIENT_SECRET`, and rejected for `PRIVATE_KEY_JWT`. Encrypted at rest and never returned.
```
JuliaMongo, line 111: active voice re who encrypts and who never returns. 

not clear about the rejected part -- might need a few more details.

docs/resources/metric_integration.md
```diff
@@ -93,6 +97,37 @@ Required:
...
+- `scopes` (Set of String) Optional OAuth 2.0 scopes requested on the token, sent as a space delimited `scope` parameter. Applies to both client authentication methods.
```
JuliaMongo, line 112: which "both methods" are implied here? 
Name them?

Are there any other params that apply to both? Existing params that haven't changed with this PR but that apply to both?

docs/resources/metric_integration.md
```diff
@@ -93,6 +97,37 @@ Required:
...
+- `token_request_params` (Map of String) Optional provider-specific parameters added to the token request, for example a resource indicator. Applies to both client authentication methods.
```
JuliaMongo, line 113: same here: name both auth methods to which this applies.

docs/resources/metric_integration.md
```diff
@@ -93,6 +97,37 @@ Required:
...
+
+Read-Only:
+
+- `signing_key_info` (Attributes) Read-only metadata for the Atlas-managed signing key used by `PRIVATE_KEY_JWT`. Present only for that method. Register the jwks_uri with your identity provider. Atlas rotates the underlying key without changing this URL. (see [below for nested schema](#nestedatt--oauth--signing_key_info))
```
JuliaMongo, line 117: same comments as above for similar content

docs/resources/metric_integration.md
```diff
@@ -93,6 +97,37 @@ Required:
...
+
+<a id="nestedatt--oauth--signing_key_info"></a>
+### Nested Schema for `oauth.signing_key_info`
+
+Read-Only:
+
+- `algorithm` (String) Signing algorithm of the Atlas-managed key.
+- `created_at` (String) When the currently active signing key was created. This parameter expresses its value in the ISO 8601 timestamp format in UTC.
```
JuliaMongo, line 125: same comments as above for similar content.

tools/codegen/config.yml
```diff
@@ -1662,6 +1662,17 @@ resources:
           description: *project_id_description
         headers.value:
           sensitive: true
+        oauth:
+          description: "OAuth 2.0 client credentials configuration. Required when auth_type is `OAUTH2`. Secrets are never returned."
+        oauth.client_secret:
+          sensitive: true
+          description: "Shared client secret. Required when client_auth_method is `CLIENT_SECRET`, and rejected for `PRIVATE_KEY_JWT`. Encrypted at rest and never returned."
+        oauth.signing_key_info:
+          description: "Read-only metadata for the Atlas-managed signing key used by `PRIVATE_KEY_JWT`. Present only for that method. Register the jwks_uri with your identity provider. Atlas rotates the underlying key without changing this URL."
```
JuliaMongo, line 1787: see comments above

Not clear "present only for that method" -- which method. for which method it is not present? what does "not present" mean?

Add a word definition for what is jwks_uri

When should registering happen? Is it before some action, or can it be registered at any point? Not clear.

tools/codegen/config.yml
```diff
@@ -1678,6 +1689,8 @@ resources:
         overrides:
           project_id:
             description: *project_id_description
+          oauth.signing_key_info:
+            description: "Read-only metadata for the Atlas-managed signing key used by `PRIVATE_KEY_JWT`. Present only for that method. Register the jwks_uri with your identity provider. Atlas rotates the underlying key without changing this URL."
```
JuliaMongo, line 1809: same q here re: registering -- is it a prerequisite?

JuliaMongo (approved):
Left a few comments to clarify text strings.

AgustinBettati (commented):
Changes LGTM, want to understand the CI testing and would revise examples

docs/resources/metric_integration.md
EspenAlbert: OAuth examples for both client authentication methods are deferred to CLOUDP-445734, as recorded in the PR description

internal/serviceapi/metricintegration/resource_test.go
EspenAlbert: Should we link to a CLOUDP here so we don't forget?

internal/serviceapi/metricintegration/resource_test.go
EspenAlbert: Removing `oauth` currently omits it from the PUT, so the API retains the stored OAuth configuration; add `send_null_as_null_on_update` for the block and cover removal

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -204,3 +216,267 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
 		return fmt.Sprintf("%s/%s", projectID, integrationID), nil
 	}
 }
+
+// oauthExtrasHCL indents each optional oauth attribute to the block indentation and joins
+// them on newlines. Empty attributes are skipped, so omitted fields are not rendered.
+func oauthExtrasHCL(attrs ...string) string {
```
EspenAlbert, line 386: [nit] consider moving this to the hcl helper package

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -30,8 +34,17 @@ const (
 			depends_on = [mongodbatlas_metric_integration.test]
 		}
 	`
+	// Dummy endpoints for OAuth integrations.
+	oauthTokenEndpoint = "https://192.0.2.2/oauth2/token" //nolint:gosec // Test data
+	oauthClientID      = "atlas-otel-test"                //nolint:gosec // Test data
+	oauthEndpoint      = "https://192.0.2.1/v1/metrics"
+	// Pre-rendered token_request_params HCL fragments, passed to the config builders.
+	trParamsClientSecret  = `token_request_params = { resource = "atlas:otel:test" }`
+	trParamsPrivateKeyJWT = `token_request_params = { resource = "TEST:URI:OTel" }`
```
EspenAlbert: This now configures `TEST:URI:OTel`, but the plural data-source assertion still expects `JPMC:URI:OTel` at line 368, so the feature-enabled acceptance test will fail

internal/serviceapi/metricintegration/resource_test.go
marcabreracast: Removing `oauth` produces a 400 from the API. `oauth` is required when `auth_type` is `OAUTH2`, that's why there's no test for it: removing `oauth` while staying `OAUTH2` isn't a valid state. The error message is `Invalid attribute oauth is required when authType is 'OAUTH2'`.

EspenAlbert: That 400 matches the documented `OAUTH2` behavior, but the switch to `HEADER` is still open: with `auth_type = "HEADER"` and `oauth` removed, the untagged null `oauth` object is omitted from the PUT (`resource_schema.go:163` has no null-update tag) and the update contract retains omitted OAuth fields, so the stored OAuth configuration survives and re-enters state on refresh. `send_null_as_null_on_update` for the block plus a transition test would close it

internal/serviceapi/metricintegration/resource_test.go
marcabreracast: Apologies for the confusion. The tests are gated in CI because the OAuth metric integration is not yet enabled on the CI test project, so creating an OAUTH2 integration would be rejected there. Created CLOUDP-447244 to track this and improved the comment in [c1c7b03](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4737/commits/c1c7b03d98c9f90be05ff4c831e5b30cf70042ac)

tools/codegen/atlasapispec/multi-version-api-spec.flattened.yml
```diff
@@ -18123,10 +18974,13 @@ components:
           minItems: 1
           type: array
           uniqueItems: true
+        oauth:
+          $ref: "#/components/schemas/OAuthConfigRequest"
         providerType:
           description: The provider type for the metric integration. Identifies the third-party service provider.
           enum:
             - CUSTOM
+            - DATADOG
```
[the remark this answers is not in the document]
marcabreracast: This is done as a result of fetching OpenAPI dev spec. Resolving

docs/resources/metric_integration.md
marcabreracast: Yes, this will be addressed in CLOUDP-445734

internal/serviceapi/metricintegration/resource_test.go
marcabreracast: I think it's better to keep it local for now, as it's single-use here and hardcodes the `oauth` block indentation, so I'd say it isn't general enough to move to the helper package

.changelog/4737.txt
marcabreracast: Short is the repo's changelog style, see  see [contributing/changelog-process.md](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/contributing/changelog-process.md), but agreed that the wording could be more explicit. Updated it in [328cf3d](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4737/commits/328cf3d71b67554d6e40e9ae1a7304556eedb9b9)

internal/serviceapi/metricintegration/resource_test.go
EspenAlbert: That covers an invalid `OAUTH2` configuration, but not switching an existing integration to `HEADER`. In that valid configuration, `oauth` is absent from Terraform configuration and becomes a null object. Because the generated root field has no `sendnullasnullonupdate` tag, the marshaller omits `oauth` from the PUT instead of sending `"oauth": null`.

The update contract says omitted OAuth fields retain their stored values, so Atlas receives the new `authType = "HEADER"` and headers but no instruction to delete the previous OAuth configuration. Since the API returns `oauth`, the stale values can re-enter Terraform state after refresh despite the absent configuration block. Generate `send_null_as_null_on_update` for the root `oauth` attribute and add an acceptance-test step that changes from OAuth to header authentication.

More details: 🤖
 When configuration changes from OAuth to headers, the desired Terraform configuration is conceptually:

 ```hcl
   auth_type = "HEADER"                                                                                                                                                                                          
   headers = [{ name = "Authorization", value = "..." }]                                                                                                                                                         
   # oauth is absent                                                                                                                                                                                             
 ```

 At that point oauth is a null Terraform object. The generated model at resource_schema.go:163 has no sendnullasnullonupdate tag, so the autogen marshaller omits the oauth key entirely from the update PUT
 rather than sending "oauth": null.

 The OpenAPI update contract explicitly distinguishes those cases: an omitted OAuth field retains its stored value, while an explicit null clears it. Thus the request updates authType to HEADER and supplies
 headers, but does not tell Atlas to delete the previously stored OAuth configuration.

 That leaves stale OAuth data on the server. Because oauth is returned by the API and is optional rather than computed on the resource schema, a subsequent read can put that server-side OAuth value back into  
 Terraform state even though the configuration no longer contains an oauth block. That can cause persistent plan/state drift and means switching authentication modes does not fully apply the declared
 configuration.

The fix is to generate send_null_as_null_on_update for the root oauth attribute, then add an acceptance-test step that creates OAuth, changes to HEADER while omitting oauth, and verifies the API/state no
 longer contain OAuth fields.

EspenAlbert: Thank you!

marcabreracast: Oh sorry! I misunderstood the use case, now I get it's related to making the switch from `OAUTH2` to `HEADERS`. 
Thanks for the analysis, but I checked and the API clears `oauth` natively when doing the switch from OAUTH2 to HEADER, so I don't believe the `send_null_as_null_on_update` is needed. I added a test to cover this scenario and it passed locally (see [3dd8b4b](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4737/commits/3dd8b4b911c9364cca98888da50410906111b168)). PR description has been updated to include the passing test.

marcabreracast:
RE: docs comments. Synced offline and clarified that for autogen, we streamline the descriptions from the OpenAPI spec, so these improvements should be done on the API side to avoid extra overriding. Will sync with the team in upcoming meeting to decide on what's the best approach to communicate this.

EspenAlbert (approved):

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -204,3 +216,313 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+	extraAttrs := []string{}
+	if len(scopes) > 0 {
+		extraAttrs = append(extraAttrs, "scopes = "+hcl.StringSliceToHCL(scopes))
+	}
+	extraAttrs = append(extraAttrs, tokenRequestParams)
+	return configOauth(projectID, endpoint, tokenEndpoint, clientID, "PRIVATE_KEY_JWT", withDS, oauthExtrasHCL(extraAttrs...))
+}
+
+// TestAccMetricIntegration_oauthClientSecret covers the CLIENT_SECRET OAuth path
+func TestAccMetricIntegration_oauthClientSecret(t *testing.T) {
```
EspenAlbert: [nit] we usually put acceptance tests before the helpers

internal/serviceapi/metricintegration/resource_test.go
```diff
@@ -204,3 +216,313 @@ func importStateIDFunc(resourceName string) resource.ImportStateIdFunc {
...
+			{
+				Config: configOauthClientSecret(projectID, endpoint, oauthTokenEndpoint, oauthClientID, clientSecret, scopes, "", false),
+				Check:  checkOauthClientSecret(clientSecret, scopes, 0, nil),
+			},
+			{
+				Config: headerConfig,
+				Check:  headerCheck,
+			},
+			{
+				Config:             headerConfig,
```
EspenAlbert: why non-empty plan here?

internal/serviceapi/metricintegration/resource_test.go
AgustinBettati: Thanks for covering this case. 
Just to be clear, given it is a PUT I would expect an omitted `oauth` property implies it is removed, which I understand is the confirmed behaviour given tests are passing without having `send_null_as_null_on_update`. Explicit null is relevant for PATCH operations, in which case omission is interpreted as "leave value as is".

tools/codegen/config.yml
```diff
@@ -1778,6 +1778,17 @@ resources:
...
+          description: "OAuth 2.0 client credentials configuration. Required when auth_type is `OAUTH2`. Secrets are never returned."
+        oauth.client_secret:
+          sensitive: true
+          description: "Shared client secret. Required when client_auth_method is `CLIENT_SECRET`, and rejected for `PRIVATE_KEY_JWT`. Encrypted at rest and never returned."
+        oauth.signing_key_info:
+          description: "Read-only metadata for the Atlas-managed signing key used by `PRIVATE_KEY_JWT`. Present only for that method. Register the jwks_uri with your identity provider. Atlas rotates the underlying key without changing this URL."
+        oauth.scopes:
+          request_body_usage: send_null_as_empty_on_update
+        oauth.token_request_params:
+          request_body_usage: send_null_as_empty_on_update
```
AgustinBettati, line 1791: just double checking, without this override do we get non-empty plans when attribute is omitted?

AgustinBettati (approved):

internal/serviceapi/metricintegration/resource_test.go
EspenAlbert: `ExpectNonEmptyPlan: false` asserts that the second plan is empty, which verifies refresh did not restore an OAuth value after the switch

tools/codegen/config.yml
EspenAlbert: Yes. Removing a configured map plans an update; without this override the marshaller omits the null map, Atlas retains the prior map, and read restores it to state so the next plan remains non-empty. The existing removal step checks the returned map count is zero after the override sends `{}`

marcabreracast: Correct. To my understanding, `auth_type` is a discriminator: switching it discards the previous type's block, so `oauth` is cleared even if omitted. `scopes`/`token_request_params` don't get that treatment: within a stable `auth_type` they're retained when omitted, which is why `send_null_as_empty_on_update` is needed. I can confirm with the upstream team, but I believe the API is designed as intended unless I'm missing something

internal/serviceapi/metricintegration/resource_test.go
marcabreracast: This check was to explicitly test for the empty plan after the `oauth` config is removed, but it makes sense to drop it as it's a bit repetitive, the previous step's state check already asserts `oauth` is absent. Addressed in [677b705](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/4737/commits/677b7051625df1dc3c1ab4eab64e67d666d8a918)

internal/serviceapi/metricintegration/resource_test.go
marcabreracast: Addressed in 677b705 consistently for the whole file

PR #4742, opened 2026-09-17
test: Removes acceptance test org users during org cleanup

EspenAlbert:
## Description

`TestCleanProjectAndClusters` deletes test projects but leaves their org users behind, so users created by acceptance tests keep accumulating in the org and keep org membership after their projects are gone.
Example error from today: `MAX_USERS_PER_ORG_EXCEEDED`

Adds `TestCleanOrgUsers`, which lists the org users, keeps those whose email matches a cleanup prefix, and removes them. The default prefix is `test-acc-tf-`, which is what `acc.RandomEmail()` produces (`acc.prefixName` + "-"). Override it with `MONGODB_ATLAS_CLEAN_USER_EMAIL_PREFIXES=prefix1,prefix2`, mirroring `MONGODB_ATLAS_CLEAN_PROJECT_PREFIXES`. A `RESOURCE_NOT_FOUND` on delete is skipped, since a concurrent cleanup run may have removed the user already.

The test is separate from `TestCleanProjectAndClusters` so users are still cleaned when project cleanup fails, and `make clean-atlas-org` now runs both. Like the project cleanup, it is gated on `MONGODB_ATLAS_CLEAN_ORG=true`, requires `MONGODB_ATLAS_ORG_ID`, and honours `DRY_RUN` by only counting the users it would remove.

No provider code changes, so no changelog entry.

## Type of change:

## Required Checklist:


internal/testutil/clean/org_clean_test.go
```diff
@@ -161,6 +161,55 @@ func TestCleanProjectAndClusters(t *testing.T) {
...
+		}
+		if !dryRun {
+			if _, err = client.MongoDBCloudUsersAPI.RemoveOrgUser(ctx, orgID, user.GetId()).Execute(); err != nil {
+				if admin.IsErrorCode(err, "RESOURCE_NOT_FOUND") {
+					continue // already removed by a concurrent run
+				}
+				return removed, err
+			}
+		}
+		t.Logf("removed test user %s", user.GetUsername())
```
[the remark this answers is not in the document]
EspenAlbert: Valid; the log sits after the `if !dryRun` block so it fires in both modes, and the make target defaults to `DRY_RUN=true`. Fixed: dry-run now logs `would remove test user %s` and the `removed test user %s` line only fires after a successful `RemoveOrgUser`

internal/testutil/clean/org_clean_test.go
```diff
@@ -161,6 +161,55 @@ func TestCleanProjectAndClusters(t *testing.T) {
...
+	if !cleanOrg {
+		t.Skip("skipping test; set MONGODB_ATLAS_CLEAN_ORG=true to run")
+	}
+	dryRun, _ := strconv.ParseBool(os.Getenv("DRY_RUN"))
+	orgID := os.Getenv("MONGODB_ATLAS_ORG_ID")
+	require.NotEmpty(t, orgID, "MONGODB_ATLAS_ORG_ID must be set")
+	prefixes := userEmailPrefixes()
+	t.Logf("cleaning users with email prefixes %v (DRY_RUN=%t)", prefixes, dryRun)
+	removedUsers, err := removeTestUsers(t.Context(), t, dryRun, acc.ConnV2(), orgID, prefixes)
+	require.NoError(t, err)
```
EspenAlbert, line 179: unauthorized responses fail the test instead of being handled as transient skips like the other clean helpers. `removeClusters` and the other list helpers wrap their calls with `clean.SkipUnauthorizedErr`, and `TestCleanProjectAndClusters` treats `clean.ErrUnauthorized` as a transient skip with a "will retry next run" log. Here a 401 on `ListOrgUsers` or `RemoveOrgUser` reaches `require.NoError` and fails `make clean-atlas-org` before the `SUMMARY` line is printed. Wrap the list error with `clean.SkipUnauthorizedErr` inside the `dsschema.AllPages` callback (the helper drops the `*http.Response`, so the wrap must happen there) and skip the test on `errors.Is(err, clean.ErrUnauthorized)`

internal/testutil/clean/org_clean_test.go
```diff
@@ -161,6 +161,55 @@ func TestCleanProjectAndClusters(t *testing.T) {
...
+
+// defaultUserEmailPrefixes mirrors the emails built by acc.RandomEmail (`acc.prefixName` + "-").
+var defaultUserEmailPrefixes = []string{"test-acc-tf-"}
+
+// removeTestUsers removes every org member whose email starts with one of the given prefixes.
+// In dry-run mode it only counts the users it would remove.
+func removeTestUsers(ctx context.Context, t *testing.T, dryRun bool, client *admin.APIClient, orgID string, prefixes []string) (int, error) {
+	t.Helper()
+	users, err := dsschema.AllPages(ctx, func(ctx context.Context, pageNum int) (dsschema.PaginateResponse[admin.OrgUserResponse], *http.Response, error) {
+		return client.MongoDBCloudUsersAPI.ListOrgUsers(ctx, orgID).PageNum(pageNum).Execute()
```
EspenAlbert: nit: set `.ItemsPerPage(itemsPerPage)` on the `ListOrgUsers` call like `readAllProjects` and the other paginated lists in this file, so the page size does not depend on the API-side default

lantoli (approved):

